PR #280, opened 2021-03-26
Add support for Macos

baothientran:
Fix some minor build config and compilation errors.

### DIFF
--- a/CesiumForUnreal.uplugin
+++ b/CesiumForUnreal.uplugin
@@ -20,13 +20,13 @@
 			"Name": "CesiumRuntime",
 			"Type": "Runtime",
 			"LoadingPhase": "Default",
-			"WhitelistPlatforms": [ "Win64" ]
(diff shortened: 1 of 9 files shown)


kring:
@baothientran this PR looks good, let's get it merged.

I looked a little into setting up CI for the macOS version, but I don't see a good way to do it. The technique I used to install UE on Travis/Windows isn't supported on macOS. Maybe there's some other way to install UE and then copy it to the Travis macOS VM to do the build. Or at least automate the install process.

Plan B is to do builds on a dedicated instance that we manually install UE on. But only dedicated hardware (not a VM) is available from AWS. Which means a single mac mini on AWS costs about $800/month.

Plan C is we manually do builds for releases. 🤷 

In any case, these changes should be safe to merge into main.

baothientran:
@kring we can merge it. Travis CI can be in a different PR

PR #281, opened 2021-03-26
CesiumSunSky now automatically attaches to the CesiumGeoreference

nithinp7:
Just makes it a little easier, all that is needed is to drag-drop the `CesiumSunSky` into the scene now. Also made longitude and latitude invisible in the editor since they should be strictly controlled by the `CesiumGeoreference`.

Edit: 
TODO before merge:
-implement Kevin's suggestion of sticking to the same georeference paradigm of implementing ICesiumGeoreferenceable.


kring:
Is it possible to make SunSky work the way Cesium3DTileset works, where instead of the Georeference having a reference to the SunSky, the SunSky implements `ICesiumGeoreferenceable`, registers with the Georeference, and updates itself when it's notified that the georeference has changed?

I'm not sure it makes a big difference, but it seems better to stick to one paradigm unless there's a good reason to switch it up.

nithinp7:
@kring That's a great idea! Since we created our own version of SunSky we should totally be able to do that now. If it doesn't work entirely in blueprint for some reason, we can give CesiumSunSky a C++ parent class that will definitely work.

nithinp7:
Will be tackled properly as part of the [georeference refactoring](https://github.com/CesiumGS/cesium-unreal/issues/330).

PR #297, opened 2021-04-02
Log selection stats at Verbose instead of Display level.

kring:
Combined with CesiumGS/cesium-native#193, this greatly reduces the amount of information written to the Output window by default, making it more likely that we (and users) notice real problems.


cesium-concierge:

Issue #11, opened 2020-06-09
CesiumJS-style camera


kring:
Implement a camera similar to the one in CesiumJS, making it easy to navigate a global scene.

xuelongmu:
I have a basic dynamic camera system working in Unreal, here's a video: https://drive.google.com/file/d/1p9wNMR5QQ4MUAXm2_njX0v3qAW1BQK-Q/view?usp=sharing

I am doing a "infinite" line trace (largest floating point number I think) down in the Z direction of the pawn and using the distance as the new camera speed, with a minimum clamp of 8000 and a large number as the maximum clamp. I suspect that this will break when you move far away from the world origin, as the direction from pawn -> terrain will diverge from Unreal's Z axis AFAIK. I will try finding a function that will return the globe-aware Z vector.

xuelongmu:
I used the `InaccurateTransformRotatorEnuToUe` to create a Down pointing vector (in East North Up ref frame) and convert it to Unreal ref frame. Screenshot below of my code:
![image](https://user-images.githubusercontent.com/10490807/114108162-8c3e1400-98a0-11eb-93ce-c1d2848701de.png)

This seems to work correctly near the Georeference origin, but it breaks down when moving far away from the origin. I recorded [it as a video here](https://drive.google.com/file/d/1SNMReh-JC1Mkg3DXqroVMK7rteXANeBn/view?usp=sharing). The red line is the Enu-converted line trace. For comparison, the blue line is the ordinary Unreal world space Z axis vector. Interestingly, it seems to hold up better than the red line, which by the end of the video is shooting off into space...

@nithinp7 @kring Do you have any ideas for what may be happening here?

nithinp7:
@xuelongmu That may be related to a problem I ran into recently (and probably introduced): https://github.com/CesiumGS/cesium-unreal/issues/265
`TransformRotatorEnuToUe` is implemented using `ComputeEastNorthUpToUnreal`, I remember looking for a while and not being able to find why it wasn't working. 

Also have you tried using the ellipsoid height, what kind of results does that give? That may be simpler than the line-trace, but line-trace probably works more consistently near the ground across the world. It may be worth doing a smaller line-trace and then switching to ellipsoid height once you hit a certain altitude (or just if the line-trace fails for any reason). That way, you wouldn't have to do raycasts over huge distances. 

[...]

Issue #284, opened 2021-03-26
Zooming out indefinitely causes flickering of position


javagl:
Looking down and then moving backwards quickly brings the whole world into the viewport. However, it glitches out at some point (*Edit: apparently, only in 'Play' mode*). It *looks* like the camera position is switching between "-large value" and "+large value", causing the globe to flicker and the navigation to behave erratically. 

This is accompanied by log messages
```
LogLevel: WORLD TRANSLATION BEGIN {2117412412, 2140242875, 2108424402} -> {2131858716, 2117999429, 2126970786}
LogLevel: WORLD TRANSLATION END {2131858716, 2117999429, 2126970786} took 0.3395 ms
...
LogLevel: WORLD TRANSLATION BEGIN {2131858716, 2117999429, 2126970786} -> {-2145834550, 2129722430, -2143266186}
LogLevel: WORLD TRANSLATION END {-2145834550, 2129722430, -2143266186} took 0.3437 ms
```
A wild guess is that the casts to `int` around https://github.com/CesiumGS/cesium-unreal/blob/c0d01255a669143e281337516ec9afae0e65d62f/Source/CesiumRuntime/Private/CesiumGeoreference.cpp#L522 are not the right thing to do, but that has to be analyzed further.

(An aside: Of course, there *is* a limit for zooming out. But for a "normal" unreal level, this seems to be orders of magnitude later, and in the case of Cesium, the observed behavior happens at a point where one could reasonably expect a "view of the full globe" to still work)

kring:
Yep @javagl's diagnosis is spot-on. The origin can only be rebased out to a distance of about 21 million meters. Once we hit that limit, we should just stop rebasing (and let the local coordinates grow) rather than letting the coordinates wrap around.

javagl:
This is one case of an issue that *might* possible to fix with reasonable effort, but that might have some conflict with https://github.com/CesiumGS/cesium-unreal/issues/330 (depending on when/how the latter is tackled). I can have a look at that later (and assign this issue to me **prelimiarily** (as a reminder)). **Iff** this is really fixable by adding some

    if (value < limit) rebase();
    else dont();

then I'd do that (usually, it's not that easy, but as such, "triaging" and estimating the effort is the first step here)

javagl:
> usually, it's not that easy

Indeed. After adding some debug output...
```
      int32 newX = static_cast<int32>(cameraLocation.X) +
              static_cast<int32>(originLocation.X);
[...]

PR #319, opened 2021-04-08
CesiumJS-style camera

### LINKED ISSUE #344: InputAction not configured in empty project
When using the plugin with a clean empty project with no template, there are no axes or actions configured by default. This means that inputs such as these on the FloatingPawn` cannot be accessed until the input action is created in the project settings.

image

The alternative is to use hardcoded keybinds like so:
image

It's more brittle, and not the best practice for how inputs should be normally set up for applications. But this may be a reasonable exception, as only the Cesium for Unreal Samples project would have these input actions set up without user intervention.

### LINKED ISSUE #362: FloatingPawn not getting spawned at correct location when using PlayerStart
Currently the FloatingPawn / GlobeAwareDefaultPawn is not getting spawned correctly on begin play. However, other types of pawns (DefaultPawn, ThirdPersonCharacter) _are_ getting spawned in the correct place.

I tested this, using a PlayerStart instead of manually drag-dropping FloatingPawn into the scene, and saw that the player pawn was getting spawned to 0,0,0 whereas the PlayerController, CWT, etc. were at something like `(X=-1500870.000000,Y=5105614.000000,Z=637592064.000000)`. As a result I couldn't see anything, just a black screen.

Adding a CesiumGeoreference caused the FloatingPawn to spawn in the correct place. I'm not sure if this is the best solution. (I get a warning about moving the PlayerStart collision capsule, and because it's part of the engine source I can't edit it easily)

Another way would be to teleport the FloatingPawn to the PlayerController is, probably on BeginPlay, as the PlayerController spawns in the correct place. (tested by manually copy-pasting the coordinates to FloatingPawn)

I suspect something about the GlobeAware part of GlobeAwareDefaultPawn is causing this issue. @nithinp7 looping you in here

xuelongmu:
Addresses #11 . Closes #362, closes #344 

Still debugging the down vector behavior noted in the issue, just wanted to make it visible for others to try out.


cesium-concierge:

xuelongmu:
@nithinp7 I modified the system as per [your suggestion](https://github.com/CesiumGS/cesium-unreal/issues/11#issuecomment-816333822) and am now using the unreal Z axis up until 100km above the surface, then switching to WGS84 altitude beyond that.

There may still be value in manually controlling the pawn speed, so perhaps we can repurpose the scroll wheel speed controls as a manual override? Trying to think of an intuitive UX for this.

xuelongmu:
> There may still be value in manually controlling the pawn speed, so perhaps we can repurpose the scroll wheel speed controls as a manual override? Trying to think of an intuitive UX for this.

Middle scroll now changes a multiplier to the base (dynamic) movement speed. Middle mouse click resets this multiplier.

xuelongmu:
@nithinp7 Let me know when you get a chance to review this or if you have any questions, thanks! This PR is for the FloatingPawn only.

nithinp7 (commented):
@xuelongmu Great job it already feels a ton smoother and easier than the old controls! Just a few comments and ideas:

* I think the line-trace part should be something like this, the unreal negative-Z direction is not necessarily down with respect to the globe. It is pretty close when somewhat near the georeference origin, but we can't assume that. Without this change, if you go well past the horizon and then test flying up and down from the ground you will sometimes notice the speed being unresponsive to the altitude.
line-trace-down

* Maybe this could be simplified by using the InaccurateUeToLongitudeLatitudeHeight directly.
ue-to-longlatheight

* I noticed that when I'm flying forward past mountains, the speed suddenly slows right as I cross over a peak, which makes complete sense given how the line-trace works right now. The one way to fix this is of course to just use the WGS84 height all the time instead of line-trace. However, I really like how the line-trace response feels right now near the ground, so I would rather keep it in. So maybe we could do the line-trace in the direction of movement instead? If you decide to switch the line-trace to use the velocity direction, I _think_ you should switch the altitude calculation to solely use the WGS84 height. A combination of a forward line-trace and the ellipsoid height should probably be enough to parameterize a super responsive camera.
* Going along with the last comment, do you think we could use FloatCurves to adjust the responsiveness of the speed at different altitudes / line-trace distances? We may be able to fine-tune some pretty smooth behavior if we had FloatCurves we could manipulate. This would let us slow players down that are getting too far from the Earth (about to cause a glitch) and it will help us cushion the camera right before it hits the ground (the camera slows nicely right now near the ground but it still feels like a "face-first" landing sometimes). If wanted, the FloatCurves could be exposed so users can customize them, but the default one should be fine-tuned to feel good at least.
* Also this is slightly orthogonal but we should create default FloatCurves for the fly-to on the FloatingPawn. Currently users have to manually create them, otherwise the fly-to behavior is very low to the ground and pretty ugly-looking without the curves.

kring:
As a point of reference, in CesiumJS, when you right-click-drag to zoom, the zoom speed is controlled by the distance from the camera to the point that you right-clicked on. I actually don't know offhand if the speed changes during the course of the drag as the camera gets closer or farther from that original point.

When you zoom with the mouse wheel, the speed is controlled by the distance to the point the mouse is over.

Similarly, when you left-click-drag, the pan speed is dependent on the distance to the point you originally clicked.

xuelongmu:
I've updated the behavior based on discussions with @nithinp7:

Now the floating pawn does two line traces, one towards the center of the Earth (0,0,0 ECEF) and one that extends 2km outwards from the player camera view. Whichever line trace returns a valid, shorter value is stored, and we can call this `LineTracedPawnSpeed`.

When you start moving from a standstill, the `LineTracedPawnSpeed` is determined _once_ and set as the pawn's max speed. This means that if you're stationary and looking at the top floor of a skyscraper, you won't zip past the building suddenly upon moving, as long as it's at the center of your view.

If you continue to move at a high input value (>0.7 of the input axes, which is more relevant for controllers as keyboards max out at 1.0 immediately upon a key press) for over one second, DynamicSpeed mode gets engaged. In DynamicSpeed mode, the `LineTracedPawnSpeed` is recalculated on tick and based on the following criteria, a new pawn max speed is set:

- If the camera line trace hits, then the `LineTracedPawnSpeed` is used.
- Otherwise, `LineTracedPawnSpeed` is used _only_ if it returns a greater value than the current max speed.

This prevents slowdowns when passing over mountains, because (ostensibly) the player is looking out at the horizon and is not interested in exploring the mountains. If they were interested, then the camera would be pointed at the mountains, which would register a hit with the camera line trace, and the max speed would be lowered accordingly.

I still need to add in float curves to customize the function for max pawn speed <-> `LineTracedPawnSpeed`, as it's currently just linear with a min/max clamp.

xuelongmu:
@nithinp7 The float curves have now been added, as well as an `EnableDynamicSpeed` parameter to control whether the player wants dynamic or manual speed control.

@argallegos Do you also want to give it a spin and see if it feels good to move around with this new system?

xuelongmu:
@nithinp7 Do you think it would be worth adding mouse control with the parameters Kevin mentioned? It might be simple enough for me to add...

> As a point of reference, in CesiumJS, when you right-click-drag to zoom, the zoom speed is controlled by the distance from the camera to the point that you right-clicked on. I actually don't know offhand if the speed changes during the course of the drag as the camera gets closer or farther from that original point.
> 
> When you zoom with the mouse wheel, the speed is controlled by the distance to the point the mouse is over.
> 
> Similarly, when you left-click-drag, the pan speed is dependent on the distance to the point you originally clicked.

nithinp7:
@xuelongmu Good job so far on this! This is getting pretty complicated but I think that's a little inevitable. Just a few thoughts after playing around with this and comparing it to the old FloatingPawn:
- I'm noticing a bit of a stutter in either the FPS or movement I can't really tell which, do you get anything like that on your side? It seems to happen more noticeably when zooming out long distances. It may be possible that it's caused by the overhead of the line-traces since the search distance is huge, but then again there's only two so I doubt this is the case. The other possibility I thought of was that maybe it's caused by floating-point precision issues, since the speed is set from a sometimes large floating-point line-trace, but I'm not so confident that's the case either.
- The dynamic speed mode is a little uncontrollable sometimes. When it feels like I'm going a bit too slow and I scroll the mouse-wheel a bit, sometimes it speeds up nicely, but sometimes it shoots out aggressively fast. Also when in dynamic mode, the camera doesn't seem to slow down when flying towards the ground. 
- In this video, I'm flying straight near the ground, it's going a little slow so I scroll to increase speed and I enter dynamic-speed-mode, so far so good. But then I tilt the camera a little down while I'm still moving forward, and suddenly it takes off and zips right to the hillside.

https://user-images.githubusercontent.com/6706316/115083187-025ffd80-9ed5-11eb-93fd-b8a065fca805.mp4

nithinp7:
Also for the player view line-trace direction, the "Get Actor Eyes View Point" node on the FloatingPawn isn't necessarily the viewpoint of the camera. In fact, the GlobeAwareDefaultPawn tries to never orient itself away from the identity rotation. The controller determines a local look direction in the controller-space, the pawn determines the location, and the camera combines these with a georeferenced transform to establish a full transform. So yeah long story short, you should probably look for a similar function on the camera itself.

xuelongmu:
@nithinp7 Ready for review!

- Fixed georeference conversion matrices being zeroed out in CesiumGeoreference
- Removed duplicate OnBeginPlay code for GlobeAwareDefaultPawn
- Below a minimum height, the FloatingPawn will slow down even in dynamic speed mode
- Speed multiplier resets to 1.0 when you stop moving pawn
- Max pawn acceleration set to `2000000`

As for the stutter, I'm noticing it in both manual and dynamic speed mode and I believe it's due to loading in tiles, or perhaps changing the GeoReference and updating all transforms accordingly.

javagl:
Just curious: Will this affect https://github.com/CesiumGS/cesium-unreal/issues/284 in any way? (It might be that this is about the origin rebasing, and thus unrelated to the camera, but I was just wondering).

xuelongmu:
@javagl The new FloatingPawn will taper off the speed once the player is a certain distance away from the ground, preventing the them from getting to the point where you see the flickering described in #284. But this behavior is something very specific I coded into a float curve, so if the user chooses a different float curve or a different pawn altogether, they will still be able to travel far enough to where flickering occurs.

nithinp7:
@xuelongmu I'm still testing it out and looking through the blueprints, but the issue of the camera slowing down when passing over a hill or building still seems to exist. Also the speed gets capped at too low of an altitude, it's hard to fly around the world reasonably quickly. This second part might just be a matter of tweaking the float curve.

nithinp7:
@xuelongmu The input for the time of day and profiling widget should probably not be bound to specific keys. I think it makes sense to keep them as input actions like they were before. The only problem was that the config file that was initially used to define those input actions was in the cesium-unreal-demo repo. I would think there's probably a way to define the input actions in some sort of config file within the plugin itself right? If there isn't a way to do that, we can keep these as specific keys but it feels a bit rigid to me.

xuelongmu:
> @xuelongmu I would think there's probably a way to define the input actions in some sort of config file within the plugin itself right? If there isn't a way to do that, we can keep these as specific keys but it feels a bit rigid to me.

Can't define within the config file unfortunately.

> I'm still testing it out and looking through the blueprints, but the issue of the camera slowing down when passing over a hill or building still seems to exist. Also the speed gets capped at too low of an altitude, it's hard to fly around the world reasonably quickly. This second part might just be a matter of tweaking the float curve.

Mind taking some video clips of what you're seeing?

nithinp7:
It's pretty much the same as before: the minimum line trace gets chosen as the line-trace speed and it also returns the override flag - so even though it's less than the current speed, the dynamic mode gets overriden and slowed down.

argallegos:
@xuelongmu This is fantastic, the new movement overall feels super smooth and satisfying. I've noticed a little bit of occasional slowdowns at certain points, presumably the same as Nithin is. I'm going to keep testing around to see if I can reliably replicate it. 

My main thought is that I also agree that the speed tapers off too early. At the very least, the pawn should be able to move quickly up until the world is about this big on screen - 
image
(Sorry for that unit of measurement, I'm not sure a more efficient way to convey that amount of distance.)
At that altitude, the player should be able to move around the entire circumference of the world in around 7 seconds. I know that's probably a whole other level of complexity with what you've already created, but I think it's important that users can quickly manipulate the entire globe when very zoomed out. 

That being said, awesome job!

nithinp7:
The min height check seems to only happen when the forward camera trace returns 0, is that correct? 
dynamic-speed-min-height

nithinp7:
Also when selecting manual mode (disabling dynamic mode in the details panel) it doesn't seem to go as fast as the normal floating pawn, it's hard to fly it into the atmosphere. If it would make it easier you could totally have both the new and old one as separate pawns.

xuelongmu:
> The min height check seems to only happen when the forward camera trace returns 0, is that correct?
> dynamic-speed-min-height

@nithinp7 This is intentional - the check is only to determine whether the Override Speed flag should be set, if the camera trace does not hit. If the camera trace hits, then the Override Speed flag is always set. 

image

xuelongmu:
@nithinp7 @argallegos I realized that the issue with the pawn moving slowly when very zoomed out was due to a max acceleration cap I'd put in, which worked well when close to the ground but throttled very high speeds. The max acceleration is now set at a minimum floor and then increases as the speed increases.

I adjusted the movement curve so that movement speed increases up to a certain (very high) point, then decreases beyond that point, to prevent the user from traveling too far out (which breaks the system)

I also set a higher cap on manual pawn movement max speed.

@nithinp7 To your point about rubberbanding around hills, I haven't seen that behavior when the camera is pointing at the horizon. When looking at the blueprint execution flow, it is not hitting the `Set Max Move Speed` either. I just made a change that should hopefully prevent accidental "dips" of the camera below the horizon from triggering a movement speed update, by dynamically changing the length of the camera trace depending on the current max speed. From testing it out, this seems to improve usability.

nithinp7:
Great work @xuelongmu! Let's merge this once CI passes. I added the old FloatingPawn back in and renamed the new one to avoid a breaking change until this pawn gets its kinks sorted out. A few loose ends to eventually address:
* The manual mode still caps out the speed at too low an altitude, which is why I added the old FloatingPawn back in.
* We should probably add in some movement-direction-aware logic so that we don't slow down when passing over something. If we aren't moving downward then there shouldn't be a need to slow down even if the downward line trace becomes smaller (e.g., passing over a building).
* Float curves should be added for the FlyTo.
* We should do some sort of profiling on the line-traces. It's tracing against a massive amount triangles per frame in some cases. I still feel like there's a slight lag that's happening, but it's too subtle for me to know for sure. Earlier you mentioned it might be tile loading, I don't think that's the case since tile loading should be asynchronous. We should profile the old vs new floating pawns for comparison.

nithinp7:
Hm something doesn't seem to be working right after the merge from master, I'll look into it.

Issue #317, opened 2021-04-08
Allow Cesium 3D tileset materials to be set to Material Instances as well as materials


argallegos:
The "Material" parameter in the Cesium/Rendering category of a Cesium3DTileset currently only accepts base materials. Material instances are an important part of efficient material design in Unreal. This parameter should be updated to accept instanced materials as well.

xuelongmu:
Changing the parameter type from `UMaterial` to `UMaterialInterface` should let you use Material Instances and base Materials. I started taking a crack at this, and discovered that there were a bunch of other functions/classes that contain `UMaterial` references, that will also have to be changed. Before I continue, @kring are there any reasons I might be missing on why `UMaterial` is being used instead of `UMaterialInterface`?

xuelongmu:
I went ahead and made the changes. Let me know if this works for you @argallegos

PR #320, opened 2021-04-09
Change UMaterial references to UMaterialInterface

xuelongmu:
Closing #317 

@argallegos I quickly checked this by creating a material instance of `GltfMaterialWithOverlays` and setting that as the Material in CesiumWorldTerrain, and the terrain rendered normally (same as if there was no reference set there). Are there other tests to make sure this is working correctly?

### DIFF
--- a/Source/CesiumRuntime/Private/CesiumGltfComponent.cpp
+++ b/Source/CesiumRuntime/Private/CesiumGltfComponent.cpp
@@ -1416,7 +1416,7 @@ UCesiumGltfComponent::CreateOffGameThread(
     AActor* pParentActor,
     std::unique_ptr<HalfConstructed> pHalfConstructed,
     const glm::dmat4x4& cesiumToUnrealTransform,
-    UMaterial* pBaseMaterial) {
+    UMaterialInterface* pBaseMaterial) {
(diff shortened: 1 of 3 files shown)


cesium-concierge:

argallegos:
This works perfectly. I tested with a few modified materials just to be sure, and it responded as expected. I don't imagine this needs an in depth code review, it's very straightforward, but I'll wait to merge just to make sure. Thanks, Xuelong!

javagl:
I have not tested it, but from a glance at the code, it seems to be a form of the usual OO-best-practice of coding against the interface, which basically means to use the most generic class/interface that is sufficient to do accomplish the given task. (As such, we might also have a look at, for example, we're taking an `UStaticMeshComponent` as a parameter where one of its ancestors (`UMeshComponent`, UPrimitiveComponent`...`UObjectBase`) would suffice...)

(This might have implications in the Unreal world, for example, for some Blueprint magic like "Create an instance of the class that is required here", which may then require the user to pick an appropriate subclass. But that wouldn't necessarily be wrong or bad...)

kring:
Thanks @xuelongmu!

PR #324, opened 2021-04-09
Custom editor-mode

nithinp7:
So far this is just the skeleton for a custom editor-mode. In order of priority, these are the things that need to be added to it:
- Globe aware camera controls when in the editor mode: https://github.com/CesiumGS/cesium-unreal/issues/325
- More convenient controls and UI for georeference placement. Although I'm not a big fan of the CesiumJS camera, something along those lines could be useful for placing the georeference or placing sublevels. Mainly the click-and-drag to rotate the globe would be nice to have. This does not necessarily have to be the same as the above camera, but it can be.  
- A straightforward UI for georeferenced sublevels: https://github.com/CesiumGS/cesium-unreal/issues/326

### DIFF
--- a/Source/CesiumEditor/Private/CesiumEditor.cpp
+++ b/Source/CesiumEditor/Private/CesiumEditor.cpp
@@ -7,6 +7,7 @@
 #include "CesiumIonPanel.h"
 #include "CesiumIonRasterOverlay.h"
 #include "CesiumPanel.h"
+#include "CesiumGeoreferenceModeTool.h"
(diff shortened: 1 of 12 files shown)


cesium-concierge:

kring:
@nithinp7 this looks really useful as a starting point if and when we want to add a custom editor mode, but I don't think there's anything actionable in the meantime. So I'm going to close this just to keep our list of PRs cleaner, but reopen if you disagree.

PR #338, opened 2021-04-13
Watermask

nithinp7:
The cesium-unreal part of the watermask implementation. The cesium-native side to it is here: https://github.com/CesiumGS/cesium-native/pull/207

MiamiWater


cesium-concierge:

WellyWonder:
This looks insanely beautiful <3 can't wait for it to be in Unreal Engine....will that be native to Cesium actor?! or similar to Cesium Sun & Sky where the plugin uses the engines own water plugin?!!

nithinp7:
@WellyWonder It will require Unreal's Water plugin to work correctly, but it shouldn't require any extra effort from users just like the CesiumSunSky.

nithinp7:
@kring Can you have a look? I _think_ it's only failing CI because [this change](https://github.com/CesiumGS/cesium-unreal/pull/380) isn't in master yet, but I'm not sure.

I still need to fix the option to enable/disable the water mask from the tileset actor. In fact, the way we handle property updates on the `Cesium3DTileset` actor should probably be overhauled a bit. Certain updated properties can be passed on into `TilesetOptions` in tick with no problem as we are doing, but other changes like the override material and water mask require a reloading of the tileset altogether.  

I can ask @argallegos to help review the new materials in detail tomorrow as well. The only thing I'm concerned about on that front is the fact that I have to scale the raster overlay uv coordinates by 0.25 for it to work for the water mask, which felt arbitrary so I think I might be missing something. Also there were seams that appeared on the South and East edges of water-masked tiles, so I clipped those corners a bit by multiplying by 0.23 instead of 0.25. This weird magic number is in the `WaterMaskSample.uasset`.

kring:
@nithinp7 It took me awhile, but I think I see why the water mask texture sampling is so dodgy. There are two reasons.

The first (and major) reason is that CesiumGltfComponent assumes that all textures have four channels. I believe that will be true when the image comes from STB, but the water mask doesn't, and it only has one channel. So effectively only every 4th pixel is used properly, and there's a bunch of uninitialized memory. This is why you need to multiply by ~0.25.

Another smaller problem is that the texture coordinates aren't quite right. The water mask requires a geographic projection, but the overlay texture coordinates use a web mercator projection. This won't be too noticeable when zoomed in, but if you can see the effect of the water mask when zoomed out, you'll notice that the mask is misaligned inside tiles (it will be correct at the edges).

nithinp7:
@kring that makes a lot of sense, I'll add a switch in the gltf parsing to pass on the number of channels. Thanks for finding that! 🙏

I'll also create geographic projected uv coords for the watermask.

nithinp7:
@kring Can you have a look, I think the water mask is done both here and on cesium-native. I think @argallegos can help us make a dedicated pass of optimizing all our materials including the new water materials here, but otherwise I think everything should be done here except for anything you find while reviewing!

argallegos:
@nithinp7 
Created a branch here for water optimization - https://github.com/CesiumGS/cesium-unreal/tree/watermask-material-optimization
If those changes look good, feel free to merge it in to this branch, or save it for later.

I removed caustics and reflection. Eventually, we will likely want most of that back in, but they aren't really serving a purpose for us right now since you can't go under the water anyways. I've commented the materials in each of those areas so it should be straightforward to hook everything back up again in the future if we need. 

This is a pretty marginal improvement, since most of the overhead is coming from the texture selection, which is essentially doubled for the version with the water. I'll look in to ways to optimize this down the line. This shader will perform a little better, though, which is a plus, and shouldn't come with a meaningful drop in visual quality - though, if I've removed something you were relying on, let me know.

Original stats:
image

Modified material stats:
image

kring:
@nithinp7 I noticed some strange rendering. Hopefully I didn't introduce this with the small changes I made to cesium-native or by merging in master 😬.

It seems like tiles that have a water mask as rendered differently from tiles that are all land:

Watermask on:
image

Watermask off:
image

This is a spot on the harbour south of Melbourne.

Just to show how the tiles align:
image

kring:
I also noticed that the water effect in CesiumJS and Cesium for Unreal are tuned for viewing at _very_ different scales. For example, here's CesiumJS:
image

And here's a similar scale in Cesium for Unreal - no water visible:
image

In Cesium for Unreal, the water isn't really apparent until we get in very close, like here-ish:
image

That looks better than CesiumJS:
image

But really both look pretty bad because the CWT water mask just isn't high enough resolution to use at this scale. So do you think it would make sense to tune the water effect for viewing from farther away?

Source/CesiumRuntime/Public/Cesium3DTileset.h
```diff
@@ -31,20 +45,32 @@ class CESIUMRUNTIME_API ACesium3DTileset : public AActor,
   ACesium3DTileset();
   virtual ~ACesium3DTileset();
 
+  /**
+   * The type of source from which to load this tileset.
+   */
+  UPROPERTY(EditAnywhere, Category = "Cesium", meta = (DisplayName = "Source"))
+  ETilesetSource TilesetSource = ETilesetSource::FromCesiumIon;
```
kring, line 52: Adding and honoring this property will break non-ion tilesets. They'll stop working after upgrading to the new version of Cesium for Unreal because this property defaults to `FromCesiumIon`.

I think we can fix this by implementing backward compatibility for our property serialization. Some docs here:
https://docs.unrealengine.com/en-US/ProgrammingAndScripting/ProgrammingWithCPP/UnrealArchitecture/VersioningAssetsAndPackages/index.html

I'll take a crack.

kring: Hey that turned out to be pretty easy, done in 427ef46.

Source/CesiumRuntime/Private/Cesium3DTileset.cpp
```diff
@@ -587,29 +592,43 @@ void ACesium3DTileset::LoadTileset() {
...
+    // check if the tileset source selection itself changed
+    bool tilesetSourceChanged = this->TilesetSource != this->_lastTilesetSource;
+    this->_lastTilesetSource = this->TilesetSource;
+
+    // check if the current tileset changed
+    switch (this->TilesetSource) {
+    case ETilesetSource::FromUrl:
+      tilesetSourceChanged =
+          pTileset->getUrl() ? TCHAR_TO_UTF8(*this->Url) != pTileset->getUrl()
+                             : this->Url.Len() > 0;
```
kring, line 610: The way this is written, switching from `FromCesiumIon` to `FromUrl` won't make the Cesium ion tileset immediately disappear if the Url is blank. Is that intended? It seems like it might be a little confusing.

nithinp7:
@kring That was a good find regarding the difference in color between the only-land and the land part of watermask tiles. It is especially glaring when there is a sunset or sunrise. I played around with the materials a good bit today and the reason seems to be that only-land uses a default lit shading model (to save shader computation) while watermask and only-water tiles use the single-layer water shading model. So far, I haven't found a way to switch between shading models within the same material. 

There might still be a way to tweak the settings in watermask tiles to minimize if not eliminate the discrepancy, I'll take a look at what the difference between shading models is behind the scenes. A hacky solution though could be to use the single-layer water for everything when the watermask is enabled? It doesn't seem to cause any issues, except for a slight lighting difference.

nithinp7:
@kring Regarding the water effect visibility by distance, for that height I think the water effect should be pretty subtle and the satellite imagery should do most of the heavy lifting. I was comparing it side-by-side earlier today. My main takeaway was that the water effect definitely feels a bit too flat and repetitive, so I am currently trying to overlay more coarse waves on top that can break apart the symmetry and give some slight swelling behavior, and can also provide extra visibility from a height. 

cesium-hawaii-aerial

real-hawaii-aerial

water-reflection3

nithinp7:
@kring As discussed offline, I fixed the last remaining small issue and I think this is ready to merge. I wrote issues for the remaining unaddressed concerns so that we can come back to them when it fits our priority:
- https://github.com/CesiumGS/cesium-unreal/issues/405
- https://github.com/CesiumGS/cesium-unreal/issues/406

kring:
Thanks @nithinp7!

PR #339, opened 2021-04-13
Added Collision Settings and Exclusion Zones

AlbanBERGERET-Epic:
Added Bodyinstance Property to define tiles collision settings on a per-actor basis
Added ExclusionZones concept to cull OSM Buildings when displaying photogrammetry assets
Added BueprintReadOnly flag to CesiumIonToken to be able to check that tokens have been set...

### DIFF
--- a/Source/CesiumRuntime/Private/Cesium3DTileset.cpp
+++ b/Source/CesiumRuntime/Private/Cesium3DTileset.cpp
@@ -14,6 +14,7 @@
 #include "CesiumGeospatial/Ellipsoid.h"
 #include "CesiumGeospatial/Transforms.h"
 #include "CesiumGltfComponent.h"
+#include "CesiumGltfPrimitiveComponent.h"
 #include "CesiumRasterOverlay.h"
 #include "CesiumRuntime.h"
 #include "CesiumTransforms.h"
@@ -26,6 +27,7 @@
 #include "IPhysXCookingModule.h"
 #include "LevelSequenceActor.h"
 #include "Math/UnrealMathUtility.h"
+#include "Misc/EnumRange.h"
 #include "PhysicsPublicCore.h"
 #include "UnrealAssetAccessor.h"
 #include "UnrealTaskProcessor.h"
@@ -915,7 +917,27 @@ void ACesium3DTileset::Tick(float DeltaTime) {
   for (Cesium3DTiles::Tile* pTile : result.tilesToRenderThisFrame) {
     if (pTile->getState() != Cesium3DTiles::Tile::LoadState::Done) {
       continue;
+		}
+     
+    // Consider Exclusion zone to drop this tile... Ideally, should be considered in Cesium3DTiles::ViewState to avoid loading the tile first...
+    if (ExclusionZones.Num() > 0) {
+        const CesiumGeospatial::BoundingRegion* pRegion = std::get_if<CesiumGeospatial::BoundingRegion>(&pTile->getBoundingVolume());
+        bool culled = false;
+        for (FCesiumExclusionZone ExclusionZone : ExclusionZones)
+        {
+            CesiumGeospatial::GlobeRectangle cgExclusionZone = CesiumGeospatial::GlobeRectangle::fromDegrees(ExclusionZone.West, ExclusionZone.South, ExclusionZone.East, ExclusionZone.North);
+            if (cgExclusionZone.intersect(pRegion->getRectangle()))
+            {
+                culled = true;
+                continue;
+            }
+        }
+
+        if (culled) {
+            continue;
+        }
     }
+    
 
     // const Cesium3DTiles::TileID& id = pTile->getTileID();
     // const CesiumGeometry::QuadtreeTileID* pQuadtreeID =
@@ -933,6 +955,22 @@ void ACesium3DTileset::Tick(float DeltaTime) {
       continue;
     }
 
+    // Apply Actor-defined collision settings to the newly-created component. 
+	UCesiumGltfPrimitiveComponent* PrimitiveComponent = static_cast<UCesiumGltfPrimitiveComponent*>(Gltf->GetChildComponent(0));
+	if (PrimitiveComponent != nullptr)
+	{
+		const UEnum* ChannelEnum = StaticEnum<ECollisionChannel>();
+		if (ChannelEnum)
+		{
+			for (int32 ChannelValue = 0; ChannelValue < ECollisionChannel::ECC_MAX; ChannelValue++)
+			{
+				ECollisionResponse response = BodyInstance.GetCollisionResponse().GetResponse(ECollisionChannel(ChannelValue));
+				PrimitiveComponent->SetCollisionResponseToChannel(ECollisionChannel(ChannelValue), response);
+			}
+		}
+	}
+    
+
     if (Gltf->GetAttachParent() == nullptr) {
       Gltf->AttachToComponent(
           this->RootComponent,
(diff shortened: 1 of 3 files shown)


cesium-concierge:
Thanks for the pull request @AlbanBERGERET-Epic!

* :x: Missing CLA.

Source/CesiumRuntime/Private/Cesium3DTileset.cpp
```diff
@@ -915,7 +917,27 @@ void ACesium3DTileset::Tick(float DeltaTime) {
   for (Cesium3DTiles::Tile* pTile : result.tilesToRenderThisFrame) {
     if (pTile->getState() != Cesium3DTiles::Tile::LoadState::Done) {
       continue;
+		}
+     
+    // Consider Exclusion zone to drop this tile... Ideally, should be considered in Cesium3DTiles::ViewState to avoid loading the tile first...
+    if (ExclusionZones.Num() > 0) {
```
kring, line 923: We need a feature like this! And it's coming! But I don't think we want to ship something like this in the meantime, just like we didn't ship the previous rectangle-based clipping system I hacked in for Project Anywhere. The problems are:

* As you noted, it's inefficient to exclude tiles here because they've already been loaded. It needs to happen in cesium-native instead.
* This mechanism only works for tiles that use a "BoundingRegion" as their bounding volume. Lots of tilesets won't, especially photogrammetry. Also, as currently written, such tilesets will cause a crash if the user tries to define an exclusion zone on them (`pRegion` will be nullptr).
* It excludes way too much, especially for a tileset like Cesium OSM Buildings that uses additive refinement. Because the biggest buildings will be in tiles covering large areas, and those areas are more likely to intersect an exclusion zone, it's going to be fairly common for the largest buildings to be mysteriously missing just cause there's an exclusion zone elsewhere.

I know you're relying on this heavily for the HoloLens stuff, though, so maybe there's a middle ground. Could we include it in the shipped plugin but behind an `EXPERIMENTAL_EXCLUSION_ZONES` #define or something perhaps?

Source/CesiumRuntime/Private/Cesium3DTileset.cpp
```diff
@@ -933,6 +955,22 @@ void ACesium3DTileset::Tick(float DeltaTime) {
...
+	UCesiumGltfPrimitiveComponent* PrimitiveComponent = static_cast<UCesiumGltfPrimitiveComponent*>(Gltf->GetChildComponent(0));
+	if (PrimitiveComponent != nullptr)
+	{
+		const UEnum* ChannelEnum = StaticEnum<ECollisionChannel>();
+		if (ChannelEnum)
+		{
+			for (int32 ChannelValue = 0; ChannelValue < ECollisionChannel::ECC_MAX; ChannelValue++)
+			{
+				ECollisionResponse response = BodyInstance.GetCollisionResponse().GetResponse(ECollisionChannel(ChannelValue));
+				PrimitiveComponent->SetCollisionResponseToChannel(ECollisionChannel(ChannelValue), response);
```
kring, line 968: Is it expected that the collision response channel will change at runtime? Or will it pretty much be configured on the tileset and left alone from then on (other than in the editor perhaps)?

In other words, I'm trying to understand if this needs to be done on Tick.

Source/CesiumRuntime/Public/Cesium3DTileset.h
```diff
@@ -49,7 +51,7 @@ class CESIUMRUNTIME_API ACesium3DTileset : public AActor,
    * The access token to use to access the Cesium ion resource.
    */
   UPROPERTY(
-      EditAnywhere,
+      EditAnywhere, BlueprintReadOnly,
```
kring, line 54: Why does this need to be `BlueprintReadOnly`? I saw your comment "Added BueprintReadOnly flag to CesiumIonToken to be able to check that tokens have been set..." but I'm not quite sure what you mean.

kring (commented):
Thanks @AlbanBERGERET-Epic! I have a few questions, see below.

Source/CesiumRuntime/Private/Cesium3DTileset.cpp
AlbanBERGERET-Epic: I understand. 
I did it like that because I did not want to do edit Cesium3DTiles, but only in the plugin. What is important for now is to have the feature ON by default in the release build. So I'm ok if you put it in #defines and the implementation changes afterwards. 

But from an user point of view, can you keep the same UX ? I mean in the future implementation, you'll probably still be able to also define rectangular clipping zones. So if the property is there and just the way to process it changes for a better implementation later on, that's fine.

Source/CesiumRuntime/Private/Cesium3DTileset.cpp
AlbanBERGERET-Epic: Well, for "standard" primitives components, the property can be changed during runtime. It's interesting for gameplay 
purposes. 

But I also understand that because of the large tileset count of, this could be a bad idea. 
I would say "choose what is best for you". You can decide that for Cesium3DTilesets, it's fixed once at beginplay and can't change. (And in my particular case, I don't need to change it at runtime...). 

What is important is that the property is applied to all newly loaded tiles. And I thought that was the best location to do it for each incoming tile. But you know your code better than me !

Source/CesiumRuntime/Public/Cesium3DTileset.h
AlbanBERGERET-Epic: I just wanted to make sure the property was accessible from a blueprint graph, but in readonly mode... 
In my code, I do a sanity check by iterating on all Cesium3DTilesets to check if the token is not Empty. If I encounter an empty one, I display a warning message in the app.

Source/CesiumRuntime/Private/Cesium3DTileset.cpp
kring: > What is important for now is to have the feature ON by default in the release build.

Does that mean that if we wrap it in a #define, that #define has to be enabled by default?

> But from an user point of view, can you keep the same UX ?

Maybe! Let me think about that. It's always a bit risky to commit to an API before it has been thoroughly considered, which is exactly why we removed the old opacity mask based exclusion system before shipping 1.0.

Source/CesiumRuntime/Public/Cesium3DTileset.h
kring: Oh ok, so if we make it read/write that would be just as good for your purposes, right?

javagl: Sorry if I misunderstood something here due to a lack of context, but I just stumbled over the statement *"iterating on all Cesium3DTilesets to check if the token is not Empty"* - On the one hand, I could simply assume that this is only intended for tilesets that are *not* loaded from local files, but on the other hand, this raises the question of why the distinction between "ion and local file" should be made ... (?)

AlbanBERGERET-Epic: @javagl, this is of course intended only for ion assets because the token is only available for this kind of assets.  It helps to manage access of remote applications. When using local files, there is no mechanism for access rights. 
My request was just to be able to do some sanity checks to make sure the developer put the right tokens in its UE level. This is really application-specific, and the current Cesium API prevented from doing that, the value was not accessible from within a BP graph.

javagl: Usually, there should not be a distinction between a "file-based asset" and an "ion-based asset" on that level. **If** we expose a certain functionality (like obtaining the token), then it will never be possible to remove that functionality in a later release. I'm always hesitant to ~"make something `public` just for convenience". There should be strong, technical reason, and one should be absolutely sure that this is the right thing to do, and will stay like that forever. But.... it's merged now, so ... no need to further dissect this.

PR #363, opened 2021-04-19
Update release-process.md

shehzan10:
Added steps for Marketplace. Will confirm as part of v1.1.0 release. Don't merge till then.


cesium-concierge:

shehzan10:
@kring This is ready for review.

The only change I'll suggest is around the zip file upload to a public link in Step 5.5. If the post-process travis job is some way off, then we can add instructions for uploading to S3, otherwise we should point to the zip-file merge travis job.

PR #371, opened 2021-04-22
Fix truncation of ECEF values to float.

kring:
This can cause round-trip errors on the order of tens of centimeters, when it should be many orders of magnitude less than that. Reported here:
https://community.cesium.com/t/solved-about-transform-location/13255/7


cesium-concierge:

nithinp7:
Thanks @kring!

PR #372, opened 2021-04-22
Move the BodyInstance property to the Collision category so it works in the Editor

### LINKED ISSUE #246: Move ACesium3DTileset::BodyInstance property to the "Collision" category
From @AlbanBERGERET-Epic...

It is sometimes necessary to set up custom collision behavior for Unreal Actors, including our Tileset Actors, in particular when working with the HoloLens 2. But it's currently not possible to do so without changes to the Cesium for Unreal plugin code. To solve this, we need to expose a new property on our Tileset Actors:

```
/** Physics scene information for this component, holds a single rigid body with multiple shapes. */
UPROPERTY(EditAnywhere, BlueprintReadOnly, Category = Collision, meta = (ShowOnlyInnerProperties, SkipUCSModifiedProperties))
FBodyInstance BodyInstance;
```

And then copy it to each GltfComponent that gets created. This will cause some new properties to appear on the Details panel:

image

The custom "TileSets" trace response in the screenshot above was created by adding it in project settings:

image

kring:
Fixes #246

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,11 @@
 # Change Log
 
+### v1.1.1 - ?
+
+##### Fixes :wrench:
+
+- Moved the `BodyInstance` property on `Cesium3DTileset` to the `Collision` category so that it can be modified in the Editor.
+
 ### v1.1.0 - 2021-04-19
 
 ##### Additions :tada:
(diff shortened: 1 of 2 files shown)


cesium-concierge:

Issue #335, opened 2021-04-13
Multiple callbacks are added due to property changes


javagl:
Whenever a property of an unreal object changes, `OnConstruction` is called. (This is not clearly said in the documentation, but anticipated according to https://answers.unrealengine.com/questions/165787/view.html )

For us, this means that the implementation of `OnConstruction` should be idempotent. But at least in the tileset, the implementation adds a callback/lambda at https://github.com/CesiumGS/cesium-unreal/blob/8b512a22f37b4bb611b3044894323f10a1d0f7d9/Source/CesiumRuntime/Private/Cesium3DTileset.cpp#L283 , resulting in this lambda being present many times:

![CesiumUnreal OnConstruction Issue](https://user-images.githubusercontent.com/5597569/114550213-3c759900-9c62-11eb-9abc-9e1ddc9136d1.png)

The implementations of `OnConstruction` should be reviewed and fixed (and it won't harm to add another comment to these functions, `/* Called on construction AND WHENEVER A PROPERTY CHANGES!!! */` to prevent similar errors in the future).

javagl:
I started addressing this. One could pragmatically introduce a flag, and then ask `if (!wasAdded) { add(); wasAdded=true; }`, but I'd 1. *strongly* prefer to move the functionality out of `OnConstruction`, and 2. tried to make the other `OnConstruction` implementations idempotent as well, which triggered some reading of https://docs.unrealengine.com/en-US/ProgrammingAndScripting/ProgrammingWithCPP/UnrealArchitecture/Actors/ActorLifecycle/index.html , coming to the conclusion that "this doesn't say anything - one has to try it out", and I eventually stumbled over the fact that what is currently done in `OnConstruction` of `GlobeAwareDefaultPawn` apparently ~"cannot be done anywhere else, without causing a wrong placement when starting". This might be related to https://github.com/CesiumGS/cesium-unreal/issues/362 , but I'll have to investigate that further.

PR #382, opened 2021-04-26
Add focussing callback only once

javagl:
Fixes https://github.com/CesiumGS/cesium-unreal/issues/335

Originally, the callback that causes the viewports to focus on a double-clicked tileset was added in `OnConstruction`. But this method is called on each property change. This PR changes this so that the callback is added in `PostInitProperties`, which seems to be called only once, after the constructor (and thus, after the object has been "initialized" to some extent...). 

The logging for the `LoadTileset` method has been fixed so that it only prints the log message "Loading tileset ..." when ... it's actually loading the tileset (and not each time when the method is called, ... that should be the same, but that's another issue).


cesium-concierge:

kring:
Thanks @javagl!

PR #388, opened 2021-04-27
Get UE for macOS from AWS S3 instead of Google GCS.

kring:
macOS builds don't run on Google Cloud like Windows and Linux builds do, so downloading it from GCS isn't free. And GCS costs more than S3, so let's use S3.


cesium-concierge:

PR #395, opened 2021-04-29
Fixing GeoreferenceComponent properties update 

Nodrev:
When using 'MoveToLongitudeLatitudeHeight' function, only ECEF properties are updated.
When using 'InaccurateMoveToLongitudeLatitudeHeight', only "geographics" (lat/long/height) properties are updated.
Fixing this by updating both types of properties regardless of the called function.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,6 +9,7 @@
 
 ##### Fixes :wrench:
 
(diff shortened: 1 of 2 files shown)


cesium-concierge:
Thanks for the pull request @Nodrev!

nithinp7:
@Nodrev Thanks for the fix! I updated the CHANGES.md and I confirmed this passes CI on another branch. Merging this

Issue #321, opened 2021-04-09
Improve Georeferenceable interface and make it blueprint-friendly


nithinp7:
Currently the `ICesiumGeoreferenceable` interface does two somewhat unrelated things:
- One, it allows implementations to be notified when the georeference has been updated. 
- Two, in the case that the georeference origin placement mode is set to "bounding box" (average center between all sub-boxes), the interface allows implementations to return their own bounding volume to contribute to the origin placement. This is so far only used by tilesets and is completely ignored by all other "georeferenceable" objects. 
 
Maybe we should have two different interfaces for these two functionalities. Something like `GeoreferenceListener` that is sent notifications pre- and post- georeference change, and something like `WithBoundingVolume` (hahah not actually sure what to call this one actually). 

The main benefit that would come from this is that it would make the `GeoreferenceListener` functionality much easier to use. This would allow custom user actors/components (even ones in blueprint) to subscribe and be notified of when a georeference change is about to take place. This would allow for disabling physics, creating a last-ditch georeference that can be restored after the change, etc.

kring:
This interface no longer exists.

Issue #330, opened 2021-04-12
Georeference refactoring


javagl:
There are [several issues related to `CesiumGeoreference`](https://github.com/search?q=CesiumGeoreference+OR+Georeference+repo%3ACesiumGS%2Fcesium-unreal&type=issues), about usability, glitches, or crashes, and general ideas for improvements. Considering these issues, and a technical talk that we had today after the standup, the consensus seems to be that the handling of the `CesiumGeoreference` (or GeoRef, for short) should be reviewed and probably refactored. 

---

Summary:

<sup>Disclaimer: The following is an attempt of a *short* summary of some points of my "mental model" that I find relevant. This may be incomplete or imprecise. Feel free to add/edit clarifictions or corrections. I only started to look at this topic in more detail.</sup>

- Actors like `Cesium3DTileset` and `GlobeAwareDefaultPawn` use a GeoRef for coordinate conversions between "Unreal World", Earth-Centered-Earth-Fixed (ECEF), East-North-Up (ENU), and Latitude/Longitude/Height. The actors store a reference to a GeoRef, and use this instance for these coordinate conversions. 
- The GeoRef may be assigned to the actors in the UI. In some cases, this happens automatically - for example, when adding a new tileset to an empty level. More generally: Under some conditions, a "default" GeoRef instance is created. Usually, all actors in one level will use the same GeoRef (but not necessarily)
- The GeoRef is "observable", and can inform "Listeners" about changes. This is done via the `ICesiumGeoreferenceable` interface. This interface is implemented by the actors that hold the GeoRef. 

Some "conceptual" issues that may already be mentioned elsewhere:

- Deleting a GeoRef will break references and bring the actors into an invalid state. Even when handling that, it is not clear how the actors should behave when they don't have a valid GeoRef.
- Resetting the GeoRef of a `GlobeAwareDefaultPawn` to `None` does not work, because the property change causes `OnConstruction` to be called, where a new default GeoRef is created if the current one is `null`
- Default GeoRef instances that are created are not always added in the World Outliner
- `ACesium3DTileset::PlayMovieSequencer` always re-creates a (new, local) GeoRef, and does not use the one from the tileset
- The actors have a reference to the GeoRef, and the GeoRef has references to the actors (because they are "observers"). This is a cycle in the object graph, and might cause issues with garbage collection, or make it hard to achieve a "consistent" state in general.

---

Notes from the discussion: 

[...]

PR #398, opened 2021-04-30
Break apart georeferenceable interface

nithinp7:
Now we have a very streamlined `ICesiumGeoreferenceListener` which subscribes for notifications of changes in the world georeference. We also now have a `ICesiumBoundingVolumeProvider` that can contribute its bounding volume to origin placement calculations (when the origin placement mode is "bounding volume origin").

Fixes: https://github.com/CesiumGS/cesium-unreal/issues/321


cesium-concierge:

Source/CesiumRuntime/Public/CesiumGeoreferenceListener.h
```diff
@@ -0,0 +1,28 @@
...
+ *
+ */
+class CESIUMRUNTIME_API ICesiumGeoreferenceListener {
+  GENERATED_BODY()
+
+public:
+  /**
+   * Notifies georeferenced objects that the underlying georeference transforms
+   * have been updated. The new transforms can be fetched from the
+   * `CesiumWorldGeoreference` actor.
```
javagl, line 25: This *might* be `CesiumGeoreference`, but ... this wouldn't necessarily tell the user where to obtain the transforms. Some options for "tansporting" this information range from

1, You have to store a reference to the GeoRef, and fetch the info from that reference on your own
2. The method could receive the `GeoRef*` that was updated (i.e. the one that triggered that call)
3. The method could receive ~"something like the updated transform"

For 1., the receiver of this call would not know which GeoRef was updated. For 3. it's not clear *which* information that is (it's more than a 4x4 matrix, for sure). In other contexts, I'd pragmatically go for 2. (and consider something like a `GeoRefChangedEvent` instead, if more information would have to be transported through that call).

But for 2., the caller could once again *store* this reference that he receives there, raising further questions about its validity and such. Depending on the other decisions, the method could also receive "an identifier" for the GeoRef, like `NotifyGeoreferenceUpdated("THE_DEFAULT_ONE")` vs. `NotifyGeoreferenceUpdated("MY_SPECIAL_ONE")`, where this "identifier" is used to access some sort of global/singleton-like instance.

(tl;dr: This has to be sorted out, and aligned with the (coarse-grained) GeoRef refactoring....)

xuelongmu:
I haven't yet had a chance to take a closer look at the code, but just wanted to bring up some of our discussions from #330, especially regarding the [delegates](https://github.com/CesiumGS/cesium-unreal/issues/330#issuecomment-820074706) for the `NotifyGeoreferenceUpdated` - do you think it would be a good idea to switch over to delegates instead of a C++ interface?

Some advantages would be being able to register a callback from blueprints instead of implementing the interface in a C++ class, and hiding the actual registration in a static function like what @javagl suggested. One stretch goal of this would be to handle multiple georeferences as well, through the static function library, using a string identifier perhaps.

baothientran (commented):
This looks too me @nithinp7, I have some comments regarding about the listener interface

kring:
I think we can close this because it is included in #459 and we're unlikely to merge it separately from the larger georeference refactoring. Please reopen if I'm mistaken.

Issue #187, opened 2021-03-03
Constrain user input values for properties


javagl:
I noticed that it is possible to, for example, set a *negative* value for the maximum screen space error. According to https://docs.unrealengine.com/en-US/ProgrammingAndScripting/GameplayArchitecture/Properties/Specifiers/index.html , there are property metadata specifiers like 

> `ClampMin="N"` | Used for float and integer properties. Specifies the minimum value N that may be entered for the property.

Such constraints could/should be applied to the properties where applicable.

For the speficic case of the SSE, a negative value does not severely break or crash anything, but causes the console to be filled with error messages:
```
LogCesiumNative: Error: [2021-03-03 23:45:26.437] [error] [RasterOverlayTileProvider.cpp:476] Failed to load image:
- Image response is empty.
```
which should be avoided if possible. In other cases, an invalid value could cause a crash. So the possible user inputs should be reviewed and checked in general. 

---

For now, this does **ONLY** refer to the `UPROPERTY` values, and we could constrain the SSE to `ClampMin=0.1` (assuming that it shouldn't be 0.0...?) or so. 

But we should also look at where/how certain values could be passed through to `cesium-native` otherwise. It might be worth to add such clamping in `cesium-native`, as in

    options.maximumScreenSpaceError = std::max(0.1, inputOptions.maximumScreenSpaceError);

and add some comment, like 

    struct CESIUM3DTILES_API TilesetOptions {
        /**
         * 
         * ... 
         * This value will be clamped to be at least 0.1
         */
        double maximumScreenSpaceError = 16.0;

---

The first step could be to come up with a list of values that need constraints, and what sensible constraints could be.

PR #410, opened 2021-05-04
Clamping user input to minimum valid values

javagl:
Fixes https://github.com/CesiumGS/cesium-unreal/issues/187

It only adds the `ClampMin=0` metadata where appropriate. I considered clamping latitude to -90...+90 and longitude to -180...+180, but wasn't sure whether this would be desired. If so, drop me a note, then I'll add that in this PR.


cesium-concierge:

javagl:
The latitude and logitude values are now clamped to [-90, 90] and [-180, 180], respectively.

I tried out the "possibly breaking" workflow of...

- creating a level (without the update) and setting a GeoRef to lat = 100
- doing the update
- loading the level

and it still showed the saved value (lat=100 in this case). The clamping only takes place when *editing* the value afterwards. 

(One could argue about whether this could/should still be considered as a "breaking change" ....)

xuelongmu (approved):
Looks good! Just specified 0.0 for some of the double properties, will merge once build completes.

PR #414, opened 2021-05-06
Fix an issue where foliage painting doesn't work for tileset actor

### LINKED ISSUE #411: Foliage painting not working in Cesium for Unreal 1.2.0
Foliage painting does not appear to work with Cesium World Terrain as of 1.2.0 - it may also not work in 1.1.0. Foliage placement in general is confirmed working, you can paint foliage on static meshes or landscapes, but it seems that CWT no longer registers as a surface that can have foliage.

baothientran:
Fixes #411 

Foliage editor shoot ray into the world to detect brush position and place instances accordingly. However, it only detects collision for ECC_WorldStatic. The code for the raycast can be found here https://github.com/EpicGames/UnrealEngine/blob/c3caf7b6bf12ae4c8e09b606f10a09776b4d1f38/Engine/Source/Editor/FoliageEdit/Private/FoliageEdMode.cpp#L1040

and this one too

https://github.com/EpicGames/UnrealEngine/blob/c3caf7b6bf12ae4c8e09b606f10a09776b4d1f38/Engine/Source/Runtime/Foliage/Private/InstancedFoliage.cpp#L4313

When we change bUseDefaultCollision to be false, all our components have ECC_WorldDynamic, so it stops working with foliage


cesium-concierge:

kring:
Thanks @baothientran!

PR #415, opened 2021-05-06
Optimized texture coordinate selection 

nithinp7:
I was doing some unreal material/shader research today and stumbled on the "custom" node which lets you type in a snippet of hlsl code. I was able to make this quick optimization to the texture coordinates (it's an optimization because the shader complexity view shows a very _slightly_ nicer shade of green, super scientific I know 🙂).

<em>Before</em>

<em>After</em>

But now that we know how to add custom snippets of hlsl maybe there's an even better way that I couldn't find. This is how the `SelectTextureCoordinates` material looks now:

And the hlsl snippet that I put in the "custom" node:
`float2 texCoords[8] = {TexCoord0, TexCoord1, TexCoord2, TexCoord3, TexCoord4, TexCoord5, TexCoord6, TexCoord7};`
`return texCoords[TexCoordIndex];`

Nothing was changed in the other materials, but maybe their binary content appears changed after updating the `SelectTextureCoordinates` function which they all used.


cesium-concierge:

nithinp7:
@kring @argallegos More of a potentially useful curiosity than a pertinent fix, but feel free to have a quick look if either of you have a moment!

kring:
@nithinp7 I stumbled upon the custom expression nodes awhile back, but was scared away from them (perhaps irrationally) by this note in the docs:

> Using the custom node prevents constant folding and may use significantly more instructions than an equivalent version done with built in nodes! Constant folding is an optimization that UE4 employs under the hood to reduce shader instruction count when necessary. For example, an expression chain of Time >Sin >Mul by parameter > Add to something can and will be collapsed by UE4 into a single instruction, the final add. This is possible because all of the inputs of that expression (Time, parameter) are constant for the whole draw call, they do not change per-pixel. UE4 cannot collapse anything in a custom node, which can produce less efficient shaders than an equivalent version made out of existing nodes. As a result, it is best to only use the custom node when it gives you access to functionality not possible with the existing nodes.

https://docs.unrealengine.com/en-US/RenderingAndGraphics/Materials/ExpressionReference/Custom/index.html

It seems like it might be very worthwhile in this case though!

argallegos:
I've poked around a little and it does look like this reduces the shader instructions by about 30 on average, which is solid. It's worth noting though that for the gltf material with water, it actually seems to increase shader instructions by 18. I imagine this likely has to do with how many of the node are present in the material. 

Regardless, I think that's ok. We can keep trying to optimize the water material further, but this is a good change that makes our materials cleaner.

xuelongmu:
I'll preface my comment by admitting I'm not familiar at all with the custom shader nodes, but it may be worth profiling with a packaged game or in Standalone as well as PIE. The shader optimizations @kring is talking about might only be performed during that packaging stage. It does seem that this change in particular is a great use case for the custom expression nodes.

nithinp7:
From what I read and assumed: I think constant folding is only applicable for uniforms like `time` that would be the same across the entire draw call. In that case, a constant expression that is only a function of uniforms can be computed just once before the draw call and the result can be used across the draw call. Identically you could compute only-uniform-derived calculations on the CPU and upload each as it's own uniform so it's reused across the draw call, I'm assuming that's the type of thing that's happening under the hood. 

In this case, there's no redundant uniform-derived data. We're only using texture coordinates which are different across the draw call.

PR #419, opened 2021-05-07
Use updated UE image for Windows.

kring:
It includes support for Android cross compiling.

### DIFF
--- a/travis/travis-get-ue.sh
+++ b/travis/travis-get-ue.sh
@@ -1,9 +1,9 @@
 if [[ $TRAVIS_OS_NAME == "windows" && ! -d "/c/Program Files/Epic Games/UE_4.26" ]]
 then
-    AWS_ACCESS_KEY_ID=${GOOGLE_ACCESS_KEY_ID} AWS_SECRET_ACCESS_KEY=${GOOGLE_SECRET_ACCESS_KEY} aws s3 --endpoint-url https://storage.googleapis.com cp s3://cesium-unreal-engine/2021-03-16/UE_4.26.zip .
+    AWS_ACCESS_KEY_ID=${GOOGLE_ACCESS_KEY_ID} AWS_SECRET_ACCESS_KEY=${GOOGLE_SECRET_ACCESS_KEY} aws s3 --endpoint-url https://storage.googleapis.com cp s3://cesium-unreal-engine/windows-2021-05-07/UE_4.26.zip .
     7z x UE_4.26.zip "-oC:\Program Files\Epic Games"
     rm UE_4.26.zip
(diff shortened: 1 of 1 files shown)


kring:
Still no luck figuring out why the Windows build is broken in this branch. It builds just fine locally when I use the same procedure Travis is using, as well the same copy of Unreal Engine. But of course my system is different from the Travis VM in a variety of ways.

Things I've tried so far, none of which worked:
* Including the UE engine source in the engine ZIP.
* Modifying the registry on the Travis VM to support long path names.
* Invoke RunUAT.bat with `-Verbose`. It logs heaps of stuff, but nothing at the critical moment where it breaks.
* Install the .NET 4.6.2 targeting pack (listed in the UE 4.26 changelog as being required, and Travis possibly maybe comes with a slightly older or different one or something?)

kring:
Aah here's a hint, finally! The build is failing on Travis while compiling some C# code into `C:\Program Files\Epic Games\UE_4.26\Engine\Intermediate\Build\BuildRules\UE4Rules.dll`. On my local system, it doesn't try to build this at all. By dumping the UBT log on Travis, there's a hint about why it's building it on Travis:

```
DynamicCompilation.RequiresCompilation: Compiling C:\Program Files\Epic Games\UE_4.26\Engine\Intermediate\Build\BuildRules\UE4Rules.dll: Removed source file (D:\Build\++UE4\Sync\Engine\Source\Runtime\Advertising\Advertising\Advertising.Build.cs)
```

That file doesn't exist on my local system either, so not sure why it's flagged as "removed" on Travis but not locally.

kring:
Argh! It's always something stupid and obvious, isn't it? The unzip of Unreal Engine is failing because the VM is out of disk space. Only that somehow didn't get flagged as an error by Travis, and it was hidden in a collapsed section of the Travis log. So yeah the build is failing on Travis because files are literally missing there. 😠

kring:
Shehzan has reached out to TravisCI to see if we can get more disk space. If not, here are some options I can think of:
1. Manually remove things from the ZIP file that we don't need. This will be a bit time consuming and error prone, but we can likely find enough to delete so that it fits on the Travis VM.
2. Switch to GitHub Actions and use a self-managed runner instance.
3. Unzip UE to a compressed folder, so it takes less space. See https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/compact
4. Stream the UE archive straight from the aws downloader through the decompressor, rather than storing it on disk first, saving 8-9GB. This probably isn't possible with ZIP, but we could use a different format.

Actually 3 is probably so easy I might try it right now.

kring:
That fixed the disk space problem and the Windows build is succeeding now. Travis is still failing because it's trying to build for Linux and the necessary SDK isn't installed. I'm not completely sure why it's trying to build for Linux.

kring:
I removed Linux support from the UE zip (we don't need it) and everything appears to be working. I also removed the Engine source code, because that's not needed either. I just need to clean up all the cruft from various previous attempts at fixing it and this will be ready to merge.

kring:
This should be good to go now. @shehzan10 do you want to review and merge?

shehzan10:
Thanks @kring!

Issue #242, opened 2021-03-18
Georeference and Credit actors doesn't show up when adding CWT in new level


baothientran:
- Create blank level, add Cesium world terrain from quick add -> CWT, CesiumGeoreference, and CesiumCreditSystem are created in the World Outliner.
- Delete all 3 of them, and re-add one of the quick-add assets, only CWT gets added, not geoferences or credit system

PR #424, opened 2021-05-09
Fix non-appearing GeoRef and CreditSystem after deletion

javagl:
This could fix https://github.com/CesiumGS/cesium-unreal/issues/242 

**But** this has to be aligned with https://github.com/CesiumGS/cesium-unreal/issues/330 ! The refactoring might already resolve the underlying issue, **if** the result of the refactoring is that there is one, global, static, undeletable GeoRef.

---

The reason for this issue was ... simple, in hindsight: When selecting actors in the World Outliner and hitting "Delete", then they are deleted. (That makes sense so far...). But they are only **marked** for deletion, and will only be cleaned up (and removed from the level) during a garbage collection. This means that the `GetDefaultForActor` function of the `ACesiumGeoreference` and `ACesiumCreditSystem` classes returned the respective "default" instance, but this instance was no longer valid (as of `IsValid`), due to the pending kill.

The solution is... \*ehrm\*.... I wonder whether there is a simpler, better solution. Right now, the `FindObject` call has been changed to a `findValidDefault...` function, that searches for an instance (GeoRef or CreditSystem), with the name *prefix* that indicates that it is a default instance, and that is *not* pending kill. 

A cleaner solution might be possible if we better knew 1. when the GC happens and the instances are killed, and 2. when the `GetDefault` function is called (they are called frequently, in a ~"just call it, to be sure"-manner). 

---

An aside: Although the *reason* for the bug was trivial, I spent a considerable amount of time with figuring out why the "Default" instances did **not** appear in the World Outliner. This involved pain-in-the-back-debug-stepping through this function several times:
```
    bool SSceneOutliner::IsActorDisplayable( const AActor* Actor ) const
    {
        return    !SharedData->bOnlyShowFolders &&                                                 // Don't show actors if we're only showing folders
                Actor->IsEditable() &&                                                            // Only show actors that are allowed to be selected and drawn in editor
                Actor->IsListedInSceneOutliner() &&
                ( (SharedData->bRepresentingPlayWorld || !Actor->HasAnyFlags(RF_Transient)) ||
                  (SharedData->bShowTransient && Actor->HasAnyFlags(RF_Transient)) ||            // Don't show transient actors in non-play worlds
                  (Actor->ActorHasTag(SequencerActorTag)) ) &&        
                !Actor->IsTemplate() &&                                                            // Should never happen, but we never want CDOs displayed
                !FActorEditorUtils::IsABuilderBrush(Actor) &&                                    // Don't show the builder brush
                !Actor->IsA( AWorldSettings::StaticClass() ) &&                                    // Don't show the WorldSettings actor, even though it is technically editable
                !Actor->IsPendingKill() &&                                                        // We don't want to show actors that are about to go away
                FLevelUtils::IsLevelVisible( Actor->GetLevel() );                                // Only show Actors whose level is visible
    }
```

I wonder where there any sensible reason to *not* write it as
```
    bool SSceneOutliner::IsActorDisplayable( const AActor* Actor ) const {
        // Don't show actors if we're only showing folders
        if (SharedData->bOnlyShowFolders) {
            return false;
        }
        
        // Only show actors that are allowed to be selected and drawn in editor
        if (!Actor->IsEditable() || !Actor->IsListedInSceneOutliner()) {
            return false;
        }
        ...
        if (Actor->IsTemplate())
            return false; // Should never happen, but we never want CDOs displayed
        }
        ...
        if (Actor->IsPendingKill()) {
            return false; // We don't want to show actors that are about to go away
        }
        ...
        return true;
    }
```
One could argue that "multiple `return` statements are an anti-pattern", but if the alternative is to write 13 booleans, including complex function calls, into a single `return` statement, then thats a non-argument.

On the upside: Our [`Tile::isRenderable`](https://github.com/CesiumGS/cesium-native/blob/main/Cesium3DTiles/src/Tile.cpp#L94) doesn't look so bad compared to that :-)


cesium-concierge:

kring:
@javagl given that the georeference refactoring may be a little while yet, this looks like a safe and good change that we should ship sooner rather later... maybe even this week's release? If you don't have any objections to that, please mark it ready for review.

javagl:
As mentioned in the first message: The the fix is ~"relatively simple". I think that the solution is a bit quirky, but _hope_ that it does not have undesired side-effects. So one could argue to make it part of the next release, to fix the immediate bug (and make sure that, after the refactoring, the behavior is still fixed - i.e. that there is no regression. We already kept an eye on that, but will have to review/re-check it thoroughly before merging). 

I don't have any objections for merging it, so I'll mark it as ready for review, **BUT** would appreciate if others would indeed *review* it (in an attempt to dodge at least some of the _responsibility_ if it does have undesired side-effects that I didn't consider... ).

kring:
Thanks @javagl. It seems to work well, and the code looks good to me. Merging.

Issue #175, opened 2021-03-01
ion UI: Reduce aliasing in Cesium for Unreal logo


kring:
Either by changing a texture filtering setting or by scaling it down ahead of time.

shehzan10:
@kring Is this done as part of 1.0?

kring:
No the logo on the Cesium panel is still big and scaled down so it looks a little aliased.

kring:
Fixed in #482

PR #427, opened 2021-05-10
Reorganize and rename plugin contents

argallegos:
This PR aims to clean up the plugin content folder and organize the materials a little better.

### Summary of changes:

- Materials have been moved to a new folder named Materials, with Material Functions and Water Material Functions in subfolders. 
- Every material and material function has been renamed to begin with Cesium. Some things have been renamed significantly:
   - GltfMaterialOpacityMask --> M_CesiumDefaultMasked
   - GltfMaterialWithOverlays --> M_CesiumOverlay
   - GltfMaterialWithOverlaysAndWater --> M_CesiumOverlayWater
- Per @nithinp7's suggestion, the previous body of M_CesiumOverlay/GltfMaterialWithOverlays has been collapsed into a new material function called "CesiumGltfWithOverlays". This functionality allows for more intuitive material editing and simplifies materials. 
- Every old material and material function has been removed.
- A water material and opacity mask material have been exposed in the details panel along with the base material for users to edit.
- Material parameters have been categorized into "Cesium" or "Cesium DoNotEdit" based on whether or not editing them will actually do anything, for users who will be creating and modifying material instances. 
- Every material function has been given a description and added to the "Cesium" library category.
- The default textures have been moved to a new folder called "Textures". 
- The original water material has been replaced with an updated water material - the original is still there, but the new one is less complex and should perform significantly better, and hopefully not crash macs.


argallegos:
Water material has been updated - by default, the material no longer requires things from the water plugin. It does still make use of some of the functions from that plugin that have been copied into the Cesium plugin, though. The old water is still there, as M_CesiumOverlayComplexWater. It can be switched back by changing the relevant code or by copying and pasting the complex node graph into the simple one.

The simple shader is also much more performant than the old one - 326 instructions compared to 981. 

Here's what the graph looks like. 
image

Most notably, the shader now only uses the Default Lit shading model - so the depth, scattering, etc is gone. I think that's an okay sacrifice, since I'm guessing that most people won't need that anyways, and if they do they can always switch it back. 

The other notable change is that it doesn't blend the imagery layer with a base color at all anymore. I think it's actually better that way, since it makes the blocky edges of the watermask much less visible. 

I looked into getting larger wave detail when far away, but still haven't really figured it out. The issue is that when the camera is at the heights previously discussed, the normal map as a whole seems to be faded out by the atmosphere. It should be possible to get it working, but I figured there was more value in getting this out sooner rather than later - we can always tweak it down the line. 

Here's some pictures:
image
image
image
image

Not blending the color does result in some imagery seams like this:
image
But the water color actually looks better along the coast now up close, which is a benefit. 
Plus, it's cool to be able to see the seafloor detail. 
image

This is almost ready to review (I'm going to do the reorganization of the other plugin contents in a different branch since this ended up being so involved). I'd still like to see if I can set up an easy switch between the simple and complex water shaders, and  maybe add a spot on tilesets to specify a watermask material as well as a normal material. Also might sneak https://github.com/CesiumGS/cesium-unreal/issues/175 into here.

argallegos:
Cesium Tilesets now accept two additional materials in the Rendering section, a Water material and an Opacity Mask material. For tilesets with watermasks available, it can now be set in-editor here. 

image

Per @nithinp7 's suggestion, eventually it would be fantastic to auto-set these material slots to the default materials, and always draw from whatever material is placed in this slot.

argallegos:
I think this is done. I tested it on several old projects after deleting the old materials, and as far as I can tell it won't break anything for users. 

@kring , would you check that the descriptions and comments I've added to the new materials and material functions are correct? 

@nithinp7 can you take a look at the water material in particular and confirm that it's working right with the watermask/not causing anything unexpected?

kring:
The descriptions and comments look good to me @argallegos!

nithinp7:
@argallegos This looks great to me, good work on the water shader! We can look into macro variation for bigger swelling waves when we have time to come back to it, but this water material and the more expensive old one are great for now. Can you just add to the Changes.md? Then I will merge this.

argallegos:
@nithinp7 I updated the changelog - I didn't want to get too granular, but I can add more detail if you think I should. I created a small merge conflict with CHANGES.md, I can resolve that and merge everything so you don't have to if you think the updated CHANGES.md looks good.

nithinp7:
Thanks @argallegos!

PR #430, opened 2021-05-13
Blueprint editable Cesium3DTileset properties

### LINKED ISSUE #342: Allow Cesium3DTileset properties to be get/set in Blueprints
Can we add BlueprintReadWrite to url and other properties?

xuelongmu:
Closes #342 

When settings are edited from blueprints, `OnConstruction` does not get automatically called. So for properties that require a `LoadTileset()`, I created specific setter functions. For other properties that are updated on Tick, I simply added a `BlueprintReadWrite` flag.

One potential issue is that Blueprints doesn't support uint32 type, only int32 and float. I changed the `IonAssetID` property to int32 and cast it to uint32 when passed into Tileset on line 701. Would we anticipate Ion Asset ID's exceeding 2,147,483,647 (max for int32, max for uint32 is double of this)?

I also added a UPROPERTY metadata specifier to the `IonAssetID` to ensure that it's set to a non-negative value (as well as a few other fields)


cesium-concierge:

xuelongmu:
Here is a blueprint PlayerController class I created to test these functions at runtime using key presses:

https://blueprintue.com/blueprint/t_j2lsul/

(first time trying to use this Blueprint pastebin site, let me know if it works)

Source/CesiumRuntime/Public/Cesium3DTileset.h
```diff
@@ -69,9 +74,12 @@ class CESIUMRUNTIME_API ACesium3DTileset : public AActor,
    */
   UPROPERTY(
       EditAnywhere,
+      BlueprintReadOnly,
```
kring, line 77: Rather than making this read-only and adding a separate Set function, could we use the `BlueprintSetter` specifier?
https://docs.unrealengine.com/en-US/ProgrammingAndScripting/GameplayArchitecture/Properties/Specifiers/index.html
(I've never tried, so the answer could very well be no)

Source/CesiumRuntime/Private/Cesium3DTileset.cpp
```diff
@@ -887,7 +912,7 @@ void ACesium3DTileset::Tick(float DeltaTime) {
   }
 
   Cesium3DTiles::TilesetOptions& options = this->_pTileset->getOptions();
-  options.maximumScreenSpaceError = this->MaximumScreenSpaceError;
+  options.maximumScreenSpaceError = (double)this->MaximumScreenSpaceError;
```
kring, line 915: Is the explicit cast to double needed? If so, it'd be better to use a static_cast instead of a C-style cast.

kring (commented):
Looks good @xuelongmu, just a couple small comments.

xuelongmu:
@kring Apparently BlueprintGetter / BlueprintSetter can be used in conjunction with private C++ UPROPERTY variables. So the URL, TilesetSource, and IonAssetID can be moved to private.

https://answers.unrealengine.com/questions/939907/view.html

Source/CesiumRuntime/Private/Cesium3DTileset.cpp
```diff
@@ -1096,6 +1125,26 @@ void ACesium3DTileset::Serialize(FArchive& Ar) {
   }
 }
 
+#if WITH_EDITOR
+void ACesium3DTileset::PostEditChangeProperty(
+    FPropertyChangedEvent& PropertyChangedEvent) {
+  const FName PropName = (PropertyChangedEvent.Property)
+                             ? PropertyChangedEvent.Property->GetFName()
+                             : NAME_None;
+  if (PropName == GET_MEMBER_NAME_CHECKED(ACesium3DTileset, TilesetSource) ||
```
kring, line 1134: Is this necessary? Because when the property is set from Blueprints, it'll go through the setter anyway, right? Which marks the tileset dirty. Or have I missed something?

kring (commented):
Looks good @xuelongmu! A few comments below.

I dont have any backward compatibility concerns. As we discussed on a call the other day, the properties are no longer accessible from C++, but they wouldn't have worked correctly before anyway, so it's not a concern. Can you please add a "breaking" section to CHANGES.md and mention it though?

Source/CesiumRuntime/Private/Cesium3DTileset.cpp
xuelongmu: The PostEditChangeProperty method override ensures that editor-time (GUI) changes to these properties will mark the dirty flag. I believe the setters won't get called in that scenario, and the OnConstruction method will call LoadTileset() but the dirty flag will still have been set to false from the previous load.

Source/CesiumRuntime/Private/Cesium3DTileset.cpp
```diff
@@ -594,53 +651,14 @@ void ACesium3DTileset::LoadTileset() {
...
-
-    bool opacityMaterialChanged =
-        this->OpacityMaskMaterial != this->_lastOpacityMaterial;
-    this->_lastOpacityMaterial = this->OpacityMaskMaterial;
-
-    if (tilesetSourceChanged || waterMaskEnabledChanged || materialChanged ||
-        waterMaterialChanged || opacityMaterialChanged) {
+    if (_tilesetIsDirty) {
       this->DestroyTileset();
+      _tilesetIsDirty = false;
```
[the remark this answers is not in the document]
xuelongmu: Great catch, I updated the code to remove the check.

xuelongmu:
@kring Updated the changelog 👍

Source/CesiumRuntime/Private/Cesium3DTileset.cpp
kring: Oh ok, for some reason I expected the editor would go through the blueprint getters/setters too.

PR #435, opened 2021-05-14
Fix zooming limit only

javagl:
This is the core of what was originally done in https://github.com/CesiumGS/cesium-unreal/pull/409 : 

It only makes sure that the rebased world origin stays in the integer range.

(EDIT: Fixes https://github.com/CesiumGS/cesium-unreal/issues/284 )

PR #409, opened 2021-05-04
Fix zooming limit

javagl:
Fixes https://github.com/CesiumGS/cesium-unreal/issues/284 ...

... as far as reasonably possible. I noticed that the controls are still wonky when moving too far and e.g. rotating with WASD. But this fix until now avoids the "flickering" of the position, by clamping the world origin (during rebase operations) to the valid 32bit integer range.

PR #436, opened 2021-05-14
Georeference tick function cleanup 

javagl:
(To be merged **after** https://github.com/CesiumGS/cesium-unreal/pull/435 ) 

This was originally done in https://github.com/CesiumGS/cesium-unreal/pull/409 , but after some discussion during the standup, has been separated from the actual zooming limit fix.

This breaks the ~200-line `Tick` implementation of the `CesiumGeoreference` into several blocks:

- The code to show the sublevel load radii (in editor mode) is in `_showSubLevelLoadRadii`
- The "edit origin in viewport" functionality (in editor mode) is in `_handleViewportOriginEditing`
- The origin rebasing and sublevel loading ... are split

Based on a short discussion, the code for handling the sublevel loading and the origin rebasing had somehow been mixed. I hope that I understood the purpose of both parts correctly, and tried to separate them:

- In `_updateSublevelState`, it checks whether the camera is in a sublevel. The sublevels that contain the camera are loaded. The other ones are unloaded. (That's also said in the doxygen comments now...)
- In `_performOriginRebasing`, the possible conditions for actually peforming the origin rebasing are checked, and this is done only when "in-game", when there is a `WorldOriginCamera`, and when origin rebasing is enabled via `KeepWorldOriginNearCamera`
  - When the camera is in a sublevel, but `OriginRebaseInsideSublevels` is **false**, then the origin is set to (0,0,0)
  - Otherwise, if the distance to the origin is too large, the origin is set to the camera view target location (clamped to the int range)

@nithinp7 If there's anything wrong on this (non-implementation) level, just drop me a note.


cesium-concierge:

Source/CesiumRuntime/Private/CesiumGeoreference.cpp
```diff
@@ -378,181 +378,225 @@ void ACesiumGeoreference::PostEditChangeProperty(FPropertyChangedEvent& event) {
...
+  if (!this->KeepWorldOriginNearCamera) {
+    return;
+  }
+  if (this->_insideSublevel && !this->OriginRebaseInsideSublevels) {
+    // If we are not going to continue origin rebasing inside the
+    // sublevel, just set the origin back to zero since the sublevel
+    // will be centered around zero anyways.
+    this->GetWorld()->SetNewWorldOrigin(FIntVector(0, 0, 0));
+    return;
+  }
```
nithinp7, line 567: In this case it will keep redundantly setting the world origin to (0, 0, 0) every Tick. I don't think that's necessarily a problem but it will trigger a bunch of delegates needlessly when it's not supposed to be origin rebasing. I know this is probably the same as the logic beforehand, but we might as well fix it up. I think we should only set origin to zero here if it is not already zero, so maybe we can add a check for that.

CHANGES.md
```diff
@@ -9,6 +9,7 @@
 - Make origin rebasing boolean properties in `CesiumGeoreference` and `CesiumGeoreferenceComponent` blueprint editable.
 - Improvements to dynamic camera, created altitude curves for FlyTo behavior.
 - Constrained the values for `UPROPERTY` user inputs to be in valid ranges.
+- Fixed a bug that caused rendering- and navigation problems when zooming too far away from the globe when origin rebasing is enabled.
```
nithinp7, line 12: I think this is the change description for the origin rebase integer overflow fix, can you add your changes from this PR?

nithinp7 (commented):
This makes things a lot cleaner, thanks @javagl! I had just two minor comments below and a merge conflict needs to be resolved as well. Otherwise this will be ready to merge.

javagl:
@nithinp7 The change in `CHANGES` was due to this PR building on the other one (i.e. there is no longer a change after merging with `main`). Regarding *this* PR: I think that adding an entry like: *"Refactored a bit, but hopefully didn't change anything"* is not worth being put into the changelog.

(And maybe I should mention that it's hard to verify whether the change in the origin rebasing does *not* change the behavior in any case. Specifically: I did not do a dedicated test in the Engine, verifying that switching the "Rebase in sublevels"-flag on/off has *exactly* the same effect before and after the change (and I think we'd need a sublevel with radius >10000 to *actually* test this). I'm reasonably sure that it does *not* change the behavior, based on my understanding of the code, but ... I could be mistaken here, that's why I asked whether the "high-level" description did in fact describe the intended behavior).

However: I have integrated your comment for setting the origin to zero only when necessary for the sublevel case.

PR #438, opened 2021-05-19
Added teleport flag for cesium georeference transform update

### LINKED ISSUE #307: CesiumGeoreferenceComponent not using teleport flag
When the world georeference changes, the `CesiumGeoreferenceComponent` is notified and it moves the owner actor such that the actor doesn't appear to move wrt the globe (even though this may be a large movement wrt Unreal coordinates). This is normally fine for "movable" objects but not "static/stationary" objects, and we have sufficiently made that distinction in the documentation. However, the current implementation fails to properly move "fully simulated skeletal meshes" as well since we fail to set a teleport flag when moving the object. 

I think this should be very simple, in `CesiumGeoreferenceComponent::_setTransform`, in the `SetWorldTransform(...)` call, we should provide the extra parameters to disable sweeping and enable teleporting.

This was caught by someone on the forum here:
https://community.cesium.com/t/coordinate-systems-dont-align-perfectly/12901/9

xuelongmu:
Closes #307 

I added a bool UPROPERTY which determines whether the teleport setting is set to None or TeleportPhysics. Perhaps it'd be a good idea to expose all three options here instead of a boolean?

UE documentation regarding this flag:
```
/** Whether to teleport physics body or not */
UENUM()
enum class ETeleportType : uint8
{
	/** Do not teleport physics body. This means velocity will reflect the movement between initial and final position, and collisions along the way will occur */
	None,

	/** Teleport physics body so that velocity remains the same and no collision occurs */
	TeleportPhysics,

	/** Teleport physics body and reset physics state completely */
	ResetPhysics,
};
```

Also found this page from the UE blog helpful: https://www.unrealengine.com/en-US/blog/moving-physical-objects

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -12,6 +12,7 @@
 - Added `M_CesiumOverlayWater` and `M_CesiumOverlayComplexWater` materials for use with water tiles.
 - Exposed all tileset materials to allow for changes in editor.
 - Fixed a bug that caused rendering- and navigation problems when zooming too far away from the globe when origin rebasing is enabled.
+- Added teleport boolean property to CesiumGeoreferenceComponent.
 
 ##### Fixes :wrench:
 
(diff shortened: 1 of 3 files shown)


cesium-concierge:

nithinp7:
@xuelongmu Looks good, can you add the fix to CHANGES.md?

xuelongmu:
> @xuelongmu Looks good, can you add the fix to CHANGES.md?

Thanks for the reminder, just updated it 👍

PR #443, opened 2021-05-19
Android Target Platform Support

shehzan10:
This draft PR is a work in progress. The current state is that Cesium Native and the Cesium for Unreal plugin compile for Android, but cannot be packaged as an APK with Cesium for Unreal Samples, or even for Android debugging mode.

See full notes in #190.

Before packaging Cesium for Unreal Samples, In File -> Package Project -> Project Settings, please set the following (or edit the `Config/DefaultEngine.ini` file)

```
+[/Script/AndroidRuntimeSettings.AndroidRuntimeSettings]
+MinSDKVersion=26
+bBuildForArmV7=False
+bBuildForArm64=True
+bSupportsVulkan=True
+bBuildForES31=False
+InstallLocation=Auto
+PackageName=com.Cesium.CesiumForUnrealSamples
+TargetSDKVersion=26
+ApplicationDisplayName=Cesium_For_Unreal
```

#### Preprocessor Definitions

Based on the suggestion in https://github.com/CesiumGS/cesium-unreal/issues/190#issuecomment-843468951, I added `-D__ANDROID__ -DANDROID_LOGGING` to both the `unreal-android-toolchain` file and to the `CesiumRuntime.Build.cs` files (using appropriate syntax). This however didn't help so I haven't pushed it up.

Patch file: [preprocessor.diff.txt](https://github.com/CesiumGS/cesium-unreal/files/6510992/preprocessor.diff.txt)

`__ANDROID__` is supposed to be a built-in definition in the compiler, but I haven't verified that it was coming in for cross compilation.

When I tried to run ` C:/Users/smohammed/AppData/Local/Android/Sdk/ndk/21.1.6352462/toolchains/llvm/prebuilt/windows-x86_64/bin/clang++.exe  -E -dM - < /dev/null`, it did not print the Android preprocessor definitions.

#### MikkTSpace and `genTangSpaceDefault`

This is a function we use from the Unreal Engine source modules. It includes a library for aarch64, but for some reason the build system is not picking the library up, resulting in a undefined reference error.

PR #444, opened 2021-05-20
Use a local copy of MikkTSpace instead of UE's

kring:
Because UE's isn't available on Android.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,6 @@
 [submodule "extern/cesium-native"]
 	path = extern/cesium-native
 	url = ../cesium-native.git
+[submodule "extern/MikkTSpace"]
+	path = extern/MikkTSpace
+	url = https://github.com/mmikk/MikkTSpace.git
--- a/Source/CesiumRuntime/CesiumRuntime.Build.cs
+++ b/Source/CesiumRuntime/CesiumRuntime.Build.cs
@@ -69,6 +69,7 @@ public class CesiumRuntime : ModuleRules
             "CesiumJsonReader",
             "CesiumUtility",
             "draco",
+            "MikkTSpace",
             "modp_b64",
             "spdlog",
             "sqlite3",
@@ -112,7 +113,6 @@ public class CesiumRuntime : ModuleRules
                 "MeshDescription",
                 "StaticMeshDescription",
                 "HTTP",
-                "MikkTSpace",
                 "LevelSequence",
                 "Projects"
             }
--- a/Source/CesiumRuntime/Private/CesiumGltfComponent.cpp
+++ b/Source/CesiumRuntime/Private/CesiumGltfComponent.cpp
@@ -233,7 +233,7 @@ static void computeTangentSpace(TArray<FStaticMeshBuildVertex>& vertices) {
   SMikkTSpaceContext MikkTContext;
   MikkTContext.m_pInterface = &MikkTInterface;
   MikkTContext.m_pUserData = (void*)(&vertices);
-  //MikkTContext.m_bIgnoreDegenerates = false;
+  // MikkTContext.m_bIgnoreDegenerates = false;
   genTangSpaceDefault(&MikkTContext);
 }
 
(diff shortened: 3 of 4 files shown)


cesium-concierge:

kring:
I originally started going down the path of moving tangent generation into cesium-native _properly_. Which tentatively meant introducing a `CesiumGltfProcessing` library which would have these capabilities to start:
* Draco decode
* Image decode
* Data URI decode
* Generate normals
* Generate tangents (with MikkTSpace)
* Turn an indexed triangle list into a non-indexed triangle list.
* Potentially lots more in the future.

But it's complicated. For example, turning an indexed triangle list into a non-indexed triangle list is ~10 lines of code in cesium-unreal. But if we do it in cesium-native and in terms of glTF, it's a _lot_ more complicated because glTF is so flexible. We need to rewrite buffers, deal with interleaved and non-interleaved attributes, watch out for a buffer/bufferView being shared between multiple primitives, and probably much more.

So in the interest of spending time on the most important stuff, I bailed on all that and just added MikkTSpace as a second extern alongside cesium-native and swapped it in. Simple. Maybe.

shehzan10:
We had an offline discussion about whether MikkTSpace should be a submodule to Cesium for Unreal or Cesium Native, even without the larger changes @kring Mentioned above. So the options were
* Submodule to Cesium for Unreal (this PR)
* Submodule to Cesium Native, but unused in Native directly. It gets installed in the ThirdParty Directory in Cesium for Unreal.

We decided that we should test this with #443 first before dedicating more time to its build system. Will comment here once we have an update.

shehzan10:
@kring looks like Mac and Linux are failing to build because of CMake errors. I believe we need to add `C` under project languages in externs/CMakeLists.txt.

shehzan10:
I can confirm that my local build with MikkTSpace compiled using NDK removes the undefined reference issue for `genTangSpaceDefault`.

kring:
This is building successfully on all platforms now. Now we need to make sure we can still package a shipping build with this. As [Marco mentioned](https://github.com/CesiumGS/cesium-unreal/issues/190#issuecomment-845067978), we may have problems with this in shipping builds because UE piles all the code (engine and plugins) into a single executable. Which means we may get UE's MikkTSpace instead of our own. Like SQLite, UE's version is _mostly_ the same. Diffing the headers, the only difference is that `SMikkTSpaceContext` has an extra field in the UE version: m_bIgnoreDegenerates. Which means that if we compile against our header, and then end up linked against UE's lib, we'll be sending in a version of this struct that is smaller than UE's implementation expects. UE will access memory past the end of the struct while attempting to read that field. And then one of these things will happen:

1. It'll be an access violation, and we'll crash. Probably not super likely.
2. It'll be non-zero, and we'll get different behavior than we intend (I'm not sure how much it'll matter).
3. It'll be zero, everything will work fine, and we'll be blissfully unaware there's even a problem.

Just like with SQLite, the solution is hard. Ideally when UE includes modified versions of third-party libraries they would rename the symbols, because the current situation makes it pretty much impossible to (reliably) use the same third-party libraries in a plugin because of how Shipping builds work. But that ship has sailed.

Other options include:
1. Compile against the UE mikktspace.h header even when we link against our own mikktspace.lib (at least on Android). This should work fine with the current version of mikktspace in the engine, but there's no guarantee it retains the right kind of ABI compatibility in the future.
2. Hackily allocate extra space for the struct and initialize it to zero. So hacky, and additional mikktspace changes in UE in the future could cause new and extremely subtle breakage.
3. Rename all the mikktspace symbols on our side so we're certain to get ours instead of UE's in all situations.

(3) is probably the least terrible, even though it's still terrible.

kring:
I addressed this in #443 by only using our MikkTSpace on Android, and using UE's on other platforms. In both cases the header should be matched to the library, so the compiler should help us pick up any (hopefully unlikely) divergence between the official MikkTSpace and UE's in the future. I'm closing this because it will be merged with the Android PR.

PR #445, opened 2021-05-20
Update CesiumSunSky directional light intensity

argallegos:
This PR updates CesiumSunSky's Directional light and changes the default intensity from 75,000 lux to 111,000 lux. This is based on the "Bright Sunlight" value from this page - https://en.wikipedia.org/wiki/Daylight. It's a more accurate default value.


cesium-concierge:

xuelongmu:
Thanks @argallegos

Issue #47, opened 2020-07-31
Support for overlaying one 3D Tiles tileset with another (clipping)


kring:
So that we can have a "base globe" tileset and then overlay smaller, more detailed areas from another tileset. For this issue, we only need something good enough for the primary use-case of "show a globe, let the user zoom in to a more detailed area" and we should aim to keep it simple. But if it's not much more time consuming to implement the latest approach in CesiumJS, we should do that instead.

kring:
Talking to some potential users yesterday, it became clear that it's probably not sufficient to just clip away the model. It's also necessary to blend between the two models so that we don't get cliff faces or holes where they meet. Adjusting vertices so that the two datasets line up probably requires assuming that the datasets are 2.5D-ish. The same folks were saying that in the rare case where the datasets aren't 2.5D at the place they meet, they sometimes (in the past) create custom geometry in 3ds max or whatever to join them, and then blend the bigger dataset with the custom geometry further back.

argallegos:
One more thought on this - the ideal solution will allow you to inverse the clipping mask. This will be important for projects using just one section of Cesium World Terrain, or any other large tileset, and so they should be able to define a volume where only tiles within the volume are loaded. It would be awesome if we could make it as simple as a checkbox on the mask/volume to invert it.

I don't know how the previous solution worked, but if we could prevent tiles outside the volume from loading at all, that would be optimal and solve the collision issue. 

I also think that ideally this would look like rectangular prism/cuboid (or spherical, or whatever is easiest) volume that players could drag in and scale to fit their needs. Ideally they'd be able to drag in multiple, and select which tilesets got clipped on a per-clipping actor basis. I'm kind of picturing Unreal's [Geometry Brush Actor](https://docs.unrealengine.com/en-US/Basics/Actors/Brushes/index.html)s, which are used for prototyping levels. You can set these 3D brushes to be additive or subtractive, which is very cool, though I don't think we would be able to implement clipping volumes in exactly the same way.

kring:
UE has a thing called Procedural Meshes. We can create one from a normal static mesh, and then do a [Slice Procedural Mesh](https://docs.unrealengine.com/en-US/BlueprintAPI/Components/ProceduralMesh/SliceProceduralMesh/index.html) on it to cut it with a plane and optionally cap the sliced edge with geometry.

Pretty fun demo using it here: https://www.youtube.com/watch?v=oIdKxYYQBdw

[...]

PR #446, opened 2021-05-20
Clipping 2.5d

nithinp7:
Partially solves: https://github.com/CesiumGS/cesium-unreal/issues/47

Builds upon the cesium-native changes here: https://github.com/CesiumGS/cesium-native/pull/252 

Allows cartographic polygon selections that will be rasterized onto chosen texture names applied to the tilesets. Returns the polygons as multiple rasterized textures in the gltf that can be used to apply mask-based shader effects (culling, water, displacement, etc.).

Instructions to use:
1. Search for the `CesiumCartographicSelection` actor and drop it in the scene.
2. There is a spline on the actor which can be expanded and manipulated using the typical Unreal spline controls.
3. There are currently two options in the `Cesium` section of the details panel: `Is For Culling` and `Target Texture`. 
    * `Is For Culling` is a boolean to let cesium-native know if it should optimize by culling tiles that are entirely within the selection (should not be checked for non-culling uses). 
    * `Target Texture` is a string letting cesium-native know what texture name to target when rasterizing the selection. This same texture name can be later used on a texture object parameter in the tileset material to access the result of the rasterization. Multiple selections can use the same target texture name to be rasterized together into the same texture. The default texture target is "Clipping".
4. Once the selections are made, go to the `Cesium3DTileset` actor you wish to use the selection on. In the `Cesium3DTileset` details panel, go to "Cesium-->Rendering", and add the selections you want to in the "Cartographic Selections" array.
5. Once you're done, click "Refresh Tileset" to apply the polygon masks. Any future adjustments to the selections can be applied by hitting the "Refresh Tileset" button again.
6. To try it out, use the `OverlayAndClipping` example material in `CesiumForUnreal Content->Experimental->Materials->Example`. The material can be dragged into the first material slot on the CWT tileset to test it. Go to the Bing overlay component and type in "Overlay0" in the new name textbox. Notice in the example material how the texture target names are used for the translation/scale and overlay texture.

###### The Cesium World Terrain and OSM Buildings tilesets culled against the Aerometrex Denver photogrammetry tileset.
culled-denver

###### The same culled selection shown without the Denver tileset.
culled-denver2

###### A lake with a custom water mask applied using the selection capability. 
denver-lake


cesium-concierge:

nithinp7:
@kring I think this is ready for a first pass review at least! I think there are a few kinks to be sorted out in this PR, but the cesium-native part is complete except for anything caught in review.

kring:
This is really cool!

I suspect the hardest (remaining) part of this is going to be figuring out the UX. It's really powerful, but also a little overwhelming, and will probably be hard to explain to users. I think the key will be in making "easy" paths for the common things, like clipping, while still making the full complexity available as an option for more advanced users.

Not a full review at all, but some quick thoughts:

* I initially found it really awkward to define a 2D boundary with a 3D spline, but once I realized I can press END to snap spline points to the surface it was much better. It's still a little fiddly though. Rendering drop lines from the spline would make a huge difference I think. But how far down do we drop them?
* Would it make sense for the spline to be a Component attached to the Tileset, rather than a separate Actor?
* Thinking about "the material problem", here's a slightly crazy and perhaps terrible idea... What if every tileset had its own  _embedded_ material used to render that tileset? When you create a new Tileset, it would create a new Material to go with it, and that Material would be useful for more tilesets while also having lots of nodes and comments around that explain and facilitate wiring it up for more advanced scenarios. Perhaps it could even use static switch parameters and we could create multiple instances of it with different switch parameter values at editor time. 😬  I guess this would probably lead to an explosion in the number of shaders.

shehzan10:
> * I initially found it really awkward to define a 2D boundary with a 3D spline, but once I realized I can press END to snap spline points to the surface it was much better. It's still a little fiddly though. Rendering drop lines from the spline would make a huge difference I think. But how far down do we drop them?

Another idea, but probably more complex: Can we derive an outline from an 3D Tileset? I assume that is going to be the use-case 80-90% of the time, to inset a high-res tileset into another 3D Tileset. If we can click a button to get its 2D footprint outline, that might go a long way.

kring:
> Can we derive an outline from an 3D Tileset?

Well we'd need to download all the tiles, right? At least the ones on the edges. It's probably more practical to do this in the tilers.

shehzan10:
> > Can we derive an outline from an 3D Tileset?
> 
> Well we'd need to download all the tiles, right? At least the ones on the edges. It's probably more practical to do this in the tilers.

Wouldn't we have a sense of the corners from the bounding volume in the root tile? And then go from there? I know it might get tricky with something like OSM Buildings.

I agree its complex, but could be post initial merge.

kring:
> Wouldn't we have a sense of the corners from the bounding volume in the root tile?

Yeah, we'd try to guess at where the edges are based on the bounding volume, but then we'd have to go download every leaf tile on the edge, iterate to find the real edge when we guess wrong, and compute the precise boundary from the individual tiles.

shehzan10:
> Yeah, we'd try to guess at where the edges are based on the bounding volume, but then we'd have to go download every leaf tile on the edge, iterate to find the real edge when we guess wrong, and compute the precise boundary from the individual tiles.

Yeah makes sense. Maybe one for the future.

xuelongmu:
> * I initially found it really awkward to define a 2D boundary with a 3D spline, but once I realized I can press END to snap spline points to the surface it was much better. It's still a little fiddly though. Rendering drop lines from the spline would make a huge difference I think. But how far down do we drop them?

One quick workflow hack to make this process a bit easier is to use the Top camera view:

image

yateam:
Any update on this PR? When is it going to be merged?

kring:
@yateam any updates will be posted here and in the corresponding cesium-native issue (https://github.com/CesiumGS/cesium-native/pull/252). We're not currently expecting to ship it in the July release, but it will very likely make it into the August release. In the meantime you can download a plugin package of this branch from our CI system and try it out. Click "Show All Checks" above and then click the "Details" link next to "plugin-package-combined". Extract it to your Unreal Engine plugins directory (you may want to uninstall the official plugin via the Marketplace first).

yuz1:
@nithinp7  It works perfectly in editor,but unable in PIE and packeted project.

yuz1:
> @nithinp7 It works perfectly in editor,but unable in PIE and packeted project.
https://github.com/CesiumGS/cesium-unreal/pull/579#issue-706332868

kring:
Still to do:

* [ ] Fix broken Travis build.
* [ ] Write tests for BoundingVolume's getGlobeRectangle and evaluate if it's "good enough" for now (produces accurate-enough bounding rectangles).
* [ ] If getGlobeRectangle is not sufficient, implement (projected) bounding rectangle computation from actual vertices.
* [ ] Verify that combining quadtree raster overlay tiles together for a geometry tile works correctly when some of the quadtree tiles successfully load and some fail.
* [ ] Make the quadtree tile cache size configurable.
* [ ] Make the water mask a raster overlay.
* [ ] Sort out projections and texture coordinate IDs. Is the move to names instead of IDs justified?
* [ ] Do projections need to be an unordered_set? There will be very few of them.

nithinp7:
This looks good, thanks @kring!

Issue #437, opened 2021-05-17
Make Cesium content easily discoverable


javagl:
<sup>This could be considered as a "generalization" of https://github.com/CesiumGS/cesium-unreal/issues/277</supe>

Right now, certain elements do not appear in the content browser. The **Quick**(!)Start provides elaborate guidance at https://cesium.com/docs/tutorials/cesium-unreal-quickstart/#step-3-add-sunsky-lighting-and-a-floatingpawn : ~*"Click here, check there, look, like this image, search for this, change this setting, ignore most of the information that you're flooded with, scroll around, until you find that name"*. We should not force newbie users to go through this procedure (to cover up for our lack of understanding of what's wrong).

We should investigate what is necessary for objects to appear in the "Place Actors" panel. This might already be sufficient. 

As an additional step, we could consider extending the Cesium "Quick Add" panel. Right now, the "CesiumForUnreal" root folder in the Content Browser contains ~20 elements, among them being things like a "PointInBox", a "transparent1x1" image, the profiling blueprint. The C++ classes contain important things like the Tileset, but also things like the "Globe Anchor Parent" (what is that? what is it used for?). 

Having the *important* classes/elements that are necessary to build a level directly in the "Quick Add" panel (maybe as some sort of "small 'content browser'", but with the option to add some explanatory text) could make things **much** more useable (as indicated by the tutorials that could be *much* shorter...).

argallegos:
It would be really nice to have a few actors (CesiumSunSky, DynamicPawn at the very least) available in the Quick Add panel. If that is relatively straightforward, it would be great to add. If not, there may be a way to force the objects to appear in the Place Actors panel - I agree that the current process is too complicated to add actors like that.

kring:
It should be really easy to add this to the Cesium panel. The hardest part is deciding where to put the buttons.

kring:
Fixed in #451.

PR #451, opened 2021-05-22
Add Cesium assets to Quick Add panel

javagl:
May fix https://github.com/CesiumGS/cesium-unreal/issues/437 ...

Although the QuickStart tutorial elaborates how to make things like the `CesiumSunSky` visible in the content browser, it's a bit unfortunate (for newbies) that this process is so complicated. Even if all Cesium assets appeared in the Place Actors panel, things could still be offered in a simpler form.

This PR adds an "AssetDataList" that can be filled with a set of predefined assets, and that can then be dragged-and-dropped into the viewport (or into the Details View, for components like the `CesiumIonRasterOverlay`). 

It currently looks like this:

CesiumUnreal asset list

Of course, styling, fonts, layout, yadda yadda, but I think that could already *greatly* simplify and shorten the QuickStart tutorial (and others, maybe).

I'm going out on a limb with this one, because ... I don't have a clue about Unreal. There are very likey things that are "not good" in terms of Unreal idioms (asset registry, slate, drag+drop, ...), so any feedback by unreal experts saying what is wrong about the general approach would be appreciated.

If this is not totally off, one could consider separating the "Actors" (that can be dragged and dropped into the viewport) and the "Components" (that can be dragged and dropped into the Details View of a selected Actor), but ... right now, I think `CesiumIonRasterOverlay` is the only component that should be available for users, so that may not be worth an own list.


cesium-concierge:

javagl:
Still TODO: I think that a line like `FModuleManager::LoadModuleChecked<FAssetRegistryModule>("AssetRegistry")` may require the `AssetRegistry` do be declared in the build file as a "loaded module". (Maybe it's not necessary for this case, because it's an internal module that is part of the editor anyhow, but it should be checked)

argallegos:
Is there a chance we might be able to get this in before the next release? It would definitely simplify the upcoming learning content about CesiumSunSky.

argallegos:
@javagl 

There will be a dedicated tutorial for CesiumSunSky, it's currently in progress.

Part of the page will include setting up a scene and connecting the georeference with the sunsky. That section is already created, but can be easily revised later on when we do iron out exactly how the Sunsky and the Georeference should work together. Since it's already written, I think it's ok to publish for now and update later. The rest of the page is essentially reference about different settings on the directional light, sky atmosphere, etc. 

I unfortunately don't know very much about editor coding, so I can't comment on some of your points, but I think it's looking very good so far. 

Design-wise, I might put the new Cesium Assets section under a header that looks like the "Quick Add" header above, and change phrasing from "Cesium Assets" to "Cesium Actors".  We also may want to add plus signs to add these entries as well, or have some way of indicating that it's a drag-and-drop instead.

javagl:
> Design-wise, I might put the new Cesium Assets section under a header that looks like the "Quick Add" header above, and change phrasing from "Cesium Assets" to "Cesium Actors". We also may want to add plus signs to add these entries as well, or have some way of indicating that it's a drag-and-drop instead.

- "Cesium Actors" does not apply to the raster overlay, because that's a Component (there's still the option to leave that out, or put it into a dedicated "Cesium Components" section). I think we currently don't have a tutorial where the "Cesium ion Raster Overlay" (or another component) is added to an (otherwise untextured) Tileset, but assume that this will be created at some point. (A technical aside: The Slate component that is used there is a "List View", and a generic one for that matter. Leaving out (or changing) the Header row should be doable, but may not be totally trivial - I have to look into that...)

- The "Plus Sign" Button: I think the "usual" ("idiomatic") way of how things work in the Unreal Editor is via Drag-And-Drop, and having to use a Button is somewhat unusual. But a caveat with dragging+dropping a tileset into the viewport right now is is that it will end up where it is dropped (i.e. its position will not be (0,0,0)). That could be a justification to leave out the Tileset there as well. (I'd like to avoid writing something like "Drag and drop the tileset, and then (manually) change its position to (0,0,0)" in the tutorial). But in any case: Having such a button to add the respective Actor (at (0,0,0)) could certainly make sense and should be easy to add.

kring:
Thanks @javagl, definitely a huge improvement here! I think we need to think through the UX a bit here, though. For example, there's a dedicated button in the toolbar to add a blank tileset. Maybe other Actors could be accessible from a dropdown attached to that button? Or maybe they could go in the Quick Add section as well and follow the same UI paradigm? Having a "quick add" for the CesiumSunSky or DynamicPawn makes a lot of sense to me, especially if adding them automatically wires the CesiumSunSky to the Georeference and avoids creating either object if one already exists in the level (it would just select the existing one instead).

As for the Georeference, it can never really be removed, right? It just gets removed from the Level? Would it be crazy for the Tileset and anything else that needs a georeference to re-add it to the level on Tick if it's not already there? 😬 I haven't totally thought through the implications of that.

I don't think we need FloatingPawn to be on the panel at all, since it's deprecated.

I think that just leaves the overlay components. Maybe the best way to approach that is with a small custom add-an-overlay UI on the Tileset Actor's details panel. That can be left for a future PR since it's probably a fair bit of new work.

javagl:
A sketch:

    [add] [upload] [learn] [help] [signout]       
    
    --- Quick add -------------------------
    
    CWT + Bing                        [ + ]
    CWT + BingLabels                  [ + ]
    ...
    OSM                               [ + ]
    
    
    --- Default actors --------------------

    (Blank) Tileset                   [ + ] 
    CesiumSunSky                      [ + ]
    DynamicPawn                       [ + ]
    
    
The "Add Blank" button is removed. The "Quick Add" section remains unchanged. The "Default Actors" contains the (Blank) Tileset (that was previously added with the "Add Blank" button), as well as `CesiumSunSky` and `DynamicPawn`. These "Default actors" (or however we call them) could be added with the "[+]" buttons at (0,0,0), or dragged-and-dropped.

(Just a sketch, to approach some consensus - have to figure out how to make this nicely in Slate).

---

> As for the Georeference, it can never really be removed, right? It just gets removed from the Level? 

Selecting and deleting it will ... well, *delete* it. That's the problem that caused https://github.com/CesiumGS/cesium-unreal/issues/242 . A new GeoRef has to be created in that case. This was addressed (somehow) in https://github.com/CesiumGS/cesium-unreal/pull/424 , where a "new *default* GeoRef" was created in that case, when it was requested. (tl;dr: The whole GeoRef lifecycle issue, you know, but ... of course, we can omit that entry from the list).

kring:
That looks good Marco, though I think I'd just put the three items under the Quick add header rather than creating a new section. The way I see it, the Quick Add list is just convenient presets to add to your level, and that's basically what these new items are, too. For example, CesiumSunSky shouldn't just be a default instance of that actor, it should be wired up to the CesiumGeoreference. DynamicPawn should be configured to auto-possess the player.

javagl:
> CesiumSunSky shouldn't just be a default instance of that actor, it should be wired up to the CesiumGeoreference. DynamicPawn should be configured to auto-possess the player.

If the `CesiumSunSky` is added *before* a tileset has been added (i.e. before the default GeoRef is created), then there will be nothing to connect to. If it is added *after* a tileset has been added, it will be connected. Explaining this difference will be difficult. But this _might_ no longer be an issue, **if** we change the `CesiumSunSky` to also use the `GetDefaultForActor` mechanism for the GeoRef.

The case that somebody adds multiple `DynamicPawn` instances (and they all auto-possess the player) does not seem to cause any trouble, though, but one might have to analyze waht is happening there under the hood (e.g. whether only the last one will be auto-possessed or so...)

kring:
> If the CesiumSunSky is added before a tileset has been added (i.e. before the default GeoRef is created), then there will be nothing to connect to. If it is added after a tileset has been added, it will be connected.

Adding a CesiumSunSky should also add a CesiumGeoreference, if there isn't one already in the level. Maybe that's what you meant with "But this might no longer be an issue, if we change the CesiumSunSky to also use the GetDefaultForActor mechanism for the GeoRef."

> The case that somebody adds multiple DynamicPawn instances

If there's already a DynamicPawn in the level, clicking the DynamicPawn item in Quick Add should select the existing one rather than add a new one. Just like it does with tilesets.

javagl:
Most of this PR has become obsolete. I'll consider creating a new one.

Issue #413, opened 2021-05-05
Investigate compatibility with sky plugins


argallegos:
There has been interest on the forums in increased compatibility with plugins like TrueSky, e.g. making CesiumSunSky share its directional light with these other systems. It may be good to research how these other plugins work and if it is straightforward to better expose the directional light in Cesium SunSky.

TrueSky, for example, references a directional light in the scene. CesiumSunSky has a directional light component as part of the blueprint. It would involve changing the base SunSky functionality somewhat, but it may be possible to modify CesiumSunSky so that it also references a directional light in a scene, rather than its own directional light component.

WellyWonder:
Yes please

WellyWonder:
Any chance for a tutorial or just advice on how to to modify CesiumSunSky so that it also references a directional light in a scene?

argallegos:
Hi @WellyWonder,

We are currently working on a tutorial that covers CesiumSunSky in detail. In the meantime, I tested out a few things and I was able to modify CesiumSunSky to use a scene directional light instead. You can find those changes in [this branch](https://github.com/CesiumGS/cesium-unreal/tree/refactor-sunsky). 
This may be merged into main and included in a future release soon, but I haven't done thorough testing yet so I'm not sure what the timeline on that would look like. 

If you want to replicate the changes yourself, here's what I did:

- Added two new editor variables, a boolean and a Directional Light Object Reference. You can create a directional light in the scene and assign it to the Level Directional Light object reference.
![image](https://user-images.githubusercontent.com/39537389/119851106-23f0d380-bedc-11eb-82ad-dbefce1e5399.png)
- In the Construction Script, add a branch to turn off and on the directional lights based on which one you'd like to use
![image](https://user-images.githubusercontent.com/39537389/119851420-61edf780-bedc-11eb-8fed-77b52f597fe6.png)
- In the Update Sun function, "Orient Sun and Directional Light" section, add a Select node to pick between the two directional lights and replace the original directional light reference. All the nodes above the comment box were added for this.
![image](https://user-images.githubusercontent.com/39537389/119851719-a37ea280-bedc-11eb-8101-9fc3107013ec.png)

Please let me know if this works for you!

PR #460, opened 2021-05-27
Expand Cesium Sun Sky functionality

argallegos:
This PR revises a few things about CesiumSunSky, and also adds functionality for alternative Directional Lights to resolve https://github.com/CesiumGS/cesium-unreal/issues/413. 

In short:

Editor variables have been reordered to clean them up a little. 

- Tooltips were added to Latitude and Longitude to warn players that they are read-only. 
- The Year property has been exposed again (I don't know why it was removed, and in my opinion it should stay, but if there was a specific reason that it was hidden please let me know). 
- The Date category and the Time category were combined, because the only public variable in Time was Solar Time. I'm not totally committed to this, so I can put these back if we want. 
image

You'll also see two new variables in the "Sun" category. This allows users to choose whether to use the default Directional Light or use a different one placed in the level, which means that it should become compatible with other sky plugins. The "Use Level Directional Light" hides and unhides the two lights as needed, which I _think_ means that it should have approximately the same performance impact as the original directional light setup, but I may be wrong about that.


cesium-concierge:

kring:
This looks good to me, thanks @argallegos! I updated CHANGES.md.

PR #465, opened 2021-05-28
Disable support for water mask on Mac

nithinp7:
I don't have a Mac to test this on, but this should prevent a horrible crash on Mac that seems to be caused by using the water material. In the future once we figure out what's causing it, we may be able to come up with a fix that allows water mask to be supported on Mac.

### DIFF
--- a/Source/CesiumRuntime/Private/Cesium3DTileset.cpp
+++ b/Source/CesiumRuntime/Private/Cesium3DTileset.cpp
@@ -697,7 +697,11 @@ void ACesium3DTileset::LoadTileset() {
   this->_startTime = std::chrono::high_resolution_clock::now();
 
   Cesium3DTiles::TilesetOptions options;
+  // TODO: figure out why water material crashes mac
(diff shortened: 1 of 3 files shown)


cesium-concierge:

xuelongmu:
I just pushed a way to check for Mac platform by storing the Platform Name in a UPROPERTY and doing a check in the `EditCondition` to see if PlatformName = TEXT("Mac"). If true, the property check box should be (hopefully) disabled.

The other way to do it (which might be more typical UE4 programming style) is to wrap the entire `EnableWaterMask` in an ifdef, so that the property does not show up at all on Mac. Any usages of `EnableWaterMask` property would then also have to be wrapped.

shehzan10:
Please make sure you open an issue to add the support back.

kring:
I don't have a Mac to test this on either, but it looks ok. I'm going to merge it to work toward a final release candidated we can test (including on Mac for real).

PR #384, opened 2021-04-26
Add developer setup instructions

javagl:
<sup>This was originally posted as a comment in https://github.com/CesiumGS/cesium-unreal/issues/296 , but now mored here.</sup>

This may fix https://github.com/CesiumGS/cesium-unreal/issues/296 and (when it's fleshed out and extended accordingly), also https://github.com/CesiumGS/cesium-unreal/issues/164 

---

I'm aware that "the real feedback" may come when somebody tried to set up the whole thing, based on this description, *from scratch*. Also, it's currently obviously Windows/VisualStudio-centric, but that's the main development platform for now.

NOTES: 

- The "Generate Visual Studio project files" step only works only after a

      git checkout add-source-folder

  in `C:\Dev\cesium-unreal-samples`

- There currently is no sensible way to to trigger breakpoints in `cesium-native`. Rumours say that it's possible by copy-and-pasting paths from one VS instance to another, but not in a way that would do the term "IDE" (as in "*Integrated* Development Environment") any justice. It *should* be possible to essentially "merge" multiple VS solutions (i.e. open them at the same time), but during my first test, this caused VS to behave erratically. Maybe there is a solution for that that I haven't found yet.

Fixes #152

PR #471, opened 2021-05-30
Update development-guide.md

shehzan10:
* Regorganized the development guide for a better flow.
* Added steps to compile and package the plugin using the `RunUAT` command.

### DIFF
--- a/Documentation/development-guide.md
+++ b/Documentation/development-guide.md
@@ -6,105 +6,144 @@ The following steps detail how to build the plugin and use it as part of your pr
 
 Cesium for Unreal depends on Cesium's high-precision geospatial C++ library - [Cesium Native](https://github.com/CesiumGS/cesium-native), which is included as a submodule.
 
-## Compiling Cesium for Unreal
+**Table of Contents**
+
+0. [Prerequisities](#0-prerequisities)
+    * [Windows](#windows)
+    * [Mac OSX](#mac-osx)
+    * [Linux](#linux)
+1. [Clone Cesium for Unreal](#1-clone-cesium-for-unreal)
+2. [Compiling Cesium Native](#2-compiling-cesium-native)
+    * [Windows](#windows-1)
+    * [Mac OSX](#mac-osx-1)
+    * [Linux](#linux-1)
+    * [Android (Cross Compiling on Windows)](#android-cross-compiling-on-windows)
+3. [Cesium for Unreal](#3-cesium-for-unreal)
+    * [Using Cesium for Unreal from Local Directory](#using-cesium-for-unreal-from-local-directory)
+    * [Packaging Cesium for Unreal Plugin](#packaging-cesium-for-unreal-plugin)
+
+## 0. Prerequisities
 
 ### Windows
-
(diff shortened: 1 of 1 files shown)


cesium-concierge:

kring:
@shehzan10 this is divering even more from #384. How can we reconcile these sets of changes?

shehzan10:
@kring Just took a look at it. The detail in there is great for an advance reading of the compilation configurations and developing Cesium for Unreal in a more active manner.

IMO These two guide, while having some parts overlap, should be separate with current guide providing a more concise single-path compilation workflow. For example, someone who wants to build and test with a custom branch will be able to use this, or someone who may want to compile the plugin for UE5 EA.

The guide #384 has a lot more instructions for actively adding new features/developing with the plugin, which is also very useful. I feel like combining them into a single guide could lead to longer reading than might be required for most users.

Maybe we should rename this guide to something like "Build Guide"? What do you think?

kring:
I'm not sure, @shehzan10. I guess some of the CLI vs Visual Studio vs CMake GUI vs VS Code stuff in #384 is redundant and we could cut it down to just one prescribed workflow. But the development-guide mostly just seems to be missing steps and information that people will need to know about in order to reliably be successful. #384 is missing Android instructions entirely, though. It also doesn't have anything about packaging the plugin from the commad-line, though this seems to me like the thing our users are least likely to need among all this content.

Ok let's break it down. If a user wants to...

1. _Build and test a custom branch_: Why?? They can just get the build plugin from our CI.
2. _Use a custom branch_: It's simple, but we should still provide instructions for installing a plugin built by our CI in the Engine plugins directory. Those instructions are missing entirely right now AFAIK.
3. _Compile for some other platform or Unreal Engine version_: Arguably the instructions in this PR would be useful for this, except that inevitably it's not going to work on that new platform/UE version and will need to be fixed. And that means having an actual development environment and knowing how to use it. So they really need #384.
4. _Build and test main or a custom branch with the intention of then modifying Cesium for Unreal in some way_: The user will eventually need #384, not sure that starting with the simplified instructions here offers any actual economy.

Have I missed anything?

Long story short, I think my preference is to roll some of the information in here into #384 (specifically the Android and plugin packaging bits) and delete the development-guide.md. We can restructure things as necessary in #384 to make it easier to focus on the most important information. For example, we might put only the CLI instructions on the main page, and have links to separate pages for the other environments.

shehzan10:
@kring Can't say I disagree with most of what you have written. The one thing that was missing for me was that we had not documented how to run the `RunUAT` commands, even for internal use. So I decided to do a little bit of cleaning of the documentation and add it in.

If we want to have #384 be the draft of what becomes the final version, we can certainly go with that. I do think we need to have a "short-form" version of the detailed instructions. Where and how that is added, we can discuss once #384 is closer to the finish line.

javagl:
I gave it a stab, and tried to merge *this* PR and https://github.com/CesiumGS/cesium-unreal/pull/384 into https://github.com/CesiumGS/cesium-unreal/pull/481

PR #475, opened 2021-06-03
Break apart tileset tick function

javagl:
While testing/debugging in the context of https://github.com/CesiumGS/cesium-native/issues/186 , I tried to break apart the `Tick` functions, into building blocks that have **names** and **meanings** and **comments**. Maybe, one day, I'll either be able to ignore things like that, or not have to do them in the first place.


cesium-concierge:

javagl:
I'll just mark this as "ready for review", even though it does not "*have* to be" merged.

kring:
Thanks @javagl!

PR #482, opened 2021-06-06
Extended quick add panel

javagl:
Based on the discussion in https://github.com/CesiumGS/cesium-unreal/pull/451 :

- Removed the "Blank Tileset" button from the Toolbar
- Added a "Blank Tileset" entry in the QuickAdd-List instead
- Added an entry to add one "CesiumSunSky"
- Added an entry to add one "DynamicPawn" (which is auto-possessed to Player 0)

CesiumUnreal QuickAdd again

I took the liberty to add tooltips to the entries. There may be some wordsmithing to be done, at https://github.com/CesiumGS/cesium-unreal/commit/4fcf6f7ccd848b8d4e1086032a2c5ca744dad0f4#diff-327861e3f97a3a851d4277e846070d9a98999a100aab925bc8072fb464a027a7 (*edit:  for example, "Bing Maps Road imagery" is not satellite data...*), but it may be a start.

---

<sup>(An aside: I don't like the style of this implementation, compared to the generic, `AssetData`-based one with Drag-And-Drop from the other PR, but maybe *this* implementation is closer to what is expected...).</sup>


cesium-concierge:

javagl:
Converting to Draft again: It does not make sense that the newly added entries are *not* shown when the user is *not* logged in. (They should be available regardless of the login state).

shehzan10:
Drive by suggestion: Let draw a separation (Like Quick Add + horizontal line) between the Quick-add Global 3D content, and the UE stuff like SunSky, Pawns etc.

javagl:
I have separated the panels into "Quick Add Assets" and "Quick Add Actors". When logged in, it looks like that:

CesiumUnreal QuickAdd LoggedIn

When logged out, the "Login panel" takes up quite some (fixed) space, due to manual `Padding`s that have been inserted... :-/ it could be OK, but one may have to scroll down on smaller screens....:

CesiumUnreal QuickAdd LoggedOut01

CesiumUnreal QuickAdd LoggedOut02

I don't particularly like this, but assume that people have strong opinions about what "The Right Layout" should be here, so leave that to others to decide.

---

<sup>(Another aside: Now that the Panels are separated, people of the future might even more wonder about why certain things are implemented the way they are, but I hope that it will then be sufficient to point to 451 where necessary)</sup>

javagl:
Assuming that, regarding the layout, there either will be feedback .. or not .. I'll mark this as Ready For Review.

kring:
I'm getting errors at startup in this branch, but not in main:
image

kring:
Not sure if some/all of these problems flow from the one above, but some other problems I'm seeing are:
* Clicking the `+` for "Cesium SunSky" or "Dynamic Pawn" does nothing if these objects are already in the level. It should select the existing object the way the other quick add buttons do.
* The Actor created by the "Cesium SunSky button" has the wrong extensions. It should have these:
image
But instead it looks like this:
image

kring:
The above aside, I don't think we've quite nailed the UX here yet. I don't think I agree with Shehzan's suggestion to create a separate section for these other actors. For one thing, they're _all_ Actors, including the previous quick add options. Also, I hadn't thought about the case where the user is not logged in to Cesium ion, and it's pretty awkward.

If I can put on my ill-fitting UX hat (ask @philipgrimmett how afraid you should be), I think we need our main panel to be something _very roughly_ like this:
image

Geeze that looks terrible, but stay with me here.... The idea is that, like now, there's a list of all the awesome things you can add to your level, and you can add them with one click. Ideally this would be pretty and eye catching: nice, desaturated images of the assets you're adding, even. We'd also move the confusing and kinda hidden "Add" and "Upload" buttons to be on this main panel as well alongside all the assets (see the buttons at the bottom).

Now the key thing is that this panel is pretty much the same no matter whether you're logged in to ion or not. The only difference is at the top there's an indication of the state and the option to sign in or out depending on the current state. But if you're not signed in and you click on one of the ion assets, that's our chance to tell people how cool ion is and give them a chance to sign in or sign up. Maybe we use a subtle indication on the panel (a background color or something) to distinguish the ion assets from the non-ion assets so that users who can't or don't want to use ion don't have to click blindly to learn what they can and can't use.

As the list of assets grows, we can have some filtering controls as well. For example, a Category that defaults to "Featured" or something like that.

This is all a good amount of work, though, and really ought to have some legit UX design go into it rather than just my hand waving. Meanwhile having these new Actors on the panel is much better than not having them. So maybe the current state is workable for now.

javagl:
That last one sounds like a considerable restructuring (even though it's "technically trivial", in some way). I'm not a UX expert, but to reiterate/summarize some thoughts:

- There may be certain "idioms" how such a UI is usually presented in UE. I thought that a list with drag-and-drop from the first PR was such an idiomatic UE thing. But the caveat that things usually should be placed at (0,0,0) justifies deviating from that, and using these [+]-buttons instead.

- Subjectively: Part of a good UX is that the UI does not change too often or too dramatically. (Sometimes, I wonder whether they just want to keep developers busy and users confused). So whatever we'll come up with, it should be "stable".

- For the login part, one option could be: Put the new items at the top. Below that could be the login area. After logging in it shows the ion-quick-add items. If it was possible to *always* show the buttons for the ion assets, but leaving them greyed out when not logged in, then this could have some "teaser effect" - particularly when they have some nice Tumbnails showing what people would get (a textured globe, that is). 
  <sup>For the thumbnails: These would probably be fixed size, and not configurable-size as in the content browser</sup>

- I thought about handling the login via some modal dialog window. Right now, for example, there is no way to cancel the sign-in-process. But that would also require some restructing and advise from someone who as certain UX/workfow in mind.

---

@kring The fact that the existing elements should be selected was an oversight. I can add this.

But more importantly: 

> I'm getting errors at startup in this branch, but not in main:

Of course, I don't get these (otherwise I wouldn't have marked this as Ready For Review). I can only guess that this may be caused by [loading the respective BP classes](https://github.com/CesiumGS/cesium-unreal/pull/482/files#diff-406aadf30a953e9dc85be4d57d33d245031092be84f45ada9a080a352867c426R212), but ... if that failed, it should print a different message. It looks like it could not load one of the classes that are used in the blueprint, Do these link-like function names (with the magnifying glass) lead somewhere when clicking on them?

(I assume that this causes the missing components. For me, it shows the CompassMesh/SkyLight/DirectionalLight/SkyAtmosphere, as usual. But... *missing* components is one thing. What's more confusing is: Where does that BillboardComponent suddenly come from?!?)

javagl:
I think the issue of the order and the placement of the "Login" panel could rather easily be resolved, by having the "Basic actors" at the top...
```
[add] [upload] [learn] [help] [signout]       

--- Add basic actors ------------------

(Blank) Tileset                   [ + ] 
CesiumSunSky                      [ + ]
DynamicPawn                       [ + ]

```
and below that, showing *either* the "Login" panel with the logo, *or* the list with ion assets and the title "Add Cesion ion assets"

---

BTW: I tried moving the "Connection status" into a modal dialog. I think that could certainly make sense, so that people don't mess around in the editor while the login process is not finished (and so that they can *cancel* the login process), but that's not so straightforward.

<sup>
(1. the connection does not proceed, because the AsyncSystem waits for it to be executed in the main thread, and 
2. there is no functionality for cancelling the connection, apparently)
</sup>

It could, however, make sense to *replace* the login panel (that with the logo) with the connection status panel during the login process.

---

And I *still* think that showing the elements of the "Cesium ion quick add" list even when the user is logged out (!), could serve as some sort of "teaser". This could be done by just disabling that list. Where and how it could be displayed... yeah, certainly not *exactly* like that:

CesiumUnreal QuickAdd Overlay

... but maybe in a similar way, as in

    [add] [upload] [learn] [help] [signout]       
    
    
    --- Add basic actors ------------------

    (Blank) Tileset                   [ + ] 
    CesiumSunSky                      [ + ]
    DynamicPawn                       [ + ]

       ________________________________
      |Log in to Cesium ion for all the|
      |fancy things below:             |
      |                                |           
      |          [ Connect ]           |
      |________________________________|


    --- Add Cesion ion assets -------------   (disabled)
    
    CWT + Bing                        [ + ]   (disabled)
    CWT + BingLabels                  [ + ]    ....
    ...
    OSM                               [ + ]

(If there wasn't the logo that takes up so much space... )

kring:
Thanks for all your work on this @javagl! I tweaked a few things and I'm going to merge it as soon as CI passes. Certainly there's a _lot_ more we can do here, but let's get this clear improvement shipped. Here's what I changed:

* Fixed the SunPosition errors I was seeing. The Cesium for Unreal plugin seems to be loaded before the SunPosition plugin on my system. I don't entirely understand why, but it's easy (and arguably an improvement) to only get the CesiumSunSky and DynamicPawn classes when we need them rather than at module startup, and that fixes the problem.
* Attempting to add a CesiumSunSky or DynamicPawn when one already exists in the level will not select the existing ones in the Outliner.
* When we attempt to add a CesiumSunSky, the new or existing CesiumSunSky is now set as the Georeference's designated SunSky.
* Made the Cesium for Unreal logo smaller, less pixelated, and less padded.
* Used the titles and section order from your latest screenshot.

Here's what it looks like before signing in:
image

And after signing in:
image

I really like the idea of showing the list of ion assets before signing in, but I'll write a separate issue for that.

Issue #485, opened 2021-06-08
Cesium icons disappear when the "small icons" setting is enabled


javagl:
As reported in https://community.cesium.com/t/small-icon-support/13988 , when enabling Edit -> Editor Preferences -> "Use Small Tool Bar Icons", the Cesium Icons in the toolbar (including the Cesium Panel toolbar) just disappear.

PR #486, opened 2021-06-08
Add support for small tool bar icons

javagl:
Fixes https://github.com/CesiumGS/cesium-unreal/issues/485


cesium-concierge:

javagl:
CesiumUnreal icons!!!

CesiumUnreal icons but small!!!

shehzan10:
@javagl Is this changing the default icon size, or it is only if a user has edited the setting in their editor that we are adding support for?

javagl:
It does not change the default icon size. The first screenshot shows the appearance with the default setting (i.e. with the global "Use Small Tool Bar Icons" being *disabled*), and this has not changed. The second screenshot shows the appearance after the setting was *enabled*. Then, it shows the smaller icons. (Before this change, enabling this setting just caused all Cesium icons to disappear).

kring:
👍

Issue #491, opened 2021-06-11
Cesium OSM Buildings spams the output log with "does not have a Gltf" messages


kring:
I think this started with #475, but I didn't notice before I merged it.

![image](https://user-images.githubusercontent.com/924374/121679439-ab089300-cafb-11eb-8a9c-db0bba5756d3.png)

Cesium OSM Buildings uses a lot of external tilesets, so there are (by design) a lot of tiles without a glTF.

PR #492, opened 2021-06-14
Omit warning about missing glTFs

javagl:
Fixes https://github.com/CesiumGS/cesium-unreal/issues/491

The original comment there said

   TODO: Not-yet-renderable tiles shouldn't be here.

*If* these tiles should not be in the "tiles to render" vector in the first place, then this TODO could be converted into an issue. 

Further actions then depend on whether it is sensibly possible to avoid this case, or to detect the case that a tile **will** have a glTF and (unexpectedly) does **not** have it (despite the LoadState being checked first). Maybe some check based on the tile content or so...?

However, for now, the warning is turned into a comment, saying under which conditions the "tiles to render" do not really have something to render.


cesium-concierge:

baothientran:
let's merge this one as I don't see anything wrong with it

PR #499, opened 2021-06-22
Load tracing and performance optimization

kring:
The cesium-unreal side of CesiumGS/cesium-native#267.

In addition to adding some tracing for the load process, this PR reduces load time by only generating tangents using MikkTSpace for models that have a normal map. There's also a new `AlwaysIncludeTangents` UProperty on tilesets that allows the user to request tangents even without a normal map, for example if the user needs them for a custom material. Tangent generation was a surprisingly large part of the model load time, especially for photogrammetry models with lots of geometry (like Melbourne), so this is a significant performance improvement.

It also will allow us (in a future PR) to avoid duplicating vertices for each triangle that they're in. THat was done primarily for the sake of tangents, and not doing that will save both memory and time.


cesium-concierge:

baothientran:
@kring removing the mikk tangent (and maybe using indexed triangle as well) is a huge performance boost. I got unreal render the San Diego CDB dataset now, and turning on tangent again makes that dataset frozen like before.

baothientran:
This looks good to me. I need to review cesium-native part and will merge after that

kring:
@baothientran told me offline that he was happy for me to merge this PR once it was updated to use the latest cesium-native. So, merging!

PR #503, opened 2021-06-27
Blueprint function library removals

javagl:
I *think* this has already been merged. There seems to be one commit on top of this branch that does not *explicitly* appear in the target branch, but from comparing the file, the respective change seems to be added there anyhow.


cesium-concierge:

PR #506, opened 2021-06-28
Fix one-frame holes in models.

kring:
Fixes CesiumGS/cesium-native#186.

The one-frame holes happen when we switch between LODs and the timing is just right so that Unreal Engine ends up using previously cached (and incorrect) occlusion information for the new LOD. A new occlusion query sorts it out quickly for the next frame, but not before the user can notice a flash of part of the model missing.

There doesn't appear to be a way to clear the stale occlusion information, so instead I've implemented a bit of a hack. On each Tick, we get a new list of tiles to hide this frame. But we don't actually hide them. Instead, we copy the list to a separate list, and we hide the tiles in that list _next_ frame. So effectively there's a one-frame delay for all tile hiding. That will mean briefly showing two LODs together for one frame, but it isn't noticeable to my eye (certainly much less noticeable than the holes!). We have to be careful to make sure that the tiles hidden last frame aren't meant to be visible this frame, though.

More details of the problem on UDN: https://udn.unrealengine.com/s/question/0D54z0000708yGxCAI/one-frame-delay-after-setvisibilitytrue-true

### DIFF
--- a/Source/CesiumRuntime/Public/Cesium3DTileset.h
+++ b/Source/CesiumRuntime/Public/Cesium3DTileset.h
@@ -584,4 +584,5 @@ private:
   bool _beforeMoviePreloadSiblings;
   int32_t _beforeMovieLoadingDescendantLimit;
   bool _beforeMovieKeepWorldOriginNearCamera;
+  std::vector<Cesium3DTiles::Tile*> _tilesToNoLongerRenderNextFrame;
 };
--- a/Source/CesiumRuntime/Public/Cesium3DTileset.h
+++ b/Source/CesiumRuntime/Public/Cesium3DTileset.h
@@ -584,5 +584,18 @@ private:
   bool _beforeMoviePreloadSiblings;
   int32_t _beforeMovieLoadingDescendantLimit;
   bool _beforeMovieKeepWorldOriginNearCamera;
+
+  // This is used as a workaround for cesium-native#186
+  //
+  // The tiles that are no longer supposed to be rendered in the current
+  // frame, according to ViewUpdateResult::tilesToNoLongerRenderThisFrame,
+  // are kept in this list, and hidden in the NEXT frame, because some
+  // internal occlusion culling information from Unreal might prevent
+  // the tiles that are supposed to be rendered instead from appearing
+  // immediately.
+  //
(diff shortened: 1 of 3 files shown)


cesium-concierge:

Source/CesiumRuntime/Public/Cesium3DTileset.h
```diff
@@ -584,4 +584,5 @@ class CESIUMRUNTIME_API ACesium3DTileset : public AActor,
   bool _beforeMoviePreloadSiblings;
   int32_t _beforeMovieLoadingDescendantLimit;
   bool _beforeMovieKeepWorldOriginNearCamera;
+  std::vector<Cesium3DTiles::Tile*> _tilesToNoLongerRenderNextFrame;
```
javagl, line 587:
```
/**
 * This is used as a workaround for cesium-native#186
 *
 * The tiles that are no longer supposed to be rendered in the current 
 * frame, according to ViewUpdateResult::tilesToNoLongerRenderThisFrame,
 * are kept in this list, and hidden in the NEXT frame, because some
 * internal occlusion culling information from Unreal might prevent
 * the tiles that are supposed to be rendered instead from appearing 
 * immediately. 
 * If we find a way to clear the wrong occlusion information in the
 * Unreal Engine, then this field may be removed, and the 
 * tilesToNoLongerRenderThisFrame may be hidden immediately.
 */
 ```

kring: Good idea, added. I only made it a regular comment rather than a Doxygen one because Doxygen comments are only meant to document the API, right?

javagl: It's a private member, so by default, does not appear in the Doxygen anyhow. I don't see a reason to not use `/* */`, but the difference is not sooo important right now. 

(If we wanted to document private stuff, then it would appear accordingly if it was a `/** comment */` (and it might even be that the number of asterisks is relevant here, as in `/**` vs. `/*`, but am not sure)), but that's not the case now).

nithinp7:
Thanks @kring, this seems to cover up the single-frame holes and the brief existence of two LODs at once isn't noticeable. The changes look straightforward and reasonable to me!

Issue #498, opened 2021-06-21
Spawning two georeferenced actors in the same tick moves one of them to (0, 0, 0)


getinthedamnbox:
I'm creating polylines using two `AActor` classes: `APoint` and `ASegment`. I also have two classes that derive from them: `ACesiumPoint` and `ACesiumSegment`. The only difference between the base and derived classes is that the derived ones have a `UCesiumGeoreferenceComponent`. The goal is to ensure that the polylines work even when the world origin is rebased.

When the user starts drawing a polyline, two `ACesiumPoint`s are spawned at the same location. The first one is supposed to stay there, whereas the second one starts getting its location updated by wherever the user is pointing. Everything works fine if I comment out the `UCesiumGeoreferenceComponent`:

```
//  in CesiumPoint.h
private:
  // UPROPERTY()
  // UCesiumGeoreferenceComponent* _georeferenceComponent;

// in CesiumPoint.cpp
ACesiumPoint::ACesiumPoint() {
  // _georeferenceComponent = CreateDefaultSubobject<UCesiumGeoreferenceComponent>(TEXT("Georeference"));
}
```

![point georeference bug_pc_good](https://user-images.githubusercontent.com/18287035/122776862-7c1bbb00-d279-11eb-82d4-462f067492eb.gif)

![point georeference bug_vr_good](https://user-images.githubusercontent.com/18287035/122777122-bbe2a280-d279-11eb-8310-c0523e4d8ce4.gif)

But if I uncomment those lines, adding the `UCesiumGeoreferenceComponent`, the first point gets moved to `(0, 0, 0)` when the second point is spawned. After that, placing points works normally again. It seems like the problem stems from two points getting spawned on the same tick.

![point georeference bug_pc_bad](https://user-images.githubusercontent.com/18287035/122777045-aa00ff80-d279-11eb-99b8-1a296fb4eb19.gif)

![point georeference bug_vr_bad](https://user-images.githubusercontent.com/18287035/122777143-bf762980-d279-11eb-86ca-17f31f1a8a55.gif)

Note that `KeepWorldOriginNearCamera` is off.

Another clue is that when I use my own tileset (not Cesium World Terrain) and try this in VR, not only does the second point get moved, but the camera gets moved as well.

Possibly related to #422. cc #330

nithinp7:
[...]

Issue #494, opened 2021-06-17
Constructing a `UCesiumGeoreferenceComponent` and then changing `RootComponent` appears to cause crash


getinthedamnbox:
I had a constructor that looked like this:

```
ACesiumTestPawn::ACesiumTestPawn() {
  PrimaryActorTick.bCanEverTick = true;

  _camera = CreateDefaultSubobject<UCameraComponent>(TEXT("Camera"));
  _georeferenceComponent = CreateDefaultSubobject<UCesiumGeoreferenceComponent>(TEXT("Georeference"));
  _floatingPawnMovement = CreateDefaultSubobject<UFloatingPawnMovement>(TEXT("Movement"));

  RootComponent = _camera;
  _georeferenceComponent->SetupAttachment(_camera);

  bUseControllerRotationYaw = true;
  bUseControllerRotationPitch = true;
}
```

I started getting this crash:

```
Unhandled Exception: EXCEPTION_ACCESS_VIOLATION reading address 0x00000000000000d0
UE4Editor_Engine
UE4Editor_CesiumRuntime!UCesiumGeoreferenceComponent::_setTransform() [C:\Repo\point-to-point-measurements\Plugins\cesium-unreal\Source\CesiumRuntime\Private\CesiumGeoreferenceComponent.cpp:399]
UE4Editor_CesiumRuntime!ACesiumGeoreference::UpdateGeoreference() [C:\Repo\point-to-point-measurements\Plugins\cesium-unreal\Source\CesiumRuntime\Private\CesiumGeoreference.cpp:340]
```

After investigating with @nithinp7, it seems like the problem stemmed from constructing the `UCesiumGeoreferenceComponent` and then changing `RootComponent` afterward. Switching the order seems to have fixed the problem.

Strangely, the bug didn't happen if I avoided changing the value I passed in as `FName SubobjectName` in `UObject::CreateDefaultSubobject(FName SubobjectName, bool bTransient)` ([doc](https://docs.unrealengine.com/4.26/en-US/API/Runtime/CoreUObject/UObject/UObject/CreateDefaultSubobject/2/)). I only started getting it when I tried to rename components.

getinthedamnbox:
Got the crash again, even after changing the order so that the `UCesiumGeoreferenceComponent` is constructed at the end of the constructor.

I changed this:

```
_georeferenceComponent = CreateDefaultSubobject<UCesiumGeoreferenceComponent>(TEXT("Pawn Georeference Component"));
```

[...]

PR #518, opened 2021-06-29
Georeference component refactoring / fixes

### LINKED ISSUE #422: Moving the pawn and another georeferenced actor in the same tick causes problems
I have an Unreal application written in C++ (not Blueprints) using Cesium for Unreal. When my application starts, it moves the player pawn to an ECEF location (plus an offset). It also spawns a non-player actor (which has a `CesiumGeoreferenceComponent`) and moves it to that same ECEF location. If these two steps happen on different ticks, everything works fine. But if they happen in the same tick, the non-player actor's location doesn't stay properly rooted to the spot on Cesium World Terrain.

The non-player actor is often (but not always) initially moved to the correct location on CWT, but then when I move the player pawn, the non-player actor moves in the wrong direction. It seems to be attempting to move with the tileset as the world's origin is rebased, but the actor's frame of reference has been rotated incorrectly. This also happens if I move the actor to an ECEF location and then immediately rotate it with `SetActorRotation`.

I need to be able to support pawns that derive from `AGlobeAwareDefaultPawn` as well as pawns that don't derive from it but have a `CesiumGeoreferenceComponent`. I move the globe-aware pawn with `AGlobeAwareDefaultPawn::SetECEFCameraLocation()`, and I move the globe-unaware pawn with `UCesiumGeoreferenceComponent::MoveToECEF()`. I move the non-player actor with `UCesiumGeoreferenceComponent::MoveToECEF()`, too.

As a side note, after I move the globe-unaware pawn, I have to force the Georeference to update immediately after, or else the orientation of the world ends up wrong. I do it like this:

```
glm::dvec3 lonLatHeight = georeference->TransformEcefToLongitudeLatitudeHeight(averageEcef);
georeference->SetGeoreferenceOrigin(lonLatHeight);
```

I'm assuming the problem with moving both actors in the same tick occurs because the world origin is rebased in `Tick()`, after it checks the distance between the camera and the origin. If I don't wait for that tick before moving other actors, I get the bug.

I don't understand the problem well enough to fix it, but I was able to work around it for now.

cc #330

xuelongmu:
Potentially fixes #422, #494, #498
Copying @javagl's notes:
- The `CesiumGeoreferenceComponent` now an ActorComponent, and no longer a SceneComponent. So it does not have a "transform" on its own. (This transform had to be ignored in the old state).
- It's now using the Ellipsoid of the GeoRef instead of the fixed WGS84
- The representation of the "position" is now solely encoded in the ECEF_X/Y/Z properties. Originally, it had at least 3 (but I think maybe even 5 or 6) representations of "the position", and keeping them consistent was hard.

The goal is to merge into `bp-function-library` after verifying a) correct behavior of georeferenced objects, and b) backwards compatibility with georeferenced actors created with the old version of this component.

@nithinp7 mentioned some issues with the sun sky, there is another PR #497 which refactors the `CesiumSunSky`, which will get merged into `bp-function-library`. Probably best to wait to debug SunSky specific issues until this PR and #497 get merged into `bp-function-library`.

### DIFF
--- a/Source/CesiumRuntime/Private/CesiumGeoreferenceComponent.cpp
+++ b/Source/CesiumRuntime/Private/CesiumGeoreferenceComponent.cpp
@@ -70,6 +70,7 @@ void logMatrix(const std::string& name, const glm::dmat4& matrix) {
 } // namespace
 
 UCesiumGeoreferenceComponent::UCesiumGeoreferenceComponent()
+    : _updatingActorTransform(false)
 // NOTE: _actorToECEF is initialized from property
 {
   this->bAutoActivate = true;
@@ -110,26 +111,28 @@ void UCesiumGeoreferenceComponent::SnapLocalUpToEllipsoidNormal() {
 }
 
 void UCesiumGeoreferenceComponent::SnapToEastSouthUp() {
-  /* TODO GEOREF_REFACTORING Snapping is not refactored yet
+  UE_LOG(
+      LogCesium,
+      Verbose,
+      TEXT("Called SnapToEastSouthUp on component %s"),
+      *this->GetName());
+
   if (!IsValid(this->Georeference)) {
     UE_LOG(
         LogCesium,
         Warning,
         TEXT(
-            "CesiumGeoreferenceComponent %s does not have a valid
-  Georeference"), *this->GetName()); return;
+            "CesiumGeoreferenceComponent %s does not have a valid Georeference"),
+        *this->GetName());
+    return;
   }
 
-  const glm::dvec3 ecef = glm::dvec3(this->ECEF_X, this->ECEF_Y, this->ECEF_Z);
-  glm::dmat4 ENUtoECEF = glm::dmat4(
-      this->Georeference->ComputeEastNorthUpToEcef(ecef));
-  ENUtoECEF[3] = glm::dvec4(ecef, 1.0);
+  const glm::dvec3 ecef(this->ECEF_X, this->ECEF_Y, this->ECEF_Z);
+  const glm::dmat3 newActorRotation =
+      this->Georeference->ComputeEastNorthUpToEcef(ecef);
+  const glm::dvec3 relativeLocation = _computeRelativeLocation();
 
-  this->_actorToECEF = ENUtoECEF * CesiumTransforms::scaleToCesium *
-                       CesiumTransforms::unrealToOrFromCesium;
-
-  this->_updateActorTransform();
-  */
+  this->_updateActorTransform(newActorRotation, relativeLocation);
 }
 
 void UCesiumGeoreferenceComponent::MoveToLongitudeLatitudeHeight(
@@ -236,7 +239,8 @@ void UCesiumGeoreferenceComponent::_updateFromActor() {
     return;
   }
   const glm::dvec3 absoluteLocation = _getAbsoluteLocationFromActor();
-  const glm::dvec3 ecef = this->Georeference->TransformUnrealToEcef(absoluteLocation);
+  const glm::dvec3 ecef =
+      this->Georeference->TransformUnrealToEcef(absoluteLocation);
 
   _setECEF(ecef, false); // TODO GEOREF_REFACTORING true or false?
 }
@@ -265,6 +269,32 @@ glm::dvec3 UCesiumGeoreferenceComponent::_getAbsoluteLocationFromActor() {
   const USceneComponent* ownerRoot = owner->GetRootComponent();
   const FVector& relativeLocation = ownerRoot->GetComponentLocation();
   return worldOriginLocation + VecMath::createVector3D(relativeLocation);
+}
+
+glm::dmat3 UCesiumGeoreferenceComponent::_getRotationFromActor() {
+  const UWorld* world = this->GetWorld();
+  if (!IsValid(world)) {
+    UE_LOG(
+        LogCesium,
+        Warning,
+        TEXT("CesiumGeoreferenceComponent %s is not spawned in world"),
+        *this->GetName());
+    return glm::dmat3(1.0);
+  }
+  const AActor* const owner = this->GetOwner();
+  if (!IsValid(owner)) {
+    UE_LOG(
+        LogCesium,
+        Warning,
+        TEXT("CesiumGeoreferenceComponent %s does not have a valid owner"),
+        *this->GetName());
+    return glm::dmat3(1.0);
+  }
+  const USceneComponent* ownerRoot = owner->GetRootComponent();
+  const FMatrix actorTransform =
+      ownerRoot->GetComponentTransform().ToMatrixWithScale();
+  const glm::dmat3 actorRotation(VecMath::createMatrix4D(actorTransform));
+  return actorRotation;
 }
 
 void UCesiumGeoreferenceComponent::OnComponentCreated() {
@@ -330,7 +360,9 @@ void UCesiumGeoreferenceComponent::ApplyWorldOffset(
 
   // Compute the absolute location based on the ECEF
   const glm::dvec3 ecef = glm::dvec3(this->ECEF_X, this->ECEF_Y, this->ECEF_Z);
-  const glm::dvec3 absoluteLocation = this->Georeference->TransformEcefToUnreal(ecef);;
+  const glm::dvec3 absoluteLocation =
+      this->Georeference->TransformEcefToUnreal(ecef);
+  ;
 
   // Apply the offset to compute the new absolute location
   const glm::dvec3 offset = VecMath::createVector3D(InOffset);
@@ -340,7 +372,8 @@ void UCesiumGeoreferenceComponent::ApplyWorldOffset(
 
   // Convert the new absolute location back to ECEF, and apply it to this
   // component
-  const glm::dvec3 newEcef = this->Georeference->TransformUnrealToEcef(newAbsoluteLocation);
+  const glm::dvec3 newEcef =
+      this->Georeference->TransformUnrealToEcef(newAbsoluteLocation);
   _setECEF(newEcef, false); // TODO GEOREF_REFACTORING true or false?
   /*
   //if (this->FixTransformOnOriginRebase) // TODO GEOREF_REFACTORING Figure out
@@ -438,8 +471,41 @@ void UCesiumGeoreferenceComponent::SetAutoSnapToEastSouthUp(bool value) {
   }
 }
 
+glm::dvec3 UCesiumGeoreferenceComponent::_computeRelativeLocation() {
+  const UWorld* const world = this->GetWorld();
+  if (!IsValid(world)) {
+    UE_LOG(
+        LogCesium,
+        Warning,
+        TEXT("CesiumGeoreferenceComponent %s is not spawned in world"),
+        *this->GetName());
+    return glm::dvec3();
+  }
+  const AActor* const owner = this->GetOwner();
+  if (!IsValid(owner)) {
+    UE_LOG(
+        LogCesium,
+        Warning,
+        TEXT("CesiumGeoreferenceComponent %s does not have a valid owner"),
+        *this->GetName());
+    return glm::dvec3();
+  }
+  // Compute the absolute location from the ECEF
+  const glm::dvec3 ecef(ECEF_X, ECEF_Y, ECEF_Z);
+  const glm::dvec3 absoluteLocation =
+      this->Georeference->TransformEcefToUnreal(ecef);
+
+  // Compute the (high-precision) relative location
+  // from the absolute location and the world origin
+  const glm::dvec3 worldOriginLocation =
+      VecMath::createVector3D(world->OriginLocation);
+  const glm::dvec3 relativeLocation = absoluteLocation - worldOriginLocation;
+
+  return relativeLocation;
+}
+
 void UCesiumGeoreferenceComponent::_updateActorTransform() {
-  const UWorld* world = this->GetWorld();
+  const UWorld* const world = this->GetWorld();
   if (!IsValid(world)) {
     UE_LOG(
         LogCesium,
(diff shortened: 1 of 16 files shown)


cesium-concierge:

Source/CesiumRuntime/Private/CesiumGeoreferenceComponent.cpp
```diff
@@ -253,223 +476,164 @@ void UCesiumGeoreferenceComponent::HandleGeoreferenceUpdated() {
...
-  // Update component properties
-  this->_updateDisplayECEF();
+    const glm::dvec3 currentRelativeLocation =
+        _computeRelativeLocation(_currentEcef);
+    const glm::dmat3 currentUnrealToEnu =
+        glm::inverse(this->Georeference->ComputeEastNorthUpToUnreal(
+            currentRelativeLocation));
+    const glm::dmat3 newEnuToUnreal(
+        this->Georeference->ComputeEastNorthUpToUnreal(newRelativeLocation));
+    newActorRotation = newEnuToUnreal * currentUnrealToEnu * oldActorRotation;
```
xuelongmu, line 625: Would it be more consistent to compute a glm::rotation here? Similar to how it's done in line 107

xuelongmu:
@javagl Just did a test with creating a georeference component from the old `bp-function-library` state, migrating to this branch, and opening it. Seems like UE is crashing on startup, with the attached log.
When running in debug mode, it stops executing on the following method in `MulticastDelegateBase.h`
```
	/**
	 * Removes all functions from this multi-cast delegate's invocation list that are bound to the specified UserObject.
	 * Note that the order of the delegates may not be preserved!
	 *
	 * @param InUserObject The object to remove all delegates for.
	 * @return  The number of delegates successfully removed.
	 */
	int32 RemoveAll( const void* InUserObject )
	{
		int32 Result = 0;
		if (InvocationListLockCount > 0) // Stops here
		{
			for (TDelegateBase<UserPolicy>& DelegateBaseRef : InvocationList)
			{
				IDelegateInstance* DelegateInstance = DelegateBaseRef.GetDelegateInstanceProtected();
```

[CesiumDevTest2.log](https://github.com/CesiumGS/cesium-unreal/files/6735434/CesiumDevTest2.log)

Source/CesiumRuntime/Private/CesiumGeoreferenceComponent.cpp
javagl: Intuitively, it sounds reasonable to compute the ellipsoid normals at these points, and use the rotation between these (as computed with `glm::rotation`) for the update. I'm not yet 100% sure whether this will really retain the "full" orientation (meaning that I'm not sure whether it prevents the object from appearing to be rotated around the ellipsoid normal direction), but ... after all, I think this should work, becasue this should be exactly "the orientation change" between both points. I'll try to think this through more thoroughly (or just try it out...).

javagl:
A general note, for the review process: The `ApplyWorldOffset` implementation currently is not implemented properly. (I'm trying to figure out a solution there - the function is called before the origin has been changed, so some computations are simply happening on the old origin where they should already use the new one - I hope that this can be fixed easily, but ... finding an easy and clean solution is the hard thing right now...).

xuelongmu:
> The `ApplyWorldOffset` implementation currently is not implemented properly.

I was actually wondering about this...I think custom logic for `ApplyWorldOffset` should only be needed if `FixTransformOnOriginRebase = true`. IIRC the method is only called at runtime when world origin rebasing occurs, and UE4 by default will shift actors' positions during origin rebasing so that they're in the same consistent location after the new origin has been set. `FixTransformOnOriginRebase = true` takes it a step further by assuming that UE4's default handling is inaccurate due to floating point errors, and "corrects" it with the double precision ECEF position.

xuelongmu:
Another startup crash when migrating from the old georeference component implementation to the new:
[CesiumDevTest2.log](https://github.com/CesiumGS/cesium-unreal/files/6735536/CesiumDevTest2.log)

```
Unhandled Exception: EXCEPTION_ACCESS_VIOLATION reading address 0x0000000000000540

UE4Editor_CesiumRuntime!glm::operator*<double,0>() [C:\Users\Xuelong\Dev\CesiumDevTest2\Plugins\cesium-unreal\Source\ThirdParty\include\glm\detail\type_mat4x4.inl:568]
UE4Editor_CesiumRuntime!UCesiumGeoreferenceComponent::OnComponentCreated() [C:\Users\Xuelong\Dev\CesiumDevTest2\Plugins\cesium-unreal\Source\CesiumRuntime\Private\CesiumGeoreferenceComponent.cpp:322]
UE4Editor_Engine!USCS_Node::ExecuteNodeOnActor() [D:\Build\++UE4\Sync\Engine\Source\Runtime\Engine\Private\SCS_Node.cpp:116]
UE4Editor_Engine!USimpleConstructionScript::ExecuteScriptOnActor() [D:\Build\++UE4\Sync\Engine\Source\Runtime\Engine\Private\SimpleConstructionScript.cpp:645]
UE4Editor_Engine!AActor::ExecuteConstruction() [D:\Build\++UE4\Sync\Engine\Source\Runtime\Engine\Private\ActorConstruction.cpp:765]
```

javagl:
I was also wondering about this, as indicated by the commented-out
```
  /*
  // TODO GEOREF_REFACTORING Figure out how to handle this...
  if (this->FixTransformOnOriginRebase)
  {
    this->_updateActorTransform();
  }
  */
```
...

But I just started working on that: I it should be possible to *not* call `_setECEF` there at all, because the rebasing should not affect the ECEF, and thus, should not affect the core state of the component (which now is only the ECEF). But I assume that when `FixTransformOnOriginRebase==true`, the actor position has to be updated with a new "relativePosition" that is computed from the current ECEF and the _new_ world origin.

javagl:
The new crash looks like a `Georeference==null` case, but ... it should be initialized directly before that. I'll try to have a look later (focussing on the world offset right now - I have another task, but should be able to figure out a solution a bit later today).

javagl:
And sometimes, a small "improvement" may lead to hours of debugging. https://github.com/CesiumGS/cesium-unreal/pull/518/commits/7ccd6fa1a31aa1e877828a71b716433a5fdd2562 was not a good idea. The matrices are in absolute space. The conversion functions are in relative space. The fact that

    const glm::dvec3 absoluteLocation =  this->Georeference->TransformEcefToUnreal(ecef);

is "wrong" but

    const glm::dmat4& ecefToUnreal = this->Georeference->GetEllipsoidCenteredToUnrealWorldTransform();
    const glm::dvec3 absoluteLocation = ecefToUnreal * ecef;

is "right" is ... subtle. 

Sure, I should have had that on the radar, I'm stupid, and all this is purely my fault. But... we *reeeally* should get these method names sorted out somehow.

I'm currently trying to clean that up, along with some comments like

``` 
  // Do NOT use TransformEcefToUnreal, because it will return the 
  // position relative to the current world origin!
```
to prevent future maintainers from making the same mistake again.

javagl:
I hope that the wrong change that I mentioned above is properly un-done now, and origin rebasing *seems* to work according to my basic test: Loading a level that was created in `main` and hitting "Start" causes the GeoRef'ed Actor to remain at the right position, also when moving around and the unreal-internal origin rebasing messages appear in the console.

I'll do further tests ASAP, try to reproduce the startup crashes, and see whether the "maintain orientation" thing can be done with a `glm::rotation` (maybe later today, after another task, or early tomorrw).

Source/CesiumRuntime/Private/CesiumGeoreferenceComponent.cpp
javagl: Thanks, done in https://github.com/CesiumGS/cesium-unreal/pull/518/commits/d5845b4f94ee4a27e83c54c8b8c244dc320b0066 . It indeed looks much cleaner and simpler (and seems to work well, although I "only" tried it out, though, and didn't do the maths...)

Unfortunately, I'm a bit held up with some other tasks that came in unexpectedly, but will try to focus on the crashes next. I'm regularly testing this with a level that I created in the `main` branch, though, and didn't see crashes (that hadn't been fixed in the meantime) until now. If you have any steps to reproduce them, that would be appreciated, otherwise, I'll try to "provoke" them in some way, maybe based on what I can derive from the traces...

javagl:
Based on some discussion, we removed the "high-precision" version functions from the `CesiumGeoreference` in https://github.com/CesiumGS/cesium-unreal/pull/518/commits/62cc0112873da4bbab976626a0039c5e4c76d3bc . 

A quick summary: Before this, most of the functions of the georeference had been offered in two flavors:

1. one function that operated on `glm` data structures with double precision
2. one function that operated on the Unreal data structures (like `FVector`), with single precision.

The second one was mainly offered so that the functions could be used from Blueprints, being declared as `BlueprintCallable`. Generally, these functions had the same name, except for the second one being prefixed with `Inaccurate` - for example:

  glm::dvec3 TransformLongitudeLatitudeHeightToEcef(
      const glm::dvec3& LongitudeLatitudeHeight) const;

  FVector InaccurateTransformLongitudeLatitudeHeightToEcef(
      const FVector& LongitudeLatitudeHeight) const;

The actual implementations of the high-precision functions had already been moved to a dedicated class (currently called `GeoTransforms` -  the reasoning for this is explained in https://github.com/CesiumGS/cesium-unreal/issues/330#issuecomment-859057502 ). So the implementations of these functions had usually been in the pattern of

  glm::dvec3 TransformLongitudeLatitudeHeightToEcef(const glm::dvec3& LongitudeLatitudeHeight) const {
    return _geoTransforms.TransformLongitudeLatitudeHeightToEcef(LongitudeLatitudeHeight);
  }

These functions have been removed now. Somebody who wants to use these high-precision functions via the C++ API can obtain the `GeoTransforms` instance from the georeference, and call the function directly.

javagl:
There may be one exception for the removal of the high-precision functions: Some of the functions use the current world origin. Originally, the world origin had been obtained from the GeoRef internally, causing the call to look like

    glm::dvec3 ecef = georeference->TransformUeToEcef(relativeUnreal);

But now, the same computation requires something like

    const FIntVector& worldOrigin = GetWorld()->OriginLocation;
    const glm::dvec3 origin = glm::dvec3(worldOrigin.X, worldOrigin.Y, worldOrigin.Z);
    glm::dvec3 ecef = georeference->getGeoTransforms().TransformUnrealToEcef(origin, relativeUnreal);

which is clumsy (also considering that the `GetWorld` would have to be wrapped into `IsValid` checks...) 

As mentioned [above](https://github.com/CesiumGS/cesium-unreal/pull/518#issuecomment-870866609) : The fact that `TransformEcefToUnreal` does something completely different than the matrix that is obtained with `GetEllipsoidCenteredToUnrealWorldTransform` already caused confusion, and will likely cause confusion in the future, so one could argue about this function in the first place. But I'd at least like to avoid the clumsiness that comes from the world origin part, and offer these functions, maybe renamed to emphasize the difference a bit:

    glm::dvec3 ecef = georeference->TransformRelativeUnrealToEcef(relativeUnreal);
    //                                       ^ Replace "Ue" with "RelativeUnreal"

Yes, it's verbose, but when there is a (buggy) line like

    glm::dvec3 ecef = georeference->TransformRelativeUnrealToEcef(absoluteLocation);

then the terms "relative" and "absolute" appearing in that line might be a valuable hint that something is wrong there...

javagl:
After some hesitation, I restored several (most of) the high-precision functions that had previously been removed, for the reasons mentioned in the last comment.

Some functions require the world origin and others don't. Some transforms use the absolute Unreal coordiantes and some functions use the relative ones. Sometimes, the conversion between relative and absolute coordinates is done by passing in the relative position and the origin, and doing the `rel+origin` part (or a `abs-origin`) at the very end. It's probably not possible to make this totally consistent, but maybe it's possible to somehow limit the confusion that this will cause at the user side.

javagl:
To keep track of migration steps that might be necessary when the refactoring is merged:

- Some functions have essentially just been renamed, to replace the `Ue` part with `Unreal`. This refers to functions like `TransformUeToEcef`  and `TransformEcefToUe` which are now called `TransformUnrealToEcef` and `TransformEcefToUnreal`, respectively. 
  (I'm still considering to extend that to `RelativeUnreal` before the merge, to make the crucial distinction between "absolute Unreal world" and "relative Unreal world" clearer...)
 
- The function 

      UCesiumGeoreferenceComponent::MoveToECEF(
        const glm::dvec3& targetEcef,
        bool maintainRelativeOrientation)

  receives a `maintainRelativeOrientation` parameter (and sometimes this is passed in from other functions, so this will affect the calling functions as well, and *which* ones these are still has to be listed somewhere). This parameter has not been properly taken into account in all cases. (The exact _conditions_ under which this happened _could_ be analyzed further, but may not be relevant). Some people seem to have anticipated this wrong behavior, and called things like

      // Move the pawn.
      georeferenceComponent->MoveToECEF(startEcef);

  with `maintainRelativeOrientation` defaulting to `true`, **but** then **expected** the relative orientation to **not** be maintained. (And "expected" indeed refers to the level of unit tests here...). These call sites will have to call

      // Move the pawn.
      georeferenceComponent->MoveToECEF(startEcef, false);

  _explicitly_ in the future.

xuelongmu:
Thanks for the valuable insights here @javagl. I agree with the call you made to put high precision functions back into CesiumGeoreference, keeping the user-facing API clean and easy to use would be the priority.

> Some callers may have used the functions that have been removed together with the CesiumGeoreferenceable interface

Given that the bounding volume is used rarely, I think that the refactored code is still OK despite its clumsiness.

Source/CesiumRuntime/Private/CesiumGeoreferenceComponent.cpp
```diff
@@ -135,90 +138,293 @@ void UCesiumGeoreferenceComponent::InaccurateMoveToECEF(
...
+      *this->GetName());
+
+  // If the Georeference is modified, detach the `HandleGeoreferenceUpdated`
+  // callback from the current instance
+  FName propertyName = PropertyThatWillChange->GetFName();
+  if (propertyName ==
+      GET_MEMBER_NAME_CHECKED(UCesiumGeoreferenceComponent, Georeference)) {
+    if (IsValid(this->Georeference)) {
+      this->Georeference->OnGeoreferenceUpdated.RemoveAll(this);
+      _updateActorTransform();
```
xuelongmu, line 413: Assuming that `this->Georeference` is pointing to the old Georeference pre-change, should we wait to call `_updateActorTransform` until `PostEditChangeProperty`?

Source/CesiumRuntime/Private/CesiumGeoreferenceComponent.cpp
```diff
@@ -135,90 +138,293 @@ void UCesiumGeoreferenceComponent::InaccurateMoveToECEF(
...
+
+  // Compute the position that the world origin will have
+  // after the rebase, indeed by SUBTRACTING the offset
+  const glm::dvec3 oldWorldOriginLocation =
+      VecMath::createVector3D(world->OriginLocation);
+  const glm::dvec3 offset = VecMath::createVector3D(InOffset);
+  const glm::dvec3 newWorldOriginLocation = oldWorldOriginLocation - offset;
+
+  // Compute the absolute location based on the ECEF
+  // Do NOT use TransformEcefToUnreal, because it will return the
```
xuelongmu, line 379: I would be in favor of renaming `TransformEcefToUnreal` to `TransformEcefToUnrealRelative`, `GetEllipsoidCenteredToUnrealWorldTransform` to `GetEllipsoidCenteredToUnrealAbsoluteTransform` (the meaning of `World` is ambiguous), and so forth.

xuelongmu:
Great work @javagl! I only had a minor suggestion, otherwise this is ready to merge.

You can hold off on reviewing the `cpp-sun-sky` branch until this PR is merged, as there are a couple improvements I want to make there based off the `CesiumSunSky` in this branch.

Source/CesiumRuntime/Private/CesiumGeoreferenceComponent.cpp
javagl: The intention behind calling `_updateActorTransform` was to ~"have an up-to-date state, _even if_ the Georeference is set to `None`/`null`". In an ideal world, this should not be necessary, because the state should be ~"up-to-date _at any point in time, anyhow_", but ... some of the Unreal lifecycle method calls are ... surprising and unexpected. (E.g. it does call `OnUnregister` on the component when editing a property - why?!?!). 

It might be possible to remove that `_updateActorTransform` call there, but I'd dedicatedly check whether it might lead to an inconsistent state when setting it to `None`. (Maybe even cover that with a unit test later...)

xuelongmu: @javagl That makes sense, thanks for clarifying. Perhaps it may be possible to remove, but this can be a question for later. For now, a simple comment on that line would be great.

javagl: I have added a comment accordingly. 

The "broader" question (that we occasionally talked about, but where a final, definite answer is difficult) is: Which class should be able to cope with a `None`/`null` Georeference, and _how exactly should it behave in this case_?

Right now, the GeoRefComponent does lots of `IsValid` checks, and usually tries to handle the `else`-case with "default values" (like identity matrices). But I'd hesitate to establish an strong guarantee along the lines of "This works and makes sense in all cases".

xuelongmu:
Thanks @javagl!

PR #520, opened 2021-06-29
Added formatting documentation

xuelongmu:
Feel free to suggest a different place to put this documentation, but I just added in a note on how developers can run clang on their code before checking it in. I didn't know the exact commands until last week, so I'm sure this will be helpful for future contributors. 

As something else to think about, Matt A suggested running `npm run format` on commit, as is done in the Cesium JS repos.


cesium-concierge:

kring:
Thanks @xuelongmu! We have an issue to add automatic formatting on commit:
https://github.com/CesiumGS/cesium-native/issues/265
(and for folks within Cesium there's a link there to an internal project that we can steal it from)

PR #525, opened 2021-06-30
Generate tangents for water tiles

### LINKED ISSUE #524: If the mesh doesn't include tangents, the water/wave effect doesn't work
After #499, we only generate tangents for glTFs with a normal map. But that means that the water effect now looks shiny but not wavy:
image

The wave effect is implemented as a normal manipulation via a custom material (i.e. the glTF itself doesn't have a normal map), so this isn't too surprising, but I hadn't realized it until now.

There are two possible solutions that I know of:

1. Automatically enable tangent generation when the water mask is enabled.
2. Set `M_CesiumOverlayWater` to use world-space normals instead of tangent space:

image

The trouble with option 2 is that it makes the "normal" pin on the material output node be in world space rather than tangent space, which in turn makes the "no normal map" texture that we use when we don't have a real normal map be interpreted in world space. So that breaks lighting. I think if we could just not output to the normal pin at all when there's no normal map we'd be fine, but I don't know of a way to do that (other than by adding separate materials for with/without a normal map).

Long story short, I think we just need to enable tangent generation when the water effect is enabled.

kring:
Fixes #524 

Note that the skip-the-tangents optimization in #499 are still a significant win even though we're rolling them back a bit here. We still don't need to generate tangents for: photogrammetry models (which have much more complicated geometry than terrain), and terrain tiles without any water.

### DIFF
--- a/Source/CesiumRuntime/Private/CesiumGltfComponent.cpp
+++ b/Source/CesiumRuntime/Private/CesiumGltfComponent.cpp
@@ -796,6 +796,43 @@ static void loadPrimitive(
     }
   }
 
+  // Initialize water mask if needed.
+  auto onlyWaterIt = primitive.extras.find("OnlyWater");
+  auto onlyLandIt = primitive.extras.find("OnlyLand");
+  if (onlyWaterIt != primitive.extras.end() && onlyWaterIt->second.isBool() &&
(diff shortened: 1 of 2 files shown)


cesium-concierge:

nithinp7:
This fixes the issue, thanks @kring!

PR #526, opened 2021-06-30
Get camera parameters using GetCameraViewPoint and GetFOVAngle instead of ViewTarget

### LINKED ISSUE #487: Utilizing "Set view target with blend" function does not update view frustrum
When utilizing the "Set view target with blend" [function](https://docs.unrealengine.com/4.26/en-US/BlueprintAPI/Game/Player/SetViewTargetwithBlend/) in Unreal engine to blend my camera's position, Cesium does not take into account the view frustum changing while the camera is interpolating.

The camera does update when I possess a new pawn, but I am using "set view target with blend" to interpolate the player's view to the new pawn before I possess that pawn.  Presumably this means cesium is not tracking the low-level player view, but instead the high-level active pawn's camera, for the active view?

- using latest release, Windows & Unreal 4.26, RTX 2080 Ti, drivers up to date

kring:
Fixes #487 
~(maybe)~ (the original reporter has confirmed it fixes the issue)

### DIFF
--- a/Source/CesiumRuntime/Private/Cesium3DTileset.cpp
+++ b/Source/CesiumRuntime/Private/Cesium3DTileset.cpp
@@ -817,10 +817,10 @@ ACesium3DTileset::GetPlayerCamera() const {
     return std::optional<UnrealCameraParameters>();
   }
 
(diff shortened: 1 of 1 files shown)


cesium-concierge:

PR #533, opened 2021-06-30
Small fixes to the quick-add panel 

### LINKED ISSUE #532: Adding CesiumSunSky through Quick Add Panel does not update longitude and latitude of CesiumSunSky
As of https://github.com/CesiumGS/cesium-unreal/pull/482 being merged, you can now add CesiumSunSky through the quick add panel, and it's automatically set on the georeference. However, this doesn't automatically update the longitude and latitude of CesiumSunSky, which means the lighting is no longer geospatially accurate. 

If the georeference is moved at all, this is fixed, but we shouldn't rely on the user moving the georeference after adding CesiumSunSky. We should make sure that CesiumSunSky's coordinates are set when adding it using the quick add panel.

It would be great if this fix could be squeezed in before the next release, @kring or @javagl

nithinp7:
fixes #532 

The SunSky was not initializing to the correct georeference when spawned through the quick add panel. I'm sure the SunSky handling will be improved drastically once the georeference and sunsky refactoring are done, but here is a fix for now.

@argallegos Can you review?

### DIFF
--- a/Source/CesiumEditor/Private/CesiumPanel.cpp
+++ b/Source/CesiumEditor/Private/CesiumPanel.cpp
@@ -90,7 +90,7 @@ TSharedRef<SWidget> CesiumPanel::LoginPanel() {
 TSharedRef<SWidget> CesiumPanel::MainIonQuickAddPanel() {
   TSharedPtr<IonQuickAddPanel> quickAddPanel =
       SNew(IonQuickAddPanel)
-          .Title(FText::FromString("Add Cesium ion Assets"))
+          .Title(FText::FromString("Quick Add Cesium ion Assets"))
           .Visibility_Lambda([]() {
             return isSignedIn() ? EVisibility::Visible : EVisibility::Collapsed;
           });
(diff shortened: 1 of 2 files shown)


cesium-concierge:

* :grey_exclamation: There was an error checking the CLA! If this is your first contribution, please send in a [Contributor License Agreement](https://github.com/CesiumGS/cesium/blob/master/CONTRIBUTING.md#contributor-license-agreement-cla).
    * Maintainers, this was the error I ran into while attempting to process the CLA check. Please resolve it to continue CLA checking.
    * You'll need to [manually check for a submitted CLA](https://github.com/CesiumGS/cesium/blob/master/Documentation/Contributors/CLAs/README.md)
    ```
    Error: The service is currently unavailable.
    ```

argallegos (approved):
Thanks for the fix, @nithinp7! I also added a very quick update to CesiumPanel.cpp - I changed "Add Cesium ion Assets" to "Quick Add Cesium ion Assets", and similar for the Basic Actors section. This is largely because the learning content frequently refers to this panel as the "Quick add panel", and I'd like to keep it consistent.

kring:
Thanks guys, I'll include this in today's release.

PR #537, opened 2021-07-02
Include Samples instructions in release process

argallegos:
Rough steps for updating Cesium for Unreal Samples.

I've never done the marketplace side of the release, so if there are specific instructions for that, someone please add them.


cesium-concierge:

kring:
Thanks @argallegos! Do you know how we create the ZIP file that we submit to the Marketplace? I think there's a "Zip Up Project" or similar on the File menu - is that it?

shehzan10:
> Thanks @argallegos! Do you know how we create the ZIP file that we submit to the Marketplace? I think there's a "Zip Up Project" or similar on the File menu - is that it?

@kring Yes, thats it, and submit that zip file to the marketplace.

argallegos:
Just added a quick addition to the Samples update process. It details how to test the 4th sample level, which is more prone to hidden errors due to the nature of sublevels.

PR #543, opened 2021-07-09
Use C++14 instead of 17 for CesiumRuntime and CesiumEditor.

kring:
Requires CesiumGS/cesium-native#280

Currently this PR uses C++14 for all platforms. I _may_ change back to C++17 for Windows, macOS, and Linux and only use C++14 for Android (and eventually iOS), but that's still TBD.


cesium-concierge:

kring:
We no longer need this because Epic accepted #544.

PR #544, opened 2021-07-09
Sneakily force C++17 mode in Android builds

### LINKED ISSUE #472: Unreal Marketplace Testing Failure for Android for v1.3.0
For v1.3.0, we added Android to the whitelist platform in the uplugin file. This is causing them to run into C++17 compilation issues similar to https://github.com/CesiumGS/cesium-unreal/issues/190#issuecomment-831522727.

[Cesium Errors.txt](https://github.com/CesiumGS/cesium-unreal/files/6579898/Cesium.Errors.txt)

If we don't need Android support through marketplace for this release, they are recommending that we submit the plugin by removing `Android` from the uplugin Supported and Target platforms, and instead move it to `BlacklistPlatforms` for this release.

Regarding the errors: As mentioned in the issue comment above, I did see them when trying to build the plugin for Android at one point, but they essentially disappeared one day. I'm still not sure why they don't show up on CI. Those errors also popped up when I was trying to compile Android with UE5 EA over the weekend.

CC @CesiumGS/unreal

kring:
If you all loved my hacky approach to compiling cesium-native with C++14, wait til ya get a load of this! :laughing: 

So the problem is that, prior to UE 4.26.2, the Unreal Build Tool ignores the CppStandard option on the plugin and instead just always passes `-std=c++14` to clang. Well, after tracing through the `AndroidToolChain.cs` file in the UE source where the Android build process is implemented, I realized that there's an `AdditionalArguments` property passed into there on a `CompileEnvironment` object. And I also learned that clang, conveniently, doesn't care if you pass multiple -std=c++X arguments and it just uses the last one. So if I could just figure out how to set that `AdditionalArguments` property, we could force clang to compile in C++17 mode even though UE is trying to specify C++14.

Well, there might be control of that at the project level, but we don't control the project when Epic is building our plugin for the Marketplace (or when users are building our plugin, for that matter). But our plugin's .build.cs file receives a `ReadOnlyTargetRules` instance, and that instance has a `AdditionalCompilerArguments` property. Perfect! Except it's read-only. Hmmm...

Well, the `ReadOnlyTargetRules` has an `Inner` field which is the non-read-only `TargetRules` instance it wraps. And `TargetRules` has a read/write version of `AdditionalCompilerArguments`. Small problem: the Inner property is protected.

No problem, I'll just bust out my C# hacking skills (😆) and use reflection to read the protected property. And once I have that, I can add the `-std=c++17` to the `AdditionalCompilerArguments`. All from CesiumRuntime.Build.cs.

This PR is the result. And I think it works! I don't have a copy of UE 4.26.0 or .1 (I wrote https://udn.unrealengine.com/s/question/0D54z000073a47JCAQ/where-to-get-4260-or-4261 to try to get it), which I need in order to _really_ test this, but I mostly tested it by setting CppStandard to 14, and then overriding it with my cheeky hack. And it compiles with C++17 as expected!

Ok, who wants to be first in line to tell me this is a horrible idea?

Fixes #472

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,13 +12,8 @@ jobs:
     - npm install
     - npm run format -- --dry-run -Werror
   - stage: Build OS Packages
-    name: Windows + VS2017
+    name: Windows
     os: windows
-    cache:
-      timeout: 1000
-      directories:
-      - /C/Program\ Files/Epic\ Games/UE_4.26
-      - /C/Program\ Files\ \(x86\)/Epic\ Games/Launcher
     install:
     - choco install 7zip.portable
     - choco install python --version 3.9.2
@@ -51,8 +46,6 @@ jobs:
     cache:
       timeout: 1000
       directories:
-      - /C/Program\ Files/Epic\ Games/UE_4.26
-      - /C/Program\ Files\ \(x86\)/Epic\ Games/Launcher
       - $HOME/android-ndk-r21e
     install:
     - choco install 7zip.portable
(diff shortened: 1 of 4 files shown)


cesium-concierge:

kring:
I confirmed this works with 4.26.0, too. I'm working on getting Travis set up to build with this version.

Also, I've mentioned this approach on UDN and the Epic support person there didn't sound too horrified:
https://udn.unrealengine.com/s/question/0D54z000073a47JCAQ/where-to-get-4260-or-4261

kring:
Travis is using 4.26.0 now, and it's passing. So this is ready to review and merge. And then we'll submit it to the Marketplace and 🤞.

nithinp7:
Thanks @kring! Since this works with 4.26.0 on Travis, it looks ready.

PR #545, opened 2021-07-11
Add blueprint library to retrieve Feature metadata

baothientran:
This PR accompanies https://github.com/CesiumGS/cesium-native/pull/281 by creating blueprint library to retrieve metadata


cesium-concierge:

kring:
This is looking good! I was able to put together a Blueprint function to print out the name, type, and value of a property:
image

And call it like this from the level blueprint:
image

(I realize I could use `GetMetadataValuesAsStringForFace` to save some nodes, but I was trying to test out the more sophisticated path)

I noticed a bug where Int64 properties were always zero, and fix that in a commit above (https://github.com/CesiumGS/cesium-unreal/pull/545/commits/115db9042c337626855742ba8e2291ee17628abc). I think there may be other similar issues lurking, but I haven't looked closely yet.

But I think more generally, the conversion system could be better. Rather than "you better request the right type or your program will die", I think all the `GetX` blueprint functions should do a best-effort conversion to the requested type. They should also take a `Default` parameter to specify the default value to return if the conversion cannot be performed.

I'll try implementing this later today if I get a chance.

kring:
Here are some Blueprints I'm using to test this. Screenshots are a terrible way to share blueprints, but I honestly don't know a better way to share level Blueprints 😬.

image
image

image

kring:
I found a site that does an impressive job of making Blueprints shareable:
https://blueprintue.com/blueprint/wsbqxvkr/
https://blueprintue.com/blueprint/3r8hmgdo/

You just select the nodes you're interested in sharing in the UE editor, Ctrl-C, then go to https://blueprintue.com/, and paste it in. And if you visit the links above, you can paste my Blueprint back into the editor too. And it even works with custom nodes added by our plugin (for example). Truly magical.

kring:
This should be ready. @baothientran can you please give it a final look and merge it if you're happy?

kring:
Just to summarize the changes:
* Separated out a lot of the functionality into separate source files.
* The GetX functions are now strictly the Blueprints types: Boolean, Byte, Integer, Integer64, Float, String, Array. And they all do a "best effort" conversion of the actual value to the requested type. No more crashing if you request the wrong type.
* There's a `GetTrueType` method that tells your Blueprint the actual underlying type (e.g. Double, Uint64).
* There's a `GetBlueprintType` method that returns the lossless, "best fit" type that is supported by Blueprints for that property (e.g. if the underlying type is Double, the only lossless option in Blueprints is String).
* Deleted the redundant code (and especially comments) on the blueprint struct versus function library. All the methods are on the function libraries now.
* Added a way to get a feature ID from a (primitive, face index, feature table).

baothientran (commented):
Thanks @kring! This looks really good! I only have minor comments below. After that, I think we can merge this

kring:
Ok I've removed the unnecessary monostates. Should be ready for another look.

baothientran:
thanks @kring!

Issue #224, opened 2021-03-14
Support rendering a tileset into multiple viewports


javagl:
When using multiple viewports (via `[v]`->"Layouts" -> "Two Panes", for example), and selecting the "Lit" rendering mode in *both* of them, then the rendered tiles are always governed by only *one* of them.

For example, this is rendered when selecting the *right* panel:

![CesiumUnrealMultiView01](https://user-images.githubusercontent.com/5597569/111078530-2c00c000-84f6-11eb-9cde-549e068a2e99.png)

And this is selected when rendering the *left* panel:

![CesiumUnrealMultiView02](https://user-images.githubusercontent.com/5597569/111078541-3f139000-84f6-11eb-8d51-78421ba41427.png)

**Very** roughly speaking, the reason for that is that some tiles are not visible via the "active" viewport (i.e. the top-down one) are deemed to be "not visible", and apparently then also not rendered in the perspective view. But the details of how this can be avoided have to be investigated further.

---

Edit: For the editor mode, this *might* not be critical. **If** this also affects the case where multiple cameras are used for rendering into multiple viewports in the "game mode", it might have a slightly higher priority.

kring:
Broadly speaking, there are two ways to handle this:
1. Run tile selection once, but use all the viewports for culling and SSE computations. i.e. if it's inside the view frustum in any viewport, it's not culled. And the SSE for a tile is the smallest SSE computed from any viewpoint. The main downside to this is that tiles selected for one viewport might have too much detail for the other, causing aliasing artifacts. But it's probably pretty minor.
2. Run tile selection multiple times, and render different tiles in each viewport. They should share a tile cache, of course. I'm not sure if this is possible in UE.

javagl:
My gut feeling is that 2. is more appropriate here

- Different viewports may have literally *nothing* in common (neither size/SSE nor camera orientation)
- It's already hard to get the traversal right (and "robust") for *one* viewport. If we already had something like

      visibilityDetectors.push_back(new VisibilityDetectorFor(viewportA));
      // visibilityDetectors.push_back(new VisibilityDetectorFor(viewportB)); // Add it or leave it...
   
   then this would be a different story. But right now, we should be **very** wary of adding *any* additional complexity here.

kring:
I don't _think_ (1) should be that scary. It impacts:

* `isVisibleFromCamera`: instead of checking one ViewState, it checks several, and returns true if any of them report it visible.
[...]

PR #547, opened 2021-07-13
Support multiple frustums

nithinp7:
Collects and sends all editor/play cameras to the 3D Tiles refinement engine.

Builds upon the changes here: https://github.com/CesiumGS/cesium-native/pull/283

Fixes: #224 

MultiFrustums  DebugGame  - Unreal Editor 2021-07-13 17-24-36 (1)


cesium-concierge:

CesiumBen:
@nithinp7 Awesome work and demo!

 I've tested these updates with VR and motion and am still getting some tiles unloading, but strangely enough only the ones which are some distance from the origin. The gif below displays it as best as possible, the 1st half of the gif I'm far from the starting position and disappearing tiles occurs more frequently. While I return to the origin point (When I'm grabbing with the hands) the culling is much better and the tiles remain as they should.

Any configuration I should be aware of when testing?

clipping

Source/CesiumRuntime/Private/Cesium3DTileset.cpp
```diff
@@ -782,102 +783,187 @@ void ACesium3DTileset::MarkTilesetDirty() {
...
+      rotation = FRotator(0.0f, 0.0f, 0.0f);
+      break;
+    case LVT_OrthoNegativeYZ:
+      rotation = FRotator(0.0f, 180.0f, 0.0f);
+      break;
+    case LVT_OrthoFreelook:
+    case LVT_Perspective:
+    case LVT_MAX:
+    case LVT_None:
+      rotation = pEditorViewportClient->GetViewRotation();
```
javagl, line 942: Is there a reason why these cases are handled individually? 

E.g. when the viewport type is `LVT_OrthoXZ`, I'd expect `GetViewRotation` to exactly return `FRotator(0.0f, -90.0f, 0.0f)`. Or will `GetViewRotation` return different values than the hard-codes ones?

Source/CesiumRuntime/Public/Cesium3DTileset.h
```diff
@@ -521,21 +522,23 @@ class CESIUMRUNTIME_API ACesium3DTileset : public AActor,
...
   void MarkTilesetDirty();
-  Cesium3DTiles::ViewState CreateViewStateFromViewParameters(
-      const FVector2D& viewportSize,
-      const FVector& location,
-      const FRotator& rotation,
-      double fieldOfViewDegrees) const;
 
   struct UnrealCameraParameters {
+    bool isPerspective;
+    bool isOrthographic;
```
javagl, line 528: Can `isPerspective` and `isOrthographic` ever have the same value? The frustum type is derived from that with 
```
  if (camera.isOrthographic) {
    frustumType = Cesium3DTiles::ViewState::FrustumType::Orthographic;
  } else {
    frustumType = Cesium3DTiles::ViewState::FrustumType::Perspective;
  }
```
and the method where they are obtained is implemented as
```
bool FEditorViewportClient::IsOrtho() const
{
	return !IsPerspective();
}
```
so I think these fields might be redundant...

Source/CesiumRuntime/Private/Cesium3DTileset.cpp
```diff
@@ -782,102 +783,187 @@ void ACesium3DTileset::MarkTilesetDirty() {
...
+    pCameraManager->GetCameraViewPoint(location, rotation);
+    double fov = pCameraManager->GetFOVAngle();
+
+    FVector2D size;
+    pViewport->GetViewportSize(size);
+
+    if (size.X < 1.0 || size.Y < 1.0) {
+      continue;
+    }
+
```
javagl, line 845: There are several `if (....) continue;` blocks above. 
Are these anticipated cases? 
If so, a comment like `// This happens when (PIE, loading a level, switching views, ... whatever)` could be helpful. 
If not, may they be worth a warning like `LOG(Warning, "Whoopsie, why is there no CameraManager"`)`?

javagl:
I dropped some small nitpicks inline, and will have a closer look at the `cesium-native` side ASAP.

Source/CesiumRuntime/Private/Cesium3DTileset.cpp
nithinp7: I initially used GetViewRotation, but it seemed to return FRotator(0, 0, 0). At least when testing with orthographic editor viewports.

CesiumBen:
> @nithinp7 Awesome work and demo!
> 
> I've tested these updates with VR and motion and am still getting some tiles unloading.....

moved to [issue](https://github.com/CesiumGS/cesium-unreal/issues/559)

javagl:
Does this have to be updated for https://github.com/CesiumGS/cesium-native/commit/cc9656230ac26892ac700232bdee87937fa9aca3 ?

javagl:
I tried this out with the latest state and the matching `cesium-native` state. Of course, it is hard to "verify" any form of "correctness" in terms of the selected tiles. But the "Dynamic Pawn" preview window shows the full tileset, even when the main window shows a part that the pawn does *not* look at, which is the most apparent difference here. The things that I tried out also worked well, including the creation of multiple pawns and selecting them (showing multiple preview windows accordingly).

nithinp7:
Hey @CesiumBen, thanks for trying this PR out. I did some more digging and found that Unreal doesn't seem to use two separate cameras for VR as I had hoped, instead there is dedicated stereographic rendering logic. In the last commit I just pushed, I attempted to use the stereographic information to determine the appropriate viewports for the two eyes in VR. When you get time, please let me know if it works, since I currently cannot test it out with VR.

Source/CesiumRuntime/Private/Cesium3DTileset.cpp
```diff
@@ -782,102 +785,164 @@ void ACesium3DTileset::MarkTilesetDirty() {
...
+        static_cast<FEditorViewportClient*>(pViewportClient);
+
+    if (!pEditorViewportClient) {
+      continue;
+    }
+
+    const FVector& location = pEditorViewportClient->GetViewLocation();
+    const FRotator& rotation = pEditorViewportClient->GetViewRotation();
+    double fov = pEditorViewportClient->FOVAngle;
+    FIntPoint _;
```
baothientran, line 934: I think we should spell out variable name even though it's unused

Source/CesiumRuntime/Private/Cesium3DTileset.cpp
```diff
@@ -782,102 +785,164 @@ void ACesium3DTileset::MarkTilesetDirty() {
...
+
+  TSharedPtr<IStereoRendering, ESPMode::ThreadSafe> pStereoRendering = nullptr;
+  if (GEngine) {
+    pStereoRendering = GEngine->StereoRenderingDevice;
   }
 
-  return UnrealCameraParameters{size, location, rotation, fov};
+  std::vector<ACesium3DTileset::UnrealCameraParameters> cameras;
+
+  for (auto controllerIt = pWorld->GetControllerIterator(); controllerIt;
```
baothientran, line 834: it seems like pWorld has GetPlayerControllerIterator() too. Should we use that to remove the cast below? I'm curious what is the difference between GetPlayerControllerIterator() and GetControllerIterator()?

Source/CesiumRuntime/Private/Cesium3DTileset.cpp
```diff
@@ -782,102 +785,164 @@ void ACesium3DTileset::MarkTilesetDirty() {
...
+    return {};
+  }
+
+  TSharedPtr<IStereoRendering, ESPMode::ThreadSafe> pStereoRendering = nullptr;
+  if (GEngine) {
+    pStereoRendering = GEngine->StereoRenderingDevice;
   }
 
-  return UnrealCameraParameters{size, location, rotation, fov};
+  std::vector<ACesium3DTileset::UnrealCameraParameters> cameras;
```
baothientran, line 832: pWorld has GetNumControllers() and GetNumPlayerControllers(), so they can use to reserve the size of cameras vector to minimize allocation

Source/CesiumRuntime/Private/Cesium3DTileset.cpp
```diff
@@ -1158,22 +1223,23 @@ void ACesium3DTileset::Tick(float DeltaTime) {
...
-  Cesium3DTiles::ViewState tilesetViewState =
-      this->CreateViewStateFromViewParameters(
-          camera.value().viewportSize,
-          camera.value().location,
-          camera.value().rotation,
-          camera.value().fieldOfViewDegrees);
+  glm::dmat4 unrealWorldToTileset = glm::affineInverse(
+      this->GetCesiumTilesetToUnrealRelativeWorldTransform());
+
+  std::vector<Cesium3DTiles::ViewState> frustums;
```
baothientran, line 1234: frustum's vector should be reserved since we already know the number of camera

baothientran (commented):
@nithinp7 this is looking good to me! For the VR camera, unless it can be tested and verified soon, can you separate that commit into a different PR?

CesiumBen:
@nithinp7 Tested the latest commit with no success Testing notes updated on [issue](https://github.com/CesiumGS/cesium-unreal/issues/559#issuecomment-886953723)

nithinp7:
@CesiumBen I was able to reproduce some of the weird culling issues you noticed by using the `-emulatestereo` commandline argument. I think the last commit seems to grab the correct fovs for stereoscopic rendering, so at least the introduced culling bug should be fixed (let me know if it hasn't been). Also please let me know if the issue in god-mode where the frustum only covers one eye is gone, hopefully all the tiles visible in both eyes are now being loaded. If it seems to work for XR, we will likely be trying to get this into the August 1st release.

PR #574, opened 2021-08-02
Improvements to tileset life cycle, undo/redo, and copy/paste

### LINKED ISSUE #573: Review the "undo" handling for tilesets and other classes
We recently noticed that certain "Undo" operations in the editor are not handled properly. A specific example was the https://github.com/CesiumGS/cesium-unreal/issues/557 - and the case of "undoing a deletion" is one of the most important ones. But more generally, we should review the code for proper Undo/Redo handling, because it may be relevant for all other property changes. 

We do handle property changes to some extent, by overriding `PostEditChangeProperty` in some classes. But the change that is implied by an Undo operation does not provide the expected information in the given `FPropertyChangedEvent`. Most likely, other Unreal lifecycle methods may be relevant here: There are methods in `UObject`, like 

	virtual void PreEditUndo() override;
	virtual void PostEditUndo() override;

that may have to be overridden to handle the Undo correctly. There is also a method for `AActor` that is called

	void ResetOwnedComponents();

and mentions "undo/redo" in its comment, but the details have to be investigated further.

---

Further cases where "Undo" does not work could be found by doing dedicated tests, consisting of performing and undoing certain changes. Doing these tests "systematically and exhaustively" may be difficult. They could probably be "guided", to some extent, by the knowledge about the changes that are usually handled in `PostEditChangeProperty`. But one has to keep in mind that this does exactly _not_ cover actions like "deleting something" or "dragging something around in the world outliner", so these actions should be checked dedicatedly.

---

To mention the known cases that have to be investigated here fore now:

- Undoing the deletion of a tileset: 
  The fix in https://github.com/CesiumGS/cesium-unreal/pull/572 prevents the crash (and that's *important* as an immediate fix - everybody can imagine how frustrating it must be for a user who accidentally deletes a tileset, panics and presses "CTRL+Z", and then everything crashes and is lost irrecoverably - ouch :-( )
  But as mentioned in the issue: It could make sense to remove the `_tilesetIsDirty` flag, to make the state space smaller and clearer - basically by saying that the tileset is "dirty" **iff** (as in "if and only if") it is `nullptr`. (A detail that has to be taken into account is whether it is safe and sensible to delete the tileset and set it to `nullptr` in the method where we originally set `_tilesetIsDirty=true`, but I don't see an immediate reason why this should not be safe...)

- Undoing the deletion of other actors
  (I'm a bit scared of what this might mean for Georeferences, although we already _tried_ to address that to some extent)

- Undoing the change of the "asset ID" of a tileset. Right now, this does *not* trigger a reload of the tileset for the restored asset ID. 

Further cases that are found during the tests may be added here as comments, or by editing the list above.

kring:
Fixes #573

* Remove the `_tilesetIsDirty` is flag. Instead, just destroy the Cesium3DTiles::Tileset when necessary.
* Create all of our components and materials with flags `RF_Transient | RF_DuplicateTransient | RF_TextExportTransient`. This prevents all the transient glTF components a tileset creates on the fly from being included when doing a cut/copy/paste, which in turn prevents warnings when pasting a tileset.
* Implemented `PostEditUndo` and `PostEditImport` to recreate the tileset after undo/redo and paste, respectively.

Things I tested (and found to work):
* Delete a tileset, then undo (it comes back).
* Edit a tileset's asset ID, then undo the change (the original asset ID is loaded successfully).
* Edit a tileset's SSE, and then undo the change (the tileset is recreated with the new SSE. It would be nice to avoid recreating, but I can't immediately see how to do it with UE's Transaction system and it's not worth spending a lot of time on.)
* Add a tileset using the Quick Add, undo it, then redo (it comes back).
* Cut a tileset (it goes away), paste it back in the same level (it comes back).
* Copy a tileset and paste it in the same level (you get two, and they both work).
* Copy a tileset and paste it into a new, blank level (georeference and credits get created automatically, and the tileset works in the new level)
* Copy a tileset and paste it into a different level that already has a tileset (it works, and uses the existing georeference and credits)
* Switching between levels with Tilesets.
* Saving a level with a tileset.

### DIFF
--- a/Source/CesiumRuntime/Private/Cesium3DTileset.cpp
+++ b/Source/CesiumRuntime/Private/Cesium3DTileset.cpp
@@ -1240,6 +1240,13 @@ void ACesium3DTileset::PostEditUndo() {
   // recreate the tileset.
   this->DestroyTileset();
 }
+
+void ACesium3DTileset::PostEditImport() {
+  Super::PostEditImport();
+
+  // Recreate the tileset on Paste.
+  this->DestroyTileset();
+}
 #endif
 
 void ACesium3DTileset::BeginDestroy() {
--- a/Source/CesiumRuntime/Private/Cesium3DTileset.cpp
+++ b/Source/CesiumRuntime/Private/Cesium3DTileset.cpp
@@ -110,7 +110,7 @@ void ACesium3DTileset::PostInitProperties() {
 void ACesium3DTileset::SetTilesetSource(ETilesetSource InSource) {
   if (InSource != this->TilesetSource) {
     this->TilesetSource = InSource;
-    this->MarkTilesetDirty();
+    this->DestroyTileset();
   }
 }
 
@@ -118,7 +118,7 @@ void ACesium3DTileset::SetUrl(FString InUrl) {
   if (InUrl != this->Url) {
     this->Url = InUrl;
     if (this->TilesetSource == ETilesetSource::FromUrl) {
-      this->MarkTilesetDirty();
+      this->DestroyTileset();
     }
   }
 }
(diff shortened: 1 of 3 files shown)


cesium-concierge:

javagl:
I quickly looked at the code changes, and it seems like an improvement from that point of view. I don't know the exact implications and effects of the `RF_` flags, but have to assume for now that they make sense in this combination. I tested the steps described in the PR comment, and they worked as intended.

PR #576, opened 2021-08-04
Round feature IDs to the nearest integer so they work in Draco-compressed meshes

### LINKED ISSUE #575: Feature IDs can be mis-registered in Draco-compressed meshes
Feature ID attributes are usually of type FLOAT, even though they must have integer values. In a mesh that is quantized with Draco, the feature ID floats will be quantized as well. Once decompressed, those floating point feature IDs may not be _quite_ the original integer value we expect. For example, feature ID 54.0 may become 53.9956055. Cesium for Unreal, expecting these values to be whole numbers, was just casting them to integers, which truncates the value rather than rounding. So that `53.9956055` becomes `53` instead of the expected `54`. End result: we sometimes show the wrong feature metadata! 😱 

CesiumJS avoids this by looking up feature IDs via a batch texture, and it arranges for the integer value (e.g. 54.0) to land in the center of the pixel that corresponds to the desired batch ID. In a very roundabout way, it rounds the batch ID to the nearest integer. Cesium for Unreal needs to round as well.

kring:
Fixes #575

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,11 @@
 # Change Log
 
+### Next Release - ?
+
+##### Fixes :wrench:
+
+- Fixed a bug that could cause mis-registration of feature metadata to the wrong features in Draco-compressed meshes.
+
 ### v1.5.0 - 2021-08-02
 
 ##### Additions :tada:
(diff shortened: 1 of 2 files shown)


cesium-concierge:

baothientran:
Thanks @kring! this looks good to me

PR #581, opened 2021-08-10
Small tweaks to the georeference PR

kring:
This is a PR into #459 with some small changes I made while reviewing:

* Renamed the various "Ue" parameters to "Unreal" and added the necessary CoreRedirects to prevent this from breaking any existing Blueprints.
* `getGeoTransforms` -> `GetGeoTransforms` for consistency with other public methods and with UE's conventions.
* Changed the function names to say "AbsoluteUnreal" rather than just "Unreal" in all cases where it's the absolute world origin rather than the relative one. This was only in `GeoTransforms` and none of the functions are exposed to Blueprints, which is good.
* Removed `GeoTransforms` destructor and copy constructor because the defaults are pefectly good (and identical to the hand-written ones other than a small bug where the copy constructor would ignore the ellipsoid).
* Removed `ACesium3DTileset::HandleGeoreferenceUpdated` because it is undefined and unused.
* Added and edited a bit of doc.


cesium-concierge:

javagl:
The renamings (to clarify the "relative vs. absolute" issue) had been considered for a while now, so with this PR, the decision is made. I don't know much about these "Redirects", but assume that ~"this is how renaming BP-parameters is handled" in the Unreal Engine. Beyond that, I think that the Doxygen clarifications and the removal of the `GeoTransforms` con/destructor make sense.

PR #590, opened 2021-08-15
Add dedicated callback for georeference initialization

javagl:
Details being discussed in https://github.com/CesiumGS/cesium-unreal/pull/459

PR #591, opened 2021-08-16
Experimental Georeference component improvements

kring:
This is an alternative to #590 that also fixes the orientation bug Nithin observed in https://github.com/CesiumGS/cesium-unreal/pull/459#issuecomment-894515102.

First, it involves only a minor (and hopefully not too scary) change to the `CesiumGeoreference` Actor: `_geoTransforms` is updated on the `Serialize` method on load, so that it is always in sync with the loaded `OriginLongitude` etc. Previously it could be out of sync until either `OnConstruction` or `BeginPlay` was called.

On the Georeference Component side, there are more changes:

* We keep track of an `_ecefIsValid` flag, which indicates whether the ECEF_X/Longitude/_currentEcef properties are valid.
* This flag only goes true (indicating our ECEF representation is trustworthy) in a few scenarios:
  * The ECEF values are loaded (i.e. PostLoad)
  * The ECEF values are set explicitly.
  * The ECEF values are computed from the Actor Transform.
* If `_ecefIsValid` is still false in `OnRegister`, the ECEF values will be computed from the Actor Transform.
* It's also set to `false` on `OnComponentCreated`. This handles the case that a georeference component is pasted onto another Actor. The old ECEF cannot be trusted because it potentially refers to a different Actor.

I certainly haven't tested every scenario, and it still needs a fair bit of cleanup (especially of the doc comments, some of which are probably wrong now), but I _think_ the overall approach is sound.

### DIFF
--- a/Source/CesiumRuntime/Private/CesiumGeoreference.cpp
+++ b/Source/CesiumRuntime/Private/CesiumGeoreference.cpp
@@ -598,17 +598,20 @@ void ACesiumGeoreference::Tick(float DeltaTime) {
   _performOriginRebasing();
 }
 
-void ACesiumGeoreference::PostLoad() {
-  Super::PostLoad();
+void ACesiumGeoreference::Serialize(FArchive& Ar) {
+  Super::Serialize(Ar);
 
-  glm::dvec3 center(0.0, 0.0, 0.0);
-  if (this->OriginPlacement == EOriginPlacement::CartographicOrigin) {
-    center = _geoTransforms.TransformLongitudeLatitudeHeightToEcef(glm::dvec3(
-        this->OriginLongitude,
-        this->OriginLatitude,
-        this->OriginHeight));
+  // Recompute derived values on load.
+  if (Ar.IsLoading()) {
+    glm::dvec3 center(0.0, 0.0, 0.0);
+    if (this->OriginPlacement == EOriginPlacement::CartographicOrigin) {
+      center = _geoTransforms.TransformLongitudeLatitudeHeightToEcef(glm::dvec3(
+          this->OriginLongitude,
+          this->OriginLatitude,
+          this->OriginHeight));
+    }
+    _geoTransforms.setCenter(center);
   }
-  _geoTransforms.setCenter(center);
 }
 
 /**
--- a/Source/CesiumRuntime/Private/CesiumGeoreference.cpp
+++ b/Source/CesiumRuntime/Private/CesiumGeoreference.cpp
@@ -598,6 +598,19 @@ void ACesiumGeoreference::Tick(float DeltaTime) {
   _performOriginRebasing();
 }
 
+void ACesiumGeoreference::PostLoad() {
+  Super::PostLoad();
+
+  glm::dvec3 center(0.0, 0.0, 0.0);
+  if (this->OriginPlacement == EOriginPlacement::CartographicOrigin) {
+    center = _geoTransforms.TransformLongitudeLatitudeHeightToEcef(glm::dvec3(
+        this->OriginLongitude,
+        this->OriginLatitude,
+        this->OriginHeight));
+  }
+  _geoTransforms.setCenter(center);
+}
+
 /**
  * Useful Conversion Functions
  */
--- a/Source/CesiumRuntime/Private/CesiumGeoreference.cpp
+++ b/Source/CesiumRuntime/Private/CesiumGeoreference.cpp
@@ -127,11 +127,6 @@ void ACesiumGeoreference::PostInitProperties() {
       *this->GetName());
 
   Super::PostInitProperties();
-
-  // Initialize the GeoTransforms with the state from the
-  // deserialized properties
-  _geoTransforms.setEllipsoid(CesiumGeospatial::Ellipsoid(
-      glm::dvec3(_ellipsoidRadii[0], _ellipsoidRadii[1], _ellipsoidRadii[2])));
   UpdateGeoreference();
 }
 
@@ -336,14 +331,7 @@ void ACesiumGeoreference::OnConstruction(const FTransform& Transform) {
 }
 
 void ACesiumGeoreference::UpdateGeoreference() {
-  glm::dvec3 center(0.0, 0.0, 0.0);
-  if (this->OriginPlacement == EOriginPlacement::CartographicOrigin) {
-    center = _geoTransforms.TransformLongitudeLatitudeHeightToEcef(glm::dvec3(
-        this->OriginLongitude,
-        this->OriginLatitude,
-        this->OriginHeight));
-  }
-  _geoTransforms.setCenter(center);
+  this->_updateGeoTransforms();
 
   UE_LOG(
       LogCesium,
@@ -586,6 +574,23 @@ void ACesiumGeoreference::_performOriginRebasing() {
   }
 }
 
+void ACesiumGeoreference::_updateGeoTransforms() {
+  glm::dvec3 center(0.0, 0.0, 0.0);
+  if (this->OriginPlacement == EOriginPlacement::CartographicOrigin) {
+    center = _geoTransforms.TransformLongitudeLatitudeHeightToEcef(glm::dvec3(
+        this->OriginLongitude,
+        this->OriginLatitude,
+        this->OriginHeight));
+  }
+
+  this->_geoTransforms = GeoTransforms(
+      CesiumGeospatial::Ellipsoid(
+          this->_ellipsoidRadii[0],
+          this->_ellipsoidRadii[1],
+          this->_ellipsoidRadii[2]),
+      center);
+}
+
 void ACesiumGeoreference::Tick(float DeltaTime) {
   Super::Tick(DeltaTime);
 
@@ -603,14 +608,7 @@ void ACesiumGeoreference::Serialize(FArchive& Ar) {
 
   // Recompute derived values on load.
   if (Ar.IsLoading()) {
-    glm::dvec3 center(0.0, 0.0, 0.0);
-    if (this->OriginPlacement == EOriginPlacement::CartographicOrigin) {
-      center = _geoTransforms.TransformLongitudeLatitudeHeightToEcef(glm::dvec3(
-          this->OriginLongitude,
-          this->OriginLatitude,
-          this->OriginHeight));
-    }
-    _geoTransforms.setCenter(center);
+    this->_updateGeoTransforms();
   }
 }
 
--- a/Source/CesiumRuntime/Private/CesiumGeoreferenceComponent.cpp
+++ b/Source/CesiumRuntime/Private/CesiumGeoreferenceComponent.cpp
@@ -36,6 +36,12 @@ void UCesiumGeoreferenceComponent::SnapLocalUpToEllipsoidNormal() {
     return;
   }
 
+  // If our ECEF position isn't valid, we need to compute it from the Actor's
+  // Transform first.
+  if (!this->_ecefIsValid) {
+    this->_updateFromActor();
+  }
+
   // Compute the local up axis of the actor (the +Z axis)
   const glm::dmat3 currentActorRotation = _getRotationFromActor();
   const glm::dvec3 actorUpUnreal = glm::normalize(currentActorRotation[2]);
@@ -71,6 +77,7 @@ void UCesiumGeoreferenceComponent::SnapToEastSouthUp() {
         *this->GetName());
     return;
   }
+
   const UWorld* world = this->GetWorld();
   if (!IsValid(world)) {
     UE_LOG(
@@ -80,6 +87,13 @@ void UCesiumGeoreferenceComponent::SnapToEastSouthUp() {
         *this->GetName());
     return;
   }
+
+  // If our ECEF position isn't valid, we need to compute it from the Actor's
+  // Transform first.
+  if (!this->_ecefIsValid) {
+    this->_updateFromActor();
+  }
+
   const glm::dmat3 newActorRotation =
       this->Georeference->ComputeEastNorthUpToUnreal(_currentEcef);
   const glm::dvec3 relativeLocation = _computeRelativeLocation(_currentEcef);
@@ -105,6 +119,7 @@ void UCesiumGeoreferenceComponent::MoveToLongitudeLatitudeHeight(
 
   this->_setECEF(ecef, maintainRelativeOrientation);
 }
+
 void UCesiumGeoreferenceComponent::InaccurateMoveToLongitudeLatitudeHeight(
     const FVector& targetLongitudeLatitudeHeight,
     bool maintainRelativeOrientation) {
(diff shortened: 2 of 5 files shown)


cesium-concierge:

Source/CesiumRuntime/Private/CesiumGeoreference.cpp
```diff
@@ -586,6 +574,23 @@ void ACesiumGeoreference::_performOriginRebasing() {
...
+void ACesiumGeoreference::_updateGeoTransforms() {
+  glm::dvec3 center(0.0, 0.0, 0.0);
+  if (this->OriginPlacement == EOriginPlacement::CartographicOrigin) {
+    center = _geoTransforms.TransformLongitudeLatitudeHeightToEcef(glm::dvec3(
+        this->OriginLongitude,
+        this->OriginLatitude,
+        this->OriginHeight));
+  }
+
+  this->_geoTransforms = GeoTransforms(
```
javagl, line 586: Is there a reason to create a new instance here? Originally, the `_geoTransforms` was supposed to be created only once, and only modified via the setters.

Source/CesiumRuntime/Private/CesiumGeoreference.cpp
```diff
@@ -598,6 +603,15 @@ void ACesiumGeoreference::Tick(float DeltaTime) {
   _performOriginRebasing();
 }
 
+void ACesiumGeoreference::Serialize(FArchive& Ar) {
+  Super::Serialize(Ar);
```
javagl, line 607: I find that scary :-) More in a final comment

Source/CesiumRuntime/Private/CesiumGeoreferenceComponent.cpp
```diff
@@ -236,17 +259,17 @@ void UCesiumGeoreferenceComponent::_updateFromActor() {
 }
 
 glm::dvec3 UCesiumGeoreferenceComponent::_getAbsoluteLocationFromActor() {
-  const UWorld* world = this->GetWorld();
-  if (!IsValid(world)) {
+  const UWorld* pWorld = this->GetWorld();
```
javagl, line 262: Mixing cleanups like this with the actual fix makes it difficult to sort that out later...

Source/CesiumRuntime/Private/CesiumGeoreferenceComponent.cpp
```diff
@@ -83,6 +87,13 @@ void UCesiumGeoreferenceComponent::SnapToEastSouthUp() {
         *this->GetName());
     return;
   }
+
+  // If our ECEF position isn't valid, we need to compute it from the Actor's
+  // Transform first.
+  if (!this->_ecefIsValid) {
```
javagl, line 93: Can this function ever be called when the flag is `false`?

javagl:
Some comments that may be due to the non-polished state (or maybe oversights, that's why I'm mentioning them):

- The `PreEditUndo` and `PostEditUndo` are overridden, but not dedicatedly implemented
- Some of the checks (for the new `_ecefIsValid`, or for `this->IsRegistered()`) appear to be at places where I wouldn't expect them to be necessary
- The `PreEditChangeProperty` was removed, and with it the `OnGeoreferenceUpdated.RemoveAll` call: The callbacks are now only added, but never removed (although, admittedly, actually changing the GeoRef instance in the UI is certainly a "testing path" that I didn't cover extensively yet - e.g. setting it to `None` causes warnings to be printed, and that could be cleaned up, but we never sorted out that question of "What should happen with a null-GeoRef?" thoroughly...)

---

Many of the issues that triggered the refactoring - and in particular, the rotation issue that is addressed here - stems from a lack of understanding of certain aspects of the behavior of Unreal ("Something is wrong? Yeah, let's introduce some `updateWhateverIsWrong` method, and call it in some of the lifecycle methods, that will teach it..."). 

Maintaining the `_ecefIsValid` flag adds new complexity (even though it is somewhat documented, assuming that the initial comment of this PR would be turned into a doxygen comment). 

Some of the other changes raise additional questions here. For example, `_initGeoreference();` is now called in `OnRegister`. That **sounds** reasonable, but we know that `OnRegister` is called frequently. Originally, I _intentionally_ called that function (which was supposed to do the *one-time-**init**ialization(!)* of the "default" georeference) in the `OnComponentCreated` and `PostLoad` methods, which are supposed to be called exactly once (when creating the component in the editor, or when loading it). When it is called multiple times, then there will be multiple `OnGeoreferenceUpdated.AddUniqueDynamic` calls - sure, the callbacks are maintained *uniquely*, but I think that this may be hard to manage in the long run.

The overridden `Serialize` method of Georeference is scary as well. At least to my understanding, "one-time-initializations" of values that are derived from properties should be done in `PostActorCreated` or `PostLoad`. (Specifically: Is there a reason to *not* to the initialization of the `_geoTransforms` in an overriden `PostLoad`?)

---

There are some changes (added `.md` file, `world`->`pWorld` renaming etc.) that could probably go into the refactoring branch. But I have some doubts about the approach for the overall approach, regarding the intention of "solving the rotation issue". 

I think that we should strive towards a state of _fewer_ flags, _smaller_ state spaces, and _fewer_ overridden lifecycle methods (and that this could largely be achieved by overriding the *right* methods, ones, and doing the *right* things in these functions, but ... that's easier said than done, for sure...).

Source/CesiumRuntime/Private/CesiumGeoreference.cpp
kring: Originally I was calling `setEllipsoid` followed by `setCenter`. But that recomputes the transforms twice. This way only does it once. We can probably just rid of the setters.

Source/CesiumRuntime/Private/CesiumGeoreferenceComponent.cpp
kring: Yes, sorry, I meant to undo that before opening this PR.

kring:
> The PreEditUndo and PostEditUndo are overridden, but not dedicatedly implemented

Yep, undo/redo are not handled yet. Nor do they work right in main or in the bp-function-library branch. These functions are my reminder to add that.

> Some of the checks (for the new _ecefIsValid, or for this->IsRegistered()) appear to be at places where I wouldn't expect them to be necessary

It's true, some may not be necessary. As you know, the UE Component lifecycle is... complicated... so the code is intentionally defensive.

> The PreEditChangeProperty was removed, and with it the OnGeoreferenceUpdated.RemoveAll call: The callbacks are now only added, but never removed (although, admittedly, actually changing the GeoRef instance in the UI is certainly a "testing path" that I didn't cover extensively yet - e.g. setting it to None causes warnings to be printed, and that could be cleaned up, but we never sorted out that question of "What should happen with a null-GeoRef?" thoroughly...)

Yes, this is probably not right.

> Specifically: Is there a reason to not to the initialization of the _geoTransforms in an overriden PostLoad?

Yep. The `CesiumGeoreference::PostLoad` gets called _after_ the `CesiumGeoreferenceComponent::PostLoad`, at least in some cases. So if we update _geoTransforms in `CesiumGeoreference::PostLoad`, `CesiumGeoreferenceComponent::PostLoad` will see a wrong value for it. Essentially by doing it in `PostLoad`, we're allowing our class invariant to be violated for some unknown amount of time in between when UE calls Serialize and when UE calls PostLoad. That gap is an easy source of bugs.

I can't see much downside to doing it in Serialize? Is there something in particular you're afraid of?

Source/CesiumRuntime/Private/CesiumGeoreferenceComponent.cpp
kring: Probably not very likely, but any C++ code with a pointer to the component could call it at any inopportune time it wants. I'm not sure if similar could be done from Blueprints.

kring:
> The PreEditChangeProperty was removed, and with it the OnGeoreferenceUpdated.RemoveAll call: The callbacks are now only added, but never removed (although, admittedly, actually changing the GeoRef instance in the UI is certainly a "testing path" that I didn't cover extensively yet - e.g. setting it to None causes warnings to be printed, and that could be cleaned up, but we never sorted out that question of "What should happen with a null-GeoRef?" thoroughly...)

I just checked and this isn't a problem in the editor, at least. OnUnregister gets called before PreEditChangeProperty with the old value and OnRegister gets called after with the new value, so the delegation subscription is updated correctly. If it were changed from Blueprints that would be a different story, but it currently can't be changed from Blueprints (it's not even accessible from Blueprints). If you were to change it from C++, it wouldn't have worked correctly before or after this PR.

kring:
@javagl I'm a bit confused by your overall feedback. In particular:

> I think that we should strive towards a state of fewer flags, smaller state spaces, and fewer overridden lifecycle methods (and that this could largely be achieved by overriding the right methods, ones, and doing the right things in these functions, but ... that's easier said than done, for sure...).

It seems to me that this PR represents our closest approximation of this yet. Do you disagree? The next best option is #590, and it has serious conceptual/state space problems (that you've described), plus a much higher chance of impacting things outside the GeoreferenceComponent.

Whereas in this PR I think the `_ecefIsValid` flows naturally from the job that the Component is meant to do. It's entire purpose is to maintain an accurate ECEF position for an Actor and adjust the Actor's Unreal Location/Rotation accordingly as the Georeference Actor and OriginLocation change. It can't ensure at all times that its ECEF mirror of the Actor's position/orientation is accurate, but it can keep track at any given time _whether_ it's accurate. That's all the flag is.

I guess what I'm coming to is that if there's as much disagreement between us is it seems from your review, then I think we may have to just close the bp-function-library PR unmerged and move on. It seems like it will take a lot of time to resolve that (on top of the huge amount of time we've already spent), for a rather modest user-facing improvement. And furthermore this entire system is likely to become obsolete in the next year or less (if it's not obvious why, ask me offline).

Source/CesiumRuntime/Private/CesiumGeoreference.cpp
javagl: The `setEllipsoid` and `setCenter` only update the transforms when the respective value actually *changed*. But even then, the call is not performance critical: It happens in `UpdateGeoreference`, and **if** this was called "in each tick", then 1. there would be something wrong, and 2. we'd have to look at all other implications that this may have: We don't know for sure *who* will listen to `OnGeoreferenceUpdated`, and *what* the listener will do, but I assume that whatever is done *there* will be more time critical than what is done in the `GeoTransforms`.

(And if we ran a benchmark in a profiler and noticed that more than 0.1% of the frame time is spent with computing that single matrix, we could still add a `setCenterAndEllipsoid` method and see whether this increased the overall performance, which I doubt).

One thought related to the `_geoTransforms` was **roughly**: There could/should be the "default `GeoTransforms` instance" that serves as the basis for the "default `Georeference`" (that is obtained with the `getDefault...` method). The "workflow" that was considered there was:

- Somebody obtains the default `Georeference`
- Somebory modifies it
- Somebody deletes it
- Some new default `Georeference` is created. It should have the state of the default one that was previously deleted (which was stored in the global, undeletable "default `GeoTransforms`").

(The `_geoTransforms` would have to be some `shared_ptr` for that to work, but that's an implementation detail). 

However, I'm fine with either solution, because the behavior of the  `getDefault...` method in view of deletions, undos, etc. has not been thought through in all depth.

javagl:
> I just checked and this isn't a problem in the editor, at least. OnUnregister gets called before PreEditChangeProperty with the old value and OnRegister gets called after with the new value, so the delegation subscription is updated correctly.

I don't understand what "updating the subsciption" means here. Roughly speaking: You have two `Georeference` objects, `A` and `B`. Georeference `A` is assigned to the component. The callback of the component is attached to the `OnGeoreferenceUpdated` of Georeference `A`. Now, in the editor, you assign Georeference `B` to the component. The callback of the component is *still* attached to the `OnGeoreferenceUpdated` of Georeference `A`, and *additionally*, to the the `OnGeoreferenceUpdated` of Georeference `B`.

(At least, that's my understanding from the code). Am I overlooking something here? (I haven't tried it out, but in doubt, I'll do that).


> The next best option is #590, and it has serious conceptual/state space problems (that you've described), plus a much higher chance of impacting things outside the GeoreferenceComponent.

As discussed during the sync: Both PRs introduce a boolean flag. The `_georeferenceInitialized` flag can be documented as `// Says whether OnConstruction has been called`. The `_ecefIsValid` flag is a `UPROPERTY`, and it has to be checked and set under not-so-obvious conditions that require extensive documentation. I think that it is harder to understand and harder to maintain. 

Usually, I'd classify two alternative approaches as follows:

- When one approach is correct and the other one is not, then there is nothing to hesitate about.
- When both approaches are correct, I'd choose the one that is easier to understand, has less impact on other code paths, has a smaller surface area, is easier to test, to debug, and to document.

The problem now is: I don't know if the approach of tracking the `_ecefIsValid` does *not* have drawbacks that the other approach has. Or to put it that way: I don't know whether one of the approaches is "more correct than the other".

kring:
> I don't understand what "updating the subsciption" means here. Roughly speaking: You have two Georeference objects...

Yes, I know. Please just try it before you write a novel about it. Updating the subscription means "removing the OnGeoreferenceUpdated handler from the old Georeference Actor, and then adding it to the new Georeference Actor." I understand the (potential) problem, and it is handled.

kring:
> The _ecefIsValid flag is a UPROPERTY, and it has to be checked and set under not-so-obvious conditions that require extensive documentation.

Anytime you use the ECEF, you need to check it's valid first. Any time the ECEF becomes valid or becomes invalid, you have to change the flag. How is that difficult to understand or require extensive documentation?

javagl:
> Please just try it before you write a novel about it.

I tried it out, and this is the output: 
```
LogCesium: Verbose: Broadcasting OnGeoreferenceUpdated for Georeference CesiumGeoreference /Game/GeoRefCompTestNew.GeoRefCompTestNew:PersistentLevel.CesiumGeoreference_5
LogCesium: Verbose: Called HandleGeoreferenceUpdated for CesiumGeoreference which uses georeference CesiumGeoreference /Game/GeoRefCompTestNew.GeoRefCompTestNew:PersistentLevel.CesiumGeoreference_3
```
The component receives an update from `CesiumGeoreference_5`, although its current one is the `CesiumGeoreference_3`. 

> How is that difficult to understand [...] ?

kring:
You're right @javagl, sorry to be both rude and wrong! I had confirmed that when you edit the Georeference property in the editor, it calls OnUnregister while the Component still has the old Georeference Actor, and then calls OnRegister after it has been set to the new one. I expected a symmetric call to `OnGeoreferenceUpdated.AddUniqueDynamic` in OnRegister and to `OnGeoreferenceUpdated.RemoveAll` in OnUnregister, and I had looked (too quickly) and thought I saw that. But what I actually saw was the similar calls for `TransformUpdated` rather than `OnGeoreferenceUpdated`.

So the commit above should fix that. Thanks!

javagl:
There's never a need to apologize, particularly not towards me. Apologies are useless. Corrections of errors are useful. In that manner, the symmetry of `OnRegister`+`AddUniqueDynamic` and `OnUnregister`+`RemoveAll` now "looks right", enough so that I'm sure that it will solve that problem, but I can/will try it out depending on the path forward.

PR #594, opened 2021-08-18
Avoid duplicating vertices when possible

### LINKED ISSUE #510: Provide option to ignore the glTF spec and use smooth normals instead of flat normals
The [glTF spec says](https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#meshes):

> When normals are not specified, client implementations should calculate flat normals.

Calculating flat normals requires duplicating every vertex that is shared between multiple triangles, which increases memory usage, loading time, and rendering time. In many cases, smooth normals -  where each vertex's normal is the weighted average of the normals of its faces - would be just as good or maybe even better. We can easily provide a Tileset option to toggle between smooth normals (faster) and flat normals (strictly spec compliant).

### LINKED ISSUE #511: Do not duplicate vertices unnecessarily
Currently, we duplicate every vertex that is shared between multiple triangles. But we only _need_ to do this when we're generating tangents or normals. After #499, we only generate tangents when they're actually needed. After #510, we might not even need to do it when generating normals. Duplicating vertices increases memory usage, loading time, and rendering time, so it's worth avoiding when we can.

nithinp7:
Only duplicates indexed vertices when either:
- Tangents are missing, but required or
- normals are missing and flat-normals need to be generated.

In [this corresponding PR](https://github.com/CesiumGS/cesium-native/pull/304), a new option has been added on the cesium-native side to generate smooth normals when the original gltf was missing normals. This can help avoid the flat-normal generation (and with it the vertex duplication) that would be otherwise required on the cesium-unreal side.

Fixes #511 
Fixes #510


cesium-concierge:

* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium-unreal/blob/use-index-buffer/CHANGES.md) was not updated.

kring:
Thanks @nithinp7, looks good now! I also confirmed collisions work both with and without vertex duplication, because I believe if the winding order were flipped on the physics meshes than you'd get stuck inside buildings rather than being prevented from entering them. All looks good, though. Merging!

PR #596, opened 2021-08-20
Add missing UFunction annotation in metadata blueprint library

baothientran:
While looking into metadata tutorial, I just realized we miss UFunction annotations, so blueprint library doesn't display those

### DIFF
--- a/Source/CesiumRuntime/Public/CesiumMetadataFeatureTable.h
+++ b/Source/CesiumRuntime/Public/CesiumMetadataFeatureTable.h
@@ -105,13 +105,21 @@ public:
    *
    * @param FeatureID The ID of the feature.
    */
-  TMap<FString, FString> static GetPropertiesAsStringsForFeatureID(
+  UFUNCTION(
+      BlueprintCallable,
+      BlueprintPure,
+      Category = "Cesium|Metadata|FeatureTable")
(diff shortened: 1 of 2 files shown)


cesium-concierge:

* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium-unreal/blob/miss-ufunction/CHANGES.md) was not updated.

Source/CesiumRuntime/Public/CesiumMetadataFeatureTable.h
```diff
@@ -95,7 +95,7 @@ class CESIUMRUNTIME_API UCesiumMetadataFeatureTableBlueprintLibrary
       BlueprintCallable,
       BlueprintPure,
       Category = "Cesium|Metadata|FeatureTable")
-  static TMap<FString, FCesiumMetadataGenericValue> GetPropertiesForFeatureID(
+  static TMap<FString, FCesiumMetadataGenericValue> GetMetadataValuesForFeatureID(
```
baothientran, line 98: I discussed with @argallegos offline, and `GetPropertiesForFeatureID` name is not suitable since it returns a map between property name and MetadataGenericValue. We have another similar method name `GetMetadataValuesForFace` in `CesiumMetadataUtilityBlueprintLibrary`. So I think this change will be consistent with the Utility's method. But this may introduce breaking changes here

PR #604, opened 2021-08-23
Cesium ion login panel updates

argallegos:
In this PR:

- Phrasing of login panel was updated to reflect the new ability to log in with an Epic Games account
- The "Connect" button has received a visual upgrade

image

@javagl did the legwork in figuring out how to change the color of a button.

As far as I can tell, if we want to upgrade the style of all the Cesium panels, we'll need to create style sets in `CesiumEditor.cpp`.  

We will want to decide if we want rounded corners on our buttons (and if so, how rounded?). Then, we need to create an image file with the appropriately rounded corners and set it as a SlateBrush. 

I think it is possible to change a font, but I can't figure out how to actually do it. If we want to, we should try to use a font like [Lato ](https://fonts.google.com/specimen/Lato)to keep it looking "Cesium".


cesium-concierge:

* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium-unreal/blob/editor-button-style-draft/CHANGES.md) was not updated.

shehzan10:
@argallegos and (@javagl) thanks for the good work here. Definitely a good upgrade.

Maybe the only change I would recommend is for the button to say "Connect to Cesium ion with your Epic Games Account".

Would it be too long for one button?

argallegos:
@shehzan10 Tried it out and actually it's not too bad with the new style (it definitely didn't fit with the old style.)
image

From a design standpoint, it's ok, though maybe not very standard. I think it feels weird because users don't have to use an Epic account, and having the text on the button feels kind of like misdirection. Like we are trying to imply that they can ONLY use an Epic account? I think keeping the text to "Connect to Cesium ion" is the best option from a UX standpoint. We can bold or otherwise emphasize the option to use an Epic games account above or below the button. 

On the other hand, the button with the full text works, and gets the point across. If that's what we definitely want, let me know and I'll push that change.

shehzan10:
@argallegos Great progress!

> I think it feels weird because users don't have to use an Epic account, and having the text on the button feels kind of like misdirection.

I see it the other way. If I am Unreal Engine user, I have an (existing) Epic Games account, and I will be much more inclined to clicking the button if I know it has SSO with Epic Games account.

> We will want to decide if we want rounded corners on our buttons (and if so, how rounded?). Then, we need to create an image file with the appropriately rounded corners and set it as a SlateBrush.

Is this complicated? I think rounded corners will help a little bit to make it look more like a button.

javagl:
>  Tried it out and actually it's not too bad with the new style

What happens when you make the panel narrower? Does the button overflow, or does it truncate the button text, or wrap it automatically? (Each of that _could_ look odd...)

> Is this complicated? I think rounded corners will help a little bit to make it look more like a button.

Even at the risk of sounding snarky: There are buttons. Default buttons. They can be created with `SNew(SButton)`. These look what buttons look like in the given environment. Everybody who sees such a thing knows: "This is a button. I can click that". Life _can_ be simple.

However: There are many possible "templates" that one could use as a basis for own buttons, and I think there was also some "template" with rounded corners that could probably be used here. (The "effort" is usually in finding the template and how to adjust and configure it - the actual result may just be a few lines of code. But I've been looking at that recently, and can have another look).

javagl:
Answering my first question: It's truncated... :-/ 

CesiumUnreal Connect Button

argallegos:
Yeah, @shehzan10 Marco explained it in better words than me - my main concern is that the text implies that there is no other option than to use an epic account. We might alienate users if they think that they might HAVE to use an epic account, and can't use their pre-existing ion account. Not everyone is going to want to use their Epic account, and if they've been using the plugin up to this point and notice that suddenly the phrasing has changed to imply that it's an Epic account only sort of thing, they might be unhappy. So we just need to be sure that the benefits of this button phrasing outweigh the disadvantages.

There is also the problem with truncated text. I'm actually fairly certain that there's no way to make a button scale based on the size of its panel, though I'd love to be proven wrong on that. But with the default editor layout, which it's safe to assume many users have, the panel is quite narrow: 
image
Which might mean that our best option is to keep the button text shorter than ~26 characters.

I'm going to try one other thing on that to see if I can otherwise make the sign in options clear without putting the text in the button itself - stand by. 

Regarding styling as a whole, I think there are templates. In `UnrealEngine\Engine\Source\Editor\EditorStyle\Private\SlateEditorStyle.cpp`, there's an `FButtonStyle` called `SimpleRoundButton` that might be what we want - but I don't know how to apply it. I'm also not sure if that's a template that we can use right out of the box or if we have to reconstruct it in `CesiumEditor.cpp`. It's not as easy as changing the `FButtonStyle` on line 166 from "FlatButton.Default" to "SimpleRoundButton", I tested that and it does nothing. But it might be something similarly simple that I just don't know how to implement precisely.

Like most of the documentation I've been able to find online, this is from 2015, but it might be a relatively viable option for us - https://nerivec.github.io/old-ue4-wiki/pages/templateslate-style-sets-part-2.html
In that tutorial, it loads a StyleSet from a .uasset that we can create in editor, rather than hardcoding the entire styleset. That would be nice for future edits to the styleset, and allow us to preview changes to styles much more quickly since we wouldn't have to rebuild the entire solution. The two caveats for this tutorial are:

a) it's designed for an in-game menu, not an in-editor menu. I'm not sure if that changes the functionality that is available to us.
 
b) It's from 2015. I've noticed that a lot of Slate has changed since 2015 and the code snippets from then don't always exactly apply nowadays. It's probably going to take a lot of playing around to figure out how to recreate something like this today.

argallegos:
Here's two other options for text display:
image
image

We can make the text about sign in options bold, and even move it below the button (though padding could use some work there.) The copy itself could also use some work - I'm not sure how best to concisely say "You can use your Epic Games account, or use a Cesium ion account, or create a new Cesium ion account." Any suggestions welcome. 

We could also go another route entirely, like making the bold text read something like "New! Log in to Cesium ion with your Epic Games account." Which might get the point across, but would have to be revisited once that feature is no longer new enough.

javagl:
I had a short stab at the "rounded button". Although I think that this is not important _in particular_, it's to some extent just a subjective "Meh, let's see how that stuff works" thing. In that manner...:

> It's not as easy as changing the FButtonStyle on line 166 from "FlatButton.Default" to "SimpleRoundButton", I tested that and it does nothing.

Yeah, tried that as well 😁 (but I hope it's clear that I have no idea (yet) about how all this works, so that `FlatButton` thing was only a first guess, and may very well have made no sense in the first place...). 

As you mentioned: A round button involves creating this special "template image", which will be used as the basis for this "FSlateBoxBrush": It takes an image like this (enlarged here) 

CesiumButton-Large

and divides that into 3x3 "tiles". (The size of the tiles is based on the `FMargin(4 / 16.0f)`, meaning that the corners will be 4x4 pixels, and the edges will be 4 pixels wide/high). This is then used to paint the button, and the template image can be "dyed" with a linear color.

I tried this out in https://github.com/CesiumGS/cesium-unreal/tree/editor-button-style-draft-round-2

CesiumUnreal Connect Button Round 2

(The template image has slightly rounded corners, and a slightly darker border, but of course, there are `width*height*(2^24)` different ways of how this may look...)

shehzan10:
Thanks @argallegos and @javagl. I don't intend to push this discussion too long as we have other things to get to as well.

I think we can go with something like this image with @javagl's rounded corners.
 
image

I would recommend changing the text to "You can now sign in with your Epic Games account!"

----------------------------------------------------------------------------------------------------

This looks good too :+1:. If its basically a copy of the style and easy to get in, lets do it.

image

argallegos:
Sounds good. @javagl thanks for looking into the rounded corners! If it's alright with you, I'll just merge your branch into the first and use your rounded corners + this blue style for all the buttons I can find. PR should be completed shortly.

argallegos:
Here's how the panels are looking now: 
(All images are using the default Unreal Editor layout)

Cesium login panel
CesiumPanelNew

Cesium ion panel
CesiumIonPanelNew
CesiumIonPanelNew2

I think this is all we need for now, so this should be good to review/merge.

javagl:
From glancing over the code, I noticed that there still was this `EditorStyle` dependency that I used in the first commit, but that has become unnecessary in the meantime, so I removed that. 

For the other changes, I could nitpick (e.g. whether the "You can now sign in with your Epic Games account!"-text has to be disabled while connecting, via https://github.com/CesiumGS/cesium-unreal/pull/604/files#diff-89074f326869710f358a9f2f49589d2e948635a591cc5afa0488d32faa2e543fR114 , but that may not be important).

Functionally and in terms of the appearance, everything seems to be OK, so I'll merge as soon as the checks are completed, and update https://github.com/CesiumGS/cesium-unreal/pull/605#issuecomment-904782094 based on that.

Issue #332, opened 2021-04-12
Add a refresh button to Cesium ion Assets Panel


shehzan10:
Closing and opening is not the best workflow. If we can add a refresh button to query the assets, that will be great.

argallegos:
Adding this would at least partially solve https://github.com/CesiumGS/cesium-unreal/issues/546.

PR #605, opened 2021-08-23
Added refresh button for asset list

javagl:
Fixes https://github.com/CesiumGS/cesium-unreal/issues/332 

<sub>(Also, removed some unused fields ... one reason of why I'm a fan of documenting even _private_ members is that the attempt to write such a comment for `_refreshInProgress` would have led to `/** This flag is not used */`, and thus, its removal...)</sub>

Tried it out with adding and deleting assets:

CesiumUnrealRefreshButton

(An aside: The cesium ion web UI doesn't refresh automatically either :D )


shehzan10:
@javagl Is it possible to use an icon button like 🔄?

shehzan10:
Font Awesome icons: https://fontawesome.com/v5.15/icons?d=gallery&p=2&q=refresh

javagl:
@shehzan10 Using an icon (and to answer the question "What does this button do?", also added a tooltip)

CesiumUnreal Refresh icon

shehzan10:
@javagl Looks great! If easy enough, some styling consistency like #604 will wrap this up. Either this can go in first and #604 adds styling, or vice versa. No strong preference from me.

javagl:
It's easy to apply the button style to that button as well (and I'll update the arrows icon to be light gray instead of dark gray then...), so in doubt, 604 should be merged first.

javagl:
Using the "Cesium Button" style, it now looks like this:

CesiumUnreal Refresh with color

(Using "white" to match the text color for "Cesium Buttons")

(It should be 16x16, the standard "small icon" size, but apparently, it's displayed as 18x18 - but this also applies to all other "small icons". I already reduced the padding for the button, but if desired, it can still be made a tad smaller...)

kring:
Thanks @javagl!

kring:
I added a mention of this to CHANGES.md in main.

PR #608, opened 2021-08-25
Raster overlay refactoring

kring:
This is the Unreal side of CesiumGS/cesium-native#325.


cesium-concierge:

* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium-unreal/blob/raster-overlay-refactoring/CHANGES.md) was not updated.

Source/CesiumRuntime/Public/Cesium3DTileset.h
```diff
@@ -242,6 +243,15 @@ class CESIUMRUNTIME_API ACesium3DTileset : public AActor,
...
   float CulledScreenSpaceError = 64.0;
 
+  /**
+   * Refreshes this tileset, ensuring that all materials and other settings are
+   * applied. It is not usually necessary to invoke this, but when
+   * behind-the-scenes changes are made and not reflected in the tileset, this
+   * function can help.
+   */
+  UFUNCTION(CallInEditor, BlueprintCallable, Category = "Cesium")
+  void RefreshTileset();
```
javagl, line 253: What is "refreshing"? (Cold softdrinks....?!). **Who** is going to call this function, to achieve **what**? Why is this public and `BlueprintCallable`?  

Right now, it does not seem to be called at all. It only calls `DestroyTileset`, and `DestroyTileset` is already called at 16 (!) different places. Many of these calls essentially convey the message "Something changed, we don't know how to handle that, let's start again from scratch" (and these calls could reasonably be changed to a `RefreshTileset` call, even if that was a private method...). 

More broadly: **If** it is known that something can change (like setting some material parameter value), then that change should be tracked and trigger the "smallest possible update" _internally_. This might, initially, well be a `DestroyTileset` call. But mabe it would be possible to improve that later...

    void layerOpacityChangedOrSo() {
        //RefreshTileset(); // remove this
        // do this instead:
        tileset->someMaterial->someLayer->setOpacity(newOpacity);
    }

nithinp7: It's not used anywhere because it's meant to be used by the user in the editor or blueprint, although it may be nice to use it internally in place of DestroyTileset (as you recommended). But either way I think it's useful as a catch-all for now. There are a lot of cases where the refresh _may_ be needed, such as adding/removing/editing a CartographicPolygon, editing material layers, editing the blend functions.

javagl:
When deleting the overlay component from a tileset (e.g. when removing the Bing Maps from the CWT that was Quick-Added), then the textures are not properly removed. It still displays the textures that had been loaded (with the specific level), and zooming in or out (changing the level) makes then disappear or re-appear. Maybe a missing call to `RefreshTileset` somewhere? 😁

javagl:
I tried it out in a branch that is "close" to main (not exactly main, but without these updates, and that's the main point), and there, deleting the overlay causes the tileset to become gray/untextured.

javagl:
I'll probably need to read more about Material Layers in the first place in order to actually try that out (or is there already some sort of instructions/draft tutorial for that part that I could use?). 

Until then, a detail: `ML_CesiumRasterOverlay` seems to contain an unused part:

CesiumUnreal ML_CesiumRasterOverlay BP

kring:
> When deleting the overlay component from a tileset (e.g. when removing the Bing Maps from the CWT that was Quick-Added), then the textures are not properly removed. It still displays the textures that had been loaded (with the specific level), and zooming in or out (changing the level) makes then disappear or re-appear.

Good catch @javagl! There were two problems here (both now fixed).

First, when we detached raster overlay texture from a geometry tile, we weren't resetting the material's texture to its default (a 1x1 transparent pixel). So it whould just stay set to whatever overlay texture until set to something else. The texture itself was handled by UE's garbage collector so no major drama there, but it caused the symptom you described until adding another overlay or explicitly calling RefreshTileset.

The second (and more serious) problem was that RasterOverlayTile was never being deleted, ever. I don't think that was causing any symptoms beyond a memory leak, but it was a quite serious leak.

Source/CesiumRuntime/Public/Cesium3DTileset.h
kring: > If it is known that something can change (like setting some material parameter value), then that change should be tracked and trigger the "smallest possible update" internally.

Yes, of course, whenever possible. But it can get quite complicated. My tileset references a material which uses a material function which uses another material function which uses a texture which I've just re-imported. We might even handle that one, but I'm not sure and in any case there are probably some (rare?) cases that we don't. As the Doxygen says, it's usually not necessary to use RefreshTileset, but it's nice to have the option.

kring:
> Until then, a detail: ML_CesiumRasterOverlay seems to contain an unused part

Fixed.

Source/CesiumRuntime/Public/Cesium3DTileset.h
javagl: It is not clear under which conditions the function should be called. It is not clear what the function does. We will never be able to remove that function, or change its semantics in any way.

People will call this function, because ~"something does not work, and after calling the function, it does work". They will not hesitate to do this, because "refresh" sounds nice and innocent. Consider calling it `CompletelyDestroyAndReInitializeTileset` - because _that's what it does_ (one could argue to just make `DestroyTileset` a `public` function - the fact that the tileset is re-initialized automatically on the fly after it was destroyed is a subtle implementation detail).

nithinp7:
Thanks @kring!

PR #614, opened 2021-08-31
Fixed "Drape Over Terrain Tileset" deleting existing overlays.

### LINKED ISSUE #613: "Drape over Tileset" button deletes existing overlays
In the Cesium Ion Asset panel, there is a convenient button for raster overlays that allows you to immediately drape the chosen overlay over the first quantized-mesh tileset in the world. Previously, since tilesets only supported one raster overlay at a time, this button deleted all existing overlays and added just the new one. After #608 however, it is now possible to have multiple overlays with a variety of blend settings for how exactly they should be combined. So this behavior should now be removed.

nithinp7:
Fixes #613 

It is still awkward to add new overlays with the drape button since new overlays require a different material layer key to be entered. Since the default material layer key is "Overlay0", the new overlay may appear to replace parts of the existing overlay. This PR only makes sure that the old overlays aren't outright deleted from the tileset's components list.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -24,6 +24,7 @@
 ##### Fixes :wrench:
 
 - Fixed indexed vertices being duplicated unnecessarily in certain situations in `UCesiumGltfComponent`.
+- Fixed an issue where the "Drape Over Terrain Tileset" button in the Cesium Ion Asset panel deleted all existing overlays.
 
 ### v1.5.1 - 2021-08-09
 
--- a/Source/CesiumEditor/Private/CesiumEditor.cpp
+++ b/Source/CesiumEditor/Private/CesiumEditor.cpp
@@ -374,16 +374,6 @@ UCesiumIonRasterOverlay* FCesiumEditorModule::AddOverlay(
     ACesium3DTileset* pTilesetActor,
     const std::string& name,
     int64_t assetID) {
-  // Remove any existing overlays and add the new one.
-  // TODO: ideally we wouldn't remove the old overlays but the number of overlay
-  // textures we can support is currently very limited.
-  TArray<UCesiumRasterOverlay*> rasterOverlays;
-  pTilesetActor->GetComponents<UCesiumRasterOverlay>(rasterOverlays);
-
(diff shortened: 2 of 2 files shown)


cesium-concierge:

javagl:
The code change looks straightforward (and removes a `TODO` - yay!). I'd try it out tomorrow (if it is urgent for the release, it could probably be merged and only _verified_ in the release build, because I don't see how this could break anything, but for the sake of the usual process, I'll test it ASAP).

javagl:
<sup>Nah, what does "tomorrow" mean anyhow.</sup>

I tried it out, and it has the intended effect: The components are _added_ instead of replacing the (formerly usually only one) component.

---

One could argue about the wording in the changelog: "Fixed an issue" sounds like a problem/bugfix. I'd say it is just a changed behavior to accommodate the (now newly supported) **multiple** overlays. 

> The "Drape Over Terrain Tileset" button now adds the raster overlay component to the tilesets, instead of replacing all existing raster overlay components, to better support workflows where multiple raster overlays are combined

or so...

---

A technical aside: I actually added

      UE_LOG(
          LogCesiumEditor,
          Warning,
          TEXT("Now there are %d"),
          pTilesetActor->GetInstanceComponents().Num());

in the changed function (and log outputs in `OnComponentCreated` and `OnComponentDestroyed`), because - despite the correct state in the editor UI - I didn't believe that these components could not be added _multiple times_. But apparently, the concept of "**Instance** components" is exactly that: When pressing the button multiple times, the old instance is destroyed, and replaced with the new one. I'll still have to figure out what the "black magic" is behind that: How are two "instance components" determined to be "equal" in that regard?

---

Excessively clicking buttons and deleting components caused some "odd rendering state" to appear that also could **not** be fixed with `RefreshTileset`, but for now, I have to assume that this is a general problem stemming from the default layer identifier being `Overlay0` for all of them, and the resulting textures to be essentially "random". 

(Thiese are things that I'd like to stress-test further, but ... ... ... time ... ... and this is unrelated to this PR in any case...)

PR #615, opened 2021-09-01
Fix missing includes and namespaces.

kring:
By default, UE compiles multiple CPP files at once by concatenating them together, called a unity build. This makes the build go faster, but can mask missing includes and namespace references. Prompted by https://community.cesium.com/t/updated-to-latest-1-5-1-plugin-and-it-wont-compile-on-4-26-or-4-26-2-source-engine-in-visual-studio/15083, I turned off this concatenation by adding the following to `C:\Users\[my user name]\AppData\Roaming\Unreal Engine\UnrealBuildTool\BuildConfiguration.xml`:

```
<?xml version="1.0" encoding="utf-8" ?>
<Configuration xmlns="https://www.unrealengine.com/BuildConfiguration">
    <BuildConfiguration>
        <bUseUnityBuild>false</bUseUnityBuild>
    </BuildConfiguration>
</Configuration>
```

Then I compiled the plugin and fixed any errors that were reported. This PR is the result.


cesium-concierge:

* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium-unreal/blob/missing-includes/CHANGES.md) was not updated.

PR #616, opened 2021-09-01
Improved UX for adding raster overlays with the Cesium panels

kring:
Following on from #614...

When using the Quick-Add panel, users almost certainly want to _replace_ the base raster overlay rather than adding another on top. When using the Cesium ion Assets panel, either possibility is pretty likely. So in this PR, I've made the Quick Add panel replace `Overlay0` with the new asset. And I've added a second button to the Cesium ion Assets panel allowing the user to choose:

image

The first button replaces `Overlay0`, just like the Quick-Add does.

The second button assigns the first unused `OverlayN` MaterialLayerKey to the new component.

This still isn't perfect UX-wise, but it's a big improvement I think.


cesium-concierge:

PR #621, opened 2021-09-01
Include Engine.ini in plugin package.

kring:
Without this change, the Engine.ini with our CoreRedirects doesn't get included in the plugin package. Which means a broken level blueprint in the Metadata sample and elsewhere.


cesium-concierge:

* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium-unreal/blob/include-engine-ini/CHANGES.md) was not updated.

PR #622, opened 2021-09-01
Add Travis builds for Unreal Engine v4.27

kring:


cesium-concierge:

* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium-unreal/blob/ci-427/CHANGES.md) was not updated.

kring:
It has taken _much_ longer than expected to get cesium-native building for Linux using Unreal's cross-compiling toolchain on a Windows host. There are a bunch of boring reasons, but here are some things worth mentioning...

* The [Unreal cross compiling version of clang](https://docs.unrealengine.com/4.27/en-US/SharingAndReleasing/Linux/GettingStarted/) still targets the host system (Windows) by default, despite its entire purpose being to cross-compile. Maybe this is typical of clang, but it took me ages to realize. Solution: add this the CMake toolchain:

```
set(CMAKE_C_COMPILER_TARGET x86_64-unknown-linux-gnu)
set(CMAKE_CXX_COMPILER_TARGET x86_64-unknown-linux-gnu)
```

* With the above, I was able to build locally, but on Travis the CMake configure step was (still) failing because it was unable to find pthread.h. After adding the `-v` option, I could see that clang was still targeting Windows when doing the test compile to check for the existing of pthread.h. Eventually I figured out that if I downgraded my local CMake from v3.20.x to v3.17.3, it started behaving the same locally as as it did on Travis (i.e. targeting Windows, failing to find pthread.h). `CMAKE_CXX_COMPILER_TARGET` isn't new (at least according to the CMake docs), and nothing jumped out at me in the changelog, so I can't entirely explain this. Solution: add ` -target x86_64-unknown-linux-gnu` to the CMAKE_CXX_FLAGS and CMAKE_C_FLAGS explicitly in the toolchain file, rather than counting on `CMAKE_C_COMPILER_TARGET` to do it for us.

kring:
I think this is ready. There's no mac build yet, but that can come later (when someone with access to a Mac can help).

One potential problem is that we're no longer building Linux Editor binaries for our plugin. That's because the BuildPlugin command refuses to do so when cross-compiling. I confirmed this by reading the source. This line:
https://github.com/EpicGames/UnrealEngine/blob/4.27/Engine/Source/Programs/AutomationTool/Scripts/BuildPluginCommand.Automation.cs#L154
is where the UE4Editor configuration is built. Clearly Linux would have to exist in the HostPlatforms list in order to build that configuration for Linux. The only place HostPlatforms is set is here:
https://github.com/EpicGames/UnrealEngine/blob/4.27/Engine/Source/Programs/AutomationTool/Scripts/BuildPluginCommand.Automation.cs#L109
where it adds (at most) the platform that we're running the build on (i.e. Windows). So, no way to do it short of changing or replacing this code.

And even if we could convince UBT to try to build UE4Editor for Linux, it almost certainly wouldn't work because Linux versions of other Editor binaries aren't included in the Windows install. (If it did, the install would be almost certainly be too big and we wouldn't be able to run it on Travis anyway).

Had I realized all this sooner I may not have worked so hard to get cross-compilation of the Linux version of the plugin working. 🤷 

But actually this may not be an issue at all. Missing binary editors should just mean the Editor will compile them on first run. That requires a compiler, of course, which is not ideal. But I think a compiler would actually be required on Linux even if we provided Editor binaries. The reason is that there are no official builds from Epic on Linux. Which means that everyone using it on Linux will have their own unique build ID. Which means a built plugin will never have the right build ID, and will always need to be rebuilt. At least, this is my current understanding. I'm trying to think how to confirm this, or perhaps I should just ask on UDN. It's telling that when we install our plugin from the Marketplace, it includes editor binaries for Windows and Mac but not Linux. 

Long story short, I think this is "good enough" and perhaps even "as good as it's going to get."

kring:
I'm thinking no one is particularly interested in reviewing this one. Merging.

PR #625, opened 2021-09-02
Choose nearest sublevel in the case of overlap

### LINKED ISSUE #611: Georeferenced sublevels not behaving when two sublevels are close together
Brought to our attention by this user on the forum: https://community.cesium.com/t/overlapping-georeference-datasets/15067

If two different sublevel locations are near enough for their level streaming bounds to overlap, but not near enough to combine in to one sublevel, ping-ponging between the two locations can occur, resulting in neither sublevel being located properly. 

The main action item here is to determine what, if any, appropriate measures can be taken in these situations. In general, it's likely not a super common issue, as it's very dependent on the specific size and location of each level. But if there is a solution, or we can easily modify the plugin to address it (such as potentially allowing users to optionally define a [Level Streaming Volume](https://docs.unrealengine.com/4.26/en-US/BuildingWorlds/LevelStreaming/HowTo/StreamWithVolumes/) rather than just a load radius) we should make sure this option is available and documented for users. 

Some testing is needed to determine if disabling World Bounds Checks helps in this scenario. If so, that's likely the easiest route.

nithinp7:
Fixes #611

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,11 @@
 # Change Log
 
+### v1.6.1 - ????
+
+##### Fixes :wrench:
+
+- Fixed strange behavior when two sublevels overlap each other. Now the closest sublevel is chosen in that case.
+
 ### v1.6.0 - 2021-09-01
 
 ##### Breaking Changes :mega:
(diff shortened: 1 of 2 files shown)


cesium-concierge:

* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium-unreal/blob/overlapped-levels/CHANGES.md) was not updated.

kring:
Thanks @nithinp7!

PR #627, opened 2021-09-03
Add point cloud rendering (support glTF POINTS geometry)

nithinp7:
Uses the Lidar Point Cloud Plugin to render glTF primitives that use the POINTS geometry mode.

Builds on: https://github.com/CesiumGS/cesium-native/pull/334


cesium-concierge:

* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium-unreal/blob/pnts/CHANGES.md) was not updated.

Source/CesiumRuntime/Private/CesiumGltfComponent.cpp
```diff
@@ -1488,10 +1544,70 @@ bool applyTexture(
...
+      *loadResult.pointCloudBuffer,
+      false,
+      FLatentActionInfo(),
+      asyncPointCloudResult,
+      progress,
+      pPointCloud);
+
+  pMesh->SetPointCloud(pPointCloud);
+  pMesh->PointSize = 1000.0f / pMesh->Bounds.SphereRadius;
+  pMesh->PointShape = ELidarPointCloudSpriteShape::Circle;
```
javagl, line 1587: This one apparently became `private` in 4.27, and this line then has to be

    pMesh->SetPointShape(ELidarPointCloudSpriteShape::Circle);

javagl:
When there is only a single point in a point cloud, it cannot compute bounds, and eventually triggers an error

> Provided bounds are incorrect ...

(More generally, this is not only related to the the number of points, but to the actual bounds, which may not be (0,0,0))

cesium-concierge:
Thanks again for your contribution @nithinp7! 

cesium-concierge:
Thanks again for your contribution @nithinp7! 

cesium-concierge:
Thanks again for your contribution @nithinp7! 

cesium-concierge:
Thanks again for your contribution @nithinp7! 

cesium-concierge:
Thanks again for your contribution @nithinp7! 

cesium-concierge:
Thanks again for your contribution @nithinp7! 

j9liu:
I've been trying to bring this up to par so we can have basic point cloud rendering in Unreal, but it's been tricky working with the LIDAR plugin.

Some pros of using the LIDAR plugin include getting point size and physics for free. There is also some classification support built in that we could potentially leverage for feature styling. It's great to have a point-rendering pipeline already figured out for us on our behalf.

However I've run into weird issues with LIDAR that I can't fix well. The plugin never shows all of the points right away. Weirdly enough, they all show up if I tab out of Unreal Engine, and it looks great when that happens:

image

But when I tab back in / move the camera around, it becomes extremely sparse again. Sometimes I zoom in and the level of detail actually gets _worse_:

image

I thought the Lidar LOD system is interfering with the visuals of higher quality tiles, and I'm doing what I can from the public API to minimize this, but it's not enough. Supposedly, the only real way to disable the LOD system is by changing the plugin's Project Settings -> Max Bucket Size to a large number. (See [Changing Data at Runtime](https://docs.unrealengine.com/4.27/en-US/WorkingWithContent/LidarPointCloudPlugin/LidarPointCloudPluginOverview/).) However, doing this doesn't fix the issue at all -- at least, not for all tiles. I get this weird effect where some of the tiles look much better than the others:

image

There's also a "global point budget" that limits the maximum number of points that can be displayed at one time. (See [Performance](https://docs.unrealengine.com/4.27/en-US/WorkingWithContent/LidarPointCloudPlugin/LidarPointCloudPluginOverview/).) This is by default set to 1,000,000 and can only be changed via console. Since Melbourne Point Cloud contains ~300 million points it would make sense if this was the main culprit behind the sparse point rendering. However, I don't know if we can change this setting from code. I also am apprehensive to force changes on the user's development environment because it could have undesired consequences if they're using the LIDAR plugin for something else.

Another prominent issue is that the LIDAR point cloud brings up a notification window every time it does something relatively intensive (re-optimizing the data, creating collision meshes, etc.). In the current implementation, a `CesiumGltfComponent` creates a `CesiumPointCloudComponent` (the points counterpart to `CesiumGltfPrimitiveComponent`) and also creates a new `ULidarPointCloud` for that component. So every single points primitive in a glTF is getting its own LIDAR point cloud object. This didn't seem to have performance implications from a glance, but when I was experimenting with point cloud settings I had the unfortunate consequence of spamming myself with notification windows, which would freeze my Unreal Editor for seconds at a time. It also might have been responsible for the memory leak that made my entire OS severely lag yesterday :grimacing: I'm not 100% certain you can get away with not triggering any functions that cause these windows to appear. Even after I turned off the setting that made it spam notifications, I would still sometimes see hitches in the editor as I panned to view new points tiles.

The last resort I considered was consolidating these multiple smaller point clouds into one large one. It seems _possible_ but it would require a concerted effort of managing which points were visible / invisible / existent / nonexistent in the point cloud. And the constant editing of the point cloud will probably drop performance, even if it is optimized for dynamic editing.

I know I argued for continuing the use of this plugin in offline discussions, but I think taking a step back at the components you'd need to get this to work... it's not possible. Not in a way that's fair to the user, and not in a way that justifies using 3D Tiles to load point clouds instead of the LIDAR point plugin itself. So I think the way to go about it is to write a custom mesh component with a custom scene proxy, vertex factory, etc. to get this to work without fighting the consequences / overhead of code that isn't ours.

PR #629, opened 2021-09-03
Fix fly-to crash when pawn isn't possessed

### LINKED ISSUE #303: FlyToLocation should check if the pawn is possessed first to avoid crashing
There is some functionality in the `GlobeAwareDefaultPawn::FlyToLocation` functions (and maybe other places in `GlobeAwareDefaultPawn`) that assumes there is a controller possessing the pawn. This should be explicitly checked in case `FlyToLocation` is triggered without first possessing the pawn, currently this causes a crash.

nithinp7:
Fixes #303

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,7 @@
 ##### Fixes :wrench:
 
 - Fixed strange behavior when two sublevels overlap each other. Now the closest sublevel is chosen in that case.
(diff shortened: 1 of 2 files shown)


cesium-concierge:

kring:
Thanks @nithinp7!

PR #631, opened 2021-09-06
Do clipping last, after water.

### LINKED ISSUE #630: Clipping does not work correctly in tiles that are partially water
image

I think this is caused by the Water layer being above the clipping layer in the layer stack, and the water layer ignoring the opacity mask when rendering water.

kring:
Our standard "with water" material applied the water effect last, after the clipping. And the water effect only uses the color from the bottom later, replacing all other material attributes with the top (water) layer. As a result, the opacity mask gets clobbered and the clipped effect doesn't work.

The approach implemented here is to simply move the clipping layer to the top of the layer stack, so the opacity mask is applied last.

Fixes #630


cesium-concierge:

kring:
Before:
image


After:
image

nithinp7:
Thanks @kring!

Issue #538, opened 2021-07-03
DynamicPawn causes a divide by zero


javagl:
<sup>I thought that I had already mentioned that, but did not find it right now, so even at the risk of creating a duplicate:</sup> 

The DynamicPawn causes a divide by zero when starting to play:
```
PIE: Play in editor total start time 0,099 seconds.
LogScript: Warning: Script Msg: Divide by zero detected: 83550.765625 / 0
Script call stack:
    Function /CesiumForUnreal/DynamicPawn.DynamicPawn_C:ReceiveTick
    Function /CesiumForUnreal/DynamicPawn.DynamicPawn_C:ExecuteUbergraph_DynamicPawn
    Function /CesiumForUnreal/DynamicPawn.DynamicPawn_C:UpdatePawnSpeed
    Function /CesiumForUnreal/DynamicPawn.DynamicPawn_C:SpeedLineTrace
```

This seems to be due to the "Max Speed Pre Multiplier" being 0 initially:

![CesiumUnreal Pawn Div By Zero](https://user-images.githubusercontent.com/5597569/124365913-a42cf600-dc4b-11eb-80bb-0e9d5c897dc6.png)

I'm not sure whether it can be solved "pragmatically", by initializing it to 1.0 or  adding some epsilon-check at this point, or whether more thought has to be put into that...

PR #636, opened 2021-09-08
Fixed divide by zero in DynamicPawn

javagl:
Supposed to fix https://github.com/CesiumGS/cesium-unreal/issues/538 . My last comment from there:

> I'm not sure whether it can be solved "pragmatically", by initializing it to 1.0 or adding some epsilon-check at this point, or whether more thought has to be put into that...

The latter. The latter was the case. The "Max Speed Pre Multiplier" is initialized with 0.0. The division by zero caused the `True` branch of the condition to be taken, because the result of dividing a value by zero is 0.0, apparently. This branch **had** to be taken, because otherwise, the other properties of the movement component had not been initialized properly at other places (it's complicated). So neither setting it to 0.000001 nor to 10000000.0 solved the issue, and it was indeed necessary to explicitly check for this floating point value to be exactly 0.0 (aka "uninitialized"). 

Before:

CesiumUnreal PawnZero01

After:

CesiumUnreal PawnZero02

If somebody better knows what the purpose of this box is, what the "issue" with loading/unloading was about, or what the unit of `height/speed` is (except for `1/seconds`, but what is that 100?), then I'd appreciate any inputs, suggestions, or PRs with **better** solutions (i.e. solutions that make things *simpler* instead of *more complicated*).

But for now, this should cause the divide-by-zero message to be no longer printed, and still work as before...


cesium-concierge:

* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium-unreal/blob/pawn-divide-by-zero-fix/CHANGES.md) was not updated.

kring:
@xuelongmu can I bug you to take a quick look at this one?

kring:
I removed a Print Debug String node that was spamming the output log and screen, but otherwise this looks good to me. @xuelongmu would still appreciate your thoughts on Marco's questions at the bottom when you have time to breathe from your other project, but it doesn't need to hold up merging this. Thanks @javagl!

Issue #617, opened 2021-09-01
Failure with very long tile URLs because FName length limit exceeded


kring:
As reported here:
https://stackoverflow.com/questions/68992967/ue4-increasing-name-size-limit

We create an ActorComponent for each tile, and we give that ActorComponent a Name derived from the URL / tile ID that was used to load that tile's content. But Unreal Engine apparently has a length limit for FNames, and when we exceed that because a tileset uses a very long URL (more than 1023 characters or so), bad things happen.

To start, we should truncate the name rather than allowing this error to occur. Or maybe we can put the URL somewhere other than the name.

PR #637, opened 2021-09-08
Make sure that FNames are not too long.

javagl:
Or at least, make it "reasonably sure"...

Supposed to fix https://github.com/CesiumGS/cesium-unreal/issues/617

---

The long name for the components is caused by integrating the [`Cesium3DTiles_TileUrl`](https://github.com/CesiumGS/cesium-unreal/blob/9673fad58a0166b900cf0591119730dfb45d94ff/Source/CesiumRuntime/Private/CesiumGltfComponent.cpp#L497) into the name. 

The "name" of a raster overlay that arrives [here](https://github.com/CesiumGS/cesium-unreal/blob/9673fad58a0166b900cf0591119730dfb45d94ff/Source/CesiumRuntime/Private/CesiumGltfComponent.cpp#L1845) is (with some back and forth) derived from the "Material Layer Key" that somebody can enter in the UI. This means that when somebody enters a "Material Layer Key" like `ThisIsSomeVeeeee[...1024 characters...]eeeryLongKey0`, then things might break apart there (Disclaimer: I don't know exactly under which conditions this path is taken, i.e. under which conditions the `pCesiumData` is `nullptr`). Just truncating the string could cause ambiguities, because he might create another key that ends with `...Key1`. But that is so unlikely that I think it won't have to be handled (and if this causes an issue, I'll own it).

---

Two questions for how to handle the length limit in general:

1. Which part of the string should be removed? 

The URL is integrated into the name for information purposes. Truncating it at the end would cause URLs like `https://api.cesium.com/v1/assets/0/1/2/` to be truncated to a useless `https://api.cesium.com/v1/assets` (omitting the relevant differences). But the URL could also be long due to some `&queryParameter`. 

2. What should be the maximum length for `FName` objects that we create?

Yes, 1024, but it's not that simple: When we create an `FName` for some component, then we don't know what Unreal will do with this name. Even if we *dedicatedly* generate a name that has only 1000 characters, then Unreal will use our `FName` to generate a new `FName` internally, as `"/Cesium/Data/Whatever/" + ourName`, causing the length limit to be exceeded again (and they don't check and handle this, by the way...). 

---

Given the fact that we don't know how exactly what causes the long URLs, or long the prefixes will be that Unreal will attach our names to, this PR does two things:

- It will truncate the URL to 256 characters, by removing characters at the *start* of the URL if necessary.
- It will make sure that no `FName` that we are creating is longer than 256 characters. Because our `FNames` are often composed of two components, like `name + "_Texture"` (and we don't necessarily know what `name` is...), this is also done by removing characters at the *start* of `name` to make sure that the limit of 256 characters is obeyed.

Now, Unreal could still append (768+1) characters to our 256-character-`FName`, which could cause a crash, but there is no specified upper limit for the length of an `FName`, so this is somewhat arbitrary anyhow...

---

<sub>Edit</sub>

An example output with some "artificial URL" (where I just added a prefix to make it way longer than 1024 characters): 

CesiumUnreal FName length


cesium-concierge:

* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium-unreal/blob/obey-fname-length-limit/CHANGES.md) was not updated.

kring:
> Which part of the string should be removed?

My thinking (which has probably been influenced by the [original user-reported issue](https://stackoverflow.com/questions/68992967/ue4-increasing-name-size-limit) is that the most likely reason for an excessively long URL will be a key included as a query parameter. And therefore we'd be better off removing the overflow characters from the end rather than the beginning. Of course it's possible to include a key as part of a path instead, or to have a silly-long path for other reasons, but that seems much less likely.

Maybe the _best_ approach would be to replace the middle with `...`, if it's not too much trouble.

> What should be the maximum length for FName objects that we create?

256 sounds fine to me.

> But that is so unlikely that I think it won't have to be handled (and if this causes an issue, I'll own it).

Yeah I'm honestly not fussed about this case at all. What you've done seems reasonable.

javagl:
> Maybe the *best* approach would be to replace the middle with `...`, if it's not too much trouble.

"Better" than computing the longest common subsequence and analyzing the Kolmogorov complexity of the string segments? 😁 

But seriously: The last commit uses a `prefix + "..." + suffix` approach, which I tested locally 



```
void testConstrainLength(const std::string& input, const size_t maxLength) {
	std::cout << "Input    : \"" << input << "\"" << std::endl;
	std::cout << "maxLength: "<< maxLength << std::endl;
	std::cout << "Output   : \"" << constrainLength(input, maxLength) << "\"" << std::endl << std::endl;
}

void testConstrainLength() {

	testConstrainLength("A", 0);
	testConstrainLength("A", 1);
	testConstrainLength("A", 2);
	testConstrainLength("Example", 0);
	testConstrainLength("Example", 1);
	testConstrainLength("Example", 2);
	testConstrainLength("Example", 3);
	testConstrainLength("Example", 4);
	testConstrainLength("Example", 5);
	testConstrainLength("Example", 6);
	testConstrainLength("Example", 7);
	testConstrainLength("Example", 8);
	testConstrainLength("http://example.com/0/1/2/3/4?q=5678", 10);
	testConstrainLength("http://example.com/0/1/2/3/4?q=5678", 20);
	testConstrainLength("http://example.com/0/1/2/3/4?q=5678", 30);
	testConstrainLength("http://example.com/0/1/2/3/4?q=5678", 40);
}
```
having the output
```
Input    : "A"
maxLength: 0
Output   : ""

Input    : "A"
maxLength: 1
Output   : "A"

Input    : "A"
maxLength: 2
Output   : "A"

Input    : "Example"
maxLength: 0
Output   : ""

Input    : "Example"
maxLength: 1
Output   : "E"

Input    : "Example"
maxLength: 2
Output   : "Ex"

Input    : "Example"
maxLength: 3
Output   : "Exa"

Input    : "Example"
maxLength: 4
Output   : "E..."

Input    : "Example"
maxLength: 5
Output   : "E...e"

Input    : "Example"
maxLength: 6
Output   : "Ex...e"

Input    : "Example"
maxLength: 7
Output   : "Example"

Input    : "Example"
maxLength: 8
Output   : "Example"

Input    : "http://example.com/0/1/2/3/4?q=5678"
maxLength: 10
Output   : "http...678"

Input    : "http://example.com/0/1/2/3/4?q=5678"
maxLength: 20
Output   : "http://ex...4?q=5678"

Input    : "http://example.com/0/1/2/3/4?q=5678"
maxLength: 30
Output   : "http://example.../2/3/4?q=5678"

Input    : "http://example.com/0/1/2/3/4?q=5678"
maxLength: 40
Output   : "http://example.com/0/1/2/3/4?q=5678"
```

kring:
>"Better" than computing the longest common subsequence and analyzing the Kolmogorov complexity of the string segments? 😁

Yes, because doing anything like that would require us to keep track of multiple strings so that we can find the longest common subsequence of them, and that would almost certainly be "worse" overall even if in the end it resulted in "better" names. 😆 

Thanks for adding the ellipsis.

PR #641, opened 2021-09-10
Don't get cameras from editor viewports in PIE mode.

### LINKED ISSUE #640: Editor viewports are used for tile selection even in Play-in-Editor mode
Basically, `GetEditorCameras` gets called in any build compiled with Editor support, and it uses any Editor viewports that it finds. This causes some weird LOD behavior, especially when origin rebasing and sublevels come into play because the editor cameras are not affected by either of these things.

AFAICT the reliable way to check if a tileset is running "in the editor" is to check `GetWorld()->IsGameWorld()`. It will return true for a PIE world and for a production game, but false for the regular level editor world. Do not be fooled by `IsEditorWorld()`. That method will return true for both the regular level editor and for PIE.

@nithinp7 this might be the cause of the increased aliasing you mentioned seeing awhile ago.

kring:
Fixes #640 

With this change, Play-in-Editor mode no longer considers the (usually invisible) Editor viewport cameras for tile selection, which is almost certainly the right thing to do. However, if you Eject (or use Simulate-in-Editor mode), then tile selection is _still_ not driven by the Editor viewport, even though now the default Editor viewport is controlling that rather than the player pawn. So, perhaps unexpectedly, tiles won't be selected or shown based on the viewport.

While this may be surprising, I think this is a feature not a bug. When Simulating in Editor, we're showing the world as it is in the game but with the camera ejected so we can view the game world from anywhere we choose. That phantom camera should only observe the world, not change it by influencing tile selection.


cesium-concierge:

nithinp7:
Thanks @kring!

PR #644, opened 2021-09-11
Match Unreal Engine's struct alignment

kring:
This is a fix for an AsyncSystem-related crash we saw on an internal project. The problem turned out to be a disagreement about the location of fields in an Async++ internal struct between the parts of Async++ compiled into cesium-native and the parts compiled directly into the particular Unreal Engine project via inlining. As a result, the wrong memory was read and we'd crash.

Specifically, Unreal Engine includes the `/Zp8` option on 64-bit system, forcing 8-byte struct alignment instead of the 16 bytes that is the usual default on 64-bit Windows. More about this option here:
https://docs.microsoft.com/en-us/cpp/build/reference/zp-struct-member-alignment?view=msvc-160

There's a nice warning in there:
> The C/C++ headers in the Windows SDK assume the platform's default alignment is used. Don't change the setting from the default when you include the Windows SDK headers, either by using /Zp on the command line or by using #pragma pack. Otherwise, your application may cause memory corruption at runtime.

Unreal Engine uses Windows SDK headers (just like every other Windows application, I have to imagine), so using this option seems very ill-advised, but I'm sure there's some reason Epic decided to change a global option like this rather than just setting the alignment on their own structs. 🤷  Here's the line in the UBT that sets the option:
https://github.com/EpicGames/UnrealEngine/blob/a47b87395132f0454c285dc6ec488dece4d45c9c/Engine/Source/Programs/UnrealBuildTool/Platform/Windows/VCToolChain.cs#L489

This PR adds this option for builds of cesium-native for Unreal Engine and that fixes that crash on the internal project.

It's not completely clear to me why projects other than this one internal one didn't have any trouble before. There's certainly a bit of luck involved in terms of what gets inlined and possibly the order of inclusion of Async++ headers or somesuch.

FYI @ggetz @nithinp7 @baothientran I've already merged this change into the `vrd` branch.


cesium-concierge:

kring:
By popular (?) request, here are a few more details about exactly what was causing a crash here...

In Async++, a task continuation function can return another task. Just like our Future continuations can return another Future, if you happen to be familiar with that (ours just wraps the Async++ one). This is called Task Unwrapping, and it's described here:
https://github.com/Amanieu/asyncplusplus/wiki/Tasks#task-unwraping

The critical part in how this is implemented is here:
https://github.com/Amanieu/asyncplusplus/blob/v1.1/include/async++/task_base.h#L483

This function is called when the outer continuation (called the parent in the Async++ code here) completes and returns a task. Async++ has to arrange for any continuations that are supposed to happen after the parent task to run when the inner task (called the child) completes. It does this by putting the parent task into a special "unwrapped" state, and adding a special continuation function (with .then) to the child task that flips the parent to the `completed` state and runs its continuations. The special continuation function has type `unwrapped_func<Result, Child>` and is constructed here.

Now, it is in the `operator()` for this `unwrapped_func` continuation that we will later crash, because we will call it with the wrong `this` address. The address is wrong because of (spoiler alert!) a difference in struct alignment, but let's examine why.

Ok so in this `unwrapped_finish` function we've created an `unwrapped_func` to use as the continuation for the child task. So far so good. Then we pass it to the call to `child_task.then` to actually install it as the continuation. That eventually calls basic_task<Result>::then_internal](https://github.com/Amanieu/asyncplusplus/blob/v1.1/include/async++/task.h#L63) which has this line of code:

```
set_internal_task(cont, task_ptr(new task_func<Sched, exec_func, cont_internal_result>(std::forward<Func>(f), std::forward<Parent>(parent))));
```

Here `f` is our `continuation_func` that will later be a problem. It's being wrapped up in a [task_func<Sched, Func, Result>](https://github.com/Amanieu/asyncplusplus/blob/v1.1/include/async++/task_base.h#L383). This class has a _private_ base class of type `func_holder<Func>`. As it says on the tin, this private base class will be used to hold our (problematic) function. Why not just store the function in a field? A comment explains:

```
// Using private inheritance so empty Func doesn't take up space
```

A lambda with no data has zero size, but a field always has a non-zero size. So Async++ cleverly uses the `func_holder` template to hold the continuation function, and [specializes it](https://github.com/Amanieu/asyncplusplus/blob/v1.1/include/async++/task_base.h#L364) for a function with zero size to not have any field at all.

But in our problematic case, there's certainly some data assocatied with the continuation function, so we get the specialization that [actually stores the Func](https://github.com/Amanieu/asyncplusplus/blob/v1.1/include/async++/task_base.h#L346). And rather than just storing it in a field of type `Func`, it stores it in this rather interesting field:

```
typename std::aligned_storage<sizeof(Func), std::alignment_of<Func>::value>::type func;
```

And then moves it into that field with placement new:

```
new(&func) Func(std::forward<Args>(args)...);
```

Why? I don't know, I haven't dug into this enough to understand why these gymnastics are needed. I suspect it has something to do with avoiding trying to default-construct a continuation function type, which is usually not possible. That's mostly a guess, though. I think understanding this bit deeply could be a long trip down the rabbit hole, but I also think we can safely assume that the Async++ designers didn't do this because they didn't _think of_ using a simple field, and that there's a good reason for this approach even if we don't know exactly what it is. 😆 

Ok so now we can see that our problematic continuation function is being stored in a block of memory reserved with some clearly-alignment-dependent magic from the standard library. `std::aligned_storage` basically provides a block of memory that is big enough to hold `Func` and that meets any of `Func`'s alignment requirements. The problem is, the alignment requirements change when Unreal adds the `/Zp8` compiler option.

So in my debug session going right now, here are the relevant pointer values:

* `this` inside `task_func` == 0x0000029438027e40
* `this` inside `func_holder` == 0x0000029438027e80 (yes, `this` values for the same instance are different in base vs derived methods, C++ is fun)
* `&func` inside `func_holder` = 0x0000029438027e80 (same as `this`)

The last address as where we're storing the problematic continuation function. So far, no problem, though.

Time passes... And finally the inner/child task finishes and Async++ must invoke its continuations, which should include our friend the `unwrapped_func`. The task gets scheduled and is run by [task_base::run](https://github.com/Amanieu/asyncplusplus/blob/v1.1/include/async++/task_base.h#L394). The task is cast to a `task_func<Sched, Func, Result>` with a pointer value of 0x0000029438027e40. That matches the task_func `this` value above, so so far all is well.

Then it calls `get_func` on that instance in order to get the `unwrapped_func` out of that `std::aligned_storage`. Unfortunately this part is inlined, so it's hard to see exactly what is happening. What we do know for sure is that when we finally call operator() on the `unwrapped_func`, its `this` value is 0x0000029438027e70. That's exactly 0x10 (16) bytes _before_ the place in memory where we really stored the function. So we're calling `unwrapped_func` methods on an invalid instance and we crash.

That inlining might seem like it's getting in the way of seeing what's going on, but actually it's more like the _cause_ of what's going on. See, cesium-native here is built in the debug configuration. There's no inlining. The host application, however, is an Unreal Engine application in the Development configuration, so things do get inlined. Therefore, we can conclude that we're currently executing code that was built as part of the UE project. When the function was stored, however, there was no inlining in sight. We were very likely executing code built with cesium-native there.

So, we wrote the function into a field whose address had been determined by cesium-native code with default alignment rules, and we tried to read it with Unreal-built code with non-standard (`/Zp8`) alignment rules. Even though the types are all identical, the underlying memory layout is different and we crash.

Why isn't this problem more widespread? Luck, mostly. With C++ templates, an implementation gets compiled into each compilation unit (CPP file) for each unique template instanstantiation. At link time, the implementations for identical instantations are de-duplicated. There's a separate copy in each object file, which one does the linker use? It..uhh.. just picks one I guess? It shouldn't matter, they're all identical anyway, right? Nope, not if your compiler settings are different. But if the different compiler settings don't affect the type system (and therefore the symbol names), the linker can't tell they're incompatible. If it choose some code from cesium-native and some code built by Unreal, they'll disagree on the alignment as we see here. If it only picks code from one or the other, or if the code generation happens to not be affected by the alignment in a significant way, then it'll work fine.

kring:
Just a bit more to add here. The documentation for the `/Zp8` option led me to believe that any field with a size greater than 8 would result in 16-byte alignment by default and 8-byte alignment with `/Zp8`. But based on a few tests, this isn't true. For example, a struct like this:

```
struct Foo {
  char whatever[40];
};
```

Will have only one byte alignment. That seems pretty sensible, even though the docs don't make it sound that way. Long story short, the number of things affected by `/Zp8` is pretty small. I don't know of any primitive data types that are larger than 8 bytes (long double maybe? I think even that is 8 bytes on x64). So really that only leaves things that have a larger alignment explicitly specified using the `alignas` specifier or similar. And _that_ includes various Async++ types including `task_base`. So that's why the mismatched alignment option didn't cause more problems. Within our code base, it's pretty much only going to affect Async++, and even then only when certain inlining planets align.

nithinp7:
Thanks for the detective work here @kring, great work! Even though this issue has somehow been difficult to reproduce, this explanation comprehensively addresses and resolves the internal issue where this first popped up. Hopefully there are not too many more painful surprises like this that Unreal will throw at us!

PR #645, opened 2021-09-14
Include 4.26 macOS builds in the 4.27 package.

kring:
The 4.26 plugin binaries won't be usable in 4.27, but the cesium-native libs will be and as long as those are available the Editor will be able to recompile the plugin binaries for 4.27 on load (as long as the user has the necessary C++ toolchain installed). This is a temporary workaround until we can get our macOS toolchain sorted for 4.27.


cesium-concierge:

* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium-unreal/blob/macos-427-hack/CHANGES.md) was not updated.

PR #646, opened 2021-09-14
Add Degree-Minute-Second editing support

javagl:
This was only a "side task", because it bugged me that using the Latitude/Longitude sliders in the Editor UI caused the position to change far too quickly, and one had to type in the right value _manually_:

With this PR, in [`CesiumRuntime::StartupModule`](https://github.com/CesiumGS/cesium-unreal/blob/dms-editing/Source/CesiumRuntime/Private/CesiumRuntime.cpp#L33), we register a [`CesiumGeoreferenceCustomization`](https://github.com/CesiumGS/cesium-unreal/blob/dms-editing/Source/CesiumRuntime/Private/CesiumGeoreferenceCustomization.cpp#L29). This establishes a special editing component for the "Details View" of Latitude/Longitude. This editing component is summarized in the main class of this PR, namely the [`CesiumDmsEditor`](https://github.com/CesiumGS/cesium-unreal/blob/dms-editing/Source/CesiumRuntime/Private/CesiumDmsEditor.cpp). 

<sup>(For more context, see the [Unreal Docs about Details Customization](https://docs.unrealengine.com/4.27/en-US/ProgrammingAndScripting/Slate/DetailsCustomization/). I might have violated some "best practices" here, \*shrugs\*. If somebody knows: "Nah, that's not how you should do this", feel free to chime in...)</sup>

The `CesiumDmsEditor` creates the Details View component that allows editing the property in two different ways:

- Using the default component, namely a `SpinBox<double>`, for editing the actual property value in decimal degrees
- Using three (+1) separate components for editing the same property, with spin boxes for Degrees, Minutes, and Seconds (+ the dropdown for N/S or W/E)

Note that this is **only** a user interface change: There is still only a single `UPROPERTY` to store the decimal degrees value. The UI components are just populated using the DMS representation that is computed from this property, or writing the value of the property based on the DMS representation that is edited in the UI.

A short summary of the current appearance and interaction:

CesiumUnreal DMS editing 2

(This is currently a draft: There are still some debug outputs, that will be removed **iff** this PR is about to be merged)

### DIFF
--- a/Source/CesiumRuntime/Private/CesiumDmsEditor.cpp
+++ b/Source/CesiumRuntime/Private/CesiumDmsEditor.cpp
@@ -1,5 +1,7 @@
 // Copyright 2020-2021 CesiumGS, Inc. and Contributors
 
+#if WITH_EDITOR
+
 #include "CesiumDmsEditor.h"
 #include "PropertyCustomizationHelpers.h"
 #include "PropertyEditing.h"
@@ -63,7 +65,7 @@ void CesiumDmsEditor::PopulateRow(IDetailPropertyRow& Row) {
 
   DegreesSpinBox = SNew(SSpinBox<int32>)
                        .MinSliderValue(0)
-                       .MaxSliderValue(179)
+                       .MaxSliderValue(IsLongitude ? 179 : 89)
                        .OnValueChanged(this, &CesiumDmsEditor::SetDegrees)
                        .Value(this, &CesiumDmsEditor::GetDegrees);
 
@@ -158,7 +160,7 @@ double CesiumDmsEditor::GetDecimalDegreesFromProperty() const {
 }
 
 void CesiumDmsEditor::SetDecimalDegreesOnProperty(double NewValue) {
-  UE_LOG(LogTemp, Warning, TEXT("SetDecimalDegreesOnProperty"));
+  UE_LOG(LogTemp, Warning, TEXT("SetDecimalDegreesOnProperty %f"), NewValue);
   DecimalDegreesHandle->SetValue(NewValue);
 
   SignComboBox->SetSelectedItem(
@@ -172,7 +174,7 @@ int32 CesiumDmsEditor::GetDegrees() const {
   return static_cast<int32>(dms.d);
 }
 void CesiumDmsEditor::SetDegrees(int32 NewValue) {
-  UE_LOG(LogTemp, Warning, TEXT("SetDegrees"));
+  UE_LOG(LogTemp, Warning, TEXT("SetDegrees %d"), NewValue);
   double decimalDegrees = GetDecimalDegreesFromProperty();
   DMS dms = decimalDegreesToDms(decimalDegrees);
   dms.d = NewValue;
@@ -187,7 +189,7 @@ int32 CesiumDmsEditor::GetMinutes() const {
   return static_cast<int32>(dms.m);
 }
 void CesiumDmsEditor::SetMinutes(int32 NewValue) {
-  UE_LOG(LogTemp, Warning, TEXT("SetMinutes"));
+  UE_LOG(LogTemp, Warning, TEXT("SetMinutes %d"), NewValue);
   double decimalDegrees = GetDecimalDegreesFromProperty();
   DMS dms = decimalDegreesToDms(decimalDegrees);
   dms.m = NewValue;
@@ -202,7 +204,7 @@ double CesiumDmsEditor::GetSeconds() const {
   return dms.s;
 }
 void CesiumDmsEditor::SetSeconds(double NewValue) {
-  UE_LOG(LogTemp, Warning, TEXT("SetSeconds"));
+  UE_LOG(LogTemp, Warning, TEXT("SetSeconds %f"), NewValue);
   double decimalDegrees = GetDecimalDegreesFromProperty();
   DMS dms = decimalDegreesToDms(decimalDegrees);
   dms.s = NewValue;
@@ -225,3 +227,5 @@ void CesiumDmsEditor::SignChanged(
   double newDecimalDegreesValue = dmsToDecimalDegrees(dms);
   SetDecimalDegreesOnProperty(newDecimalDegreesValue);
 }
+
+#endif // WITH_EDITOR
--- a/Source/CesiumRuntime/Private/CesiumDmsEditor.cpp
+++ b/Source/CesiumRuntime/Private/CesiumDmsEditor.cpp
@@ -128,7 +128,7 @@ void CesiumDmsEditor::PopulateRow(IDetailPropertyRow& Row) {
 
   SecondsSpinBox = SNew(SSpinBox<double>)
                        .MinSliderValue(0)
-                       .MaxSliderValue(59.999999) // TODO This is ugly :-(
+                       .MaxSliderValue(59.999999)
                        .OnValueChanged(this, &CesiumDmsEditor::SetSeconds)
                        .Value(this, &CesiumDmsEditor::GetSeconds);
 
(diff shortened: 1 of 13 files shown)


cesium-concierge:

* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium-unreal/blob/dms-editing/CHANGES.md) was not updated.

kring:
This looks great, let's get it merged.

javagl:
I've marked it as "ready for review", but  

# before merging:

Such a review could cover the code-level as well as aspects of "usability" or UI/layout details. Right now, there's still log/debug output, but if there is no further feedback, I'd just remove that log output (or maybe make it `LogCesium VeryVerbose` or so...).

kring:
Yep I'll review the code and UI (both look good at a glance). Please remove the logging.

javagl:
I have turned the log into `LogCesium VeryVerbose` and added some minor comments.

One thing that still carried a `TODO` was the line at https://github.com/CesiumGS/cesium-unreal/pull/646/commits/971773b0ad04329a5fe5f8d1a772462eb602237a#diff-d7d6af05c0f364914d90a60d1fa369d9831f0163f978390f926e90c359324f31R131 where I set the maximum value of the 'Seconds' slider to `59.999999`, because it should offer the range `[0,60)` instead of `[0,60]` - there doesn't seem to be a "nice" way to achieve that (for obvious reasons). It should be OK, but hope that it does not cause any annoying glitches...

On a more conceptual side: The `DMS` struct and the degree<->DMS conversion functions could be (static private) in the header. I was a bit on the fence with that one: Although it would allow to avoid a few lines of redundant code in the CPP file, this is actually *only* an implementation detail - and even though the class already is in the `Private` folder, I thought it could make sense to treat it as such, and put it into the CPP for that reason...

kring:
Would it be possible to put all of this code in CesiumEditor instead of CesiumRuntime? It might be more trouble than it's worth, but at a glance it looks like it might be very straightforward, in which case I think it's well worth doing.

Source/CesiumRuntime/Private/CesiumDmsEditor.cpp
```diff
@@ -0,0 +1,290 @@
...
+ */
+double dmsToDecimalDegrees(const DMS& dms) {
+  double dd = dms.d + (dms.m / 60.0) + (dms.s / 3600.0);
+  if (dms.negative) {
+    return -dd;
+  }
+  return dd;
+}
+
+CesiumDmsEditor::CesiumDmsEditor(
```
kring, line 97: The abbreviation "dms" is ok for the private `dmsToDecimalDegrees`, and I can even live with the `DMS` struct (but should both of those be in the anonymous namespace?), but for a file and class name, `CesiumDmsEditor` really should be `CesiumDegreesMinutesSecondsEditor` even though it's a mouthful. Though we sometimes bend it, the official Cesium rule is "no abbreviations or acronyms at all."

Source/CesiumRuntime/Private/CesiumDmsEditor.cpp
```diff
@@ -0,0 +1,290 @@
...
+  double decimalDegrees;
+  FPropertyAccess::Result AccessResult =
+      DecimalDegreesHandle->GetValue(decimalDegrees);
+  if (AccessResult == FPropertyAccess::Success) {
+    return decimalDegrees;
+  }
+  // This should never be the case, if the actual prperty
+  // was a double property, so this warning indicates a
+  // developer error:
+  UE_LOG(LogCesium, Warning, TEXT("GetDecimalDegreesFromProperty FAILED"));
```
kring, line 216: Good use for an assertion, then. Especially if we're not going to do anything different when we detect this case, but just return decimalDegrees as we normally would.

javagl:
I'll have to take a look at (or just try out) the registration of the editor components that is currently done in `CesiumRuntime.cpp` at https://github.com/CesiumGS/cesium-unreal/pull/646/files#diff-410b2b68ed714723d92217e76b486fb234cdd08139c62c3f29f0ca02011f871aR32 . A *very* vague concern is that ~"if it is not done correctly, then it might cause some odd dependency from the packaged game to the editor part, which should not be there". But I'll give it a try, and if it's "wrong", this will likely be caught during the review.

---

Totally unrelated (well, to this PR, but not to the reason for this PR), just to have the link (from 2017) here: https://answers.unrealengine.com/questions/566915/view.html

Source/CesiumRuntime/Private/CesiumDmsEditor.cpp
```diff
@@ -0,0 +1,290 @@
...
+      NewValue < 0 ? NegativeIndicator : PositiveIndicator);
+}
+
+int32 CesiumDmsEditor::GetDegrees() const {
+  double decimalDegrees = GetDecimalDegreesFromProperty();
+  DMS dms = decimalDegreesToDms(decimalDegrees);
+  return static_cast<int32>(dms.d);
+}
+void CesiumDmsEditor::SetDegrees(int32 NewValue) {
+  UE_LOG(LogCesium, VeryVerbose, TEXT("SetDegrees %d"), NewValue);
```
kring, line 238: Log statements like this kind of just seem like clutter to me. If you really care about these events you could achieve the same thing without modifying the code by configuring the debugger to log for you:
image
Is there some compelling purpose that isn't obvious to me?

kring (commented):
Looks great, just a few small comments.

javagl:
> Is there some compelling purpose that isn't obvious to me?

Possible user posts/issues like "I entered 123, but it shows 123.0001, why?". But maybe that's not a strong reason for now.

> Would it be possible to put all of this code in CesiumEditor instead of CesiumRuntime? It might be more trouble than it's worth

That certainly makes sense, considering that basically all classes had been wrapped into some (ridiculously global) `#if WITH_EDITOR` anyhow. So now it's part of the `CesiumEditor`. The other inlined comments should be addressed as well.

---

One last thing that bugs me (far more than it should) is that the "Place Georeference Origin Here" button is still shown as the last component, even though it shouldn't. To be clear: It does _not_ bug me that it is shown as the last component (because that doesn't matter so much). It bugs be that it is shown as the last component **even though** it is [*explicitly* and *dedicatedly* added as the first one](https://github.com/CesiumGS/cesium-unreal/pull/646/files#diff-65d644e15de7dc42c22b1514ab95360b7a1a795e8f9c5130985923c783ad36eaR21), and 1. every source that I found said that the properties will be shown in the order in which they are added, 2. that's what one would expect (I mean, what else, seriously?!?), 3. there is no obvious way to affect the order otherwise. If somebody has a hint how to fix that without reading engine code for a few hours, then I'll do that, but I already spent more time than I could reasonably justify for this detail...

kring:
> One last thing that bugs me (far more than it should) is that the "Place Georeference Origin Here" button is still shown as the last component, even though it shouldn't. 

I looked into this briefly, and the problem is that `AddProperty` can't be used with a UFUNCTION. That call has no effect whatsoever. I think that a custom UI needs to be added for the button, and then the UFUNCTION needs to be excluded from being added automatically at the bottom by removing its `EditAnywhere` attribute where it's declared. It's really frustrating that there's no `AddFunction` equivalent to `AddProperty`, but I don't see aything of the sort.

cesium-concierge:
Thanks again for your contribution @javagl! 

cesium-concierge:
Thanks again for your contribution @javagl! 

cesium-concierge:
Thanks again for your contribution @javagl! 

cesium-concierge:
Thanks again for your contribution @javagl! 

cesium-concierge:
Thanks again for your contribution @javagl! 

javagl:
I wonder if this had been merged if there wasn't that issue of the "Place Georeference Origin Here" button.

I'll probably disable the concierge notifications soon. But I think that this editor really could have increased the usability.

cesium-concierge:
Thanks again for your contribution @javagl! 

csciguy8:
Closing this pull request, as it's been over a year since activity. Will create an issue to remind us that there is functionality here worth integrating.

PR #654, opened 2021-10-01
Don't show ion assets until we have a token.

### LINKED ISSUE #647: Token race condition?
Cesium for Unreal seems to (sometimes) use the wrong access token for the assets created through the Cesium ion UI.

I saw this while investigating https://community.cesium.com/t/vricon-3dtilesets-are-not-appearing-in-ue4-scene-editor/15277. I started out logged in to my personal ion account, which doesn't have access to the tilesets mentioned. So I clicked the Sign Out button, then "Connect to Cesium ion", then in the web browser I switched to a different ion account that did have access, hit Authorize, and switched back to Unreal. So far so good.

Then I navigated to a Vricon dataset in the Cesium ion Assets panel and added it and got a 404 error. Weird. I copied down the key that the UI had assigned to the Tileset, and then in the ion web UI I confirmed that this was not the right key for the new account. In fact, it was the key for my personal account that I had just signed out of.

So I signed out again, signed back in, re-added the dataset... and this time it worked fine. In fact, I haven't been able to reproduce the problem since.

Which reminded me that I saw a similar problem awhile ago. In that case, I was adding a new Imagery dataset as a raster overlay, and the UI didn't assign it a token at all (it was just blank). That time, too, just trying again fixed it.

So I'm not sure exactly what's going on here, but it's pretty clear that something is wrong with the token management.

kring:
Fixes #647 

Here's how to reproduce it:

1. Launch the Unreal Editor.
2. Either close the Cesium panel or make it hidden (i.e. by selecting another tab in its tab group, like Place Actors).
3. Open the Cesium ion Assets panel and make it active.
4. Close the Editor, then start it up again. The editor should remember your last state, so the Cesium ion Assets panel should be visible but the Cesium panel should not be. The Cesium ion Assets panel should have a list of assets in it.
5. Select any asset from the list and add it.

The new tileset will be created with a blank access token. If you remove and re-add it, it will have the right token the second time.

The problem was that the Assets panel wasn't ensuring we had a valid token (and loading it if necessary) before allowing assets to be added.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,11 @@
 # Change Log
 
+### v1.6.3 - ?
+
+##### Fixes :wrench:
+
+- Fixed a bug that caused the Cesium ion access token to sometimes be blank when adding an asset from the "Cesium ion Assets" panel while the "Cesium" panel is not open.
+
 ### v1.6.2 - 2021-09-14
 
 This release only updates [cesium-native](https://github.com/CesiumGS/cesium-native) from v0.7.1 to v0.7.2. See the [changelog](https://github.com/CesiumGS/cesium-native/blob/main/CHANGES.md) for a complete list of changes in cesium-native.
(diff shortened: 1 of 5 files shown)


cesium-concierge:

baothientran:
I confirm this fixes it and look good to me

PR #655, opened 2021-10-01
Compute tangents after UVs are set, and initialize LoadModelResult fields

### LINKED ISSUE #652: computeTangentSpace called before texture coordinates are set
CesiumRuntime\Private\CesiumGltfComponent.cpp  530line loadPrimitive function.
now, call computeTangentSpace before updateTextureCoordinates, wrong?
need after updateTextureCoordinates.

kring:
Fixes #652 

In addition to fix the tangent computation order, I noticed and fixed a bug where an unitialized `LoadModelResult::waterMaskTexture` pointer, if it happened to not be null, would make us erroneously think we needed to duplicate vertices and generate tangents.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -159,6 +159,7 @@ jobs:
       homebrew:
         packages:
         - p7zip
+        update: true
     install:
     - python -m pip install --upgrade pip
     - pip3 install --upgrade pip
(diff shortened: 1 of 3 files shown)


cesium-concierge:

* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium-unreal/blob/tangent-space-order/CHANGES.md) was not updated.

baothientran:
this looks good to me, so I'm gonna merge this

PR #656, opened 2021-10-01
Add homebrew update to fix macos build.

kring:
The Travis macOS image doesn't automatically update homebrew, and its p7zip record has a URL that has recently become invalid. This makes all our macOS builds fail to install 7zip, which in turn makes them fail to publish the built package. This PR fixes that by telling homebrew to update.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -159,6 +159,7 @@ jobs:
       homebrew:
         packages:
         - p7zip
+        update: true
     install:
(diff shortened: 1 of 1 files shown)


cesium-concierge:

* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium-unreal/blob/fix-macos-ci/CHANGES.md) was not updated.

Issue #653, opened 2021-09-27
Physics ObjectType not being set on Cesium3DTileset's CesiumGLTFComponent


CesiumBen:
Currently Cesium3DTilesets set collision via BodyInstance struct stored on the Cesium3DTileset, then matches collision settings on the CesiumGLTFComponent stored within:

https://github.com/CesiumGS/cesium-unreal/blob/e194139776f8a1e27d11f26ddb014428f58bc5b2/Source/CesiumRuntime/Private/Cesium3DTileset.cpp#L1204

While it matches the settings of a given ObjectType (ECollisionChannel) through get/set response to channel, it does not actually set the ObjectType of the tileset meaning you can't perform certain operations like [LineTraceMultiByObjectType](https://docs.unrealengine.com/4.27/en-US/API/Runtime/Engine/Engine/UWorld/LineTraceMultiByObjectType/) or [LineTraceSingleByObjectType](https://docs.unrealengine.com/4.27/en-US/API/Runtime/Engine/Engine/UWorld/LineTraceSingleByObjectType/) . So regardless of what object type you set on the body instance currently, the tileset is always channel 0, world static, because it's never set. 

_______________________

VR Dashboards use case, we have a tileset for survey data of a construction site, and also other tilesets that lay on top of the survey layer that represent the target design (what clients want the site to eventually look like). We do line tracing downwards for a number of reasons, and querying by object type / channel is most ideal. We have a solution in place so this isn't an urgent fix but its rather confusing that you can set the object type on the Cesium3DTileset's BodyInstance and it's not being applied to the tileset mesh itself. I'd be happy to open a PR if it's more convenient. 

cc: @nithinp7 @kring @baothientran Thanks!

xuelongmu:
Also wanted to comment about the existence of this method which might eliminate the need to iterate through each collision channel, as is done currently:

https://docs.unrealengine.com/4.27/en-US/API/Runtime/Engine/PhysicsEngine/FBodyInstance/SetResponseToAllChannels/
https://docs.unrealengine.com/4.27/en-US/API/Runtime/Engine/PhysicsEngine/FBodyInstance/GetResponseToChannels/

and this would be the method for setting object type:
https://docs.unrealengine.com/4.27/en-US/API/Runtime/Engine/PhysicsEngine/FBodyInstance/SetObjectType/

[...]

PR #659, opened 2021-10-06
Set collision object type added in

CesiumBen:
This PR relates to #653 and addresses setting physics object types on Cesium3DTilesets based on the FBodyInstance settings. 

Previously only the collision channel responses were being adjusted based on the object type, however the type itself was not set on the tileset with them remaining as WorldStatic collision type objects regardless of the Object Type set on the FBodyInstance settings. 

This would be a great in allowing users to call [LineTraceMultiByObjectType](https://docs.unrealengine.com/4.27/en-US/API/Runtime/Engine/Engine/UWorld/LineTraceMultiByObjectType/) or [LineTraceSingleByObjectType](https://docs.unrealengine.com/4.27/en-US/API/Runtime/Engine/Engine/UWorld/LineTraceSingleByObjectType/) and receive the expected results.


kring:
Thanks for chiming in @AlbanBERGERET-Epic, very helpful to have your input on this!

And thanks for the PR @CesiumBen! It looks good and I'm merging it.

PR #662, opened 2021-10-19
Set C_Cpp.default.configurationProvider VSCode setting.

kring:
Without this, Intellisense doesn't work on my system for cesium-native opened on the extern directory. It shows red squiggles for most `#include` lines because it apparently can't resolve the include directory. It used to work, so I'm not sure what changed, but we already have this setting in cesium-native itself and there should be no harm in adding it here, too.

### DIFF
--- a/extern/.vscode/settings.json
+++ b/extern/.vscode/settings.json
@@ -104,5 +104,6 @@
         "*.idl": "cpp",
         "string_view": "cpp",
         "xmemory0": "cpp"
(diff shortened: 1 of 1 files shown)


cesium-concierge:

* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium-unreal/blob/fix-vscode-intellisense/CHANGES.md) was not updated.

PR #679, opened 2021-11-03
Make raster overlay options configurable.

kring:
Most importantly the screen-space error, which is now independent of the geometry SSE.

Requires CesiumGS/cesium-native#327.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -154,7 +154,7 @@ jobs:
     env: PATH=/c/Python39:/c/Python39/Scripts:$PATH
   - name: MacOS 4.26
     os: osx
-    osx_image: xcode11.3
+    osx_image: xcode12
     addons:
       homebrew:
         packages:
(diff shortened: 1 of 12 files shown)


cesium-concierge:

* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium-unreal/blob/overlays-on-3dtiles/CHANGES.md) was not updated.

kring:
This PR also switches our Travis macOS build to use xcode 12 instead of 11.3, because 11.3 uses macOS 10.14 and is apparently not supported by homebrew anymore, so it does a bunch of compilation from source and times out. 🙄 Using xcode 12 might be ok (it's the "recommended" xcode version for UE 4.26), or it might cause our Epic Marketplace submissions to start failing because cesium-native is built against the wrong xcode. In the latter case... I guess we'll have to figure something out.

nithinp7:
Thanks @kring, great work!

PR #681, opened 2021-11-03
Added custom depth and stencil support to Cesium3DTileset

xuelongmu:
UE4 has a pretty handy custom depth buffer and stencil buffer. They're settings available on any primitive component, but so far it hasn't been exposed to the Cesium3DTileset actor. I added this setting and passed it to the `GltfComponent` so it gets applied when creating the `GltfPrimitiveComponent`.

I figured that setting it on creation would be a better idea than setting it on Tick (even though the engine calls do check if there's a change before marking the component's render state dirty) because of the number of components that might exist in a scene, and I would doubt that users would need to change it at runtime.

To test this feature, you will need to configure the project to support Custom Depth, or Custom Depth with Stencil if you want to test out the stencil feature. Here's a helpful tutorial:
https://medium.com/unreal-engine-material-guides/overlapping-custom-depth-stencils-a084aa763f10

The settings I added are available in the `Rendering` (not `Cesium|Rendering`) section of the actor. After enabling Custom Depth, you should be able to see its output by going to the viewport -> Show -> Buffer Visualization -> Custom Depth or Custom Stencil.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,11 @@
 # Change Log
 
+### ???
+
+##### Additions :tada:
+
+- `Cesium3DTileset` now has options for enabling custom depth and stencil buffer.
+
 ### v1.7.0 - 2021-11-01
 
 ##### Breaking Changes :mega:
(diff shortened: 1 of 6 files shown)


cesium-concierge:

* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium-unreal/blob/custom-depth-stencil/CHANGES.md) was not updated.

nithinp7:
Thanks @xuelongmu! The CI should pass once merged into main.

PR #687, opened 2021-11-16
Add a tile colorizer and an option to output selection stats.

kring:
- Added an ActorComponent for the tile colorizing raster overlay added in https://github.com/CesiumGS/cesium-native/pull/396
- Added a new option "Log Selection Stats" to the debug section of the Cesium3DTileset Actor. When enabled, stats about the tile selection process are printed to the output log with `Display` level. Previously these were always logged but effectively never seen because they were printed with the `VeryVerbose` level.


cesium-concierge:

nithinp7:
Thanks @kring!

PR #688, opened 2021-11-16
Use correct FOV angle for Editor viewports

### LINKED ISSUE #667: Wrong field-of-view for piloted actors
In the Editor, we appear to be accessing the camera field-of-view angle incorrectly. Instead of `FEditorViewportClient::FOVAngle` we should (apparently) be using `FEditorViewportClient::ViewFOV`. This results in an incorrect FOV angle used in LOD computations when piloting an Actor in the Editor.

To reproduce:
1. The Cesium for Unreal Samples level 1 (or any other level really), click the Dynamic Pawn, then the Camera component attached to it, and set the Field of View property to 30 degrees.
2. Right click on the DynamicPawn and choose "Pilot 'DynamicPawn'".
3. Notice that the level of detail in the piloted view is much lower than expected, because Cesium is still doing LOD computations with a FOV of 90 degrees (the value of the `FOVAngle` property) rather than 30 degrees (the actual FOV and value of the `ViewFOV` property)

kring:
Fixes #667

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,6 +9,7 @@
 ##### Fixes :wrench:
 
 - Cesium for Unreal now does a much better job of releasing memory when the Unreal Engine garbage collector is not active, such as in the Editor.
+- Fixed a bug that could cause an incorrect field-of-view angle to be used for tile selection in the Editor.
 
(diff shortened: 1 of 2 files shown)


cesium-concierge:

PR #694, opened 2021-11-29
Add missing #include.

### LINKED ISSUE #674: 'ENGINE_MAJOR_VERSION' is not defined as a preprocessor macro, replacing with '0' for '#if/#elif'
The UE Project recently upgraded from CesiumForUnreal 1.6.0 -> 1.6.3.
I was able to compile our project by using the Engine/Plugins from marketplace. 
But when I keep the CesiumForUnreal plugin in MyProject/Plugins Folder instead and try to compile the project, it fails.

`Error	C4668	'ENGINE_MAJOR_VERSION' is not defined as a preprocessor macro, replacing with '0' for '#if/#elif'	*\Source\CesiumRuntime\Private\CesiumGeoreference.cpp	290	`

I had to add this line to CesiumGeoreference.cpp to compile
`#include "Runtime/Launch/Resources/Version.h"`

kring:
Fixes #674 

It's not needed in all (most?) scenarios, but should be here regardless.

### DIFF
--- a/Source/CesiumRuntime/Private/CesiumGltfComponent.cpp
+++ b/Source/CesiumRuntime/Private/CesiumGltfComponent.cpp
@@ -30,6 +30,7 @@
 #include "MeshTypes.h"
 #include "PhysicsEngine/BodySetup.h"
 #include "PixelFormat.h"
+#include "Runtime/Launch/Resources/Version.h"
 #include "StaticMeshOperations.h"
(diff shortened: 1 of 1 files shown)


cesium-concierge:

* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium-unreal/blob/missing-include/CHANGES.md) was not updated.

PR #696, opened 2021-11-29
Expose raster overlay properties to Blueprints.

### LINKED ISSUE #683: Make raster overlay properties Blueprint-editable
It's possible to create Cesium3DTileset and CesiumSunSky actors via Blueprints at runtime:

image

Unfortunately, it's not possible to add a working raster overlay. We can add the component no problem, but we can't set the properties necessary to make it actually work. Raster overlay properties such as `IonAssetID` and `IonAccessToken` are not Blueprint-editable.

kring:
Fixes #683


cesium-concierge:

PR #697, opened 2021-11-30
Make overlay Add/Remove/Refresh methods accessible from Blueprints.

kring:
Building on #696. This isn't strictly necessary because you can just `Deactivate` and then (re-) `Activate` the component to achieve the same effect. But it's more convenient.


cesium-concierge:

* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium-unreal/blob/overlay-blueprints-functions/CHANGES.md) was not updated.

PR #699, opened 2021-12-01
Go back to xcode 11.3 (instead of 12)

kring:
Epic reported problems building our latest release on macOS:

```
Undefined symbols for architecture x86_64:

"___darwin_check_fd_set_overflow", referenced from:

httplib::detail::SocketStream::is_readable() const in libCesiumIonClient.a(Connection.cpp.o)
httplib::detail::SocketStream::is_writable() const in libCesiumIonClient.a(Connection.cpp.o)
httplib::detail::keep_alive(int, long) in libCesiumIonClient.a(Connection.cpp.o)
httplib::Server::listen_internal() in libCesiumIonClient.a(Connection.cpp.o)

ld: symbol(s) not found for architecture x86_64

clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

This is almost certainly caused by our move to xcode 12 in #682. This PR goes back to xcode 11.3 and then solves the original problem by removing the use of Homebrew entirely. We were only using it to install 7zip, and the built-in zip command is just fine for our needs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -154,12 +154,7 @@ jobs:
     env: PATH=/c/Python39:/c/Python39/Scripts:$PATH
   - name: MacOS 4.26
     os: osx
-    osx_image: xcode12
-    addons:
-      homebrew:
-        packages:
-        - p7zip
-        update: true
+    osx_image: xcode11.3
     install:
     - python -m pip install --upgrade pip
     - pip3 install --upgrade pip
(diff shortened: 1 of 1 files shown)


cesium-concierge:

* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium-unreal/blob/old-xcode/CHANGES.md) was not updated.

PR #713, opened 2022-01-04
Add build settings to support iOS

The-O-King:
CMakeList.txt properly names the iOS built libraries

Added both iOS and Mac toolchain files

Build.cs files now point to proper folder for libraries built for iOS
They also require the use of the c++17 trick used for Android builds as iOS does not currently respect the c++17 cppStandard flag

Update Cesium3DTileset.cpp to get the right file path for the sqlitedb file
**This may run into an engine bug that had to be fixed where the engine did not properly return the right ProjectPersistentDownloadDir() for iOS

Add iOS to the supported platforms list in the uplugin file


cesium-concierge:
Thanks for the pull request @The-O-King!

* :x: Missing CLA.
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium-unreal/blob/ios_support/CHANGES.md) was not updated.

shehzan10:
Thanks @The-O-King for opening this PR, this is definitely an exciting update. We'll test it and also address the questions and you have raised.

To reconfirm my understanding, iOS can only be built using Mac OS X right? And in this case, we're only concerned about x64 hosts, not Apple M1?

Can I also ask you to check if we need to make any project changes to https://github.com/CesiumGS/cesium-unreal-samples so that we can test it out end-to-end?

> ❌ Missing CLA.

I can confirm that we have a CLA signed.

The-O-King:
I just update this PR to make it work without modifying the Engine source for iOS! This involves changing the logic for creating the sqlitedb file to use ProjectSavedDir() rather than PersistentDownloadDir(), but it seemed to work on iOS, Mac, Android, and Windows when I tested things (thanks to my coworker Gabe for the help there :) )

I did run into a strange CMake issue with the iOS; the toolchain doesn't seem to actually work - what does work though is specifying the same CMake variables but through the command line, without the toolchain file:

`cmake -B build_ios_debug -S . -GXcode -DCMAKE_SYSTEM_NAME=iOS -DCMAKE_OSX_ARCHITECTURES=arm64 -DCMAKE_POSITION_INDEPENDENT_CODE=ON -DCMAKE_BUILD_TYPE=Debug`
`cmake --build build_ios_debug --target install`
`cmake -B build_ios_release -S . -GXcode -DCMAKE_SYSTEM_NAME=iOS -DCMAKE_OSX_ARCHITECTURES=arm64 -DCMAKE_POSITION_INDEPENDENT_CODE=ON -DCMAKE_BUILD_TYPE=Release`
`cmake --build build_ios_release --target install`

The specific issue is that CMake can't seem to find pthreads; which is strange - seems to be a somewhat known issue but the solutions described in this [stackOverflow thread](https://stackoverflow.com/questions/14171740/cmake-with-ios-toolchain-cant-find-threads/24380608#24380608) haven't worked for me, can you all test the CMake/these commands on your end and see what works? I was able to confirm this latest update worked in our project no problem though 

Yes you are correct that iOS can only be built with MacOS 

No I didn't have to make any project changes to get the plugin to work on the samples - but I did find that I was getting some crashing on maps other than the first map which I haven't been able to pinpoint the issue just yet (could be Out of Memory related though) - let me know if you see the same things

And do let me know if you have any other questions!

shehzan10:
Thanks @The-O-King for the update. 

> The specific issue is that CMake can't seem to find pthreads;

This is something I remember running into as well. But we'll test it again and retry.

> I did find that I was getting some crashing on maps other than the first map which I haven't been able to pinpoint the issue just yet

Good to know, we'll keep an eye out for it.

shehzan10:
> `cmake -B build_ios_debug -S . -GXcode -DCMAKE_SYSTEM_NAME=iOS -DCMAKE_OSX_ARCHITECTURES=arm64 -DCMAKE_POSITION_INDEPENDENT_CODE=ON -DCMAKE_BUILD_TYPE=Debug`
> `cmake --build build_ios_debug --target install`
> `cmake -B build_ios_release -S . -GXcode -DCMAKE_SYSTEM_NAME=iOS -DCMAKE_OSX_ARCHITECTURES=arm64 -DCMAKE_POSITION_INDEPENDENT_CODE=ON -DCMAKE_BUILD_TYPE=Release`
> `cmake --build build_ios_release --target install`

@The-O-King I was just trying this out, and I'm seeing both the `Debug` and `Release` configs building debug libraries. Would you be able to double check please?

shehzan10:
@The-O-King Got the time to test this out. Here are a few notes:

#### Compiling Instructions
```
git clone --recursive git@github.com:CesiumGS/cesium-unreal.git
cd cesium-unreal
git fetch origin pull/713/head:ios-support
git checkout ios-support
cd extern
cmake -B build_ios_debug -S . -GXcode -DCMAKE_SYSTEM_NAME=iOS -DCMAKE_OSX_ARCHITECTURES=arm64 -DCMAKE_POSITION_INDEPENDENT_CODE=ON -DCMAKE_BUILD_TYPE=Debug
cmake --build build_ios_debug --target install
cmake -B build_ios_release -S . -GXcode -DCMAKE_SYSTEM_NAME=iOS -DCMAKE_OSX_ARCHITECTURES=arm64 -DCMAKE_POSITION_INDEPENDENT_CODE=ON -DCMAKE_BUILD_TYPE=Release  --config Release
cmake --build build_ios_release --target install --config Release
cd /Users/Shared/Epic\ Games/UE_4.27/Engine/Build/BatchFiles
./RunUAT.sh BuildPlugin -plugin="/Users/smohammed/workspace/cesium-unreal/CesiumForUnreal.uplugin" -Package="/Users/smohammed/workspace/packages/CesiumForUnreal" -CreateSubFolder -TargetPlatforms=Mac+iOS
```

Note the `--config Release` needed to build the release libs for Cesium Native.

Once the plugin is compiled, I copied it into the Engine Plugins directory. (Note: I actually first copied it into the Project's plugin directory, the app compiles but the runtime doesn't like that and doesn't load the app. So always use Engine directory)

From there, it was setting up the iOS Certificates and Profiles, and packaging the project.

Since the Mac was remote, I copied the IPA back to my Windows computer and used iTunes to copy the app using copy-paste on the iPhone.

While the app worked and streamed tiles, it was nearly impossible to keep it without crashing for more than 5 seconds on the Melbourne level. Presumably its running out of memory.

CC @kring 

TODO: We need to add project compilation documentation for Mac+iOS similar to what we have for Linux, Windows and Android.

The-O-King:
Hello! Just wanted to follow up and see where exactly we were at with this landing! Is the only issue now the out of memory crashing? If so I think that's unrelated from the actual build setup and perhaps we can tackle that in a different PR?

shehzan10:
@The-O-King As far as getting the build steps to merge in and creating an iOS app - based on the above commands and my testing, it works. I don't think we would call it stable support just yet, as those issues would need to be tackled.

The biggest remainder for now is the build steps documentation - https://github.com/CesiumGS/cesium-unreal/tree/main/Documentation. Since its needed for the Mac side as well, we can add that as a separate PR and get this merged in.

@kring do you have any other concerns of things you want to test for iOS? Would it be possible to set up a CI job for iOS builds? Or did we remove iOS support from the Unreal OSX install?

kring:
This all looks ok to me @shehzan10. I'm not sure offhand if our Unreal for macOS image has OSX support built into it or not, and don't have ready access to a Mac to check.

shehzan10:
I created a branch and added documentation here: https://github.com/CesiumGS/cesium-unreal/tree/ios-support-pr713-for-ci.

However, when running CI, we are running into the same PTHREADS missing error as mentioned above https://app.travis-ci.com/github/CesiumGS/cesium-unreal/jobs/557019685#L304. Not entirely sure whats causing it, and why its not happening on the test Mac that I have access to. I'm trying to recall if I fixed this system-wide, but not entirely sure.

kring:
Sounds like it could be caused by code signing not being set up on Travis?
https://stackoverflow.com/questions/14171740/cmake-with-ios-toolchain-cant-find-threads

shehzan10:
@kring Looks like this commit https://github.com/CesiumGS/cesium-unreal/commit/1de252baa794c927180b63d69f667352d6972251 is at least getting it past the CMake Build Error (travis build in progress as of writing).

The CMake log on travis shows:
```
-- Looking for pthread.h
-- Looking for pthread.h - not found
-- Found Threads: TRUE  
```

Whereas on the test system I'm working with, its getting it as
```
-- Looking for pthread.h
-- Looking for pthread.h - found
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Success
-- Found Threads: TRUE
```

So we'll see if it ends up causing issues with the build or with app deployment.

PR #715, opened 2022-01-05
Expose CesiumGltfPrimitiveComponent.h to public

mmoedinhas:
I'm opening this pull request to ask for this component to be exposed


cesium-concierge:
Thanks for the pull request @mmoedinhas!

* :x: Missing CLA.
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium-unreal/blob/make-primitive-component-header-public/CHANGES.md) was not updated.

mmoedinhas:
I've sent the CLA

shehzan10:
Thanks @mmoedinhas for opening this pull request.

Can you please share how this component being made public helps and what goals it will accomplish that are currently not possible? I wonder if those can be accomplished by exposing the right APIs through the glTF classes.

mmoedinhas:
Thank you for the quick response!

In our project, we're raycasting the tileset to check whether the tile under the camera contains water or terrain. The raycast hits this component in the tileset actor, which has a CesiumGltf::Model in it with the information that we want. Is there any other way to approach this problem?
Before we were doing a DFS through all the Tiles for the one nearest the camera. That works, but sometimes it has buggy results. We were trying a more straighforward approach to this problem by using a raycast, but were blocked by the component being private.

shehzan10:
Thanks @mmoedinhas for elaborating. I think @kring is best positioned to either accept this or suggest alternatives. Kevin is currently on leave and will be back next week. 

If this PR represents any urgency on your end, please feel free to chime in here.

kring:
Thanks for the pull request @mmoedinhas! Your use-case makes sense, but I'm a little wary of exposing UCesiumGltfPrimitiveComponent, because it's an implementation detail that is fairly likely to change in the future.

There's already very similar (but not quite adequate) functionality in CesiumMetadataUtilityBlueprintLibrary. The `GetPrimitiveMetadata` function gets the set of feature metadata corresponding to a particular `UPrimitiveComponent` (e.g. that you get back from a line trace). That function returns FCesiumMetadataPrimitive, which is constructed with the glTF model and primitive; exactly what you need! But they're not exposed publicly. So I think a reasonable approach here is to change that: make it possible to obtain the model and primitive from a FCesiumMetadataPrimitive instance. That should solve your use-case without exposing implementation details.

If you're up for implementing that, great! If not, feel free to keep using the change here as a workaround and we'll write an issue to implement something like this ourselves (but I can't promise when).

Thanks again!

mmoedinhas:
Thank you for your answers. 

I followed your advice and ended up using FCesiumMetadataPrimitive, with a few modifications that I pushed here in case you would like to merge them.

Source/CesiumRuntime/Private/CesiumMetadataPrimitive.cpp
```diff
@@ -12,6 +12,9 @@ FCesiumMetadataPrimitive::FCesiumMetadataPrimitive(
     const CesiumGltf::ExtensionMeshPrimitiveExtFeatureMetadata&
         primitiveMetadata) {
 
+  _model = model;
```
kring, line 15: This should be set using a C++ initializer list, i.e.:

```
FCesiumMetadataPrimitive::FCesiumMetadataPrimitive(
    const CesiumGltf::MeshPrimitive& primitive,
    const CesiumGltf::ExtensionModelExtFeatureMetadata& metadata,
    const CesiumGltf::ExtensionMeshPrimitiveExtFeatureMetadata&
        primitiveMetadata) : _model(model), _meshPrimitive(primitive) {
```

Source/CesiumRuntime/Public/CesiumMetadataPrimitive.h
```diff
@@ -48,10 +48,16 @@ struct CESIUMRUNTIME_API FCesiumMetadataPrimitive {
...
 
+  const CesiumGltf::Model& GetModel() const;
+  const CesiumGltf::MeshPrimitive& GetPrimitive() const;
+
 private:
   TArray<FCesiumMetadataFeatureTable> _featureTables;
   VertexIDAccessorType _vertexIDAccessor;
 
+  CesiumGltf::Model _model;
+  CesiumGltf::MeshPrimitive _meshPrimitive;
```
kring, line 59: Holding the CesiumGltf::Model by value means that the entire model - including all its data buffers - will be copied into the FCesiumMetdataPrimitive at construction, which is going to be really expensive. A better alternative is to hold (const) pointers to these objects.

kring (commented):
Thanks for making the changes @mmoedinhas! Looks pretty good, just a couple of suggested changes.

mmoedinhas:
I've applied your code review suggested changes.

Unfortunately this is not working 100% for our use case and I can't figure out why.
I thought this was working at first, but now I notice that the metadata that is returned with GetPrimitiveMetadata is always empty... Which is weird since the pModel and pMeshPrimitive have values. I've attached a screenshot that shows this.
image

I'm thinking of doing a workaround, which is adding a GetModel to UCesiumMetadataUtilityBlueprintLibrary and stop using the  FCesiumMetadataPrimitive at all. That way I'll get the model "directly". I'll push the change here too, but if you don't want it, just say and I'll revert it. 

Thank you again for your time

mmoedinhas:
I just noticed that CesiumGltf::Model is not exposed to blueprints, which makes my suggested workaround of creating a GetPrimitiveModel in UCesiumMetadataUtilityBlueprintLibrary not make much sense, as the function wouldn't be able to be called from blueprints.
Do you think there are other alternatives?

cesium-concierge:
Thanks again for your contribution @mmoedinhas! 

j9liu:
This PR is pretty outdated (merging into an obsolete `ue4-main`!) so I'm going to close it to keep things tidy. But feel free to reopen discussion in an up-to-date PR with an explanation of the changes, and the motivations behind them.

PR #722, opened 2022-01-13
Improve CesiumSunSky details panel UI

argallegos:
This PR reorders and organizes a lot of the CesiumSunSky variables. When the actor was moved to C++ from Blueprint, the Details panel fell out of order and it became more difficult to find the desired parameters. The CesiumSunSky parameters also became mixed with the Unreal parameters, which made it hard to tell what was part of the actor code and what wasn't.  This change is helpful for user experience as well as creating documentation.

Below left is old. Collapsed categories are default Unreal actor categories that got mixed in with the Cesium categories. Below right is new. 

SunSkyComp

Other notes:

- CesiumGlobeAnchorComponent.h was modified to put it's editor-exposed parameters into a Cesium|Georeference category. This is because the component's parameters were appearing on the base actor, and they're not needed. I wasn't able to find a way to prevent them from appearing there, but by putting them in a subcategory, we can collapse them and keep things neater. This subcategory can be renamed if we want, I wasn't sure if Georeference or Globe Anchor was the better word to use.
- the UFUNCTION `AdjustAtmosphereRadius` was renamed to `UpdateAtmosphereRadius`.
- The UPROPERTY `EnableMobileRendering` was changed to `UseMobileRendering`.
- I did not add a category for anything that wasn't visible in editor. I'm not sure if we should. As far as I know, the category is only for visualization, and should not impact function, so I opted not to, but I'm happy to change that if we should.

Questions before merging:

- Should the Cesium|Georeference category containing the Globe Anchor Component properties be renamed to something else, like Cesium|Globe Anchor?
- Is it ok to leave properties that aren't `EditAnywhere` uncategorized?


cesium-concierge:

* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium-unreal/blob/sunsky-ui-fix/CHANGES.md) was not updated.

kring:
Thanks @argallegos. The changes look good, but we have to watch out for breaking changes. In particular we need core redirects for the renamed properties in order to avoid breaking old assets (they'll lose information on load) or old blueprints (they won't compile anymore). Take a look at `Config/Engine.ini`. There will also be breaking changes from a C++ perspective, but they're harder to fix and we're more willing to live with them, so we can just document them in CHANGES.md.

I don't _think_ the new categories or the property order affect backward compatibility, but we should make sure. Configure a CesiumSunkSky with some known, non-default property values in the previous version of Cesium for Unreal, then load it in a version from this branch and make sure the values all come through correctly.

argallegos:
Thanks @kring! I added the core redirects and updated CHANGES.md. 

I tested out the backwards compatibility of the recategorization. All of the non-default values that I set were preserved after updating to this version, including `UseMobileRendering`. In general, I'll try to be more proactive about UI appearance to hopefully rename things before they are merged in to main, and avoid the need for pedantic changes like this.

kring:
Nah not at all @argallegos, my fault for not paying closer attention to this when working on the new CesiumSunSky. Thanks for making the changes, it looks good!

PR #730, opened 2022-01-19
Removed extra space from engine version in .uplugin

xuelongmu:
There are two spaces in between `EngineVersion` and `4.26.0` in the uplugin file. As a result, the travis pipeline for 4.27 https://github.com/CesiumGS/cesium-unreal/blob/main/.travis.yml#L65 should break at the point where it does an in-place edit from `4.26.0` to `4.27.0`. Perhaps builds are still getting created despite this, but when running that command locally, the `sed` command doesn't succeed.

### DIFF
--- a/CesiumForUnreal.uplugin
+++ b/CesiumForUnreal.uplugin
@@ -10,7 +10,7 @@
 	"DocsURL": "https://cesium.com/learn/unreal/",
 	"MarketplaceURL": "com.epicgames.launcher://ue/marketplace/content/87b0d05800a545d49bf858ef3458c4f7",
 	"SupportURL": "https://community.cesium.com",
-	"EngineVersion":  "4.26.0",
+	"EngineVersion": "4.26.0",
 	"CanContainContent": true,
 	"IsBetaVersion": false,
 	"IsExperimentalVersion": false,


cesium-concierge:

* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium-unreal/blob/fix-uplugin-formatting/CHANGES.md) was not updated.

kring:
Huh... I have no idea why this was working before. The sed is returning success:

```
0.06s$ sed -i 's/"EngineVersion": "4.26.0"/"EngineVersion": "4.27.0"/g' CesiumForUnreal.uplugin
The command "sed -i 's/"EngineVersion": "4.26.0"/"EngineVersion": "4.27.0"/g' CesiumForUnreal.uplugin" exited with 0.
```
(https://app.travis-ci.com/github/CesiumGS/cesium-unreal/jobs/554107767#L2318)

But I just checked and it returns success even if it doesn't actually _do_ anything.

The package files coming out of CI have the correct version and do not have the extra space. My best guess is that the plugin package process is removing the extra space, and fixing the engine version as well.

In any case, good catch @xuelongmu, and thanks!

PR #731, opened 2022-01-20
Cesium ion token management and troubleshooting

### LINKED ISSUE #598: In-editor ion token manager
Continuing off of this issue in Cesium for Unreal Samples - https://github.com/CesiumGS/cesium-unreal-samples/issues/19

The entire ion token user experience needs to be improved somewhat. Some core issues:

- When tokens update or change, there's no warning for users - tilesets just don't load anymore, which leads users to think the whole plugin broke. Of course, it's still functioning exactly as intended, but that's not clear to users. 
- When tokens change, updating them is laborious, and disrupts user flow. Having to update all tokens on all tilesets and raster overlays in all scenes is a pain. 

Previously, we've discussed centralizing this a little bit, and I think it would really benefit us to push this issue. We should consider implementing a new interface (Maybe in the Cesium panel, above or below the Quick Add assets) where users can set a default token. 

Desired user experience:

- User can set a default token in a Cesium panel somewhere. (Perhaps this auto-populates with the user's token used to add new tilesets. Or it does the first time a user logs in, and users can replace it with a different token if they wish, which will then be used to add new ion assets.)
- If a tileset or raster overlay has a blank token, it will use the default token instead. If a token is set on a specific tileset or raster overlay, that token will take priority over the default token. 
- If a token is not working for a tileset, the UI will indicate this - either in the details panel (preferable) or near the default token in the Cesium panel (Or both?) We really need to make it clear to the user when a token is not working, and give them concise instructions to fix the issue.
- If the user is logged out of ion, show a warning in the Cesium panel that they are not connected to ion, and that online tilesets may not appear correctly. This will hopefully also help users of Cesium Samples to change the token.

There are a few big questions, though:

- Can we detect when a token is invalid in editor?
- Can we detect if a project is using the Cesium for Unreal samples tokens? Maybe we can code in something to compare the token to past Cesium for Unreal Samples tokens, and if it matches, show a warning. While not all plugin users use the samples, enough are building experiences off of the samples that it's worthwhile building in a failsafe to the plugin. 
- Is there a way to make a bigger warning? Like, pop up a new window in the editor to complain about invalid tokens? If users don't click on a tileset or have the Cesium panel open, they still wouldn't see a warning otherwise.

kring:
Fixes #598 

This is a major overhaul of how Cesium for Unreal manages Cesium ion tokens. Previously, when a user first logged in to Cesium ion from within the Editor, Cesium for Unreal would automatically and silently create a new Cesium ion token (named after the project) with read access to all of the users' assets. Then when tilesets and overlays were created via the UI, it would use this token for these assets.

This was... not great.

With this PR, a Cesium ion token is no longer created automatically. When you add a tileset or overlay via the UI, you are presented with a panel to select or create a token:

image

For new users, the panel will default to creating a new token with the same name as the one that was created in previous versions (as seen above). If your ion account already has a token with that name, the panel will default to using that existing token.

Using this panel to create or select a token will set the "project default token". It is stored with the project, including committed to source control (it's stored in an ini file in the project's Config directory). It is no longer necessary to set a token on every tileset and ion overlay; the project default token is used automatically when an object-specific token is not specified.

The token can also be found in the project settings: Edit -> Project Settings -> Plugins -> Cesium.

Initially, this token is created with _only_ access to Cesium World Terrain. (Ideally it would be created without access to any assets at all, but Cesium ion does not allow that). Then assets are added via the UI, the token is automatically authorized to access to the new assets, and a message informing the user of this is logged.

This panel appears automatically when no project default token is set, but it's also possible to summon it explicitly, e.g. to change the project default token, by clicking the new  "Token" button in the toolbar:

image

This PR also includes help solving token problems. Whenever a tileset or overlay tries to access a Cesium ion asset and it doesn't work, a "troubleshooting" panel appears:

image

It runs diagnostic tests on the tileset-specific access token (if there is one) and on using this project default token to access this tileset. It also checks whether the specified asset ID exists in the user's Cesium ion account. Depending on the outcome of these checks, it suggests various automatic remedies: (Note that everywhere I say tileset here, it could also be a raster overlay)

* _Use the project default token for this tileset_: When the project default token would work for this tileset, but it has its own token specified instead.
* _Authorize the tileset's token to access this asset_: This asset ID exists in the signed-in Cesium ion account, and the tileset's token is valid and associated with the same user account, but it doesn't allow access to this particular asset.
* _Authorize the project default token to access this asset_: This asset ID exists in the signed-in Cesium ion account, and the project default token is valid and associated with the same user account, but it doesn't allow access to this particular asset.
* _Select or create a new project default token_: This asset ID exists in the signed-in Cesium ion account, but the project default token is either invalid or it doesn't allow access to the asset and we can't fix that because the token is not the in same user account.

If you're not signed in to Cesium ion, the troubleshooting panel will suggest that you connect first:

image

If no automatic remedies are possible, the panel will try to help you understand why:

image

You can also explicitly open this panel by pressing the new "Troubleshoot Token" button on the Tileset and Ion Overlay Details panels:

image


cesium-concierge:

* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium-unreal/blob/token-improvements/CHANGES.md) was not updated.

shehzan10:
@kring This looks fantastic! Can't wait to try it out.

Should we also make a simultaneous update to Samples, or wait for this PR to be merged to do that?

@argallegos Please consider whether this update should be its own "Best Practices for ion Tokens in Unreal" tutorial, or if it can go into the current "Add datasets" tutorial.

argallegos:
This is a huge improvement! 

I'm inclined to say since the panels are so self-explanatory, it will fit in well to the add datasets tutorial. We can also link to the new ion token tutorial.

nithinp7:
@kring This looks great! I have no critiques on the UI, it looks clear and fits with everything else. There have been two issues I've noticed so far:

- There seems to be a non-fatal error that happens when a "Troubleshoot Token" window is open and the project default token is changed. The error only takes place the first time this happens during a session. If the editor is closed and reopened, the issue happens again if the same thing is done with the two windows.
- I haven't had any luck in authorizing a newly created token (created from the UI) to access a tileset. I see the token afterwards in my Ion account and it still shows up in the "Select Existing Token" drop down. Attempting to authorize access to a tileset prints out the console error: `Cannot grant a token to an asset because the token was not found in the signed-in Cesium ion account.`

kring:
Thanks @nithinp7, both of those problems should be fixed now.

nithinp7:
@kring Great, both those issues look fixed to me as well! One more small thing I just found:

If I select a custom project default token and then reopen the editor, the "Use an existing token" selection goes back to the token named after the project. In the project settings, under Plugins->Cesium, both fields relating to the token seem to be blank until I select a token again. So something appears not to save. I'm not sure if it's going back to the token named after the project or if there is no token, CWT still seems to work though. If I have any tileset in the level that was specifically authorized with the token that failed to save, upon reopening the editor the troubleshoot window opens up with:

token-manager-boston

kring:
That should be fixed now too @nithinp7. Saving of settings totally didn't work the way I expected. I also made the token select panel more sensible when not yet signed into Cesium ion.

nithinp7:
This looks good to merge now, thanks @kring! The CI is failing with the same issue as in main, so I am going to merge this PR and then create a separate PR to fix main.

PR #734, opened 2022-01-25
Blueprint API for Registering Custom Cameras

nithinp7:
Custom camera views into Cesium tilesets that aren't automatically picked up can be manually registered now through Blueprint or C++. Extra views can be programmatically updated each frame, in the example below the custom view simply sweeps 360 degrees. This custom camera registration ability may also be useful for AI that needs to "probe" certain sections of tileset geometry to navigate.

##### The custom camera sweeps 360 degrees.
ezgif-2-5007a8e56f

##### Registering a custom camera with the Cesium Camera Manager. A reference to the camera manager and the camera id are saved for later.
camera-manager-0

##### Updating the custom camera, referencing it using the camera id saved earlier.
camera-manager-1

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,10 @@
 
 ### v1.9.0 - 2022-01-03
 
+##### Additions :tada:
+
+- The new `FCesiumCamera` and `ACesiumCameraManager` can be used to register and update custom camera views into Cesium tilesets.
+
 ##### Fixes :wrench:
 
 - Fixed a bug that could cause incorrect LOD and culling when viewing a camera in-editor and the camera's aspect ratio does not match the viewport window's aspect ratio.
--- a/Source/CesiumRuntime/Private/Cesium3DTileset.cpp
+++ b/Source/CesiumRuntime/Private/Cesium3DTileset.cpp
@@ -84,6 +84,7 @@ ACesium3DTileset::ACesium3DTileset()
       _tilesToNoLongerRenderNextFrame{} {
 
   PrimaryActorTick.bCanEverTick = true;
+  PrimaryActorTick.TickGroup = ETickingGroup::TG_PostUpdateWork;
 
   this->SetActorEnableCollision(true);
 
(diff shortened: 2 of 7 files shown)


cesium-concierge:

* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium-unreal/blob/camera-manager/CHANGES.md) was not updated.

kring:
This looks reasonable to me @nithinp7. I suggest we road test it with the two use-cases we know about (one internal, one external) to make sure we've actually solved the problem prior to merging. I do think we'll have to be careful to ensure whatever updates the cameras happens before the tileset Ticks.

nithinp7:
@kring 

Here is a mini-map type of use case building on the sweeping custom camera shown above. The custom camera (a cube with a scene capture component) is manually registered and updated since it does not automatically get picked up. The camera's view is drawn to a UI render target.

SceneCapture  DebugGame  - Unreal Editor 1_26_2022 3_26_27 PM


This video shows this PR working for the movie render queue . Again the user manually registers the camera with the same parameters they send the MRQ. This is a workaround to automatically picking up camera parameters from the MRQ which turns out to be a huge pain. CC: https://github.com/CesiumGS/cesium-unreal/issues/680.

https://user-images.githubusercontent.com/6706316/151414977-633be58a-cc17-4a05-8d42-b02dfc6d2975.mp4

kring:
Thanks @nithinp7! This looks good and I'm going to merge it, but when you get a chance can you paste a screenshot or https://blueprintue.com/ of the Blueprint you used for the two use cases? We'll want a tutorial or at least forum post for this at some point, and those will be useful starting points.

tbrajkovic:
Do we have a tutorial with the entire on how to use Cesium Camera Manager?

kring:
We don't have a tutorial, but this community forum post walks through the process:
https://community.cesium.com/t/tiles-not-loading-at-edge-of-render-area/26884/5

PR #651, opened 2021-09-23
Add chaos collision meshes for UE5

nithinp7:
Fixes compilation errors when targeting UE5 and adds collision mesh generation when the Chaos physics engine is being used.

PR #735, opened 2022-01-25
UE5: Get the plugin working on the latest 5.0 branch

kring:
#651 gets the plugin working in the UE5 early access build that was released last year. This PR includes all of those changes, and also makes further changes that are required to build the plugin for later pre-release versions of UE5. Specifically, https://github.com/EpicGames/UnrealEngine/tree/5.0 at commit 94bd0265ed86d2a4d23542f85c16623d2d189403.

This builds on #731 because cesium-unreal's main currently does not build against cesium-native's main, so merge that first.

Here's a running inventory of what works and what doesn't:

* [x] Loading and rendering tilesets. I don't see any LOD switching artifacts or other dodgy behavior in this version of UE5. 🎉 
* [x] CesiumSunSky
* [x] Georeferencing and world origin rebasing (though it has not yet been explored whether these are still needed now that UE5 can use double precision)
* [x] Sub-levels, using the old world composition system. There's a deprecation warning, but it works fine.
* [x] Painting foliage on terrain
* [ ] Collision: currently you get stuck inside terrain, instead of being prevented from entering it. Reversed normals or winding order?
* [x] UE4: ~currently this branch only builds for UE5~ works now.
* [ ] UE5 Early Access: it may not be possible to support both the Early Access build and the later builds. Hopefully it won't be useful to support early access much longer.
* [ ] Frequent crashes related to material lifetime. But not one of our materials. Unclear if this is something we're doing wrong (we're doing dodgy things with material lifetime, no doubt) or a bug in this version of UE5.


cesium-concierge:

xuelongmu:
> UE5 Early Access: it may not be possible to support both the Early Access build and the later builds. Hopefully it won't be useful to support early access much longer.

Projects made with UE5 Early Access are incompatible with the final shipping version of UE5 anyways, whereas UE5 preview 1 can be thought more of as a beta of the final version. IMO UE5 EA can now be thought of as deprecated since preview 1 is newer and better in every way.

nithinp7:
@kring Since UE5 preview 1 is out, let me know if this is ready for review / merge! Also would it make sense to add the UE5 CI as part of this PR?

kring:
Yes please review and merge @nithinp7! I've started setting up CI for preview 1. UE5 got...bigger, which is making it take a little while.

nithinp7:
I've noticed a few issues so far, not all of which have to necessarily be solved in this PR:
- When creating a new project, I can't figure out where to find the "extended luminance range" option, it doesn't appear to be in project settings any more. I'm sure the option still exists somewhere, we'll just need to update the tutorial once we find its new location.
- I ran into a one-time issue where the dynamic pawn couldn't automatically find the georeference during play. It was just strangely floating in space (or maybe the center of the earth) and couldn't move. I haven't been able to reproduce this so it's probably nothing.
- In play, there are regularly very incorrect shadows that show up. Unlike a previous problem, these shadows don't just flash on the screen for a frame, they actually stay there persistently until the camera moves a good bit.  

ue5-shadows

nithinp7:
Forgot to circle back here, it looks like the shadow problem is related to origin rebasing (big surprise). Turning off origin rebasing seems to fix the issue. As discussed before, we almost certainly do _not_ want to use origin rebasing in UE5 since we have large world coordinates, but we can change the default to off in a future PR.

nithinp7 (approved):
Thanks @kring this looks good to me! I'll merge this now instead of waiting for CI so you can get started on releasing with main. I think the last merge should have been pretty simple, but keep an eye on that if main breaks.

PR #736, opened 2022-01-26
Enabling iOS Build 

shehzan10:
This PR builds on #713 by adding:
* Developer documentation
* Travis Build

@The-O-King Would you mind using the Travis built plugin to run a test on your end as well to make sure everything is working as expected? You can download the `AllPlatforms` build (Note: we only build 4.26 on Travis for OSX/iOS at this time, so even the build marked 4.27 uses builds for 4.26 for OSX/iOS).


cesium-concierge:

* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium-unreal/blob/ios-support-pr713-for-ci/CHANGES.md) was not updated.

Documentation/developer-setup.md
```diff
@@ -23,7 +23,7 @@ There are detailed instructions for setting up a Cesium for Unreal development e
 
 * [Windows](developer-setup-windows.md)
 * [Linux](developer-setup-linux.md)
-* macOS - coming soon
+* [Mac OS X](developer-setup-osx.md)
```
kring, line 26: Pretty sure Apple doesn't call it this anymore. It's just `macOS` now.

Source/CesiumRuntime/Private/Cesium3DTileset.cpp
```diff
@@ -708,25 +709,17 @@ class UnrealResourcePreparer
...
+      FPaths::Combine(*FPaths::ProjectSavedDir(), TEXT("Cesium"));
+  if (!IFileManager::Get().DirectoryExists(*CesiumFolder)) {
+    IFileManager::Get().MakeDirectory(*CesiumFolder, true);
+  }
+  FString CesiumDBFile =
+      FPaths::Combine(*CesiumFolder, TEXT("cesium-request-cache.sqlite"));
+  FString PlatformAbsolutePath =
+      IFileManager::Get().ConvertToAbsolutePathForExternalAppForWrite(
+          *CesiumDBFile);
+  return TCHAR_TO_UTF8(*PlatformAbsolutePath);
```
kring, line 722: Previously we stored the cache in a user-specific engine directory everywhere except Android. That meant all your Cesium for Unreal projects could share a cache, which is pretty useful for e.g. Cesium World Terrain. With this change the cache is per-project. We at least need to call that out in CHANGES.md. And we should be sure it's what we want.

kring (commented):
I'm happy to merge this as-is, but I do have two comments below. One is just a nitpick, but the other we probably need to give a little thought.

I didn't actually try the builds (no Mac or iOS devices), so we should wait for confirmation that they work before merging this, too.

Documentation/developer-setup.md
shehzan10: @kring Good catch. Fixed throughout.

Source/CesiumRuntime/Private/Cesium3DTileset.cpp
shehzan10: @kring I'll leave it to you to decide the action here. If we want to maintain status quo and use per-project cache only for mobile devices, then lets keep that (and I think thats the right move). Feel free to push an update to this branch.

shehzan10:
> we should wait for confirmation that they work before merging this, too.

@kring Agreed

shehzan10:
Made a few more changes:
* Merged in main to build with UE 4.27 (#739)
* Remove POST actions for per-platform builds. Now only `AllPlatforms` builds are reported.
* Renamed `osx` -> `macos` in the intermediate filenames

I was able to download the builds, launch the editor, and package the Samples into a iOS app.

@kring I'm satisfied with this and will turn it over to you to fix the caching path. I think we can merge this PR once thats done.

kring:
Ok, I switched back to the old cache location everywhere but Android and iOS. Merging.

PR #744, opened 2022-02-01
Revert change to Android cache DB location

kring:
#713 made a seemingly sensible change to where the request cache is stored on Android devices. But after that change, apps that use Cesium for Unreal crash on startup on Android. On my phone, the cache filename ends up being `../../../CesiumForUnrealSamples/Saved/Cesium/cesium-request-cache.sqlite`. That sure looks like a relative path to me, despite the fact that it's the result of calling `ConvertToAbsolutePathForExternalAppForWrite` 🤷. My guess is that it's relative to the wrong place or somesuch, so Android won't let us open a file for writing there.

This PR switches back to the previous location on Android. It continues to use the new one on iOS.


cesium-concierge:

PR #748, opened 2022-02-03
fix setting indices 16bit or 32 bit based on number of vertices

joseph-kaile:


cesium-concierge:

* :x: Missing CLA.
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium-unreal/blob/fix-set-indices-16-bit/CHANGES.md) was not updated.

kring:
Good catch @joseph-kaile!

I think it might be wise to disallow std::numeric_limits<uint16>::max() as well, i.e. switch to 32bit indices when the number of vertices is greater than _or equal to_ that value. I'm mostly basing this on [this note in the glTF spec](https://www.khronos.org/registry/glTF/specs/2.0/glTF-2.0.html#meshes-overview):

> The maximum values trigger primitive restart in some graphics APIs and would require client implementations to rebuild the index buffer.

Also, please update CHANGES.md.

PR #756, opened 2022-02-10
No troubleshooting panel in game worlds

### LINKED ISSUE #755: Token troubleshooting interacts poorly with Blueprint-created tilesets/overlays
As reported here:
https://community.cesium.com/t/cesium-access-token-issue-when-spawning-component-at-runtime/17253

When you create tilesets and overlays at runtime with Blueprints (or C++, probably), the Token Troubleshooting panel sometimes pops up for no good reason.

It's not totally obvious how this should all work, but I think these claims at least are hard to argue with:
- Creating a tileset or overlay via Blueprints and setting their properties should not pop up the troubleshooting panel before all the properties have been set, no matter what order you set them in (assuming they're set back-to-back, not in different ticks or something).
- The troubleshooting panel shouldn't ever pop up in Play-in-Editor mode. Or in a shipping game, obviously.

kring:
Fixes #755 

There's more we could do here, but this fixes the most important part by far. With this PR, the Token Troubleshooting panel will no longer appear when in Play-In-Editor (or any other game world). And we'll no longer try to connect to ion and pop up a troubleshooting panel for an asset ID of 0.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,11 @@
 # Change Log
 
+### ? - ?
+
+##### Fixes :wrench:
+
+- The Cesium ion Token Troubleshooting panel will no longer appear in game worlds, including Play-In-Editor.
+
 ### v1.10.1 - 2022-02-01
 
 ##### Fixes :wrench:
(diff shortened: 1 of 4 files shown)


cesium-concierge:

kring:
@joseph-kaile would you mind reviewing and merging this one?

joseph-kaile:
This looks good.

PR #762, opened 2022-02-23
Refactor CesiumGltfComponent.cpp

nithinp7:
Attempts to slightly clean up the hierarchical glTF loading in `CesiumGltfComponent.cpp`. The main goal is to make gltf loading more easily extensible as we grow support for extensions. Hopefully this means future changes / additions to the gltf loading process can be more localized to only the functionality and hierarchy level that they effect.

Main Changes:
- `CreateModelOptions` replaced by several structs for model options, node options, mesh options, and primitive options. Each struct can be extended easily in the future to contain new options (usually from glTF extensions) at that level. For example, nodes usually have a single mesh, but with the MAXAR_mesh_variants extension, an arbitrary number of meshes may be associated with a node. Similarly, in EXT_mesh_gpu_instancing, the node-level extension can indicate that its mesh (and so all the primitives coming from the mesh as well) should be instanced. These are examples of what could eventually go inside some of these options struct, even though most of them are empty for now.
- `LoadModelResult` replaced by several structs (model result, node result, mesh result, and primitive result) to retain the hierarchy of the result of the loading. Similar to the above change, this will help new additions be localized to a specific level (node, mesh, primitive, etc) and the result of that loading will be discernible conveniently at that same level.
- Changed some function / struct names that were misleading about what they were doing. E.g. `LoadModelResult` has been changed to `LoadPrimitiveResult`, `loadModelGameThreadPart` has been changed to `loadPrimitiveGameThreadPart`.
- ... hopefully more!

CC: @kring @baothientran @joseph-kaile 
Do you folks have any suggestions / objections for refactoring this code and how I'm going about it here? I'm open to more suggestions for other things to clean up / refactor in this PR related to the gltf loading.

### DIFF
--- a/Source/CesiumRuntime/Private/CesiumGltfComponent.cpp
+++ b/Source/CesiumRuntime/Private/CesiumGltfComponent.cpp
@@ -529,16 +529,17 @@ FName createSafeName(
 template <class TIndexAccessor>
 static void loadPrimitive(
     std::vector<LoadModelResult>& result,
-    const CesiumGltf::Model& model,
-    const CesiumGltf::Mesh& mesh,
-    const CesiumGltf::MeshPrimitive& primitive,
     const glm::dmat4x4& transform,
-    const CreateModelOptions& options,
+    const CreatePrimitiveOptions& options,
     const CesiumGltf::Accessor& positionAccessor,
     const CesiumGltf::AccessorView<FVector>& positionView,
     const TIndexAccessor& indicesView) {
 
   CESIUM_TRACE("loadPrimitive<T>");
+
+  const Model& model = *options.meshOptions.nodeOptions.modelOptions.pModel;
+  const Mesh& mesh = *options.meshOptions.pMesh;
+  const MeshPrimitive& primitive = *options.pPrimitive;
 
   if (primitive.mode != CesiumGltf::MeshPrimitive::Mode::TRIANGLES &&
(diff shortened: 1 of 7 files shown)


cesium-concierge:

* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium-unreal/blob/gltf-refactor/CHANGES.md) was not updated.

Source/CesiumRuntime/Private/CesiumGltfComponent.cpp
```diff
@@ -1357,21 +1322,31 @@ static std::vector<LoadModelResult> loadModelAnyThreadPart(
     // Show the default scene
     const CesiumGltf::Scene& defaultScene = model.scenes[model.scene];
     for (int nodeId : defaultScene.nodes) {
-      loadNode(result, model, model.nodes[nodeId], rootTransform, options);
+      CreateNodeOptions nodeOptions = options;
+      nodeOptions.pNode = &model.nodes[nodeId];
```
kring, line 1326: I think it'd be a little clearer and safer to construct the CreateNodeOptions from the two things explicitly, like:

```
CreateNodeOptions nodeOptions(options, &model.nodes[nodeId]);
```

kring (commented):
This looks good to me @nithinp7. I only have one nitpick below, and it doesn't even need to hold up merging this. So just move it out of the draft state and I'll merge it.

kring:
👍

PR #768, opened 2022-03-01
Check IsPendingKillOrUnreachable before modifying materials.

kring:
Prior to #129, we sometimes had a crashing at shutdown or when switching levels causes by attempting to modify material parameters while the material was being destroyed. I fixed it by checking `Material->IsPendingKillOrUnreachable()` before attempting to modify parameters. But UE5 doesn't have this method, so in #735 I replaced with with a simple `IsValid`. That's not the same thing, and caused the crash bug to come back. So this PR switches back to `IsPendingKillOrUnreachable` in UE4, avoiding the regression. There are likely still problems in UE5.


cesium-concierge:

* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium-unreal/blob/material-pending-kill/CHANGES.md) was not updated.

PR #775, opened 2022-03-04
Make sure that glTF render data can be created in the worker thread

baothientran:
Sometime glTF sent to unreal is invalid or has unsupported primitive mode (e.g LINE). The PR makes sure that the render data can be created before adding to the list of result

### DIFF
--- a/Source/CesiumRuntime/Private/CesiumGltfComponent.cpp
+++ b/Source/CesiumRuntime/Private/CesiumGltfComponent.cpp
@@ -1168,10 +1168,16 @@ static void loadMesh(
 
   for (const CesiumGltf::MeshPrimitive& primitive : mesh.primitives) {
     CreatePrimitiveOptions primitiveOptions = {&options, &primitive};
+    LoadPrimitiveResult loadPrimitiveResult;
     loadPrimitive(
-        result->primitiveResults.emplace_back(),
+        loadPrimitiveResult,
         transform,
         primitiveOptions);
+
+    // if it doesn't have render data, then it can't be loaded
+    if (loadPrimitiveResult.RenderData) {
+      result->primitiveResults.emplace_back(loadPrimitiveResult);
+    }
   }
 }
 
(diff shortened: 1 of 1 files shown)


cesium-concierge:

* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium-unreal/blob/unsupport-gltf/CHANGES.md) was not updated.

xuelongmu:
Thanks @baothientran! Tested it out and we're able to now load tilesets that were previously crashing. It would be great to expedite the merge of this into main so that our application can pull from that.

Source/CesiumRuntime/Private/CesiumGltfComponent.cpp
```diff
@@ -1168,10 +1168,13 @@ static void loadMesh(
...
-        result->primitiveResults.emplace_back(),
-        transform,
-        primitiveOptions);
+    LoadPrimitiveResult loadPrimitiveResult;
+    loadPrimitive(loadPrimitiveResult, transform, primitiveOptions);
+
+    // if it doesn't have render data, then it can't be loaded
+    if (loadPrimitiveResult.RenderData) {
+      result->primitiveResults.emplace_back(loadPrimitiveResult);
+    }
```
nithinp7, line 1177: Just a nitpick. This causes a copy when inserting into primitiveResults when the render data exists right? Do you think it might be better if we keep the `loadPrimitive(result->primitiveResults.emplace_back(), ...)` but `pop_back` the last result out of the list if we find that the render data is missing afterwards? That way, in the usual case where render data exists, we don't create a copy. The potential realloc of the list might not matter if we call `vector.reserve(...)` in `loadMesh` (which I should have done when I made these changes).

Anyways, if you know the compiler optimizes the copy away somehow or feel this change wouldn't make a difference, feel free to resolve the comment and I'll merge!

baothientran: yup that works. I updated the change in the new commit

PR #778, opened 2022-03-08
Show credits onscreen

### LINKED ISSUE #765: Option to show credits on screen instead of in a popup
CesiumJS allows certain credits to be displayed over the globe instead of in a separate popup. Cesium for Unreal should have the same option. So, a few parts to this:

1. Add an option to cesium-native to allow certain credits to be designated to be displayed on the main screen instead of in a popup.
2. Add an option to Tileset and RasterOverlay to put that objects's credits in one place or the other.
3. Modify the credit system to actually display the credits in the right place.

joseph-kaile:
Fixes #765


joseph-kaile:
I modified the display such that the credits will align to the bottom of the screen. There is no scrollbar, and if the credits get clipped, the user would just need to modify the PositionY and SizeY of the horizontal box.
image

kring:
@joseph-kaile this is really close. On my system (perhaps because of the DPI), the credits look very misaligned:

image

The alignment changes based on the size of the window, so it's probably not possible to fix this with padding. I think the best solution might be to move the Cesium ion logo and the Data Attribution button into the web browser widget, too. And then use HTML to align everything properly. It should be possible to respond to a click on the Data Attribution button as an HTML element just by giving it a fake URL and catching it in the On Before Popup callback.

The Cesium ion logo shouldn't need to be handled specially then too, right?

I also think that credits specified on the tileset and overlays directly (rather than on individual tiles within them) should be treated as attached to every tile for the purpose of counting occurrences. In other words, they should be first. Whether that's actually done by counting or by treating those credits as special, I'm not too fussed about. That should help avoid the awkward Bing logo in the middle of the text credits.

Let me know if any of this turns out to be harder than expected, though.

joseph-kaile:
I put the cesium ion logo and the data attribution button in the web browser and it does look a lot better: 
image

I'm going to see what I can do about the Bing logo.

kring:
Looks great! What do you think about moving the tileset/overlay credits (especially that Bing logo) before the per-tile credits?

joseph-kaile:
Yes, I got distracted by fixing the problem of the cesium.com popping up randomly. I think I finally found the solution. 

I think I was going to initialize, tile credits with the tileset credits, however I think just giving tileset credits higher priority is a better solution.

Source/CesiumRuntime/Private/CesiumCreditSystem.cpp
```diff
@@ -163,16 +163,43 @@ void ACesiumCreditSystem::Tick(float DeltaTime) {
...
+                        "</li>";
+        hasPopupCredits = true;
+      }
     }
     creditString += "</ul></body>";
     Credits = UTF8_TO_TCHAR(creditString.c_str());
-
+    if (hasPopupCredits) {
+      onScreenCreditString +=
+          "<span> </span><a href=\"https://data-attribution-popup\">Data attribution</a>";
```
kring, line 196: I'd suggest using a made-up protocol to make it clear this is not a real web URL. `cesium-unreal://data-attribution-popup` maybe. Unless that doesn't work for some reason that isn't obvious to me? Either way, a comment explaining the weird URL would be helpful.

kring (commented):
In your screenshot, the Bing logo is at the end, which doesn't seem right. Is that because credits on raster overlays aren't considered to be in every tile the way tileset-level credits now are?

Source/CesiumRuntime/Private/CesiumCreditSystem.cpp
joseph-kaile: > In your screenshot, the Bing logo is at the end, which doesn't seem right. Is that because credits on raster overlays aren't considered to be in every tile the way tileset-level credits now are?

I thought that was the correct location because thats the same way CesiumJS does it. The Bing logo is only added once per frame. https://github.com/CesiumGS/cesium/blob/7e70b89cd1a3ced4c57cf8ccdf05f24585043cbc/Source/Scene/GlobeSurfaceTileProvider.js#L361

I tried to imitate CesiumJS as much as possible so the credits display have almost the same behavior.

kring:
👍

PR #786, opened 2022-03-14
Manage FStruct with TUniquePtr instead of raw pointer

baothientran:
This prevents the leak where loaded result from worker thread is getting deleted in the main thread, but doesn't delete any sub resources. The bug can be reproduced by tuning the cache down to 0 and enabling ancestor and sibling preloading options. To make sure it's not the garbage collection being idle, you can use the command `gc.CollectGarbageEveryFrame 1`. You can see the memory increase after about 15 or 20 minutes of moving around. It happens more frequently with larger tileset.

To track the memory, you can run the sample with `-LLM` in the visual studio command line options. Then, in the editor, use command `stat LLM` in the console to see the memory


cesium-concierge:

* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium-unreal/blob/memory/CHANGES.md) was not updated.

kring:
Thanks @baothientran!

Issue #771, opened 2022-03-02
Autogenerate metadata materials 


nithinp7:
Can be built as part of #698 or as a separate PR. 

We should try to build an editor tool to auto-generate the basic outline of a metadata-styling material based on the metadata of tiles currently visible in the editor. The following steps would be needed:
- A way to report unique metadata schema classes of currently visible tiles from a tileset.
- For each selected, unique metadata schema class, users should be able to select specific properties that they want access to in the material.
- This selection will be named and a material can be generated that resolves all the required property values into variables. This material would sample the feature id texture / feature attribute to get the feature id and then use the feature id to index into the needed encoded property arrays to get the needed property values. The initially requested property values will now be ready to use in shader variables of the appropriate type. 
- Users can then edit the generated material with custom styling logic using the requested variables.
- Different combinations of schema classes can be used to generate different materials, the materials can exist in an atlas of sorts. Later when loading tiles, when a particular combination of metadata classes appear together, the corresponding material can be chosen and instantiated from the atlas. 
- Combining this material atlas idea with the existing material layers system may be a bit awkward, since each material in the atlas will need all the other same non-metadata-related material layers. Maybe we can resolve that with another automatic editor-time step that adds each auto-generated metadata layer onto a copy of a base material with existing non-metadata layers, before inserting into the metadata material atlas.

kring:
The first four dot points were implemented in #698. The last two have not yet been implemented. I'll update the issue title accordingly.

PR #787, opened 2022-03-14
Auto-generate material-outlines for EXT_feature_metadata styling

nithinp7:
Builds on #698 
Partially implements #771 

Add button on the `CesiumEncodedMetadataComponent` to auto-generate an incomplete material layer based on the currently selected metadata to encode. This generated material will take care of all the boiler plate code responsible for parsing feature id textures / attributes and performing the per-pixel / per-vertex feature id lookup. The requested property values will be available as the output pins of a custom shader node as plain floats or vectors.

##### After using "Auto Fill" the `CesiumEncodedMetadataComponent` should look something like this, listing the existing metadata in the tileset.
autoFill

##### After picking the required metadata properties and pressing "Generate Material", a material layer accessing the required metadata is created.
autoFill1

##### The output pins of the feature table lookup node can be used to access the metadata property values. They can be used to drive any shading effect. Here we highlight buildings built in 1931. Note that for vertex attributes, this needs to be run on the vertex shader, which means the output has to be run through a vertex interpolator. The auto-generated template material creates an unattached vertex interpolator next to feature tables using per-vertex feature id attributes, as a reminder.
autoFill2

##### Add the material layer to to a material instance, as usual.
autoFill3

##### The final output: the NYC buildings tileset, with all buildings from 1931 highlighted yellow.
autoFill4

Remaining Issues:
- [ ] There seems to be a strange issue with deletion of auto-generated materials causing crashes in some cases.
- [ ] Add auto-generated materials for feature textures. 
- [ ] Occasional issues regarding using vertex interpolators in materials with multiple layers underneath. E.g., vertex interpolator in the metadata layer causes the material to not compile, if there is a base glTF layer. 
- [ ] Auto-generate comments to the generated materials to provide hints for where and where not to edit (the auto-generated boilerplate code should not be edited).


cesium-concierge:

nithinp7:
Going to self-merge this into the base metadata branch, this will allow convenient end-to-end testing in one place. The two branches weren't modular and independent enough to justify keeping them apart.

PR #789, opened 2022-03-14
Delete emissive texture

baothientran:
Some tilesets uses emissive texture for unlit, but we don't delete it properly

### DIFF
--- a/Source/CesiumRuntime/Private/CesiumGltfPrimitiveComponent.cpp
+++ b/Source/CesiumRuntime/Private/CesiumGltfPrimitiveComponent.cpp
@@ -62,6 +62,7 @@ void destroyGltfParameterValues(
       assocation,
       index);
   destroyMaterialTexture(pMaterial, "normalTexture", assocation, index);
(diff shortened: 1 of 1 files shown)


cesium-concierge:

* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium-unreal/blob/texture-del/CHANGES.md) was not updated.

nithinp7:
@baothientran I'll merge it once we can run CI again.

But unrelated to this immediate fix, could it make sense to abstract the set of allocated resources into some sort of structure that retains explicit references to the resources, e.g. a list of texture pointers? It would let cleanup be trivial, e.g. just iterate through texture references and delete resources. And it would remove the strange code mirroring that currently needs to happen between the resource allocation in CesiumGltfComponent and the destruction in CesiumGltfPrimitiveComponent. The `metadata-textures` branch has a similar concept, but only for the metadata resources, not the regular glTF resources.

baothientran:
yeah I think that would work. I created an issue here https://github.com/CesiumGS/cesium-unreal/issues/790, so that we can do it in a different PR

PR #800, opened 2022-03-25
Update third-party.json license

baothientran:
I'm still unsure about MikkTSpace license. They don't seem to have license file on their repo

### DIFF
--- a/ThirdParty.json
+++ b/ThirdParty.json
@@ -5,9 +5,19 @@
     "license": "MIT"
   },
   {
+    "name": "Catch2",
+    "url": "https://github.com/catchorg/Catch2",
+    "license": "Boost Software License 1.0"
+  },
+  {
     "name": "cesium-native",
     "url": "https://github.com/CesiumGS/cesium-native",
     "license": "Apache 2.0"
+  },
+  {
+    "name": "cmake-modules",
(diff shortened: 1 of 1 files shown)


cesium-concierge:

* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium-unreal/blob/third-party/CHANGES.md) was not updated.

shehzan10:
@baothientran Looks like MikkTSpace uses a custom open source license with full text here: https://github.com/mmikk/MikkTSpace/blob/master/mikktspace.h.

A comparable is NVIDIA Omniverse - They simply list the license as "mikktspace License" - https://docs.omniverse.nvidia.com/con_connect/common/licenses.html#mikktspace-license

This is similar to what Unreal Engine also does - https://github.com/EpicGames/UnrealEngine/blob/release/Engine/Source/ThirdParty/Licenses/MikkTSpace_License.txt

So in our case, we should point to the header file and call the license "mikktspace license".

shehzan10:
@baothientran Can you also please sort if by alphabetical order (so some of the new licenses would go in the middle of others)?

shehzan10:
Thanks @baothientran. Looks good to me. I'm merging this without CI since its just a text file change.

PR #802, opened 2022-03-29
Fix On Screen Credits for UE5

joseph-kaile:
Before, to update the credit string it would first load a dummy url and then execute javascript...

Now, it loads the dummy url once during Begin Play:
image

Then, when updating credits, it will clear the screen, and then execute the javascript:
image

I tested it and it works in both UE4 and UE5.


cesium-concierge:

* :x: Missing CLA.
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium-unreal/blob/ue5-fix-on-screen-credits/CHANGES.md) was not updated.

kring:
@joseph-kaile the very first time I did a Play-in-Editor with this branch, I got the Cesium web page at the bottom. But every time afterward it's worked correctly. Could we set the default URL to `about:blank` instead of `https://www.cesium.com`? Would it help?

kring:
My best guess about what's happening here....  the browser is taking ages to start up and load a URL. In the meantime, the tiles for the default view are loaded (from cache), at least sufficiently so that `Credits Updated` is no longer true. Finally the browser gets organized and loads cesium.com. But because "Credits Updated" is false, the loaded page is never replaced with the actual credits. I can force it to happen consistently by adding a delay:

image

I can also trigger it by making the startup page one that will load slowly. An easy way to set that up is to go https://webhook.site , click Edit in the top right, enter `10` in "Timeout before response", and whatever you want in the response body. Then use the "unique URL" as the startup URL in the blueprint. The server will take 10 seconds to respond, and in the meantime the credits will settle down. Then they'll be replaced with the web server's response. This is why it probably happens more for me than you: it takes longer to load cesium.com from Australia.

Hopefully using `about:blank` as the startup will fix this, by loading almost instantly.

jjkaile:
Thanks, that is probably exactly what is going on. The about:blank change certainly works. Thanks for the suggestion.

joseph-kaile:
That exactly what is going on. The `about:blank` change certainly works. Thanks for the suggestion.

kring:
👍

PR #803, opened 2022-03-29
Remove dependency on EditorScriptingUtilities plugin in the metadata-textures branch

kring:
This is a PR into #698. The reason Travis CI fails in that branch is because our dodgy cut down distribution of Unreal Engine doesn't include a bunch of plugins, including EditorScriptingUtilities. But it's relatively easy to remove the dependency in this case, so that's what I've done.

I also removed the code that saves the generated material. I think it's fine to let the Editor consider it unsaved, and to let the user save in the normal way if they would like to do so. But maybe there's a more subtle reason it was necessary to save it explicitly? Everything seemed to work ok, but it's possible I missed something.

### DIFF
--- a/CesiumForUnreal.uplugin
+++ b/CesiumForUnreal.uplugin
@@ -42,10 +42,6 @@
 		{
 			"Name": "Water",
 			"Enabled": true
-		},
-    {
-      "Name": "EditorScriptingUtilities",
-      "Enabled": true
-    }
+		}
 	]
 }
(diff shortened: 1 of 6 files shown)


cesium-concierge:

nithinp7:
Thanks for looking into this @kring, the stale material issue I was worried about does not seem to appear even without auto-saving the material! So any open material editor is automatically updated after the regeneration (really it's closed and reopened). It's probably better not to save anyways, that should be up to the user.

PR #806, opened 2022-03-31
Make use of UE5's double-precision

kring:
This is a PR into the new `ue5-main` branch.

It's a bunch of small changes to take better advantage of UE5's newfound double-precision support, especially in Blueprints. There's more we can do, but this should be a pretty good start.

Also removed UE4 builds from the TravisCI configuration (because this branch is UE5 only). And some other UE5-specific tweaks and cleanup.


cesium-concierge:

PR #823, opened 2022-04-20
Compute much better axis-aligned bounding boxes for tiles, for use by Unreal Engine

kring:
When we create a static mesh, we must provide an AABB and bounding sphere. These bounding volumes aren't used for culling or tile selection in cesium-native, because it has its own (more accurate) bounding volumes, but Unreal Engine uses them for its own frustum and occlusion culling and possibly other things.

The AABB is provided in model coordinates, and then it's transformed into world coordinates as the static mesh component's transformation changes. The problem is that our tile models are often expressed in ECEF (minus a translation), and so they're really poorly aligned with the axes. Therefore, our AABBs, when expressed in model coordinates, are big and cube-like. Once they're rotated into world coordinates, the CesiumGeoreference guarantees that tiles near the origin are _much_ more axis-aligned. But Unreal Engine doesn't have enough information to produce better world-space AABBs, so it still uses the big cube-like ones, only rotated to a different orientation.

Fortunately Unreal Engine provides a virtual method on UPrimitiveComponent, `CalcBounds`, that allows us to calculate a new world-space AABB according to a new transformation in a much more accurate way. Specifically, we use our accurate oriented bounding box, bounding region, etc. to produce a much more accurate world-space AABB. This results in _much_ smaller world-space AABBs in many cases, which should make rendering more efficient.

In the screenshots below, look at the blue lines, which are the outlines of the AABBs, and notice how much smaller the volumes are in the "after" screenshot.

Before:
image

After:
image

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,11 +4,11 @@
 
 ##### Breaking Changes :mega:
 
-- Deprecated parts of the old Blueprint API for feature ID attributes from `EXT_feature_metadata`. 
+- Deprecated parts of the old Blueprint API for feature ID attributes from `EXT_feature_metadata`.
 
 ##### Additions :tada:
 
-- Added a Blueprint API to access feature ID textures and feature textures from the `EXT_feature_metadata` extension. 
+- Added a Blueprint API to access feature ID textures and feature textures from the `EXT_feature_metadata` extension.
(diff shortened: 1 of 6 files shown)


cesium-concierge:

joseph-kaile (approved):
Looks very good. Works well.

PR #828, opened 2022-04-23
More thoroughly sanitize property names for use in Materials

kring:
I'm working on a version of Cesium OSM Buildings that uses `EXT_feature_metadata`, and I tried using the new metadata styling system with it. But some of the most important properties have names like `cesium#estimatedHeight`. The `#` resulted in generation of an invalid material (compilation errors). This PR applies `createHlslSafeName` more consistently to avoid these problems.


cesium-concierge:

kring:
Also fixed a  problem where properties named `height`, `width`, and `propertyIndex` couldn't be used because of conflict with variable names in the custom expression node. Now the variables are prefixed with `_czm_` to reduce the chance of conflict.

kring:
I had originally updated CHANGES.md, but removed it because this PR improves a not-yet-shipped feature.

Source/CesiumRuntime/Private/CesiumGltfComponent.cpp
```diff
@@ -1649,7 +1650,10 @@ static void SetMetadataFeatureTableParameterValues(
        encodedFeatureTable.encodedProperties) {
 
     pMaterial->SetTextureParameterValueByInfo(
-        FMaterialParameterInfo(FName(encodedProperty.name), association, index),
+        FMaterialParameterInfo(
+            FName(createHlslSafeName(encodedProperty.name)),
+            association,
+            index),
```
nithinp7, line 1656: This is how the parameter name is created for encoded property array textures. As a reminder, encodedProperty.name is built as follows: "FTB_(feature table name)_(property name)". 

Two points about the name sanitation:
- Do parameter names need to be sanitized? I haven't seen material parameters not work due to the name (hlsl identifiers on the other hand definitely should be consistently sanitized), but it couldn't hurt to sanitize it anyways.  
- If we are going to sanitize them, we should do the same in `SetMetadataParameterValues` below as well.

Source/CesiumRuntime/Private/CesiumEncodedMetadataComponent.cpp
```diff
@@ -590,22 +569,23 @@ void UCesiumEncodedMetadataComponent::GenerateMaterial() {
       UMaterialExpressionTextureObjectParameter* PropertyArray =
           NewObject<UMaterialExpressionTextureObjectParameter>(
               this->TargetMaterialLayer);
+
+      FString propertyName = createHlslSafeName(property.Name);
+
       PropertyArray->ParameterName =
-          FName("FTB_" + featureTable.Name + "_" + property.Name);
+          FName("FTB_" + featureTable.Name + "_" + propertyName);
```
nithinp7, line 576: Here only the property name is sanitized, instead of the whole parameter name "FTB_featureTableName_propertyName". I'm not sure material parameter names need to be sanitized in general, but either way this is inconsistent with how the names are sanitized in `CesiumGltfComponent::SetMetadataFeatureTableParameterValues`, there the entire parameter name including the feature table name is sanitized.

kring:
Thanks @nithinp7. I got a bit carried away with the sanitization! I've now confirmed that material parameters don't need to be sanitized and reverted the unnecessary/conflicting changes.

nithinp7:
Thanks @kring this looks good to me! Feel free to self-merge it in after CI passes or right away if you don't want to wait.

PR #832, opened 2022-04-27
Use a `TWeakObjectPtr` to store texture pointers

kring:
This is a PR into #828 so merge that first.

#698 originally used `AddToRoot` to make sure metadata UTexture2Ds were not garbage collected, but I convinced Nithin [here](https://github.com/CesiumGS/cesium-unreal/pull/698#discussion_r842171200) that this wasn't necessary. But today I started seeing a crash at exit or level unload, caused by attempting to destroy a metadata texture that was already destroyed, so I was probably wrong.

Here is what's happening.

We create a UTexture2D and store a pointer to it in `LoadedTextureResult`. A pointer to a UObject in regular old struct like that doesn't prevent the UE garbage collector from destroying this texture. Nor will that pointer be set to nullptr when that happens. That's usually ok-ish, because we also set a material parameter with that texture, and so the material will keep the texture from being garbage collected. However, if the texture happens to not be used at all, then it will get garbage collected and we'll crash later when we try to explicitly destroy the dangling pointer in `LoadedTextureResult`.

Now I'm not 100% sure why the texture ends up unused in this particular case. Based on a very briefly look at the Engine code, setting a material parameter should keep the texture alive even if that parameter doesn't really exist in the Material. But maybe the Engine "later" realizes the parameter doesn't exist and frees the assocaited resources? It could also just be timing; the Material and therefore the texture get garbage collected earlier than the CesiumGltfComponent.

In any case, the solution implemented here is to store a `TWeakObjectPtr` in `LoadedTextureResult`. That way, we don't keep the texture from being garbage collected, but when it is, the pointer will be safely set to nullptr.


cesium-concierge:

nithinp7:
Thanks for investigating this @kring, this looks good to me! I think this bug exclusively exists in the context of metadata styling, since it is the only case (up until now) where a single texture can be referenced from multiple primitives, and more importantly referenced by none at all. So since this is a fix to a not-yet-shipped bug, it might not necessarily need to be in the changelog, what do you think?

kring:
Ok, I removed it.

PR #833, opened 2022-04-28
Verify metadata indices in Blueprint functions

kring:
Builds on #828 so merge that first.

Previously, if we passed an invalid array index or feature ID to one of the Blueprint metadata functions, it would assert if we're lucky, or more likely read undefined memory. Now it returns the default value.


cesium-concierge:

nithinp7:
@kring The indices are signed (looks like we can't have unsigned numbers in blueprint), so should we check that they are >= 0 as well?

kring:
Good idea, done!

Issue #834, opened 2022-04-28
Best practices for including sample materials


argallegos:
Currently, we have two sample materials for the new metadata styling functionality included with the plugin: https://github.com/CesiumGS/cesium-unreal/tree/ue5-main/Content/Materials/MetadataStyling

Other sample materials have usually been included as part of the Cesium for Unreal Samples project. Would the metadata styling materials be better included in the Samples? If not, should any of the Samples materials be integrated back into the plugin?

argallegos:
I'll move these materials to the Cesium for Unreal Samples tomorrow.

PR #835, opened 2022-04-29
move metadata mats out of plugin

argallegos:
This PR simply removes the new metadata example materials from the plugin folder. They will be included in the Cesium for Unreal Samples instead. 

1/2 of the fix for https://github.com/CesiumGS/cesium-unreal/issues/834. 

I wasn't sure what the new process is now that we have a UE4-main and UE5-main branch. Please rebase this PR to merge into the appropriate branch if needed.


cesium-concierge:

nithinp7:
Thanks @argallegos! I'm now second-guessing if we should remove these, even though I originally told you we should. I wonder if it would serve as a useful example for how a metadata material should look, without having to download / open up the entire samples project. @kring Do you have an opinion here.

If we remove these though, we should also remove the MetadataStyling folder since I don't think there's anything else in it.

kring:
I don't have a strong opinion either way. Creating a new material instance is pretty easy! It's just:

1. Create a new blank material isntance by right-clicking in the content browser and choosing Materials & Textures -> Material Instance.
2. Double-click to open the new material instance.
3. On the Details panel (should pop up by default), set the Parent property to M_CesiumBaseMaterial. This will automatically populate the default layers: Clipping, Overlay1, Overlay0, Background.
4. Rename one of the existing layers to "Material" and change the Layer Asset to the generated metadata material layer. Or alternatively, add a new layer by imitating one of the existing ones.

So I think with a tutorial covering this, it's not so important to have an out-of-the-box material instance built into the plugin.

I guess I'm inclined to err on the side of not shipping it, because removing it later is a breaking change. So I'm going to merge this.

kring:
> If we remove these though, we should also remove the MetadataStyling folder since I don't think there's anything else in it.

I don't think there's any way in git to remove a directory. Empty directories are automatically removed.

nithinp7:
@kring Was this intended to go into ue5-main? My understanding was we would make all changes that apply to both ue4 and ue5 in the ue4-main branch and then merge that into ue5-main, is that right?

PR #837, opened 2022-05-02
Stop encoding metadata after deleting metadata component

### LINKED ISSUE #836: Deleting EncodedMetadataComponent doesn't remove styling
Even after refreshing the tileset, deleting an EncodedMetadataComponent doesn't stop metadata from being saved into textures. This is the responsible line:
https://github.com/CesiumGS/cesium-unreal/blob/a949dc8bd8837ac7b4c8fc712d997e8d8644a302/Source/CesiumRuntime/Private/Cesium3DTileset.cpp#L803-L809

We should set the _encodedMetadataDescription to be empty if there is not EncodedMetadataComponent during LoadTileset.

nithinp7:
Fixes #836


cesium-concierge:

PR #838, opened 2022-05-06
Update ThirdParty.json

kring:
No longer includes the transitive dependencies of cesium-native. Also updated the MikkTSpace license to indicate it's Zlib, not something special to MikkTSpace.

### DIFF
--- a/ThirdParty.json
+++ b/ThirdParty.json
@@ -1,112 +1,20 @@
 [
   {
-    "name": "Async++",
-    "url": "https://github.com/Amanieu/asyncplusplus",
-    "license": "MIT"
-  },
-  {
-    "name": "Catch2",
-    "url": "https://github.com/catchorg/Catch2",
-    "license": "Boost Software License 1.0"
-  },
-  {
     "name": "cesium-native",
     "url": "https://github.com/CesiumGS/cesium-native",
-    "license": "Apache 2.0"
-  },
-  {
-    "name": "cmake-modules",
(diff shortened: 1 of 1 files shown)


cesium-concierge:

ThirdParty.json
```diff
@@ -1,112 +1,20 @@
...
-  {
-    "name": "tinyxml2",
-    "url": "https://github.com/leethomason/tinyxml2",
-    "license": "zlib"
-  },
-  {
-    "name": "uriparser",
-    "url": "https://github.com/uriparser/uriparser",
-    "license": "BSD-3-Clause"
+    "license": ["Zlib"]
```
shehzan10, line 18: @kring Can you share where you found the zlib license?

I believe MikkTSpace uses a "custom" Permissive License. See discussion in https://github.com/CesiumGS/cesium-unreal/pull/800#issuecomment-1079560416. If you have a different authoritative source, happy to pick zlib too.

kring: It doesn't say zlib anywhere, but that text in mikktspace.h _is_ the zlib license:
https://opensource.org/licenses/Zlib

shehzan10: Got it, thanks for confirming!

kring:
Oops realized I opened this PR into ue5-main instead of ue4-main. Maybe ue4-main should be the default? This is such a pain. In any case, I've now cherry picked the contents of this PR into ue4-main as well.

PR #841, opened 2022-05-09
Fix incorrect computation of bounding boxes

### LINKED ISSUE #839: Bounding boxes are incorrect away from the georeference origin
https://user-images.githubusercontent.com/29616723/167078082-133730dd-8bd5-4ff4-a698-cfac5af3c766.mp4

https://docs.unrealengine.com/4.27/en-US/TestingAndOptimization/PerformanceAndProfiling/Visualizers/OutOfBoundsPixels/

kring:
Fixes #839

I "improved" the bounding box computation in #823. It looked surprisingly right considering how actually wrong it was. 😆
Generally it worked ok near the georeference origin, but got very bad for tiles that are not well aligned to the Unreal axes.

I think it's correct now, but it would definitely be wise for someone to double-check my math when reviewing this!

### DIFF
--- a/Source/CesiumRuntime/Private/CesiumGltfPrimitiveComponent.cpp
+++ b/Source/CesiumRuntime/Private/CesiumGltfPrimitiveComponent.cpp
@@ -183,14 +183,10 @@ struct CalcBoundsOperation {
         glm::max(glm::length(halfAxes[0]), glm::length(halfAxes[1]));
     sphereRadius = glm::max(sphereRadius, glm::length(halfAxes[2]));
 
-    FVector xs(halfAxes[0].x, halfAxes[1].x, halfAxes[2].x);
-    FVector ys(halfAxes[0].y, halfAxes[1].y, halfAxes[2].y);
-    FVector zs(halfAxes[0].z, halfAxes[1].z, halfAxes[2].z);
-
     FBoxSphereBounds result;
     result.Origin = VecMath::createVector(center);
     result.SphereRadius = sphereRadius;
-    result.BoxExtent = FVector(xs.GetAbsMax(), ys.GetAbsMax(), zs.GetAbsMax());
+    result.BoxExtent = FVector(sphereRadius, sphereRadius, sphereRadius);
     return result;
   }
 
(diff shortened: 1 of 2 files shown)


cesium-concierge:

kring:
After this is merged into ue5-main, post an update to https://community.cesium.com/t/linetrace-hit-detection-causing-value-to-be-world-origin/18478/3 to ask the user to check whether it fixes their problem as well.

Source/CesiumRuntime/Private/CesiumGltfPrimitiveComponent.cpp
```diff
@@ -183,14 +183,10 @@ struct CalcBoundsOperation {
...
 
-    FVector xs(halfAxes[0].x, halfAxes[1].x, halfAxes[2].x);
-    FVector ys(halfAxes[0].y, halfAxes[1].y, halfAxes[2].y);
-    FVector zs(halfAxes[0].z, halfAxes[1].z, halfAxes[2].z);
-
     FBoxSphereBounds result;
     result.Origin = VecMath::createVector(center);
     result.SphereRadius = sphereRadius;
-    result.BoxExtent = FVector(xs.GetAbsMax(), ys.GetAbsMax(), zs.GetAbsMax());
+    result.BoxExtent = FVector(sphereRadius, sphereRadius, sphereRadius);
```
nithinp7, line 189: There's still a comment that says the sphere only has to reach the side of the box, is that still applicable / correct?

Also is all the stuff with the half-axes necessary in the BoundingSphere conversion, can't we just convert the sphere radius from meters to centimeters? Could we assume getTilesetToUnrealWorld is a rigid transform + a meters to centimeters scaling?

nithinp7:
@kring This seems to fix the issue and looks good to me! Only left one (probably unnecessary) comment.

Source/CesiumRuntime/Private/CesiumGltfPrimitiveComponent.cpp
kring: I was trying to avoid making any assumptions about the transformation, since it's just a parameter to the method. So instead I construct a bounding box around the sphere, transform the box, and then build a sphere inset in the box. The comment is referring to the fact that the sphere only needs to be inset in the box, rather than fully contain it. There could very well be a better way to do this, but I think it works reasonably well.

PR #842, opened 2022-05-09
Don't crash when destroying a material that isn't the right type

### LINKED ISSUE #717: Crash after applying a material to a tileset with drag-and-drop
Steps to reproduce:
1. Drag a material onto a tileset from the Content Browser. The material should apply to the tileset.
2. Click on the tileset and find the Material parameters in the details panel
3. Try to set the Material or Water Material via drag-and-drop or using the dropdown in the parameter
4. Editor will immediately crash.

When drag-and-dropping a material onto the tileset, it will apply to the tileset, but doesn't seem to be applying properly. It won't show up in the Details panel on that tileset, and sometimes won't apply to all tiles (random other tiles will use the old material still). So whatever the editor is doing to change the material might be fighting with what we do to change the material.

We should probably either prevent drag-and-drop material changing at all, or else make sure that when a material is drag and dropped it is applied properly in the tileset code too.

Sorry, no logs that I've been able to find for this one.

kring:
Fixes #717 

When you drag/drop a material on a tileset in the viewport, Unreal applies that material to all of the actor's static meshes. This is not very useful because Cesium for Unreal creates new static meshes all the time, and it won't use that new material when it does so. So basically the material change is just temporary.

However, if you then set the material property on the Tileset itself, our code will try to destroy the old Material. And it does so in such a way that if the old material that was applied to a static mesh isn't derived from `UMaterialInstanceDynamic` (as ours are), it will end up dereferencing a nullptr and crash.

So this PR simply guards against the crash. I looked briefly into what it would take to make dragging/dropping a material properly apply the material to the tileset, but it doesn't look to be straightforward. The editor behavior looks fairly hard-coded here.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,11 @@
 # Change Log
 
+### ? - ?
+
+##### Fixes :wrench:
+
+- Fixed a bug that could cause a crash after applying a non-UMaterialInstanceDynamic material to a tileset.
+
 ### v1.13.1 - 2022-05-05
 
 ##### Breaking Changes :mega:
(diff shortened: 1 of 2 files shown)


cesium-concierge:

nithinp7:
Thanks @kring!

PR #843, opened 2022-05-09
Update developer setup instructions to look for INSTALL at the root.

kring:
As reported on the community forum:
https://community.cesium.com/t/development-setup/16792

Our setup instructions say to look in the Visual Studio Solution Explorer for a folder called `CMakePredefinedTargets`. But that doesn't seem to exist anymore, probably as a result of a newer version of either Visual Studio 2019 or CMake. Instead, the INSTALL target is just sitting at the top level. So I updated the instructions accordingly.


cesium-concierge:

PR #847, opened 2022-05-10
Initialize fields of UStructs.

### LINKED ISSUE #846: Fix property "is not initialized properly" warnings
UE5 (and maybe earlier versions, but I never noticed) writes messages like this into the Output Log:

```
LogClass: Warning: ObjectProperty FCesiumRasterOverlayLoadFailureDetails::Overlay is not initialized properly. Module:CesiumRuntime File:Public/CesiumRasterOverlayLoadFailureDetails.h
```

If I understand correctly, these are fields in UStructs that are not given a default value, and so they risk being uninitialized. Unreal Engine always uses the release version of the C runtime library, so these uninitialized fields will very often be zeros and we won't notice anything wrong. But if a pointer, for example, happens to get a non-zero initial value, it might very well lead to a crash.

So we should fix all of these warnings.

kring:
Fixes #846


cesium-concierge:

PR #853, opened 2022-05-16
Use proper labels for self-hosted runners.

kring:
I've enabled the [runner_enable_workflow_job_labels_check](https://github.com/philips-labs/terraform-aws-github-runner#input_runner_enable_workflow_job_labels_check) option in our self-hosted Actions runner config so that it won't spin up new EC2 instances for builds that don't need the self-hosted runner (i.e. macOS, iOS, Combine). But with that option enabled, the set of labels has to match exactly. So this PR adds the necessary labels.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -2,7 +2,7 @@ name: Cesium for Unreal
 on: [push]
 jobs:
   Windows:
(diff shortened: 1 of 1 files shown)


cesium-concierge:

PR #854, opened 2022-05-17
Don't copy broken UE4 builds to S3

kring:
So we don't fill up S3 with broken builds. Because of the way the UE build process works, failed plugin builds are much larger than successful ones.


cesium-concierge:

kring:
Travis refuses to run on this branch, possibly because I originally opened it against ue5-main. I'm going to close it and start over.

PR #856, opened 2022-05-17
Don't publish broken builds.

kring:
Moves package upload to after_success. So we don't fill up S3 with broken builds. Because of the way the UE
build process works, failed plugin builds are much larger than successful ones.


cesium-concierge:

PR #868, opened 2022-06-01
Display Credits without Web Browser Widget

joseph-kaile:
Instead of using the Web Browser Widget, this uses Unreal's Rich Text Format. The images work, hyperlinks work, popup credits work.

Some things that still need to be completed:
- [X ] The new HTML Tidy library doesn't seem to be working in GitHub actions.
- [ X] Add bullets to separate credits
- [X ] Third party attribution

Some issues currently:

-  Hyperlink text doesn't wrap
- The html tidy library builds generates and installs unneeded pkg-config files 
- Html strings that are just text, i.e. "2022 Microsoft Corporation", will cause html tidy to crash. Worked around it by adding <span> around it.


cesium-concierge:

* :x: Missing CLA.

Source/CesiumRuntime/Private/CesiumCreditSystem.cpp
```diff
@@ -163,45 +175,125 @@ void ACesiumCreditSystem::Tick(float DeltaTime) {
...
+  // avoid buffer overflow by wrapping with html tags
+  html = "<span>" + html + "</span>";
+
+  tidyBufAppend(
+      &docbuf,
+      reinterpret_cast<void*>(const_cast<char*>(html.c_str())),
+      static_cast<uint>(html.size()));
+
+  std::string output, url;
+  err = tidyParseBuffer(tdoc, &docbuf);
```
kring, line 291: I think you could save some trouble by using `tidyParseString(html.c_str())`.

Source/CesiumRuntime/Private/CesiumCreditSystem.cpp
```diff
@@ -163,45 +175,125 @@ void ACesiumCreditSystem::Tick(float DeltaTime) {
...
+  if (err >= 0) {
```
kring, line 292: This doesn't look right. 0 means success, 1 means warnings, 2 means errors. I don't think it'll ever be negative.

Source/CesiumRuntime/Private/CesiumCreditSystem.cpp
```diff
@@ -163,45 +175,125 @@ void ACesiumCreditSystem::Tick(float DeltaTime) {
...
+  tdoc = tidyCreate();
+  tidyOptSetBool(tdoc, TidyForceOutput, yes);
+  tidyOptSetInt(tdoc, TidyWrapLen, 0);
+  tidyOptSetInt(tdoc, TidyNewline, TidyLF);
+
+  tidySetErrorBuffer(tdoc, &tidy_errbuf);
+  tidyBufInit(&docbuf);
+
+  // avoid buffer overflow by wrapping with html tags
+  html = "<span>" + html + "</span>";
```
kring, line 283: It's strange that this is necessary. I tried commenting it out, and it looks like it crashes while trying to build an error message that says `plain text isn't allowed in <head> elements`. Perhaps the crash is a bug in the development version of tidy-html5 we're using? I'm not really sure how it's meant to work when we parse a HTML document that isn't really a document, but is more like an HTML fragment. If that's the problem - credits are being interpreted as part of `<head>` and that's sometimes invalid - maybe the right approach is to wrap it up to be a proper document:

```
<!DOCTYPE html>
<html>
<body>
{{credit text goes here}}
</body>
</html>
```

Probably not much different from adding the `<span>`, but perhaps more justifiable?

Source/CesiumRuntime/Private/ScreenCreditsWidget.cpp
```diff
@@ -0,0 +1,241 @@
...
+    FHttpResponsePtr HttpResponse,
+    bool bSucceeded,
+    int32 id) {
+  if (bSucceeded && HttpResponse.IsValid() &&
+      HttpResponse->GetContentLength() > 0) {
+    UTexture2D* texture =
+        FImageUtils::ImportBufferAsTexture2D(HttpResponse->GetContent());
+    texture->SRGB = true;
+    texture->UpdateResource();
+    texture->AddToRoot();
```
kring, line 170: I don't think this AddToRoot (or the other one below) is needed as long as you make `_creditImages` a `UPROPERTY`. And that'll avoid a memory leak, too, because there's no RemoveFromRoot anywhere.

kring (commented):
This is a great improvement @joseph-kaile! 

A couple of things in addition to the comments below:

Can we use an official version of tidy-html5, like 5.8.0, rather than the head of the `next` branch?

I think we should be able to remove `WebBrowserWidget` from CesiumForUnreal.uplugin now, which would be good to do!

kring (commented):
Just one small problem left!

joseph-kaile:
Thanks. I updated the widget by wrapping it with a Safe Zone widget. I also changed the Image with Hyperlink to use the hyperlink widget.

kring:
Unfortunately tidy-html5 5.8.0 doesn't work in debug builds because of a bug in their CMake in that version (it ignores `CESIUM_DEBUG_POSTFIX` when attempting to install the PDB). So I'm going to update to 5.9.14. It's not as stable/reliable a release as 5.8.0, but it should be fine for our purposes.

kring:
Thanks @joseph-kaile!

PR #870, opened 2022-06-06
Sublevel property cleanup

argallegos:
Contains a few small fixes to the editor UI for georeferenced sublevels.

- Swapped the latitude and longitude parameter of sublevels in the array so that they match the latitude/longitude standard. (Tested this in editor and the change did not seem to introduce any issues.)
- Recategorized the sublevel-related parameters on the CesiumGeoreference so that the sublevels category is nested under the "Cesium" category and not as a separate category
- Changed the display name of the `CesiumSubLevels` array to read "Georeferenced Sublevels". This is for clarity and consistency of documentation. Since the property name itself is unchanged, this should not be a breaking change.

Old on left, new on right.
<p>
</a>&nbsp;
</a>&nbsp;
<br/>

I wasn't sure if the sublevel parameters had been in their own category for a particular reason. If they were, just let me know and I'll discard those changes.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,11 @@
 # Change Log
 
+### v1.???
+
+##### Fixes :wrench:
+- Swapped latitude and longitude parameters on georeferenced sublevels to match with the main georeference.
+- Adjusted the presentation of sublevels in the Cesium Georeference details panel.
+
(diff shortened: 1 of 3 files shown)


cesium-concierge:

nithinp7:
This looks good to me, thanks @argallegos!

PR #882, opened 2022-06-21
Add dithered LOD transitions

### LINKED ISSUE #508: Improve transitions between tile LODs
This has been discussed in several places but it looks like we don't have an issue yet. 

Tile pop-in can be unsightly and disorienting, and could potentially be aided by dithering between LOD levels, assuming that doesn't lead to a performance problem of its own. 

An option could be to use the DitherTemporalAA material node in the tileset material, as outlined in [this page](http://danieru.com/2018/07/04/dither-based-transparency-in-unreal-engines-material-editor/) found by @xuelongmu. A downside of that is that it requires changing the material Blend Mode to Masked, which could decrease shader performance. If users are utilizing masking already on a tileset, though, it presumably wouldn't make much of a difference.

Please add any other ideas for improving LOD transitions on this issue!

nithinp7:
Resolves #508 

Uses an efficient dither-based fading to smoothly blend between tiles during LOD changes. Builds upon https://github.com/CesiumGS/cesium-native/pull/543.

Here's a comparison without LOD dithering and with dithering. Screen recording frame hitches and gif compression hide some of the visual improvement (it is a lot better when trying it out at proper resolution and interactively).
**Before:**
Before

**After:**
After


cesium-concierge:

nithinp7:
@kring I think this is ready for review alongside https://github.com/CesiumGS/cesium-native/pull/543! Small changes, but I think it makes the overall visuals of the plugin much smoother. Tried to pick a sensible default fading period (2 seconds seemed to work nice for me, but let me know what you think). This seems to works well with photogrammetry and CWT (although CWT shows slight Z-fighting when the transition period is too high).

nithinp7:
This should be ready to review now as well!

Source/CesiumRuntime/Private/Cesium3DTileset.cpp
```diff
@@ -1817,17 +1877,34 @@ void ACesium3DTileset::Tick(float DeltaTime) {
...
   removeVisibleTilesFromList(
       this->_tilesToNoLongerRenderNextFrame,
-      result.tilesToRenderThisFrame);
-  hideTilesToNoLongerRender(this->_tilesToNoLongerRenderNextFrame);
-  this->_tilesToNoLongerRenderNextFrame = result.tilesToNoLongerRenderThisFrame;
-  showTilesToRender(result.tilesToRenderThisFrame);
+      result.tilesSelectedThisFrame);
+  removeVisibleTilesFromList(
+      this->_tilesToHideNextFrame,
+      result.tilesSelectedThisFrame);
```
baothientran, line 1891: I maybe wrong here but `_tilesToNoLongerSelectedNextFrame` should consists of `_tilesToHideNextFrame` right? If so, is the removal of `_tilesToHideNextFrame` not needed right?

argallegos:
Thanks @nithinp7! This will be a fantastic visual improvement to the plugin. I did a UX review to look at how to best present this to our users.

Given the current performance impact of the dithering, I recommend we have it off by default. This will help users with less powerful machines and internet speed. In the future if tile load speeds become more comparable, it would be great to have it on by default. A load speed of 1 second also seems to be a good balance for now. I'm also going to leave some comments about meta tags for the parameters.

I'm noticing an issue (at least on the Melbourne tileset) where some element of the dithering effect remains after the tileset has fully loaded in. It's more noticeable when LOD transition length is lower. Setting it to 0.5 results in "cobwebs" of lower LODs ghosting on the screen. This is inconsistent and seems different every time - on LOD 1 or higher it's rarely noticeable, but I am still seeing the issue.
image 
image

I'm noticing something similar on CWT, and I can only assume it's the same issue. It looks like some previously-dithered LODs are not being unloaded completely. To visualize, here's a section of CWT in Melbourne (LOD transition length 0.2):
image
Here it is with the Shader Complexity and Quads optimization viewmode:
image

There's a LOT of overlapping mesh faces here. It seems to get much worse when you're not looking - in this example, it seemed like every time I turned around a new layer would appear.

Source/CesiumRuntime/Public/Cesium3DTileset.h
```diff
@@ -545,6 +554,25 @@ class CESIUMRUNTIME_API ACesium3DTileset : public AActor {
...
+   * How long dithered LOD transitions between different tiles should take, in
+   * seconds.
+   *
+   * Only relevant if UseLodTransitions is true.
+   */
+  UPROPERTY(
+      EditAnywhere,
+      BlueprintReadWrite,
+      Category = "Cesium|Rendering",
+      meta = (EditCondition = "UseLodTransitions"))
```
argallegos, line 573:
```suggestion
      meta = (EditCondition = "UseLodTransitions", EditConditionHides))
```
No need to show this in the panel if dithering is disabled.

Source/CesiumRuntime/Public/Cesium3DTileset.h
```diff
@@ -340,7 +341,11 @@ class CESIUMRUNTIME_API ACesium3DTileset : public AActor {
    * This will cause more tiles to be loaded, but helps to avoid holes and
    * provides a more consistent mesh, which may be helpful for physics.
```
argallegos, line 342:
```suggestion
   * provides a more consistent mesh, which may be helpful for physics.
   * 
   * Note that this will always be disabled if UseLodTransitions is set to true.
```

Source/CesiumRuntime/Public/Cesium3DTileset.h
```diff
@@ -340,7 +341,11 @@ class CESIUMRUNTIME_API ACesium3DTileset : public AActor {
...
    */
-  UPROPERTY(EditAnywhere, BlueprintReadWrite, Category = "Cesium|Tile Culling")
+  UPROPERTY(
+      EditAnywhere,
+      BlueprintReadWrite,
+      Category = "Cesium|Tile Culling",
+      Meta = (EditCondition = "!UseLodTransitions"))
   bool EnableFrustumCulling = true;
```
argallegos, line 349: Is there any way to ensure that this gets unchecked in the Details panel whenever UseLodTransitions is on? Otherwise, this gets confusing because the setting is uneditable, and may be on, but under the hood it's actually off. If we can't tie it to the actual used frustrum culling setting, we may want to add EditConditionHides.

Source/CesiumRuntime/Public/Cesium3DTileset.h
```diff
@@ -351,7 +356,11 @@ class CESIUMRUNTIME_API ACesium3DTileset : public AActor {
    * of the camera above the ground, tiles that are far away (close to
    * the horizon) will be culled when this flag is enabled.
    */
-  UPROPERTY(EditAnywhere, BlueprintReadWrite, Category = "Cesium|Tile Culling")
+  UPROPERTY(
+      EditAnywhere,
+      BlueprintReadWrite,
+      Category = "Cesium|Tile Culling",
+      Meta = (EditCondition = "!UseLodTransitions"))
   bool EnableFogCulling = true;
```
argallegos, line 364: Same as for frustrum culling - mention how it gets overridden by UseLodTransitions in the comment, and update its value when UseLodTransitions is enabled, if possible.

Source/CesiumRuntime/Public/Cesium3DTileset.h
```diff
@@ -545,6 +554,25 @@ class CESIUMRUNTIME_API ACesium3DTileset : public AActor {
   UPROPERTY(BlueprintAssignable, Category = "Cesium");
   FCompletedLoadTrigger OnTilesetLoaded;
 
+  /**
+   * Use a dithering effect when transitioning between tiles of different LODs.
```
argallegos, line 558:
```suggestion
   * Use a dithering effect when transitioning between tiles of different LODs.
   * 
   * When this is set to true, Frustrum Culling and Fog Culling are always disabled.
```

Source/CesiumRuntime/Public/Cesium3DTileset.h
nithinp7: Awesome idea, hadn't heard about this specifier before!

nithinp7:
Thanks for reviewing and catching the bug @argallegos! I found the fix for it, it was a small mistake I made in Cesium Native. Try it again if you get a chance, I think both the invisible extra tiles as well as the buggy water mask should both be fixed now.

As for performance, I don't think the PR has a clearly bad performance impact after the latest fix. 
Regarding the "disabling" of load-culling: GPU memory cost and the network request count is unchanged, the only change is the order in which tiles are loaded. Previously lower priority tiles (offscreen tiles) now end up getting loaded before the parent tile refines. This causes detail on the screen to come in a bit slower, especially when first loading in (this can of course be offset in user apps by having a loading screen tied to the tileset loaded delegate).

Regarding draw calls, it's probably worth taking another look at the stats with / without lod transitions now that the bug is fixed. The increase in draw calls during transitions is somewhat inevitable to the concept of lod blending, but it will be much worse of an impact when you set the transition length to be too long (e.g., 4 seconds like you tried). So for performance measurement I would stick to 1 second or less for the transition length, that's usually the reasonable range.

nithinp7:
Thanks for the detailed review @argallegos! I agree with your comments. I defaulted the LOD transitions to off, per offline discussion.

Per my current understanding, enabling the option has an impact on latency, perhaps a bit beyond the expected latency when disabling culling, even. I set the default transition length to 0.5s, which seems to be quite responsive while still keeping a bit of transition smoothness. It is definitely a bit choppier than 1s, but still a world better than no transition at all. 

That's also a good point on the waiting for successive LODs, currently a tile fade needs to be finished before it can be refined and faded out. However, once the fade is complete LODs get skipped and the target LOD should be faded into directly if it is loaded at that point. I tried allowing multiple tiles to be fading at once, so there was no sequential dependency whatsoever. It actually looked super nice and had low latency, but unfortunately it caused spikes in draw calls that were especially bad in dense city scenes. We might want to revisit the options here, like you said maybe we could explore varying the transition speed or doing instant transitions depending on the LOD gap we are jumping between. We might even want to have a way to abort a transition and change the fade-in target.

For now, I think this is a pretty decent improvement over no transition with no huge drawbacks. We can revisit the transition smoothness vs latency tradeoff in the future when we get a chance.

@kring Bao and Alex both gave this a review, can you have one last test and merge it if you feel it's ready for the release?

Source/CesiumRuntime/Private/Cesium3DTileset.cpp
```diff
@@ -1765,9 +1795,43 @@ void ACesium3DTileset::showTilesToRender(
...
+  UCesiumGltfComponent* pGltf = reinterpret_cast<UCesiumGltfComponent*>(
+      pRenderContent->getRenderResources());
+  if (!pGltf) {
+    return;
   }
+
+  // Remap the fade percentage so that [0,0.5] --> [0,1] and [0.5,1] --> [1].
+  // This forces fading out tiles to stay opaque until the new tiles have fully
+  // faded in. This is needed for dithered fading, otherwise you will be able
+  // to partially see straight through both tiles during a transition.
```
kring, line 1829: I don't know how practical this is, because I don't know how the dither function actually works, but I wanted to jot down this idea somewhere...

Here's how I think of the dither function...  At 0.0, the dither has no pixels active. At 1.0, the dither has all the pixels active. In between, the number of pixels active increases monotonically. A greater dither value will never deactivate any pixels that were previously active, it will only add new active pixels. All that is probably true, I'm pretty sure (though I might have the 0.0 and 1.0 backwards).

Now let's add one more property that might not be true, but maybe it is, and if it's not then I don't see why it couldn't be made true, at least...

The sequence of pixels activated is fully deterministic. At a value of 0.5, a pseudo-random set of pixels will be active, but it's always the _same_ pseudo-random set. I believe the dither is in screen-space, so when I refer to sets of pixels I mean on the screen.

If all of these things are true, then there's no need to fully fade in before fading out, we can do both simultaneously. For the tile fading in, we use a dither value from 0.0 to 1.0. For the fading out, we _also_ use a dither value from 0.0 to 1.0, but then we invert the opacity mask value, so we only show the tile where the dither is _not_ active. As long as both the fading in and the fading out tile use the same dither value, one tile will be shown exactly in the screen-space pixels where the other is not, at all times.

Entirely likely I've missed something, but that seems like it would work and give a nice visual quality improvement. No depth fighting, for one thing! Faster and smoother transitions for another. What do you think?

kring:
This looks good to me! Merging!

argallegos:
@nithinp7 @kring It looks like we have some promising starting points for potential improvements to the LOD transitions. An issue has been created for further discussion and future improvements - please add to it if you have any more ideas. https://github.com/CesiumGS/cesium-unreal/issues/961. I'd love to see this get to the point that we can have it on by default!

PR #884, opened 2022-06-23
Don't crash if a tileset/overlay is destroyed while reporting error.

### LINKED ISSUE #879: Crash when switching levels with an invalid project token
1. Open Cesium for Unreal Samples.
2. In the Cesium UI, click the Token button.
3. Select "Specify a token" and then enter some random characters to create an invalid token, and click "Use as Project Default Token".
4. A panel will pop up explaining that the token is invalid and offering to help fix it. Click the X to close the panel.
5. From the Content Draw, open another level.
6. Unreal Engine crashes.

Note that this can also happen without the "token is invalid" popup at step 4 if the token is valid for the resources on the current level,  but invalid for the ones on the new level, e.g. because the token isn't authorized for those assets, or because those assets haven't been added to the user's account from the Asset Depot.

kring:
Fixes #879 

Like Bao said in that issue, I don't completely understand why UE5 (only) is unloading the _new_ level and destroying its tilesets when switching levels. But whatever the reason (or even if it's a bug), the fact remains that it's _possible_ for a Tileset or RasterOverlay to be destroyed between the time we schedule a main thread task to handle a load error and the time that main thread task is actually run. And it's not ok to crash when that happens. So now the load error structs use `TWeakObjectPtr` to guarantee we don't dereference a stale pointer while handling load errors.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,6 +11,7 @@
 - Swapped latitude and longitude parameters on georeferenced sublevels to match with the main georeference.
 - Adjusted the presentation of sublevels in the Cesium Georeference details panel.
 - We now explicitly free physics mesh UVs and face remap data, reducing memory usage in the Editor and reducing pressure on the garbage collector in-game.
+- Fixed a bug that could cause a crash when reporting tileset or raster overlay load errors, particularly while switching levels.
 
 ### v1.14.0 - 2022-06-01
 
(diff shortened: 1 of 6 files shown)


cesium-concierge:

nithinp7:
Thanks @kring!

PR #885, opened 2022-06-23
Scale viewport by pixel density

joseph-kaile:
There doesn't seem to be an easy way to find the "devicePixelRatio" in Unreal. It seems that for desktop platforms, [GetDPIScaleFactor](https://docs.unrealengine.com/4.27/en-US/API/Runtime/SlateCore/Widgets/SWindow/GetDPIScaleFactor/) returns the value that we want, but not for Android or iPhone.

For Android/iPhone, you can get the pixels per inch, and using that number I divide by 160, which number I got from [here](https://developer.android.com/training/multiscreen/screendensities).

The Pixel 2 has a DPI of 294. The iPhone 13 has a DPI of 460.

The devicePixelRatio for the devices is 2.625 and 3 respectively.

Although it is not very accurate, at least higher phone density means more scaling.


cesium-concierge:

* :x: Missing CLA.

kring:
This is looking pretty good on my Windows laptop with screens with different scaling; it comes up with just the right values. But a few tweaks will make it perfect:

1. The scaling factor should be per-camera, including editor-only cameras. I think all of the existing camera pathways already get a viewport client, so hopefully it's easy to get the scaling factor and scale the resolution when creating each camera.
2. The scaling factor should be updated every frame. This will probably happen automatically with (1). Scaling factors can change with changes in display settings and when moving windows between monitors, and it's infurating when applications get this wrong (fortunately it's increasingly uncommon). It looks like Unreal is caching the value, so it won't be expensive to get it every frame.
3. DPI scaling should be on by default. We'll have to list this as a breaking change, but I think it's worth it.

I haven't looked closely at how this operates on mobile yet, will do that next. It would also be good to check macOS and Linux.

kring:
My Pixel 6 Pro has a native resolution of 3120x1440 (512 ppi). A default Cesium for Unreal Samples deployed to it, though, renders at 1380x720. So that's a scale factor of 2 (the 3120 direction gets a little shorter in Unreal to avoid rendering over the camera cutout). However, Google Chrome on this same device uses a scale factor of 3.5. A couple of takeaways from this anecdote:

1. If we do something to compute the same scale factor as Chrome does (3.5), we'll end up drastically over scaling and reducing detail too much, because Unreal is already doing some scaling.
2. If we don't do any scaling, we're scaling less than CesiumJS does, and end up rendering about 3x the number of tiles.

So I think step 1 here is to do _nothing_ on mobile. The changes in this PR are good for desktop operating systems, but we should remove any scaling for mobile devices.

Step 2 is to think hard about how we should really approach this on mobile, probably starting with this page: https://docs.unrealengine.com/5.0/en-US/performance-guidelines-for-mobile-devices-in-unreal-engine/
For example, perhaps we should make additional scaling a function of device profiles? Or something? In any case, even though the original motivation here was to achieve better performance on an iPhone, let's treat that as a separate PR.

nithinp7:
Thanks @joseph-kaile!

PR #893, opened 2022-07-01
fix warnings when packaging ios project

joseph-kaile:
changed to branch from ue4-main


nithinp7:
Thanks @joseph-kaile!

nithinp7:
Need to get the release rolling so will just wait for CI to complete on the `ue4-main` branch.

Issue #829, opened 2022-04-25
Use (non-Material) Blueprints to pass metadata to Materials


kring:
After #698, Cesium for Unreal has the ability to pass feature metadata into Materials as textures 🎉. While the "Auto Fill" button on the `CesiumEncodedMetadataComponent` only discovers numeric properties, even string properties can be coerced to numbers (i.e.: parsed) and then used for shading.

However, it's not currently possible to do more complicated transformations of properties to prepare them for use in Materials. For example, Cesium OSM Buildings has a `cesium#color` property with string values like `"rgb(100, 65, 170)"`. There's no way to parse this string and write an actual color into the generated texture. I hacked up a way to do this in https://github.com/CesiumGS/cesium-unreal/tree/metadata-colors, but I don't think we should ship it in this form.

In CesiumJS, [3D Tiles Styling](https://github.com/CesiumGS/3d-tiles/tree/main/specification/Styling) allows the use of JavaScript-like expressions to style. Implementing this in Unreal at runtime is unlikely, but we could offer an in-editor option to turn a style description into a Material.

Perhaps the more Unreal-y way to do it, though, is to let the user write a Blueprint that takes feature metadata as input and returns values to be stored in a texture, which can then be accessed from a Material. Conceptually this is really straightforward, and a proof of concept of this would probably be quick to implement. But it raises some questions:

* Where do the Blueprints live? On the encoded metadata component?
* Will executing a Blueprint function per feature be too slow?
* Can we run a Blueprint function in a worker thread, so we don't have to block the main thread while we're processing metadata for shading?

kring:
Related to #771.

j9liu:
We considered potentially incorporating this into the metadata overhaul, but we probably won't implement it due to the time constraints and performance concerns. But to quote @kring,

> Even just the ability to write and register encoding functions in C++, and then select which one of them to use in the CesiumEncodedMetadataComponent UI, would be huge.

Inspired by this idea, I'd like to rework the encoded metadata component UI as so:

- Display the property's actual type, regardless of whether or not it is supported.
- Have some option that displays a method of coercion? Like an enum dropdown that says "None, Uint8, Float." We can even throw in "Color" based on the `metadata-color` branch, handling `rgb` values as well as hexcode colors.

This way, it more accurately reflects the data that is available, even if it's incompatible with the GPU. And if the user chooses a coercion method that isn't compatible with that type, maybe we can make use of templates to indicate that it can't be coerced?

[...]

PR #902, opened 2022-07-11
Hash metadata attribute strings via Fnv1a

installgentoo:
Provides a limited form of access to the string contents in ue
materials.


cesium-concierge:
Thank you so much for the pull request @installgentoo! I noticed this is your first pull request and I wanted to say welcome to the Cesium community!

installgentoo:
See my case for this on the forums https://community.cesium.com/t/getting-string-from-metadata-into-unreal-material/19490

nithinp7:
As per the conversation in the forum conversation, I don't think we can merge this in the current state. When we get ready to implement #829, we might be able to add this as an option (or it might become application-level logic).

Issue #774, opened 2022-03-03
Add a way to check when tiles are finished loading


argallegos:
This has been requested on the forum - [1](https://community.cesium.com/t/async-loading-callback-on-progress-of-load/16278), [2](https://community.cesium.com/t/how-to-know-when-its-fully-loaded/17555) 

It would be great to be able to access the loading status of tiles, or otherwise be able to trigger events when tiles are done loading. This is important for loading screens, or disguising tile pop-in. 

@nithinp7 suggested this might be possible by exposing these values with getter functions: https://github.com/CesiumGS/cesium-unreal/blob/ccf13b9e0b4f699fcf94ad666edd6a57d097112c/Source/CesiumRuntime/Public/Cesium3DTileset.h#L776-L778

ptrgags:
Sorry for the drive-by comment, just wanted to point out the corresponding API in CesiumJS:

* [initialTilesLoaded event](https://cesium.com/learn/cesiumjs/ref-doc/Cesium3DTileset.html?classFilter=cesium3dtileset#initialTilesLoaded) - triggered a single time when the initial view has finished loading tiles
* [allTilesLoaded event](https://cesium.com/learn/cesiumjs/ref-doc/Cesium3DTileset.html#allTilesLoaded) - trigged multiple times, whenever the current view is finished loading the required tiles.

Personally I've found the naming of these events a little non-obvious in the past, but they've certainly come in handy.

nithinp7:
@cevanno No this still has not been implemented in Cesium for Unreal, fyi I think Peter was just suggesting a possible way to implement such an API inspired by CesiumJS.

Also @argallegos, tacking onto your description:
Kevin brought up the other day that the _lastTilesLoaded...Priority values all being zero still doesn't guarantee the tiles are done loading, since it doesn't count tiles and subtrees currently in flight (ContentLoading state). In cesium native's Cesium3DTilesSelection::Tileset, there are _loadsInProgress and _subtreeLoadsInProgress that should be zero as well before we can confirm all the currently needed tiles are loaded.

Also an additional note, tiles in the _lastTilesLoadedLowPriority category are not needed for the current view, but rather might be the ancestors or siblings of currently visible tiles. These are loaded when preloadSiblings and preloadAncestors are enabled respectively. So depending on the needs of this "ready" notification, we may or may not want to check the low priority loading tiles.

csciguy8:
This is an old issue that appears solved.
[...]

PR #906, opened 2022-07-14
Add callback on tileset finished loading

miatang13:
- Depends on this PR from Cesium Native: https://github.com/CesiumGS/cesium-native/pull/525 
- Solves #774  
- This PR contains adding a new delegate `OnTilesetLoaded` that is exposed to the blueprint, which is invoked when the current tileset has finished loading 
- Currently we determine a tileset to be loaded when there are no tiles to be loaded in the queues, no tiles currently in load, and no occlusion results that we are waiting on.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ Intermediate
 ThirdParty/build
 Source/ThirdParty
 node_modules
+extern/build-helper.sh
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,5 @@
+{
+    "files.associations": {
+        "vector": "cpp"
+    }
+}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,5 +1,6 @@
 {
     "files.associations": {
-        "vector": "cpp"
+        "vector": "cpp",
+        "atomic": "cpp"
     }
 }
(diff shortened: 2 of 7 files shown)


cesium-concierge:

kring:
@nithinp7 since you merged the cesium-native side of this, can you please review this PR as well?

nithinp7:
Thanks @miatang13! Finally got a chance to try this out, it's super cool and the callback works well, even from BP. My only suggestion for this one would be to expose a (BP accessible) function on `Cesium3DTileset` to get the load progress as a percentage. 

Also related to that, the current load progress is being cast to an unsigned int, I think it might be better to leave it as a float what do you think? I should have caught that before merging the Cesium Native PR, since that would require slightly tweaking the Cesium Native api, but I think it would be worth it.

nithinp7:
Also please update the changelog here.

miatang13:
This PR currently depends on this PR from Cesium Native: https://github.com/CesiumGS/cesium-native/pull/533 

I have added a new blueprint getter function that retrieves the current loaded progress. Also I've updated the changelog. @nithinp7

nithinp7:
Perfect, looks good to me!

DSoftwareArtist:
I assumed that this had been implemented. May I know how this feature works? Is there any example? @nithinp7

Issue #904, opened 2022-07-13
Occlusion retrieval race condition?


nithinp7:
Joseph ran into an access violation when retrieving the occlusion results in `CesiumViewExtension::BeginRenderViewFamily`. This was not reproducible on subsequent runs and I have not personally run into the issue. I wonder if the occlusion info is not actually game-thread-safe during `BeginRenderViewFamily` as I assumed, we might have to go back to enqueueing render thread tasks to sync occlusion results back to the game thread. Here is the screenshot from Joseph including the crash location as well as a glimpse of the stack trace.

![bug-occlusion](https://user-images.githubusercontent.com/6706316/178849213-c9b281be-8967-4c41-8106-86ae3e47e2f6.PNG)

nithinp7:
Should be fixed by #919, but please reopen this if anyone runs into anything similar related to occlusion.

@kring Did the repo settings change recently to not automatically close issues marked as fixed by a PR?

kring:
> Did the repo settings change recently to not automatically close issues marked as fixed by a PR?

Unfortunately our `ue5-main` branch is considered the "main" one, so linked issues won't be closed until the PR's commits land in that branch. This is probably yet another reason we should make ue4-main the primary branch for now.

PR #917, opened 2022-07-21
Collect occlusion results on the render thread

nithinp7:
We are pretty sure that the current way that occlusion results are retrieved from the renderer is technically thread-unsafe (#904). This attempts to fix that by retrieving occlusion results on the render thread.

My main worry with this approach is the lack of throughput in occlusion results from the render thread to the game thread. The latency on the other hand is unavoidable since the game thread may be 1-2 frames ahead of the render thread. However, fwiw the original implementation for retrieving occlusion results (before the thread unsafe usage of the CesiumViewExtension) had the same throughput limitation and it still worked quite well. We might be able to implement some sort of 2-3 frame pipelining to increase the throughput, but it may require some more thought.

### DIFF
--- a/Source/CesiumRuntime/Private/Cesium3DTileset.cpp
+++ b/Source/CesiumRuntime/Private/Cesium3DTileset.cpp
@@ -785,7 +785,9 @@ static std::string getCacheDatabaseName() {
   return TCHAR_TO_UTF8(*PlatformAbsolutePath);
 }
 
-void ACesium3DTileset::UpdateFromView(FSceneViewFamily& ViewFamily) {
+void ACesium3DTileset::AggregateViews_RenderThread(FSceneViewFamily& ViewFamily) {
+  // This is precarious, but should be safe since ACesium3DTileset::DestroyTileset
+  // waits on a render fence that is placed after the aggregation commands.
   if (this->BoundingVolumePoolComponent) {
     const TArray<USceneComponent*>& children =
         this->BoundingVolumePoolComponent->GetAttachChildren();
@@ -799,8 +801,25 @@ void ACesium3DTileset::UpdateFromView(FSceneViewFamily& ViewFamily) {
 
       // Check that the primitive is definitely occluded in every view.
       for (const FSceneView* View : ViewFamily.Views) {
-        pBoundingVolume->UpdateOcclusionFromView(View);
+        pBoundingVolume->AggregateOcclusionFromView_RenderThread(View);
       }
+    }
+  }
+}
+
+void ACesium3DTileset::CompleteViewsAggregation() {
+  if (this->BoundingVolumePoolComponent) {
+    const TArray<USceneComponent*>& children =
+        this->BoundingVolumePoolComponent->GetAttachChildren();
+    for (USceneComponent* pChild : children) {
+      UCesiumBoundingVolumeComponent* pBoundingVolume =
+          Cast<UCesiumBoundingVolumeComponent>(pChild);
+
+      if (!pBoundingVolume || !pBoundingVolume->IsMappedToTile()) {
+        continue;
+      }
+
+      pBoundingVolume->FinalizeOcclusionResultForFrame();
     }
   }
 }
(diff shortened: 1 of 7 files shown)


cesium-concierge:

nithinp7:
@kring Can you have a quick look at the code, especially looking at if the multi-threaded logic looks sane?

kring:
First things first: I definitely agree that the previous approach was not thread safe. Elements are both added to and removed from `PrimitiveOcclusionHistorySet` on the render thread, which will (sometimes) trigger reallocation of the array backing the set. If that happens while the game thread is accessing that array, behavior is undefined.

kring:
With the new approach, it may well be correct, but it's very hard to tell. 

For example, `AggregateViews_RenderThread` calls `GetAttachChildren` from the render thread. Is that safe? One potential problem is that the component we're calling it on gets deleted, and you have a comment that addresses that. Ok... so as long as the component still exists... _then_ is `GetAttachChildren` safe? I have no idea! The official word is that it's _never_ safe to access a UObject from the render thread. But if we look at the implementation, it's just returning an array, and ... uhhh.. surely no one should be adding and removing sub-components to our bounding volume components, right? So it's safe?

We really shouldn't make evaluations like this. Even if we prove it's safe now, there's no guarantee it will be safe in the next UE release.

Also, the thread sychronization here is manual and quite complicated. I can't spot any problem with it, but it's complicated enough that it would be extremely easy to miss a subtle problem.

The best way to design a robust multithreaded system is to aggressively aim to a) share as little as possible between threads, and b) use higher-level primitives when sharing is necessary.

Here's an idea, though it's probably imperfect. Rather than using enqueued render commands and fences, let's think about this as message passing. The render thread determines occlusion results every frame, and then it mails those results off to the game thread. The game thread receives those results in its "occlusion result special delivery" mailbox and acts on them. At any given time, there could be multiple results in the mailbox.

How do we implement this mailbox? Unreal has a type called `TQueue` that looks perfect for this. It's a lock-free, FIFO queue than can be configured specifically for a single producer (render thread), single consumer (game thread), which makes it more efficient. The lifetime of the TQueue is easy to manage if it's owned by the CesiumViewExtension, because CesiumViewExtension effectively lives forever.

But what are the messages? Well, the simplest possible thing is a message struct with two fields:

1. An FSceneView pointer.
2. A _copy_ of the `PrimitiveOcclusionHistorySet` for that view.

So every invocation of `CesiumViewExtension::PostRenderViewFamily_RenderThread` adds a new item to the queue containing the two items above. Every invocation of `CesiumViewExtension::BeginRenderViewFamily` pops all the items from the queue and updates its game-thread-copy of the occlusion results for each view. Then we can check each of the occlusion bounding volumes against these game-thread-owned occlusion results.

We don't even need to worry about the validity of the `FSceneView` pointers in the messages as long as we only use them as a lookup key. When we see a `FSceneView` in `BeginRenderViewFamily`, we look for the same one in our known results.

The good thing about this is that it's easy to implement and easy to convince ourselves that it's thread safe. The bad thing about it is that copying those occlusion results _could_ be expensive. The render thread will be doing a heap allocation per view per frame.

If that seems problematic (it does a little), we can improve it with another queue. The renderer thread posts a copy of the occlusion results, as before. But after the game thread receives them, it uses another queue to send the TSet back to the renderer thread via std::move. The renderer thread can then reuse the TSets it receives rather than allocating new ones. Actually it might be better to use a TArray rather than a TSet here, because it looks like TSet does a `realloc` in its `operator=`, which is unnecessary (it'd be better if the allocated memory for the occlusion results were grow-only).

We can also go a step further: the game thread can tell the renderer thread which primitives it actually cares about, rather than just grabbing all of them. I think this will be mostly unnecessary, though.

There could be a flaw to this I haven't considered (I haven't tried to implement it), but I think it should solve the throughput problem you mentioned while also being much easier to verify its correctness.

nithinp7:
@kring Thanks a ton for the suggestions and advice here, they are super helpful! I was definitely spinning my wheels a bit on this one, I think this clarifies for me how the problem should approached.

nithinp7:
I just restarted on a new branch since there was nothing from this branch worth carrying over, will close this PR and open one for the other branch shortly.

PR #919, opened 2022-07-21
Collect occlusion results on the render thread 

nithinp7:
We are pretty sure that the current way that occlusion results are retrieved from the renderer is technically thread-unsafe (https://github.com/CesiumGS/cesium-unreal/issues/904). This attempts to fix that by retrieving occlusion results on the render thread.

Thanks to Kevin for suggesting this type of approach in [this comment](https://github.com/CesiumGS/cesium-unreal/pull/917#issuecomment-1191240481), this feels a lot safer and easier to reason about!

### DIFF
--- a/Source/CesiumRuntime/Private/Cesium3DTileset.cpp
+++ b/Source/CesiumRuntime/Private/Cesium3DTileset.cpp
@@ -785,26 +785,6 @@ static std::string getCacheDatabaseName() {
   return TCHAR_TO_UTF8(*PlatformAbsolutePath);
 }
 
-void ACesium3DTileset::UpdateFromView(FSceneViewFamily& ViewFamily) {
-  if (this->BoundingVolumePoolComponent) {
-    const TArray<USceneComponent*>& children =
-        this->BoundingVolumePoolComponent->GetAttachChildren();
-    for (USceneComponent* pChild : children) {
-      UCesiumBoundingVolumeComponent* pBoundingVolume =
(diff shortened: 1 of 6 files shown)


cesium-concierge:

kring:
@nithinp7 this looks great to me, and I'm happy to merge it. But it's marked draft, so maybe there's something you still want to do? If not, feel free to hit merge yourself.

nithinp7:
@kring Thanks for having a look Kevin, I think the thread synchronization is solid now. I think the strategy for avoiding allocation also feels clean and I've spot checked it in the debugger and it seems to be reusing the previously-allocated sets as intended. I'm going to look a little more carefully at the move calls in the debugger to confirm Unreal is in fact moving instead of copying (there is a long stack of MoveOrCopy calls that get triggered and the last one presumably uses type traits to determine what to do). 

The only other thing is I've seen a very rare case where initially refining an area sometimes causes a few previously visible tiles to unrefine to a parent (or maybe further up ancestor) for a few frames, before it then refines completely to the correct detail. I can only seem to run into the issue after minutes of flying around and carefully looking for it. 

I'm going to try reproducing it in main first thing this morning, if it is an issue there as well I'm going to merge this PR, open an issue for it, and continue looking into the problem.

nithinp7:
Was able to reproduce the unrefinement issue with the occlusion culling in main, opened #924. Self-merging this with Kevin's approval, will continue looking into the unrefinement issue.

PR #925, opened 2022-07-26
Remove supportrasteroverlay

miatang13:
- This PR removes calling `supportsRasterOverlay` which is removed in `cesium-native` in this PR: https://github.com/CesiumGS/cesium-native/pull/532


cesium-concierge:

nithinp7:
Thanks @miatang13! This looks good to me, I can merge it after I have a look at the Cesium Native PR. 

Separate from this PR, now that all tilesets support raster overlays, the `FindFirstTilesetSupportingOverlays` function no longer makes much sense. It would be better to have an editor UI that lets you select the tileset you want to apply a raster overlay to, instead of just picking the first tileset. Opened this issue for it (does not need to be addressed right now): https://github.com/CesiumGS/cesium-unreal/issues/926

Issue #827, opened 2022-04-22
Accept callbacks for DynamicPawn fly-tos


nithinp7:
It may be useful to perform latent actions upon finishing a dynamic-pawn flight, we should add a way to trigger a blueprint-friendly callback when a flight is completed.

kring:
Fixed in #927

PR #927, opened 2022-07-28
Add callback for when dynamic pawn finishes flying

miatang13:
- Solves #827 
- Adds a BP accessible delegate `onFlightComplete` in the `AGlobeAwareDefaultPawn` class. 
- Tested with the `flyTo` example following the tutorial: https://cesium.com/learn/unreal/unreal-flyto/. 
- @nithinp7

### DIFF
--- a/Source/CesiumRuntime/Public/GlobeAwareDefaultPawn.h
+++ b/Source/CesiumRuntime/Public/GlobeAwareDefaultPawn.h
@@ -14,6 +14,12 @@ class UCesiumGlobeAnchorComponent;
 class UCurveFloat;
 
 /**
+ * The delegate for when the pawn finishes flying
+ * which is triggered from _handleFlightStep
+ */
+DECLARE_DYNAMIC_MULTICAST_DELEGATE(FCompletedFlight);
+
+/**
  * This pawn can be used to easily move around the globe while maintaining a
  * sensible orientation. As the pawn moves across the horizon, it automatically
  * changes its own up direction such that the world always looks right-side up.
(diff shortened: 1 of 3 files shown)


cesium-concierge:

Source/CesiumRuntime/Public/GlobeAwareDefaultPawn.h
```diff
@@ -98,6 +104,13 @@ class CESIUMRUNTIME_API AGlobeAwareDefaultPawn : public ADefaultPawn {
   UPROPERTY(EditAnywhere, Category = "Cesium", meta = (ClampMin = 0.0))
   double FlyToGranularityDegrees = 0.01;
 
+  /**
+   * A delegate that will be called whenver the pawn finishes flying
+   *
+   */
+  UPROPERTY(BlueprintAssignable, Category = "Cesium");
+  FCompletedFlight onFlightComplete;
```
nithinp7, line 112:
```suggestion
  FCompletedFlight OnFlightComplete;
```

Small nitpick, Unreal's naming convention is usually to start with a capital letter. This is confusing since it conflicts with our typical C++ naming conventions, but we try to follow Unreal convention for variables exposed to blueprint and the Unreal editor (there are probably places where we've forgotten to do so though).

nithinp7:
I haven't tried this out yet, but it looks solid! The one thing I would consider is to add handling for "flight interruptions", see:
https://github.com/CesiumGS/cesium-unreal/blob/3402d3ac2af42e03b5d4b2f88158fe8d97d0de9d/Source/CesiumRuntime/Private/GlobeAwareDefaultPawn.cpp#L383

When flight interruptions are enabled, moving the controls can stop the flight, this allows the camera to be more responsive and is useful in some situations. We could either just trigger the "flight completed" delegate, or have a separate one for "flight interrupted", what do you think?

miatang13:
> When flight interruptions are enabled, moving the controls can stop the flight, this allows the camera to be more responsive and is useful in some situations. During interruptions, we could either just trigger the "flight completed" delegate, or have a separate one for "flight interrupted", what do you think?

I think having a separate delegate is the better option here since moving the control might be intentional or by accident, and technically speaking if the flight is interrupted, it has not been completed.

miatang13:
This PR now adds a new delegate `OnFlightInterrupt` which is triggered when a pawn's flying is interrupted. Currently tested with interrupting flight with both keyboard (WSAD) and moving the mouse.

joseph-kaile (approved):
Looks good, @miatang13 . I also tested both delegates and works as expected.

PR #928, opened 2022-07-28
Fix 1-3 frame false-occlusion bug

### LINKED ISSUE #924: Occlusion culling causing unrefining of visible tiles (pretty rare)
I've seen a very rare problem where initially refining an area sometimes causes a few previously visible tiles to unrefine to a parent (or maybe further up ancestor) for a few frames, before it then refines completely to the correct detail. I can only seem to run into the issue after minutes of flying around and carefully looking for it. It doesn't seem to appear when occlusion is disabled (but disabling the "wait for occlusion results" option doesn't seem to fix it though). This might be related to how occlusion culling and external tilesets work together or how occlusion and the kicking mechanism interact.

nithinp7:
Fixes #924. 

This turned out to be a very similar problem to the infamous flickering holes bug. It has to do with stale occlusion results for bounding volume proxies that are newly set to visible (ones that have been newly assigned to a tile). Particularly, when there was a bounding volume proxy that was previously occluded when assigned to a previous tile and then got unassigned and then much later reassigned to a new tile. The last occlusion result from the old tile assignment ended up being picked up and the selection mistakenly thinks the tile is occluded. The solution is to keep track of when a tile is assigned and ignore occlusion results from before the assignment.

### DIFF
--- a/Source/CesiumRuntime/Private/CesiumBoundingVolumeComponent.cpp
+++ b/Source/CesiumRuntime/Private/CesiumBoundingVolumeComponent.cpp
@@ -104,7 +104,8 @@ void UCesiumBoundingVolumeComponent::UpdateOcclusion(
   TileOcclusionState occlusionState =
       cesiumViewExtension.getPrimitiveOcclusionState(
           this->ComponentId,
-          _occlusionState == TileOcclusionState::Occluded);
+          _occlusionState == TileOcclusionState::Occluded,
+          _mappedFrameTime);
 
   // If the occlusion result is unavailable, continue using the previous result.
   if (occlusionState != TileOcclusionState::OcclusionUnavailable) {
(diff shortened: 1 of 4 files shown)


cesium-concierge:

kring:
This looks good to me. But I noticed that my last status update is almost always showing `Waiting For Occlusion Results 1`. That seems like a problem, doesn't it, if not all of the results are coming back?

kring:
Whether I use the cesium-native commit referenced by this branch, or the current `main`, I consistently see it stop at "Waiting for Occlusion Results 1` (or sometimes 2), in both the normal Editor and in Play-in-Editor.

So I added some code to `PostRenderViewFamily_RenderThread` to detect primitives that were frustum culled in the view and also mark them occluded. This seems to resolve the problem (the "waiting" count now goes to zero), and I can't see any negative side effects. I was slightly worried about the extra work in the render thread, so I profiled it and `PostRenderViewFamily_RenderThread` was less than 1/100th of the total time spent in the render thread, which is acceptable if not amazing.

So I committed it. It's slightly dodgy because it uses `FViewInfo`, which is a private renderer class. But that's not breaking any new ground.

I guess overall I'm just not extremely confident that the occlusion culling is going to work perfectly in all cases, on all platforms, and in all situations, especially now that I've mucked with it. 😆 So I'm going to introduce a feature flag for occlusion culling in this release and default it to off. I don't want to just default the Tileset option to false, because then any tilesets saved while it defaults to Off will have it disabled forever (until the user manually changes it). Hopefully lots of people will try it out and we'll get some feedback and can enable it quickly (even in an interim release before September 1).

I'll merge this now and open a separate PR for the feature flag.

PR #929, opened 2022-07-28
Fix Perforce issue when saving tokens

### LINKED ISSUE #776: Perforce-like source control breaks Cesium ion token operations
Some source control systems, like Perforce, mark files read-only until they're explicitly checked out by the user. If a project's config files are read-only, we won't be able to update them when the user selects a project default token or signs in to Cesium ion. Apparently it just silently fails.

As report on the forum:
https://community.cesium.com/t/need-to-authorize-cesium-every-time-unreal-engine-is-launched/17595/4

At the very least we should report when we can't save a token change. Ideally, we'd ask Unreal to check it out of source control for us.

joseph-kaile:
Fixes #776

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,10 @@
 
 ### ???? - ????
 
+##### Fixes :wrench:
+
+- Perforce marks files read only if not checked out. To save access tokens, it now asks to check out the necessary config files from source control.
+
 ##### Additions :tada:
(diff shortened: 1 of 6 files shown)


cesium-concierge:

* :x: Missing CLA.

kring:
Thanks @joseph-kaile, this looks great! I just extracted the common source control related code to a separate class.

PR #933, opened 2022-08-01
Put occlusion culling behind a feature flag.

kring:
As mentioned in https://github.com/CesiumGS/cesium-unreal/pull/928#issuecomment-1200835178, this puts the occlusion culling feature behind a feature flag, which is found in Project Settings. The reason is:

1. I'm nervous to have occlusion culling "on by default" currently given the deep renderer integration and last-minute changes. While it's working well in our tests, it wouldn't be surprising to run into issues on some platforms or in some environments, or even to run into visual artifacts with certain tilesets in certain cases.
2. If we do the simple thing of just defaulting the "Enable Occlusion Culling" property on Cesium3DTileset to _off_, then later, when we do want it to be on by default, people with existing levels will need to re-enable it manually. We won't be able to easily distinguish "intentionally disabled" from "disabled just because that used to be the default."

By adding a "feature flag" at the project level, we can disable the entire feature by default, but let people enable it to try it out. And once we have more confidence in the feature (which doesn't need to wait until September), we can remove the feature flag and make the plugin act as if it's always enabled (subject to the Tileset setting).

Unfortunately, Unreal made this a little harder than I anticipated, because UPROPERTY `EditCondition` is so limited, but this should be a workable template for future feature flags if we need them as well.

### DIFF
--- a/Source/CesiumRuntime/Private/Cesium3DTileset.cpp
+++ b/Source/CesiumRuntime/Private/Cesium3DTileset.cpp
@@ -1906,6 +1906,7 @@ void ACesium3DTileset::Destroyed() {
   AActor::Destroyed();
 }
 
+#if WITH_EDITOR
 void ACesium3DTileset::RuntimeSettingsChanged(
     UObject* pObject,
     struct FPropertyChangedEvent& changed) {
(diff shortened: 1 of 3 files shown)


cesium-concierge:

PR #935, opened 2022-08-01
Occlusion culling: handle nullptr State, disable the CesiumViewExtension when the feature is disabled

kring:
This PR fixes a crash in UE5 when opening the Project Settings panel. It's caused by a `FSceneView` having a nullptr `State`, which can apparently (according to a comment above the field) happen when the view is used for thumbnail rendering.

It also "fixes" (very much in quotes) a crash at startup in UE4.26. The crash happens when we try to copy the `PrimitiveOcclusionHistorySet`, and it may actually be a 4.26 bug. What I've done here is to more fully disable the `CesiumViewExtension` when the Occlusion Culling feature is disabled. This way the crash doesn't happen at startup, but instead happens when the user tries to enable the experimental occlusion culling feature. Disabling the CesiumViewExtension is a good idea anyway, to reduce the chances of problems caused by our interaction with the renderer.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,12 @@
 # Change Log
 
+### v1.16.1 - 2022-08-01
+
+##### Fixes :wrench:
+
+- Fixed a bug that could cause a crash when using thumbnail rendering, notably on the Project Settings panel in UE5.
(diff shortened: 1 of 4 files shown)


cesium-concierge:

PR #936, opened 2022-08-02
Make occlusion culling work in UE 4.26.

kring:
By using an explicit struct to shuttle the necessary data from the render thread, rather than trying to copy an engine data structure. It should be slightly more efficient in other engine versions, too.


cesium-concierge:

joseph-kaile:
Works well, tested in 4.26 and 5.0.

PR #937, opened 2022-08-03
Don't use dodgy Editor viewports.

### LINKED ISSUE #801: Invisible editor viewports still affect tile selection
I saw this in UE5 preview 2, so I'm not certain if it happens in UE4, too.

* Click the "Maximizes or restores this viewport" button in the top-right corner of the main viewport:
image
* The view turns into a grid with four viewports, the original main viewport plus three orthographic views.
image
* Click "Maximizes or restores this viewport" again in the main viewport window in the lower left to restore the original view.
image

Even though everything now looks the same as when we started, those additional orthographic views are now affecting tile selection. If you fly the editor camera elsewhere and look back toward where you started, you'll see high-detail tile selection in the original area, caused by those non-visible viewports.

These viewports return true for `IsOrtho()` and false for `IsRealtime()`, so we can exclude them on either or both of those results and that would probably be a good change. But it's possible to create a normal realtime, prespective view, and the we still wouldn't want it to affect tile selection when it's not visible. It's not clear from a quick look how we can tell when a viewport isn't visible, but there is most likely a way.

kring:
Now only viewports that are visible, real-time, and use a perspective projection are used for tile selection.

This is especially important in UE5 because collapsed Editor viewports seem to no longer get a 0,0 size in that version.

Probably the most controversial one here is the "visible", because Unreal considers all viewports non-visible if another application (other than the UE Editor) has the focus. That effectively means tile selection will stop when the Editor loses focus. I think we can live with this, though, especially since I wasn't able to find a better alternative.

Fixes #801

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,6 +7,10 @@
 - The translucent parts of 3D Tiles are now correctly rendered as translucent.
 - Added a `TranslucentMaterial` property to `Cesium3DTileset`, allowing a custom material to be used to render the translucent portions of a tileset.
 
+##### Fixes :wrench:
+
+- Cesium for Unreal now only uses Editor viewports for tile selection if they are visible, real-time, and use a perspective projection. Previously, any viewport with a valid size was used, which could lead to tiles being loaded and rendered unnecessarily.
+
 ### v1.16.1 - 2022-08-01
 
 ##### Fixes :wrench:
--- a/Source/CesiumRuntime/Private/Cesium3DTileset.cpp
+++ b/Source/CesiumRuntime/Private/Cesium3DTileset.cpp
@@ -1442,6 +1442,12 @@ std::vector<FCesiumCamera> ACesium3DTileset::GetEditorCameras() const {
       continue;
     }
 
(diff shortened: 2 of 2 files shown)


cesium-concierge:

javagl:
> Now only viewports that are visible, real-time, **and use a perspective projection** are used for tile selection.

<sup>(emphasis by me)</sup>

Won't that re-introduce the issue from the second image in https://github.com/CesiumGS/cesium-unreal/issues/224 ? (Maybe that's the desired behavior, I just wanted to cross-link here...)

kring:
I don't think we have ever supported orthographic views correctly. Before this PR it did "something" (used the position and orientation of the orthographic view's camera, but not its actual view shape). After this PR it will just ignore it for selection purposes. Nithin added some experimental support for orthographic views in https://github.com/CesiumGS/cesium-unreal/pull/547, but we reverted it because it wasn't reliable. If we ever add proper support for orthographic views, we can remove the perspective requirement. Until then this PR doesn't make anything worse as far as I know.

joseph-kaile:
Works well 👍

PR #942, opened 2022-08-12
Fix disappearing tiles when camera is orbiting

joseph-kaile:
This PR fixes the problem where pressing ALT and the drag left mouse button in the Editor viewport causes tiles to disappear.

### DIFF
--- a/Source/CesiumRuntime/Private/Cesium3DTileset.cpp
+++ b/Source/CesiumRuntime/Private/Cesium3DTileset.cpp
@@ -1447,18 +1447,17 @@ std::vector<FCesiumCamera> ACesium3DTileset::GetEditorCameras() const {
         !pEditorViewportClient->IsPerspective()) {
       continue;
     }
-    auto GetViewRotation = [pEditorViewportClient]() {
-      if (pEditorViewportClient->bUsingOrbitCamera) {
 
-        return (pEditorViewportClient->GetLookAtLocation() -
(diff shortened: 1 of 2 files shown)


cesium-concierge:

* :x: Missing CLA.

Source/CesiumRuntime/Private/Cesium3DTileset.cpp
```diff
@@ -1447,9 +1447,18 @@ std::vector<FCesiumCamera> ACesium3DTileset::GetEditorCameras() const {
         !pEditorViewportClient->IsPerspective()) {
       continue;
     }
-
+    auto GetViewRotation = [pEditorViewportClient]() {
```
nithinp7, line 1450: Is this lambda needed? At first I thought it was saving a copy of the FRotator, since all we really need is a const reference. But it looks like the inferred return value of the lambda function is just a plain FRotator (one of the code paths returns a const ref but the other returns a FRotator by value). I actually didn't know a temporary value (the result of the lambda call) could be assigned to a const ref (e.g., on L1461)! But I still think a copy is happening regardless. If that is the case, would it be better to turn that const ref into a temporary variable and use a straightforward conditional to choose an FRotator, instead of the lambda function?

Or is the compiler optimizing something that I don't understand?

joseph-kaile: Thanks, that does make sense.

PR #944, opened 2022-08-16
Fix Globe Anchor Component preventing Actor transform from resetting/changing.

### LINKED ISSUE #938: Globe Anchor component prevents changing/resetting actor transform in details panel
Observed here: https://community.cesium.com/t/question-dynamic-pawn-location-step-7-part-3-of-cesium-into-unreal-quickstart-tutorial/14372/5

An object with a Globe Anchor component can not be moved in editor via the transform fields in the details panel. This includes manually typing transform values or using the yellow "reset" arrow. It seems like the globe anchor component overrides changes to the actor's transform to keep it at the same location. 

- Tested in UE4 and UE5. In UE4, the object will at first appear to move, but then snap back once you stop editing the field. In UE5, the object does not move at all. 
- This seems to occur with all actors with a Globe Anchor component, including the DynamicPawn. 
- Deleting the GlobeAnchor component resolves the issue, so it's definitely caused by that.
- Actors can still be moved using the viewport gizmo (E.G. dragging the arrows in the viewport.) 
- Users noted that the location CAN be edited while in play mode. However, it should still be possible to edit the location without entering play mode.

### Steps to reproduce

1. Add a cube to the scene. Optionally, scale it up to see it better.
2. Move the cube away from 0,0,0
3. Add a Globe Anchor component to the cube
4. Try to set any of the Location values back to 0. Observe issue.
5. Try to use the yellow "Reset to Default" arrow to reset the transformation. Observe issue.

### As a temporary workaround, use either of these methods to move an actor:

- Use the Transform gizmo in the viewport to drag the actor around. 
- To reset the actor to the world's 0,0,0, click on the actor's Globe Anchor Component and set the Longitude, Latitude, and Height values to match those of the Georeference Origin.

### LINKED ISSUE #943: Globe Anchor Component - Allow for rotational offset as parameter and getting rotations from anchor component modified actors.
For the [5th VR tutorial ](https://github.com/CesiumGS/cesium-unreal-vr-tutorial/issues/5) about dragging the globe it would be great if we could pass in a rotational offset to the SnapLocalUpToEllipsoidNormal function to allow for more control. This is so we could point a VR player downward to face the globe for example. With snapping taking total control of the actors rotation, that other layer of control would he helpful.

As a utility it would also be helpful to do the inverse and get a rotation from the owning actor, which is modified by whatever rotation is applied to the actor to make it along the normal of the ellipsoid.

joseph-kaile:
Fixes #938

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,7 @@
 ##### Fixes :wrench:
 
 - Cesium for Unreal now only uses Editor viewports for tile selection if they are visible, real-time, and use a perspective projection. Previously, any viewport with a valid size was used, which could lead to tiles being loaded and rendered unnecessarily.
+- Fixed a bug in the Globe Anchor Component that prevented changing/resetting the actor transform in the details panel.
 
 ### v1.16.2 - 2022-08-04
 
--- a/Source/CesiumRuntime/Private/CesiumGlobeAnchorComponent.cpp
+++ b/Source/CesiumRuntime/Private/CesiumGlobeAnchorComponent.cpp
@@ -397,10 +397,11 @@ void UCesiumGlobeAnchorComponent::OnUnregister() {
   USceneComponent* pOwnerRoot = pOwner->GetRootComponent();
   if (pOwnerRoot) {
     pOwnerRoot->TransformUpdated.RemoveAll(this);
-    _onTransformChangedWhileUnregistered = pOwnerRoot->TransformUpdated.AddLambda(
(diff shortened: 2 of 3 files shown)


cesium-concierge:

* :x: Missing CLA.

nithinp7:
Thanks @joseph-kaile! Ah I'm surprise this one managed to hide for this long. I'm not sure if it's a regression, but I think it's more than a year old either way. This looks like it completely fixes it!

PR #947, opened 2022-08-19
Fix UE5 Line Tracing bug

### LINKED ISSUE #900: LineTrace occasionally fails to collide with tiles at a certain LOD
Originally detected in this thread: https://community.cesium.com/t/linetrace-hit-detection-causing-value-to-be-world-origin/18478

## Summary of issue:
The user is trying to access the world position height of an arbitrary point on the tileset, so that they can place another actor flush with the surface. They are using a line trace (SphereTraceForObjects) to find this height. Most of the time, the line trace hits reliably, but when a tile is at a certain LOD, some traces will not hit anything at all. 

- Missed traces are consistent: At a given longitude/latitude and tile refinement level, a trace will always miss or always hit. 
- Collision is enabled and otherwise working on the tileset. 
- This only occurs in UE5, and was not a problem in UE4.

The issue does not seem to occur on all tilesets. The affected tilesets were collated with QGIS from DSMs, then tiled with Cesium ion. They are being streamed through Cesium ion. 

## Workaround
For some reason, enabling the **Always Include Tangents** setting on the tileset seems to resolve the issue. 

Not sure if this is a tiling/tileset issue, or something Cesium for Unreal specific. Hopefully, since including the tangents helps with the issue, it's something we can nail down and fix.

## Reproducing
The user provided a test project with a tileset and blueprints that can reproduce the issue consistently. Because the tileset is not public, message me for access to the project.

joseph-kaile:
Fixes #900

When setting the Chaos triangles, the indices are set to uint16 if the number of vertices is less than 65536, and int32 otherwise.

However, for some reason, adding uint16 triangles causes the issue. 

In the Unreal Code, https://github.com/EpicGames/UnrealEngine/blob/release/Engine/Source/Runtime/Engine/Private/PhysicsEngine/Experimental/ChaosCooking.cpp#L302, they also create an array of uint16 triangles, however in the lamba function, they add int32 triangles to it (https://github.com/EpicGames/UnrealEngine/blob/release/Engine/Source/Runtime/Engine/Private/PhysicsEngine/Experimental/ChaosCooking.cpp#L254).

This PR changes the code to do the same thing.

### DIFF
--- a/Source/CesiumRuntime/Private/CesiumGltfComponent.cpp
+++ b/Source/CesiumRuntime/Private/CesiumGltfComponent.cpp
@@ -2373,12 +2373,12 @@ static void fillTriangles(
 
   triangles.Reserve(triangleCount);
 
-  for (TIndex i = 0; i < static_cast<TIndex>(triangleCount); ++i) {
-    TIndex index0 = 3 * i;
-    triangles.Add(Chaos::TVector<TIndex, 3>(
-        static_cast<TIndex>(indices[index0 + 1]),
-        static_cast<TIndex>(indices[index0]),
-        static_cast<TIndex>(indices[index0 + 2])));
+  for (int32 i = 0; i < triangleCount; ++i) {
+    const int32 index0 = 3 * i;
+    triangles.Add(Chaos::TVector<int32, 3>(
+        indices[index0 + 1],
+        indices[index0],
+        indices[index0 + 2]));
   }
 }
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,6 +11,7 @@
 
 - Cesium for Unreal now only uses Editor viewports for tile selection if they are visible, real-time, and use a perspective projection. Previously, any viewport with a valid size was used, which could lead to tiles being loaded and rendered unnecessarily.
 - Fixed a bug in the Globe Anchor Component that prevented changing/resetting the actor transform in the details panel.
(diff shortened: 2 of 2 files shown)


cesium-concierge:

* :x: Missing CLA.

Source/CesiumRuntime/Private/CesiumGltfComponent.cpp
```diff
@@ -2373,12 +2373,12 @@ static void fillTriangles(
...
   triangles.Reserve(triangleCount);
 
-  for (TIndex i = 0; i < static_cast<TIndex>(triangleCount); ++i) {
-    TIndex index0 = 3 * i;
-    triangles.Add(Chaos::TVector<TIndex, 3>(
-        static_cast<TIndex>(indices[index0 + 1]),
-        static_cast<TIndex>(indices[index0]),
-        static_cast<TIndex>(indices[index0 + 2])));
+  for (int32 i = 0; i < triangleCount; ++i) {
+    const int32 index0 = 3 * i;
```
kring, line 2377: This is the key line that fixes the bug. The rest I'm pretty sure is unnecessary (though it won't do any harm). Even if `i` is small enough to fit into a uint16, there is no guarantee that `3 * i` still will be. And when it's not, it will overflow and access the wrong source indices.

kring:
Thanks @joseph-kaile!

PR #949, opened 2022-08-23
Copy UV data to physics mesh only when using metadata textures

### LINKED ISSUE #874: Physics meshes require a lot more memory after metadata styling changes
In #698, we changed our physics meshes generation code to include an additional option: `bSupportUVsAndFaceRemap`. The discussion about it at the time was:

> I think users have to enable UV results from line traces as a project wide setting and then it shows up as part of the HitResult struct. If the above option is set to true and if the project wide setting is set to true, then I think the physX cooking saves the UV coordinates for line traces later. This will take up more CPU memory, but I doubt it should affect cook time

However, this does not appear to be true. UV and face remap data appears to get generated regardless of the project setting, probably because we're short-circuiting the process and explicitly asking PhysX to generate this data.

This new data caused a memory leak (fixed in #873), but even with that fixed it means we're using significantly more memory for physics meshes than we used to. Some ideas for improving this situation:

* Honor that project setting flag, rather than forcing PhysX to always generate this data.
* Do we need UV info for _every_ set of texture coordinates? We only actually use it for picking (not styling) with feature ID textures, right?
* Can we make this optional so that users only need to pay for it when they use it?

Also for consideration: what changes (if any) are needed for Chaos / UE5?

joseph-kaile:
Fixes #874

We need to copy texture coordinates for all three types, Feature Textures, Feature ID textures and Feature ID attributes. 

An easy way to tell if a model/primitive may need the UV data is if ```metadataTextureCoordinateParameters``` is not empty.

TODO: Only copy the UV channels associated with metadata.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,7 +11,7 @@
 
 - Cesium for Unreal now only uses Editor viewports for tile selection if they are visible, real-time, and use a perspective projection. Previously, any viewport with a valid size was used, which could lead to tiles being loaded and rendered unnecessarily.
 - Fixed a bug in the Globe Anchor Component that prevented changing/resetting the actor transform in the details panel.
-- Reduce the size of physics meshes by only copying UV data if "Support UV from Hit Results" is enabled and if the model uses feature ID textures.
+- Reduce the size of physics meshes by only copying UV data if "Support UV from Hit Results" is enabled.
 
 ### v1.16.2 - 2022-08-04
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,6 +11,8 @@
 
 - Cesium for Unreal now only uses Editor viewports for tile selection if they are visible, real-time, and use a perspective projection. Previously, any viewport with a valid size was used, which could lead to tiles being loaded and rendered unnecessarily.
 - Fixed a bug in the Globe Anchor Component that prevented changing/resetting the actor transform in the details panel.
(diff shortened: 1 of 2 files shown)


cesium-concierge:

* :x: Missing CLA.

nithinp7:
Thanks for looking into this @joseph-kaile! I think we should probably split this into two PRs. The most immediate problem is that even when we don't enable "Support UV Hit Results" in the project settings, we are still copying all the UVs into the physics mesh. The second, bigger, task is #951. 

The implementation here looks good, but it would bar users from being able to access UV coords for their own usage if the tileset _doesn't_ have metadata. [This video (3:58)](https://youtu.be/yVgpya8gamg?t=238) has a nice little example  of the types of effects users might implement by using UV hit results, unrelated to metadata.

In this PR, I think we should stick with the expected behavior, which is copy _all_ the UVs if the project setting is enabled and _none_ otherwise. So no special consideration for the existence of metadata. In the short term this will cause users that implement effects similar to the above video to copy _all_ the UVs as a result. Users that only want access to the metadata UVs will also have to copy all of them. If you would prefer you could add an extra user option on the tileset to only copy metadata UVs, on top of the existing project setting. 

In the future, we'll implement #951 and no longer require metadata UVs for picking. At that point, if a user still decides to enable the "Support UV Hit Results" project setting we will just blindly copy all UVs in the render mesh to the physics mesh, since we won't know what they are trying to do.

nithinp7:
Thanks @joseph-kaile!

PR #950, opened 2022-08-24
Globe Anchor Component - Rotation offset

CesiumBen:
This PR adds a rotation offset FRotator to the CesiumGlobeAnchorComponent so that when calling "SnapToEastSouthUp" the actor can have a rotational offset applied to it. This is particularly useful if working in blueprints with a class that has a strict un-editable root hierarchy.

I attempted to apply the offset to SnapLocalUpToEllipsoidNormal but thinking through it's purpose, it didn't make much sense to do so. A possible adjustment would be to apply the offset without tampering the actor's Yaw as that is preserved with this function and makes it very useful. Any ideas about that would be helpful, thank you!

Related issue #943 
CC: @xuelongmu, @nithinp7, @kring, @baothientran

### DIFF
--- a/Source/CesiumRuntime/Public/CesiumGlobeAnchorComponent.h
+++ b/Source/CesiumRuntime/Public/CesiumGlobeAnchorComponent.h
@@ -245,14 +245,19 @@ public:
   /**
    * Rotates the Actor so that its +X axis points in the local East direction,
    * its +Y axis points in the local South direction, and its +Z axis points in
-   * the local Up direction.
+   * the local Up direction. Rotation offset is then applied to all three axes.
    */
   UFUNCTION(BlueprintCallable, CallInEditor, Category = "Cesium")
   void SnapToEastSouthUp();
 
   /**
-   * Sets the rotation offset to be applied when snapping to east/south/up or
-   * ellipsoid normal.
+   * Gets the rotation offset applied when snapping to east/south/up.
+   */
+  UFUNCTION(BlueprintCallable, CallInEditor, Category = "Cesium")
+  FRotator GetRotationOffset();
+
+  /**
+   * Sets the rotation offset applied when snapping to east/south/up.
    */
   UFUNCTION(BlueprintCallable, CallInEditor, Category = "Cesium")
   void SetRotationOffset(const FRotator& rotationOffset);
@@ -467,7 +472,7 @@ private:
   void _updateCartographicProperties();
 
   /**
-   * Rotation offset applied to the Actor.
+   * Rotation offset applied to the Actor when snapping to east/south/up.
    */
-  glm::dmat3 _rotationOffset;
+  FRotator _rotationOffset;
 };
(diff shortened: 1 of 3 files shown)


cesium-concierge:
Thanks for the pull request @CesiumBen!

kring:
Thanks for the PR @CesiumBen! The code looks good, but I'm having a little trouble wrapping my head around the user experience. The use-case (as described in #943) is that you're periodically moving an object on the globe and using `SnapToEastSouthUp` to reorient it upright at its new globe position. Except you don't want it necessarily oriented exactly to East-South-Up but instead want an offset from that. Do I have that right?

If so, could you solve the problem by creating an empty Actor with a Globe Anchor Component, and then adding the Actor you want to offset as a child of that Actor? Then give the child a relative transform of whatever you want. That seems like it would be more intuitive than a separate property.

What do you think? I could easily be missing something here.

CesiumBen:
@kring Great points, this was something we considered but ran into some unique cases where root offset rotation would be needed. Using an ACharacter through blueprints has some limitations where this arose. 

With its component setup, it's hard to achieve kind of solution you mention above by nesting it within a separate actor. I might be missing something with the approach. For other actors or projects where you can define component hierarchy, nesting transforms would be the best approach.

CesiumBen:
@kring For some more context, if you wanted to nest a Character pawn into an Actor with a GlobeAnchorComponent so that it appears to be up facing, you would have to update the parent's position to match the child's. The Character movement hinges on the UCharacterMovementComponent which updates the world location of the pawn itself. On top of this with the Character responding to physics, I'm not sure how that would respond with its parent moving.

Source/CesiumRuntime/Public/CesiumGlobeAnchorComponent.h
```diff
@@ -458,4 +470,9 @@ class CESIUMRUNTIME_API UCesiumGlobeAnchorComponent : public UActorComponent {
    * globe transform.
    */
   void _updateCartographicProperties();
+
+  /**
+   * Rotation offset applied to the Actor when snapping to east/south/up.
+   */
+  FRotator _rotationOffset;
```
nithinp7, line 477: This should be a `UProperty` so it gets saved and restored right? Or should the offset be reset everytime like it is now? Related to that, you could make this `EditAnywhere` so that it can be set just once in the editor. What do you think?

CesiumBen: Yeah those adjustments could work well, it would be ok to set the offset variable directly as well within the code without the get/set as there's no additional functionality.

CesiumBen:
@kring @nithinp7 Going off of our discussion, this PR won't be needed as the user can apply a rotational offset to the actor after calling SnapToEastSouthUp. If you agree that the offset should be handled outside of the CesiumGlobeAnchorComponent then we can ignore this PR.

PR #962, opened 2022-09-01
Make tilesets apply collision settings to all glTF primitives

### LINKED ISSUE #742: Tileset collision settings not applying to all glTF primitives
The Tileset actor's collision settings is currently only being applied to the first `UCesiumGltfPrimitiveComponent` in each `UCesiumGltfComponent`. In reality, each glTF can have multiple primitives so this looks like a bug, we should iterate through all primitives in each glTF and apply the same logic. Thanks @xuelongmu for finding this.

https://github.com/CesiumGS/cesium-unreal/blob/3a02420b8b5066c889c798c57c291c67d4c5a2c0/Source/CesiumRuntime/Private/Cesium3DTileset.cpp#L1304-L1305

j9liu:
Fixes #742 . This PR modifies `applyActorCollisionSettings` such that the function iterates over all of the `UCesiumGltfComponent`'s children and applies the setting.

### DIFF
--- a/Source/CesiumRuntime/Private/Cesium3DTileset.cpp
+++ b/Source/CesiumRuntime/Private/Cesium3DTileset.cpp
@@ -1640,7 +1640,7 @@ void applyActorCollisionSettings(
 
   for (USceneComponent* ChildComponent : ChildrenComponents) {
     UCesiumGltfPrimitiveComponent* PrimitiveComponent =
-        Cast<UCesiumGltfPrimitiveComponent, USceneComponent>(ChildComponent);
+        Cast<UCesiumGltfPrimitiveComponent>(ChildComponent);
     if (PrimitiveComponent != nullptr) {
       if (PrimitiveComponent->GetCollisionObjectType() !=
           BodyInstance.GetObjectType()) {
--- a/Source/CesiumRuntime/Private/Cesium3DTileset.cpp
+++ b/Source/CesiumRuntime/Private/Cesium3DTileset.cpp
@@ -1640,7 +1640,7 @@ void applyActorCollisionSettings(
 
   for (USceneComponent* ChildComponent : ChildrenComponents) {
     UCesiumGltfPrimitiveComponent* PrimitiveComponent =
-        static_cast<UCesiumGltfPrimitiveComponent*>(ChildComponent);
+        Cast<UCesiumGltfPrimitiveComponent, USceneComponent>(ChildComponent);
     if (PrimitiveComponent != nullptr) {
       if (PrimitiveComponent->GetCollisionObjectType() !=
           BodyInstance.GetObjectType()) {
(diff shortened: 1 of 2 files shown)


cesium-concierge:
Thank you so much for the pull request @j9liu! I noticed this is your first pull request and I wanted to say welcome to the Cesium community!

Source/CesiumRuntime/Private/Cesium3DTileset.cpp
```diff
@@ -1635,18 +1635,24 @@ void removeCollisionForTiles(
...
-        BodyInstance.GetObjectType()) {
-      PrimitiveComponent->SetCollisionObjectType(BodyInstance.GetObjectType());
-    }
-    const UEnum* ChannelEnum = StaticEnum<ECollisionChannel>();
-    if (ChannelEnum) {
-      FCollisionResponseContainer responseContainer =
-          BodyInstance.GetResponseToChannels();
-      PrimitiveComponent->SetCollisionResponseToChannels(responseContainer);
+  TArray<USceneComponent*> ChildrenComponents;
+  Gltf->GetChildrenComponents(false, ChildrenComponents);
```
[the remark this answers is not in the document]
j9liu: I looked inside `GetChildrenComponents` and it actually calls `GetAttachedChildren`, so I think this works.

Source/CesiumRuntime/Private/Cesium3DTileset.cpp
```diff
@@ -1635,18 +1635,24 @@ void removeCollisionForTiles(
...
-    if (ChannelEnum) {
-      FCollisionResponseContainer responseContainer =
-          BodyInstance.GetResponseToChannels();
-      PrimitiveComponent->SetCollisionResponseToChannels(responseContainer);
+  const TArray<USceneComponent*>& ChildrenComponents =
+      Gltf->GetAttachChildren();
+
+  for (USceneComponent* ChildComponent : ChildrenComponents) {
+    UCesiumGltfPrimitiveComponent* PrimitiveComponent =
+        static_cast<UCesiumGltfPrimitiveComponent*>(ChildComponent);
```
kring, line 1643: static_cast'ing a `USceneComponent` to a `UCesiumGltfPrimitiveComponent` when it is not actually a `UCesiumGltfPrimitiveComponent` will result in undefined behavior. static_cast isn't smart enough to return nullptr in this scenario. In regular C++ you would need `dynamic_cast`, but in Unreal we should use Unreal's `Cast` template.

kring: The original code has this problem, too. 😆 We should fix it while we're in here, though.

kring:
Looks great! Just one more request: can you please add this to CHANGES.md? We just did a release, so you'll need to add a new section at the top and name the version/date as just `? - ?` for now.

j9liu:
@kring done!

kring:
Thanks @j9liu!

PR #963, opened 2022-09-02
Faster and smoother dithered fading

nithinp7:
The way we previously implemented dithered fading, the fading-in tile needed to be completely opaque before the fading-out tile could start fading. Otherwise, if both were dithered at the same time, you could see straight through both tiles for some fragments. @kring suggested that since the dither texture sample happens in screen-space, we could use the same fade percentage (dither threshold) for both fading-in and fading-out tiles, but flip the result of the dither sample for fading out tiles. The result is that we get a water-tight blending between the two LODs. This allows the transition to be smoother - since there are no longer two visibly distinct phases in the transitions, and faster - since the old tile doesn't need to wait for the new tile to fully fade in.


cesium-concierge:

kring:
Thanks @nithinp7! Please add this to the changelog in main.

Issue #955, opened 2022-08-26
RunUAT.sh doesn't produce a package


widarr:
I'm trying to build CesiumForUnreal on Arch Linux for Unreal Engine 5 according to the documentation: https://github.com/CesiumGS/cesium-unreal/blob/ue4-main/Documentation/developer-setup-linux.md
(sidenote: the directory $UNREAL_ENGINE_DIR/Engine/Source/ThirdParty/Linux/LibCxx doesn't exist, I used
$UNREAL_ENGINE_DIR/Engine/Source/ThirdParty/Unix/LibCxx instead)

The compilation succeeded but now I'm stuck at the packaging step described here: https://github.com/CesiumGS/cesium-unreal/blob/ue4-main/Documentation/packaging-guide.md

The $UNREAL_ENGINE_DIR/Engine/Build/BatchFiles/RunUAT.sh command just doesn't produce any output.
This is all I get:

Running AutomationTool...

Fixing inconsistent case in filenames.
Setting up Mono
Setting up bundled DotNet SDK
AutomationTool.dll is up to date
Start UAT Interactively: ./AutomationTool BuildPlugin -Plugin=/home/vidar/dev/cesium-unreal-samples/Plugins/cesium-unreal/CesiumForUnreal.uplugin -Package=/home/vidar/dev/packages/CesiumForUnreal -CreateSubFolder -TargetPlatforms=Linux

But the dev/packages folder stays empty. Am I missing something?

Addendum: I also tried to check out the ue5-main branch, got it to compile alright but the outcome is the same as above: the RunUAT.sh script doesn't produce a package.

widarr:
For the sake of completeness: I tried installing the release version of the plugin  (https://github.com/CesiumGS/cesium-unreal/releases/tag/v1.16.2-ue5)
But then I always get the dreaded "Missing Modules" error message ("Engine modules cannot be compiled at runtime. please build through your IDE.")

Nathan127:
I'm glad I'm not the only one having issues running Cesium for UE5 on Linux. I've tested this on Ubuntu 20.04 as well as Ubuntu 22.04 and I'm not able to run the release version of the plugin either. (Same "Engine modules cannot be compiled at runtime..." error).

I tried building the plugin using those same guides but I'm also running into issues on the RunUAT. For me, script fails with the following errors:

[...]

PR #966, opened 2022-09-08
Updated UE5 Linux setup documentation

Nathan127:
This workflow has been tested on Ubuntu 20.04.3 LTS with UE 5.0.3. It may not fix all issues encountered with compiling Cesium-Unreal on Linux, but it also fixes several obvious out of date instructions.

Based on troubleshooting done in #955


Nathan127:
@cesium-concierge CLA has sent!

kring:
Thanks @Nathan127! I edited the note at the end to remove the bit about "may not have full functionality" because, to my knowledge, everything works well in UE5. Let me know if you know of anything that doesn't. Thanks again for helping to diagnose this and for opening the PR with the doc fixes!

PR #968, opened 2022-09-09
Fix Rich Text Block image decorators not showing

### LINKED ISSUE #967: Rich text block image decorators don't show up in latest versions of plugin
Using a rich text block and an image decorator, you can set up a way to show images in-line with text as described in the [UMG Rich Text Block documentation](https://docs.unrealengine.com/5.0/en-US/umg-rich-text-blocks-in-unreal-engine/#decoratorclasses).

In the latest versions of C4U (v1.15 and up) the image shows in PIE and everything's working as expected. However, in packaged games (tried with shipping and DebugGame), the image does not show. Rather, the HTML tag shows, e.g. ``

In a prior version of C4U (v1.10.1) the image does show properly in both contexts.

I uploaded an [example project](https://drive.google.com/file/d/1R-ocirelq-1MpGFdMvPKst0od-J00yFC/view?usp=sharing) for reproducing the issue. If you use plugin version v1.10.1 and package the samples project, you'll see:

image

In v1.17.0, you'll see:

image

as a clue, I'm seeing the following error:
`LogSlateStyle: Warning: Unable to find Slate Widget Style 'img'. Using FTextBlockStyle defaults instead.`

In a conversation with Epic, their engineer said:

> It looks like the 'img' is being treated as a style name from the style table instead of being recognized as a decorator. It's not immediately clear how a plugin would be causing the decorator class to be missing or unrecognized, you could drop some breakpoints in URichTextBlockImageDecorator to see if it's functions are being called (meaning it exists but isn't finding the image) or if it's missing altogether. The Supports function in there will return false if a matching brush isn't found and will fall back to looking for a style, you could set bWarnIfMissing to true to log out warnings whenever the decorator looks for a specific image and doesn't find it.

joseph-kaile:
Fixes #967

The problem was naming collision, both decorators had the name `FRichInlineImage`. By renaming it to `FScreenCreditsDecorator` fixed the issue.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,7 @@
 ##### Fixes :wrench:
 
 - Fixed a bug where collision settings were only being applied to the first primitive in a glTF.
+- Fixed a bug where the Screen Credits Decorator prevented the Rich Text Block Image Decorator from working.
 
 ### v1.17.0 - 2022-09-01
 
--- a/Source/CesiumRuntime/Private/ScreenCreditsWidget.cpp
+++ b/Source/CesiumRuntime/Private/ScreenCreditsWidget.cpp
@@ -22,9 +22,9 @@
 #include <string>
 #include <vector>
 
-class SRichInlineImage : public SCompoundWidget {
+class SCreditImage : public SCompoundWidget {
 public:
-  SLATE_BEGIN_ARGS(SRichInlineImage) {}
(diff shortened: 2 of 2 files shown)


cesium-concierge:

* :x: Missing CLA.

kring:
Thanks @joseph-kaile!
CC @xuelongmu

PR #973, opened 2022-09-21
Add option to change mobility

### LINKED ISSUE #970: CesiumGltfComponent (and primitive) should inherit its mobility from the tileset and its root component
Currently Cesium for Unreal hard-codes `Gltf->SetMobility(EComponentMobility::Static);`. Usually that's what we want, but sometimes, in some applications, it's actually useful to move tilesets, and this makes it impossible. Instead, we should inherit the mobility of the Tileset's root component. That way the user can easily make a tileset movable by changing the property in one place.

This is needed for Project Anywhere XR, so would be good to do it soon.

CC @AlbanBERGERET-Epic (who reported this)

nithinp7:
We recently changed the components in Cesium3DTilesets to have static mobility (previously they were movable), allowing for several optimizations in the engine. In certain use cases however, it may still be useful to have a movable tileset. This PR adds an option in the Cesium3DTileset details panel to change the default mobility.

Fixes #970

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,7 @@
 ##### Additions :tada:
 
 - Improved the dithered transition between levels-of-detail, making it faster and eliminating depth fighting.
+- Add an option on the `Cesium3DTileset` to change the tileset's mobility. This allows users to make tileset movable at runtime, if needed.  
 
 ##### Fixes :wrench:
 
--- a/Source/CesiumRuntime/Private/Cesium3DTileset.cpp
+++ b/Source/CesiumRuntime/Private/Cesium3DTileset.cpp
@@ -108,7 +108,6 @@ ACesium3DTileset::ACesium3DTileset()
 
   this->RootComponent =
       CreateDefaultSubobject<UCesium3DTilesetRoot>(TEXT("Tileset"));
-  this->RootComponent->SetMobility(EComponentMobility::Static);
 
   PlatformName = UGameplayStatics::GetPlatformName();
 }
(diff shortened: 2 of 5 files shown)


cesium-concierge:

kring:
@AlbanBERGERET-Epic I updated this PR to include a version of the change you suggested on Slack.

If the mobility is Movable, georeference and origin rebase changes just use `SetRelativeTransform` to update the position. This call just takes care of everything, including physics.

But if the mobility is Static or Stationary, calling `SetRelativeTransform` will cause UE to print lots of warnings, and it will be very slow. It may not even update the transform at all, I'm not sure. So instead, for these mobility types, we continue to use `SetRelativeTransform_Direct` to update the transform "behind the scenes". And then make some other calls to keep other transform-dependent data in sync... including, at Alban's suggestion, calling `SendPhysicsTransform` to update the physics transform after moving.

You might be wondering why we're updating the transform of static objects at all.

In UE4, it's necessary for world origin rebasing in order to maintain position accuracy. In fact, our approach to updating position is inspired by the engine's implementation of `ApplyWorldOffset`.

In both UE4 and UE5, it's necessary to allow the CesiumGeoreference position to change at runtime. Arguably we shouldn't allow this with a static tileset, because technically the tileset is moving in UE coordinates. But the thinking is that a georeference change changes the coordinates of the whole world (just like a world origin rebase), and so the tileset's position isn't actually changing relative to the rest of the world. And so things like precomputed lighting might still be ok, and so we shouldn't force the user to use a movable tileset just to be able to change the georeference.

I'm really not very confident we have all this correct, though. So I'm very interested in your feedback, Alban. I've created a UE5 version of this branch [here](https://github.com/CesiumGS/cesium-unreal/tree/inherit-movability-ue5), so it's easier to try it out. You can download a build here:
https://github.com/CesiumGS/cesium-unreal/suites/8411981375/artifacts/372141357

AlbanBERGERET-Epic:
I just tested and this update works fine in ProjectAnywhereXR. 
Thanks @kring !

PR #733, opened 2022-01-21
Fix issues with rotator transforms

nithinp7:
Addresses: #724

Issue #265, opened 2021-03-23
Determine if ComputeEastNorthUpToUnreal has a bug


nithinp7:
I must have introduced this in #222 when originally implementing `CesiumGeoreference::ComputeEastNorthUpToUnreal` based on a previous version of the function in `GlobeAwareDefaultPawn`. I ran into the issue that `GlobeAwareDefaultPawn::MoveUp_World` wasn't working when far from the origin. This is fixed in #264 but the underlying issue that originally caused it is unclear to me, but it seems to have something to do with `ComputeEastNorthUpToUnreal`.

nithinp7:
The commented out code that mysteriously didn't seem to work:
https://github.com/CesiumGS/cesium-unreal/blob/000b33741d25ad50733b5896a3b4aab05bdb2273/Source/CesiumRuntime/Private/GlobeAwareDefaultPawn.cpp#L63-L69

kring:
@nithinp7 do you think this could have been fixed by #371?
If not, is it worth spending a bit of time to confirm that it's really a bug, and then we can prioritize a fix accordingly?

nithinp7:
I remember this giving wildly wrong results so I'm not sure if it would have been fixed there. I will try to investigate the problem again this week so that at least we can prioritize.

kring:
It's a year later and we're still not sure if this is a bug. Let's say it's not and close this. :)

PR #983, opened 2022-10-16
Fix broken euler angles and quaternion conversions

nithinp7:
Somewhat related to #265 and #733.

I think our East-North-Up <--> Unreal conversion functions in the Georeferencing API are quite broken. This PR attempts to fix some if not all of it. 

We had functions to transform rotators (pitch, yaw, roll) from ENU to Unreal and vice-versa. These were implemented by:
Step 1: Convert FRotators (pitch, yaw, roll) into quaternions.
Step 2: Compute the Unreal <--> ENU matrix and convert the rotation part into a quaternion.
Step 3: And then multiply the above two quaternions to transform a rotation between Unreal and ENU.
Step 4: And of course convert back from quaternion to pitch, yaw, roll at the end in the destination coordinate system.

Implementation mistakes fixed:
- There was a glaring mistake (made by me!) that flipped the direction of conversions between UE and ENU. So UE to ENU mistakenly converted from ENU to UE and vice-versa. In fact I had started fixing it here #733 but found that rotations still remained broken overall.
- It turns out ComputeEastNorthUpToUnreal was _actually_ computing a East-South-Up to Unreal pure rotation matrix! In fact some code calling this function from the GlobeAwareDefaultPawn and the GlobeAnchorComponent seemed to acknowledge in comments that the result was an ESU to Unreal rotation. 
- The rotator and quaternion conversion functions were built upon the above function and so were mistakenly doing ESU to UE and vice versa. In fact, an _actual_ ENU --> Unreal matrix converted to a quaternion (step 2) might be undefined (or at least incorrect to use) since ENU and Unreal have opposite handedness.
- Some code to convert back from glm quaternions to Unreal FRotators was innocently using the pitch(..), yaw(..), and roll(..) functions from the glm library. While glm is usually agnostic of coordinate system conventions (e.g., handedness, up-direction, etc.), computing pitch, yaw, and roll from a quaternion implies some assumed convention of which axis is up, right, and forward and an implied rotation-order convention for euler angles (or yaw, pitch, roll in this case). Even determining whether positive angles rotate CW or CCW for each axis comes down to convention. So it's no surprise that the assumed conventions of glm's quaternion --> pitch, yaw, roll did not match that of Unreal. 

My solution was primarily to get rid of ENU <--> UE conversions and just do ESU <--> UE instead. This will make it much easier to reason about the rotations since the rotation conventions and handedness will remain the same through the conversion.

TODOs before merge:
* [x] Address and remove TODOs in the code
* [x] Add backward compatibility via core-redirects where possible. Some of these functions were completely broken, so we'll have to weigh how likely it is somebody is relying on the broken version.
* [ ] Find a way to unit test this, even if it is ad-hoc! Or alternatively move this into Cesium Native and unit test there??

I started looking into this after seeing this community forum post recently:
https://community.cesium.com/t/correct-way-to-set-euler-angles-for-actor-in-c/20808

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -15,6 +15,11 @@
 - Significantly reduced frame-rate dips during asynchronous tile loading by eliminating thread pool monopilaztion by Cesium tasks.
 - Improved the tile destruction sequence by allowing it to defer being destroyed to future frames if it is waiting on asynchronous work to finish. Previously we would essentially block the main thread waiting for tiles to become ready for destruction.
 
+
+##### Breaking Changes :mega:
+
+- Removed a number of buggy funnctions on the `CesiumGeoreference`: `ComputeEastNorthUp`, `TransformRotatorEastNorthUpToUnreal`, and `TransformRotatorUnrealToEastNorthUp`. These functions were replaced with "EastSouthUp" counterparts.
+
 ### v1.18.0 - 2022-10-03
 
 ##### Additions :tada:
--- a/Source/CesiumRuntime/Private/CesiumGeoreference.cpp
+++ b/Source/CesiumRuntime/Private/CesiumGeoreference.cpp
@@ -911,55 +911,55 @@ ACesiumGeoreference::InaccurateTransformUnrealToEcef(const FVector& ue) const {
   return FVector(ecef.x, ecef.y, ecef.z);
 }
 
-glm::dquat ACesiumGeoreference::TransformRotatorUnrealToEastNorthUp(
+glm::dquat ACesiumGeoreference::TransformRotatorUnrealToEastSouthUp(
     const glm::dquat& UeRotator,
     const glm::dvec3& UeLocation) const {
-  return this->_geoTransforms.TransformRotatorUnrealToEastNorthUp(
+  return this->_geoTransforms.TransformRotatorUnrealToEastSouthUp(
       glm::dvec3(CesiumActors::getWorldOrigin4D(this)),
       UeRotator,
       UeLocation);
 }
 
-FRotator ACesiumGeoreference::InaccurateTransformRotatorUnrealToEastNorthUp(
+FRotator ACesiumGeoreference::InaccurateTransformRotatorUnrealToEastSouthUp(
     const FRotator& UERotator,
     const FVector& ueLocation) const {
-  glm::dquat q = TransformRotatorUnrealToEastNorthUp(
+  glm::dquat q = TransformRotatorUnrealToEastSouthUp(
       VecMath::createQuaternion(UERotator.Quaternion()),
       VecMath::createVector3D(ueLocation));
   return VecMath::createRotator(q);
 }
 
-glm::dquat ACesiumGeoreference::TransformRotatorEastNorthUpToUnreal(
-    const glm::dquat& EnuRotator,
+glm::dquat ACesiumGeoreference::TransformRotatorEastSouthUpToUnreal(
+    const glm::dquat& EsuRotator,
     const glm::dvec3& UeLocation) const {
-  return this->_geoTransforms.TransformRotatorEastNorthUpToUnreal(
+  return this->_geoTransforms.TransformRotatorEastSouthUpToUnreal(
       glm::dvec3(CesiumActors::getWorldOrigin4D(this)),
-      EnuRotator,
+      EsuRotator,
       UeLocation);
 }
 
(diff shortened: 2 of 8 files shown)


cesium-concierge:

kring:
Thanks @nithinp7, your changes all make sense to me!

This is a huge improvement that fixes the biggest problems, but I think we would still benefit from a more holistic look at how this all works. For example, the notion of converting "ESU" to Unreal is... weird. The Unreal coordinate system in this case _is_ an ESU system, right? So what are we actually doing? Well, we're transforming between an ESU system centered at globe position and an ESU system centered at another globe position. And if I'm not mistaken, both use units of centimeters (which is not 100% clear in the API).

If we switch Unreal to use the new `LocalHorizontalCoordinateSystem` in cesium-native, one way we could make this more obvious and explicit is by allowing transformations between two `LocalHorizontalCoordinateSystem` instances. So we can have one instance that represents the georeferenced Unreal world coordinate system. And another that represents an ESU system (or whatever we want) somewhere on the globe, and then we can ask for a matrix to transform between them. And then maybe we don't need all these difficult-to-name-precisely conversion methods. (they're difficult for us to name, but it's probably even more difficult for our users to decide which one to use!)

Mostly thinking out loud here, we certainly don't need to implement anything like this for this PR.

nithinp7:
From offline discussion, we have decided _not_ to provide backward compatibility with core redirects. The reasoning is that the old behavior was outright broken, so any implementation depending on the old behavior would silently break if we don't do an explicitly breaking change.

nithinp7:
@kring Regarding the ESU <--> Unreal matrix, I think it should just be a plain 3x3 rotation matrix now right? So it shouldn't have any notion of centimeters / meters? 

Otherwise I definitely agree with the ideas you mentioned of explicitly defining local-horizontal coordinate systems to convert between. Although I think there might be a lot of complexity we'll inherit trying to represent arbitrary euler angle conventions in Cesium Native. If Cesium Native can be constrained to _only_ be matrix-based, then we only really have to consider possible left-handed to right-handed CRS switches + scale differences - which is easy. Arbitrary euler angle conventions will get quite tedious to implement - each euler angle convention yields unique matrix <--> euler angles and quaternion <--> euler angle conversions. And of course finding rotations between two CRS with different conventions / handedness is probably undefined.

So anyways... I think it is still convenient to be able to say, here is another local ESU coordinate system with the exact same conventions / handedness as the Unreal world. And we can freely transform anything between this CRS and the Unreal world, without any undefined behavior. If we tried to abstract some of this in Cesium Native, I worry we might enable incorrect / undefined API usage like "convert euler angles from ENU to Unreal" - i.e., the types of things hopefully fixed in this PR.

The formula then, to minimize the pain when implementing new Cesium Native integrations might be:
1) Provide engine <--> Ecef position / direction conversion functions and provide a way to get the actual transformation matrix.
2) Define a local ESU or ENU system (choose based on engine's handedness), this CRS will implicitly have all the other conventions of the engine, including scale and rotation conventions.
3) Provide a way to get the transformation matrix and rotation matrix between this local (georeferenced?) CRS and the engine.
4) Provide a way to convert positions, directions, rotators, euler angles, quaternions, etc. between the local CRS and the engine CRS.
5) _Crucially_, do not provide any rotation conversion functions between the engine CRS and anything else (including ECEF, local tile CRS, or any other ENU / ESU with different conventions). Everything else should be handled with matrices - even 3x3 matrices between such systems should not be _called_ rotation matrices since they might change handedness / scale.

So out of that list 1) could be implemented in Cesium Native, and isn't much work at all. But 2-4 seem to be strictly tied to the particular choices of the game engine, so it makes more sense to me to just implement those conversion functions there. Even converting from engine types to glm types seems error-prone to me when we are talking about euler angles. Hopefully that wasn't too verbose - like you said most of this stuff is out of the scope of this PR anyways, but what do you think?

nithinp7:
@kring Regarding this PR itself, I think it should be good to go. I cleared some of my doubts about whether certain conversions were affected by handedness / conventions. Notably, quaternion <--> matrix conversions do not have a notion of handedness - so we can do them in glm or Unreal equivalently. I also explicitly added a note in the quaternion / matrix --> rotator conversion functions that we assume the input uses Unreal conventions.

kring:
Looks good! I just pushed a small change, to move some comments inside some methods into the doc comments. I'm going to merge it, but please take a quick glance @nithinp7 and tell me if anything looks suspect.

PR #990, opened 2022-10-28
Implement derived content caching

nithinp7:
Builds on top of https://github.com/CesiumGS/cesium-native/pull/569 (see that PR for more details)

This is the simplest implementation of derived content caching, we simply serialize / deserialize the in-memory gltfs along with all the buffers. Currently this lets us skip all the Cesium Native loading, but we still have to go through the Cesium for Unreal loading. Before merging we plan to:
* [x] Serialize / deserialize cooked physics meshes into the derived content cache (as of writing this physics cooking is the _biggest_ remaining load cost).
* [x] Serialize the glTF after `CreateOffGameThread(..)` - since the loading process may mutate the in-memory glTF, e.g., generate mipmaps.
* [ ] Use LZ compression on the bulk data before writing to disk, to save a bit of disk-space? From offline discussion we are not sure how big of a performance / memory win this will be though.
* [ ] Finalize a future-proof archive format that can be versioned and properly invalidated when settings change.

Still a good bit of work to do here, but so far this is a good show-case implementation of the Cesium Native changes. 

Some preliminary results:
In the below picture, the load tasks containing the big yellow task (physics cooking) are cache misses (fresh response data from the network with no derived cache content). The super fast ones are cache hits. The load time difference (per tile) is already incredible, a rough average improvement might be ~400ms down to ~40ms for a tile's load work. The overall apparent load time seems improved as well, but may still be slightly bottlenecked by the sqlite reader thread pool (we can experiment with increasing the number of threads there).

CachedPhysics


cesium-concierge:

j9liu:

PR #998, opened 2022-11-20
Get the plugin compiling on UE 5.1.

kring:
This is only barely tested, but the basics work.


cesium-concierge:

Issue #1007, opened 2022-12-03
Segfault in InitRHI when creating textures on Linux and macOS


maumock:
Small issues with tag v1.20.0-ue5.
1. generated lib names have "d.a" instead of ".a" in their name
2. crash when adding cesium terrain

Happy to test any hot fixes or patches if needed.

TLDR

**Machine**

- unreal engine code base: 5.0.3-release
- Ubuntu 22.04 (clean install) also tried latest pop os
- Wayland
- RTX 3060 12gb 
- AMD 12 core
- 32gb ram

**Lib Issue**
On compile all the libraries in cesium-unreal (Source/ThirdParty/lib/Linux-x64) have "d.a" in the name (i.e. libCesiumGeometryd.a) instead of ".a" (i.e. libCesiumGeometry.a). Obviously this creates problems when linking. After renaming the lib files, compilation will work. 

**Crash**
When clicking on "Cesium World Terrain + Bing Maps World imagery", a crash occurs immediately. with the following error: 
```
Caught signal 11 Segmentation fault

libUnrealEditor-CesiumRuntime.so!(anonymous namespace)::FCesiumTextureResource::InitRHI() [/home/rho/dev/packages/CesiumForUnreal/HostProject/Plugins/CesiumForUnreal/Source/CesiumRuntime/Private/CesiumTextureUtility.cpp:299]
libUnrealEditor-RenderCore.so!FRenderResource::InitResource() [/home/rho/dev/UnrealEngine/Engine/Source/./Runtime/RenderCore/Private/RenderResource.cpp:120]
libUnrealEditor-CesiumRuntime.so!TEnqueueUniqueRenderCommandType<CesiumTextureUtility::loadTextureGameThreadPart(CesiumTextureUtility::LoadedTextureResult*)::Cesium_InitResourceName, CesiumTextureUtility::loadTextureGameThreadPart(CesiumTextureUtility::LoadedTextureResult*)::$_125>::DoTask(ENamedThreads::Type, TRefCountPtr<FGraphEvent> const&) [/home/rho/dev/UnrealEngine/Engine/Source/Runtime/RenderCore/Public/RenderingThread.h:193]
[...]

PR #1009, opened 2022-12-05
Use legacy texture creation for mac and linux

nithinp7:
Hopefully this is a workaround for #1007, although it would be slightly better to get the current texture creation path to work for mac and linux. Both the Vulkan and Metal RHI implementations in Unreal do not have async texture creation implemented, so the current texture creation path on these platforms does the next best thing by avoiding a texture memcpy on the game thread (by wrapping the source texture memory into an Unreal-specific interface for direct GPU upload). But it won't be a huge loss to go back to the legacy texture creation code paths on these platforms.

TODO:
- Do we need to hardcode any other platforms to use the legacy code path, Android maybe?

### DIFF
--- a/Source/CesiumRuntime/Private/CesiumTextureUtility.cpp
+++ b/Source/CesiumRuntime/Private/CesiumTextureUtility.cpp
@@ -40,6 +40,25 @@ void legacy_populateMips(
     bool generateMipMaps) {
   uint32 width = static_cast<uint32>(image.width);
   uint32 height = static_cast<uint32>(image.height);
+
+  if (image.mipPositions.empty()) {
+    // Only the full image is available.
+    FTexture2DMipMap* pMip = new FTexture2DMipMap();
+    platformData.Mips.Add(pMip);
+    pMip->SizeX = width;
+    pMip->SizeY = height;
+
+    pMip->BulkData.Lock(LOCK_READ_WRITE);
+    void* pDest =
+        pMip->BulkData.Realloc(static_cast<int64>(image.pixelData.size()));
+
+    FMemory::Memcpy(pDest, image.pixelData.data(), image.pixelData.size());
+
+    pMip->BulkData.Unlock();
+
+    return;
+  }
+
   for (const CesiumGltf::ImageCesiumMipPosition& mipPos : image.mipPositions) {
     FTexture2DMipMap* pMip = new FTexture2DMipMap();
     platformData.Mips.Add(pMip);
--- a/Source/CesiumRuntime/Private/CesiumTextureUtility.cpp
+++ b/Source/CesiumRuntime/Private/CesiumTextureUtility.cpp
@@ -613,7 +613,6 @@ TUniquePtr<LoadedTextureResult> loadTextureAnyThreadPart(
     // data.
     pResult->textureSource = LegacyTextureSource{};
   }
-#endif
 
   return pResult;
 }
--- a/Source/CesiumRuntime/Private/CesiumTextureUtility.cpp
+++ b/Source/CesiumRuntime/Private/CesiumTextureUtility.cpp
@@ -35,8 +35,10 @@ using namespace CesiumGltf;
 
 namespace {
 #if LEGACY_TEXTURE_CREATION
-// Legacy texture creation code path. Only for testing, no safety checks are
-// done.
+// Legacy texture creation code path - creates textures using Unreal's
+// FTexture2DMipMap and Texture2D::UpdateResource(). While this is slightly
+// less efficient than other approaches we've found, this is still the only
+// approach that works on certain platforms (as far as we currently know).
 
 void legacy_populateMips(
     FTexturePlatformData& platformData,
--- a/Source/CesiumRuntime/Private/CesiumTextureUtility.cpp
+++ b/Source/CesiumRuntime/Private/CesiumTextureUtility.cpp
@@ -25,21 +25,15 @@
 #include <stb_image_resize.h>
 #include <variant>
 
-#if PLATFORM_LINUX || PLATFORM_MAC
-#define LEGACY_TEXTURE_CREATION 1
-#else
-#define LEGACY_TEXTURE_CREATION 0
-#endif
-
 using namespace CesiumGltf;
 
 namespace {
-#if LEGACY_TEXTURE_CREATION
 // Legacy texture creation code path - creates textures using Unreal's
 // FTexture2DMipMap and Texture2D::UpdateResource(). While this is slightly
-// less efficient than other approaches we've found, this is still the only
-// approach that works on certain platforms (as far as we currently know).
-
+// less efficient than other approaches we've found, the other approaches
+// are not always supported on certain platforms. Once image sources from tiles
+// and raster tiles can be safely read from the render-thread, this approach
+// should no longer be necessary.
 void legacy_populateMips(
     FTexturePlatformData& platformData,
     const CesiumGltf::ImageCesium& image,
@@ -76,7 +70,6 @@ void legacy_populateMips(
     }
   }
 }
-#endif
 
 struct GetImageFromSource {
   CesiumGltf::ImageCesium*
@@ -294,6 +287,12 @@ public:
         GetOrCreateSamplerState(deferredSamplerStateInitializer);
 
     if (!this->TextureRHI) {
+      // TODO: Currently copying from `ImageCesium` on the render-thread is
+      // unsafe. Currently we should never hit this code. Remove this in the
+      // future when images from tiles and raster tiles can be safely accessed
+      // from the render thread.
+      checkNoEntry();
+
       // Asynchronous RHI texture creation was not available. So create it now
       // directly from the in-memory cesium mips.
       CesiumGltf::ImageCesium* pImage =
@@ -593,13 +592,6 @@ TUniquePtr<LoadedTextureResult> loadTextureAnyThreadPart(
   pResult->sRGB = sRGB;
   pResult->generateMipMaps = generateMipMaps;
 
-#if LEGACY_TEXTURE_CREATION
-  // Legacy texture creation copies mip data into the FTexturePlatformData.
-  legacy_populateMips(*pResult->pTextureData, image, generateMipMaps);
-  // Mark the image source as legacy, so we later know where to look for image
-  // data.
-  pResult->textureSource = LegacyTextureSource{};
-#else
   if (GRHISupportsAsyncTextureCreation) {
     // Create RHI texture resource asynchronously.
     TRACE_CPUPROFILER_EVENT_SCOPE(Cesium::CreateRHITexture2D)
@@ -609,7 +601,17 @@ TUniquePtr<LoadedTextureResult> loadTextureAnyThreadPart(
   } else {
     // The RHI texture will be created later on the render thread, directly
     // from this texture source.
-    pResult->textureSource = std::move(imageSource);
+    // pResult->textureSource = std::move(imageSource);
+
+    // TODO: Use the above texture creation path instead when the image source
+    // becomes render-thread safe in the future. Currently that is not the
+    // case for any `CesiumGltf::ImageCesium` from tiles or raster tiles.
+
+    // Legacy texture creation copies mip data into the FTexturePlatformData.
+    legacy_populateMips(*pResult->pTextureData, image, generateMipMaps);
+    // Mark the image source as legacy, so we later know where to look for image
+    // data.
+    pResult->textureSource = LegacyTextureSource{};
   }
 #endif
 
--- a/Source/CesiumRuntime/Private/CesiumTextureUtility.cpp
+++ b/Source/CesiumRuntime/Private/CesiumTextureUtility.cpp
@@ -25,7 +25,11 @@
 #include <stb_image_resize.h>
 #include <variant>
 
+#if PLATFORM_LINUX || PLATFORM_MAC
+#define LEGACY_TEXTURE_CREATION 1
+#else
 #define LEGACY_TEXTURE_CREATION 0
+#endif
 
 using namespace CesiumGltf;
 


kring:
The next texture creation stuff seems to be working fine on my Android phone, even before this PR.
https://github.com/CesiumGS/cesium-unreal/pull/1011

I foolishly merged ue5-main into this branch, though. So I'll merge it there and then merge the one commit into ue4-main.

kring:
This is strange though isn't it, that Vulkan on Android is working fine, but Vulkan on Linux is not? Any idea why the difference?

kring:
I force-pushed to switch this back to ue4-main instead.

kring:
There's a comment: "Legacy texture creation code path. Only for testing, no safety checks are done."

After this PR, we're not using it only for testing anymore. Do we need to add some safety checks?

kring:
I can't seem to get the Vulkan preview rendering to work in UE5.1 to test this theory, but this seems odd... At the end of `loadTextureAnyThreadPart`, there are these lines:

```
  // Replace the image pointer with an index, in case the pointer gets
  // invalidated before the main thread loading continues.
  if (result && std::get_if<GltfImagePtr>(&result->textureSource)) {
    result->textureSource = GltfImageIndex{source};
  }
```

But then in `loadTextureGameThreadPart`, an `FCesiumTextureResource` is constructed with that texture source and InitRHI access it using `GetImageFromSource`. But `GetImageFromSource` only has support for GltfImagePtr and EmbeddedImageSource; all others return nullptr. So that seems sure to lead to a crash in when we try to construct `FCesiumTextureData` with a nullptr. There's a `check`, but that won't do anything in release builds. I'm probably missing some details, but it seems suspicious.

nithinp7:
@kring I probably did not arrive at the most sensible structure for this, I tried to abstract the image source concept but it probably made things more confusing.

I think maybe the answer to what you asked about is in this overload of `loadTextureGameThreadPart`:
```
UTexture2D* loadTextureGameThreadPart(
    LoadedTextureResult* pHalfLoadedTexture) {
  if (!pHalfLoadedTexture) {
    return nullptr;
  }

  GltfImageIndex* pImageIndex =
      std::get_if<GltfImageIndex>(&pHalfLoadedTexture->textureSource);
  if (pImageIndex) {
    pHalfLoadedTexture->textureSource = pImageIndex->resolveImage(model);
  }

  return loadTextureGameThreadPart(pHalfLoadedTexture);
}
```

The image index is a temporary replacement for an image pointer, so that the actual pointer can be re-resolved as needed during the game thread part of the loading. So before the actual game thread loading commences, this function is called which substitutes the image index for an image pointer again, using the model. I did it this way so the other loadTextureInGameThread didn't require a "model" since the texture could be from anywhere (e.g., raster overlay).

nithinp7:
Also I'll repeat what I mentioned offline that I tried Vulkan on Windows (with main), this time in UE5.1 and did not have any issues. So I wonder if there is something platform-specific in the Linux case. Regarding mac, there may be a more obvious RHI level issue that we can solve - since I never really tested the new-ish texture code with Metal.

kring:
@nithinp7 I finally was able to package the plugin for UE 5.1 and run the samples on my Linux system, and can confirm it worked fine even before this PR. With Vulkan:

```
Cmd: r.RHI.Name
Running on the Vulkan RHI
```

kring:
I spoke too soon. A release build installed as an engine plugin works fine. A debug build embedded in the project crashes with an error like this:

```
LoginId:75d8c4a28c1c40579a2f2a569a87447d-000003e8
EpicAccountId:

Caught signal 11 Segmentation fault

libUnrealEditor-CesiumRuntime-Linux-DebugGame.so!(anonymous namespace)::FCesiumTextureData::FCesiumTextureData(CesiumGltf::ImageCesium const&) [/home/kring/cesium/cesium-unreal-samples/Plugins/cesium-unreal/Source/CesiumRuntime/Private/CesiumTextureUtility.cpp:160]
libUnrealEditor-CesiumRuntime-Linux-DebugGame.so!(anonymous namespace)::FCesiumTextureResource::InitRHI() [/home/kring/cesium/cesium-unreal-samples/Plugins/cesium-unreal/Source/CesiumRuntime/Private/CesiumTextureUtility.cpp:299]
libUnrealEditor-RenderCore.so!FRenderResource::InitResource() [/mnt/horde/++UE5/Sync/Engine/Source/./Runtime/RenderCore/Private/RenderResource.cpp:277]
libUnrealEditor-CesiumRuntime-Linux-DebugGame.so!CesiumTextureUtility::loadTextureGameThreadPart(CesiumTextureUtility::LoadedTextureResult*)::$_108::operator()(FRHICommandListImmediate&) const [/home/kring/cesium/cesium-unreal-samples/Plugins/cesium-unreal/Source/CesiumRuntime/Private/CesiumTextureUtility.cpp:814]
libUnrealEditor-CesiumRuntime-Linux-DebugGame.so!TEnqueueUniqueRenderCommandType<CesiumTextureUtility::loadTextureGameThreadPart(CesiumTextureUtility::LoadedTextureResult*)::Cesium_InitResourceName, CesiumTextureUtility::loadTextureGameThreadPart(CesiumTextureUtility::LoadedTextureResult*)::$_108>::DoTask(ENamedThreads::Type, TRefCountPtr<FGraphEvent> const&) [/home/kring/UE_5.1.0/Engine/Source/Runtime/RenderCore/Public/RenderingThread.h:206]
libUnrealEditor-CesiumRuntime-Linux-DebugGame.so!TGraphTask<TEnqueueUniqueRenderCommandType<CesiumTextureUtility::loadTextureGameThreadPart(CesiumTextureUtility::LoadedTextureResult*)::Cesium_InitResourceName, CesiumTextureUtility::loadTextureGameThreadPart(CesiumTextureUtility::LoadedTextureResult*)::$_108> >::ExecuteTask(TArray<FBaseGraphTask*, TSizedDefaultAllocator<32> >&, ENamedThreads::Type, bool) [/home/kring/UE_5.1.0/Engine/Source/Runtime/Core/Public/Async/TaskGraphInterfaces.h:1348]
libUnrealEditor-Core.so!FNamedTaskThread::ProcessTasksNamedThread(int, bool) [/mnt/horde/++UE5/Sync/Engine/Source/./Runtime/Core/Private/Async/TaskGraph.cpp:760]
libUnrealEditor-Core.so!FNamedTaskThread::ProcessTasksUntilQuit(int) [/mnt/horde/++UE5/Sync/Engine/Source/./Runtime/Core/Private/Async/TaskGraph.cpp:648]
libUnrealEditor-Core.so!FTaskGraphCompatibilityImplementation::ProcessThreadUntilRequestReturn(ENamedThreads::Type) [/mnt/horde/++UE5/Sync/Engine/Source/./Runtime/Core/Private/Async/TaskGraph.cpp:2149]
libUnrealEditor-RenderCore.so!RenderingThreadMain(FEvent*) [/mnt/horde/++UE5/Sync/Engine/Source/./Runtime/RenderCore/Private/RenderingThread.cpp:415]
libUnrealEditor-RenderCore.so!FRenderingThread::Run() [/mnt/horde/++UE5/Sync/Engine/Source/./Runtime/RenderCore/Private/RenderingThread.cpp:566]
libUnrealEditor-Core.so!FRunnableThreadPThread::Run() [/mnt/horde/++UE5/Sync/Engine/Source/./Runtime/Core/Private/HAL/PThreadRunnableThread.cpp:25]
libUnrealEditor-Core.so!FRunnableThreadPThread::_ThreadProc(void*) [/mnt/horde/++UE5/Sync/Engine/Source/Runtime/Core/Private/HAL/PThreadRunnableThread.h:185]
libc.so.6!UnknownFunction(0x94b42)
libc.so.6!UnknownFunction(0x1269ff)
```

I also saw this one time, which might be a red herring or it might be a clue:

```
LoginId:75d8c4a28c1c40579a2f2a569a87447d-000003e8
EpicAccountId:

Assertion failed: false [File:Runtime/VulkanRHI/Private/Linux/../VulkanRHIPrivate.h] [Line: 903] A texture was marked as a shading rate source but attachment VRS is not supported on this device. Ensure GRHISupportsAttachmentVariableRateShading and GRHIAttachmentVariableRateShadingEnabled are true before specifying a shading rate attachment.

libUnrealEditor-VulkanRHI.so!FVulkanTexture::GenerateImageCreateInfo(FVulkanTexture::FImageCreateInfo&, FVulkanDevice&, FRHITextureDesc const&, VkFormat*, VkFormat*, bool) [/mnt/horde/++UE5/Sync/Engine/Source/./Runtime/VulkanRHI/Private/VulkanTexture.cpp:363]
libUnrealEditor-VulkanRHI.so!FVulkanTexture::FVulkanTexture(FVulkanDevice&, FRHITextureCreateDesc const&, FRHITransientHeapAllocation const*) [/mnt/horde/++UE5/Sync/Engine/Source/./Runtime/VulkanRHI/Private/VulkanTexture.cpp:1619]
libUnrealEditor-VulkanRHI.so!FVulkanDynamicRHI::RHICreateTexture(FRHITextureCreateDesc const&) [/mnt/horde/++UE5/Sync/Engine/Source/./Runtime/VulkanRHI/Private/VulkanTexture.cpp:938]
libUnrealEditor-VulkanRHI.so!FVulkanDynamicRHI::RHICreateTexture_RenderThread(FRHICommandListImmediate&, FRHITextureCreateDesc const&) [/mnt/horde/++UE5/Sync/Engine/Source/Runtime/VulkanRHI/Public/VulkanDynamicRHI.h:246]
libUnrealEditor-Landscape.so!FLandscapeTexture2DArrayResource::InitRHI() [/mnt/horde/++UE5/Sync/Engine/Source/./Runtime/Landscape/Private/LandscapeEditResources.cpp:122]
libUnrealEditor-RenderCore.so!FRenderResource::InitResource() [/mnt/horde/++UE5/Sync/Engine/Source/./Runtime/RenderCore/Private/RenderResource.cpp:277]
libUnrealEditor-RenderCore.so!TEnqueueUniqueRenderCommandType<BeginInitResource(FRenderResource*)::InitCommandName, BeginInitResource(FRenderResource*)::$_153>::DoTask(ENamedThreads::Type, TRefCountPtr<FGraphEvent> const&) [/mnt/horde/++UE5/Sync/Engine/Source/Runtime/RenderCore/Public/RenderingThread.h:206]
libUnrealEditor-RenderCore.so!TGraphTask<TEnqueueUniqueRenderCommandType<BeginInitResource(FRenderResource*)::InitCommandName, BeginInitResource(FRenderResource*)::$_153> >::ExecuteTask(TArray<FBaseGraphTask*, TSizedDefaultAllocator<32> >&, ENamedThreads::Type, bool) [/mnt/horde/++UE5/Sync/Engine/Source/Runtime/Core/Public/Async/TaskGraphInterfaces.h:1348]
libUnrealEditor-Core.so!FNamedTaskThread::ProcessTasksNamedThread(int, bool) [/mnt/horde/++UE5/Sync/Engine/Source/./Runtime/Core/Private/Async/TaskGraph.cpp:760]
libUnrealEditor-Core.so!FNamedTaskThread::ProcessTasksUntilQuit(int) [/mnt/horde/++UE5/Sync/Engine/Source/./Runtime/Core/Private/Async/TaskGraph.cpp:648]
libUnrealEditor-Core.so!FTaskGraphCompatibilityImplementation::ProcessThreadUntilRequestReturn(ENamedThreads::Type) [/mnt/horde/++UE5/Sync/Engine/Source/./Runtime/Core/Private/Async/TaskGraph.cpp:2149]
libUnrealEditor-RenderCore.so!RenderingThreadMain(FEvent*) [/mnt/horde/++UE5/Sync/Engine/Source/./Runtime/RenderCore/Private/RenderingThread.cpp:415]
libUnrealEditor-RenderCore.so!FRenderingThread::Run() [/mnt/horde/++UE5/Sync/Engine/Source/./Runtime/RenderCore/Private/RenderingThread.cpp:566]
libUnrealEditor-Core.so!FRunnableThreadPThread::Run() [/mnt/horde/++UE5/Sync/Engine/Source/./Runtime/Core/Private/HAL/PThreadRunnableThread.cpp:25]
libUnrealEditor-Core.so!FRunnableThreadPThread::_ThreadProc(void*) [/mnt/horde/++UE5/Sync/Engine/Source/Runtime/Core/Private/HAL/PThreadRunnableThread.h:185]
libc.so.6!UnknownFunction(0x94b42)
libc.so.6!UnknownFunction(0x1269ff)
```

nithinp7:
Thanks for looking into this @kring! I think I was finally able to reproduce this on Windows with Vulkan, probably the same underlying issue. I set the raster tile cache size to 0 and spammed the "Refresh Tileset" button and it seems to be a good way to catch that race condition, where the raster tile is freed before the render thread creation task is complete.

One option is we could create a new fence around the render thread texture creation and check that during `destroyTexture` - that way textures won't get destroyed before `FCesiumTextureResource::InitRHI()` finishes. Now in the ridiculous 0-cache size case this would cause stutters as freshly-loaded textures are immediately destroyed when the camera moves, causing the destruction to wait for the render thread (which might be 1-3 frames behind). This might also happen if way too many tiles are being loaded for a given view (e.g., 1 SSE), the cache will be filled with "required" tiles and also have constant turnover similar to the 0-cache size. In typical cases, I imagine a newly loaded texture should still be unlikely to be deleted immediately, but I still am tempted to find a non-blocking solution.

nithinp7:
@kring I can move this idea to a different issue once I work it out but:

The main motivation for the rambling below is that it would be nice to be able to do something like e.g., pass around `shared_ptr<Image>` that the renderer can use to keep an image alive during texture creation. I don't think fixing the current issue alone warrants implementing the below refactor idea, but the idea might have several other advantages long-term that make it worthwhile. Let me know what you think!

I was thinking recently about how both `ImageCesium` and `BufferCesium` are sort of odd implementation details in contrast to the rest of `CesiumGltf` which can be serialized / parsed almost entirely with auto-generated code from schema. For instance, the `CesiumGltf::Model` can own the image descriptions (`CesiumGltf::Image`), but it does not need to own the image content itself (e.g., the parsed image representation currently in `ImageCesium`). So maybe we could fully decouple those and instead provide utility functions like `readImage` that read-out to an externally defined image representation (maybe move the `ImageCesium` functionality to some `Cesium3DTilesSelection::Image`). 

The advantage of decoupling those is that we can provide a lot more implementation-specific features in the new image class. Some of these features would feel very awkward in `CesiumGltf`. These might include:
- Representing the renderer content corresponding to the image in the same image class. This partially decouples resource management concerns from the tile and rasterTile. 
- Allows the images to be separately maintained in some sort of global resource table. 
- Changing the ownership semantics of the image itself (e.g., [enables sharing of external image render content](https://github.com/CesiumGS/cesium-native/issues/497)). We might have a global resource table, individual tiles / raster tiles, and texture creation tasks that all have a `shared_ptr<Image>` to the same image resource.
- Decouple the lifecycle of individual images from the tiles or raster tiles they correspond to (this is what might be relevant here). It would awkward to have tile loading explicitly express main-thread-followup render-thread work in its state space. It would be better to decouple images so that tiles / rastertiles can be released whenever, but the underlying images simply won't be released until they have no further references.

kring:
Thanks @nithinp7!

PR #1011, opened 2022-12-06
Fix Android game package failure in UE 5.1

kring:
UE 5.1 seems to have [a bug](https://forums.unrealengine.com/t/5-1-missing-precompiled-manifest-error-when-building-code-plugin-for-android/691983) that prevents packaging plugins for Android. I worked around this by moving some files around in the  UE directory. That allows us to package the plugin for Android successfully, but unfortunately when we later try to package a _game_ using that plugin, we get errors. These errors are basically the opposite of the errors during plugin packaging: the plugin includes an `arm64` directory and the game packaging doesn't expect it. So this PR adds a step to hack up the plugin after it's packaged so that it will later work at game package time.


kring:
Seems to be working well on my Pixel 6 Pro now. There's sometimes this strange blue artifact over part of the screen, though. And sometimes it flashes partially or completely cyan.
Screenshot_20221206-151035
It goes away when I enable "Use Mobile Rendering" on the CesiumSunSky, so presumably the SunSky still isn't working quite right on Android in UE 5.1 (though it's looking pretty good other than that one artifact!).

Issue #527, opened 2021-06-30
SkyAtmosphere sometimes hides earth's surface


argallegos:
This was originally spotted by a user on the forum - https://community.cesium.com/t/resolved-issue-with-cesium-sun-sky-hiding-earth-surface/13743

The SkyAtmosphere of CesiumSunSky will occasionally hide earth's surface when viewing from space. 
![atmosphereIssue](https://user-images.githubusercontent.com/39537389/123982757-4c636600-d991-11eb-8059-238ca1a7614c.gif)

This even persists on the ground if you zoom into an affected area - the imagery layer fades to black.
![atmosphereIssueGroundShort](https://user-images.githubusercontent.com/39537389/123983810-268a9100-d992-11eb-8387-732725755fa3.gif)

From my testing, it looks like this depends on proximity to the georeference.

When the sun is not lighting the area where the georeference is, the terrain imagery will fade to black, even if it's light in the area where the camera currently is. We've noticed that lighting is funky the farther you get from the Georeference, but I don't think we've noticed this specific issue before. 
![image](https://user-images.githubusercontent.com/39537389/123985178-4078a380-d993-11eb-861f-8ac3d862992a.png)
(Looks pretty cool, though.)

Since we encourage users to use world origin rebasing if they are traveling far from the original georeference, I don't think that it's really a breaking issue for projects on the ground. However, it's definitely undesirable when viewing the earth from space. 

The forum user noted that changing the Transmittance Minimum Angle in the SkyAtmosphere component of CesiumSunSky from -90 to 90 degrees fixes the issue. The Sky Atmosphere documentation describes the Transmittance Minimum Angle as "The minimum elevation angle in degrees that should be used to evaluate the sun transmittance to the ground. Useful to maintain a visible sun light and show on meshes even when the sun has started going below the horizon. This does not affect the aerial perspective." 
Setting it to 90 essentially means that the sun transmittance is always evaluated. This may be desirable, but could also come with performance or graphical problems. It's also concerning that the parameter does seem to affect the aerial perspective, there may be a bug. 

We should do more testing to make sure there are no side effects, but if there are not and we have no better solution, we should change the Transmittance Minimum Angle of the SkyAtmosphere of CesiumSunSky to 90 in the plugin.

@xuelongmu is this still a problem in your branch with the recent changes to CesiumSunSky?

xuelongmu:
[...]

PR #1017, opened 2022-12-12
Unlit Materials

joseph-kaile:
normal1

normal2


cesium-concierge:

* :x: Missing CLA.

Source/CesiumRuntime/Private/CesiumGltfComponent.cpp
```diff
@@ -286,6 +287,21 @@ static void computeTangentSpace(TArray<FStaticMeshBuildVertex>& vertices) {
   genTangSpaceDefault(&MikkTContext);
 }
 
+static void computeFakeNormals(
+    const TArray<uint32_t>& indices,
+    TArray<FStaticMeshBuildVertex>& vertices) {
+  // Compute flat normals
```
kring, line 293: Incorrect comment.

Source/CesiumRuntime/Private/CesiumGltfComponent.cpp
```diff
@@ -286,6 +287,21 @@ static void computeTangentSpace(TArray<FStaticMeshBuildVertex>& vertices) {
...
+  for (int i = 0; i < indices.Num(); i += 3) {
+    FStaticMeshBuildVertex& v0 = vertices[i];
+    FStaticMeshBuildVertex& v1 = vertices[i + 1];
+    FStaticMeshBuildVertex& v2 = vertices[i + 2];
+
+    v0.TangentX = v1.TangentX = v2.TangentX = TMeshVector3(0.0f);
+    v0.TangentY = v1.TangentY = v2.TangentY = TMeshVector3(0.0f);
+    v0.TangentZ = v1.TangentZ = v2.TangentZ = TMeshVector3(0.0f, 0.0f, 1.0f);
```
kring, line 301: I think maybe using +Z as the normal isn't quite right. It's fine when the tileset is located near the georeference origin. But what if the origin is on one side of the globe, and the tileset is on the other? Then the tileset would only be lit when the sun is below the horizon, which is likely to surprise users.

So I think instead of +Z, the normal at each point should be set to the ellipsoid surface normal at that point, transformed to the mesh's coordinate system.

Basically you need to get each vertex position, transform it to ECEF using the `transform` passed into this method, and then call `Ellipsoid::WGS84.geodeticSurfaceNormal` to get the surface normal at that point. Then that surface normal has to be transformed from ECEF back to Unreal coordinates using the inverse of that passed-in `transform` (don't forget to construct the vector to transform as `glm::dvec4(ecefNormal, 0.0)` before multiplying it with the inverse of the transform, because it's a direction not a position).

CHANGES.md
```diff
@@ -1,5 +1,11 @@
 # Change Log
 
+### ? - ? 
+
+##### Fixes :wrench:
+
+- Disable normals for unlit materials, which caused photogrammetry to appear too dark.
```
kring, line 7: Please mention the extension name here.

nithinp7:
@joseph-kaile Sorry if I'm missing context, but a couple of comments:
- Smooth normals fix the weird triangl-ey lighting issues in photogrammetry and they seem like a more elegant solution than fake normals right? 
  - Technically that is against gltf spec when normals are missing - spec dictates clients should generate flat normals when normals are missing, but imo smooth normals look better than flat normals for photogrammetry. 
  - The only times flat normals _need_ be generated are when tangents are needed and missing e.g., to apply a normal map. When tangents are missing normal maps are relative to MikkTSpace tangents, which requires flat normals. All that being said though, normal maps are never used for unlit photogrammetry (currently).
  - So for example, we could potentially just generate smooth normals if the unlit extension is present, if this is the route to go down.
- But alternatively, why don't we just use an actually unlit material shading model instead of faking normals when the unlit extension is present? I don't necessarily love this option, as it would look even more unrealistic - but it seems to be how photogrammetry was _intended_ to be rendered. Plus that's what we do in Cesium for Unity and CesiumJS.

Source/CesiumRuntime/Private/CesiumGltfComponent.cpp
joseph-kaile: OK, I believe I implemented the above in code. I'm not sure how to visualize the normals. However the code is straightforward enough. If I invert the normals the buildings get dark. Also, I disabled the dynamic shadows:
result

Source/CesiumRuntime/Private/CesiumGltfComponent.cpp
```diff
@@ -1075,8 +1087,22 @@ static void loadPrimitive(
       }
     }
   } else {
-    TRACE_CPUPROFILER_EVENT_SCOPE(Cesium::ComputeFlatNormals)
-    computeFlatNormals(indices, StaticMeshBuildVertices);
+    if (material.hasExtension<ExtensionKhrMaterialsUnlit>()) {
+      glm::dvec3 ecefCenter = glm::dvec3(
+          transform * CesiumTransforms::unrealToOrFromCesium *
```
kring, line 1092: I don't think the `unrealToOrFromCesium` here and below are needed. I tested by using Buffer Visualization -> World Normal and comparing the color of Melbourne with Cesium World Terrain. Cesium World Terrain will have some undulations in its normals, while Melbourne will be a solid color, but the overall color of them should be the same because the normals should point in basically up in both cases. With the `unrealToOrFromCesium` in place, they were quite different, but once I removed it they were the same. This makes sense because there's no need to use the Unreal coordinate system at all. The Origin is expressed in model coordinates, then we transform to ECEF, compute the normal, and transform the ECEF normal back to model coordinates.

There seems to be something weird with normals in Cesium for Unreal _in general_, though. Everything is good near the georeference origin, but on the back side of the globe something is wonky. For example, in this screenshot, both CWT and Melbourne are dark even though the sun is high in the sky:
image
This doesn't necessarily need to be fixed as part of this PR, but maybe you have an idea what's wrong? I'm at a bit of a loss at the moment.

joseph-kaile: That makes sense, I assumed that there was code I couldn't find that automatically transformed the coordinates in the position view accessor.

kring: Could be precision related, but _in theory_ the directional light should just be a direction, so the large coordinate values shouldn't matter too much. But who knows. I'll write an issue.

kring: Oh! I think it's just this issue again:
https://github.com/CesiumGS/cesium-unreal/issues/527
And there's a workaround listed in there.

kring:
Thanks @joseph-kaile!

PR #1021, opened 2022-12-23
Declare UE modules as public dependencies, not private

kring:
This is necessary because, at least in some cases, we `#include` headers from these modules in our own public headers. If these modules aren't declared public, projects that try to include our headers will get compiler errors unless they explicitly depend on the necessary modules.

See:
https://community.cesium.com/t/unable-to-package-ue5-0-3-project-with-cesium-v1-19-0/21162


cesium-concierge:

PR #1031, opened 2023-01-30
Fix broken materials in Unreal Engine 5.1

kring:
UE 5.1 introduced breaking changes in how we obtain the names of material layers. I updated the code to compile, but didn't realize that the "Layer Name" I was now accessing was actually the name of the material layer _asset_, not the name of the material layer itself. So, for example, for this layer:

image

we wanted to get `Overlay0`, but in UE 5.1 we were instead getting `ML_CesiumRasterOverlay`.

Because various material parameters are assigned based on these names, Cesium materials pretty much completely stopped working in UE 5.1. This was most obvious for tilesets with raster overlays, which would just appear white.

You might wonder how this went unnoticed. Well, as long as the material was last saved in a _previous_ version of Unreal Engine, it worked fine even in UE 5.1. It was only if we created or saved the material in UE 5.1 that it would fail to work.

Exactly this symptom was reported on the forum here:
https://community.cesium.com/t/material-instance-error-too-many-texture-coordinate-sets-defined-on-gpuskin-vertex-input-max-4/21693

I initially thought it was a bug in UE 5.1, and reported it on UDN, and Epic in fact found and fixed a bug. But even after the fix, our materials were still not working. So today I dug deeper and noticed the problem with the layer names.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -11,6 +11,7 @@ jobs:
           fetch-depth: 0 # so that `git describe` works.
       - name: Set environment variables
         run: |
+          choco install -y visualstudio2019-workload-manageddesktop
           $ENV:CESIUM_UNREAL_VERSION=$(git describe)
           $ENV:BUILD_CESIUM_UNREAL_PACKAGE_NAME="CesiumForUnreal-50-windows-${ENV:CESIUM_UNREAL_VERSION}"
           # Make these available to subsequent steps
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -11,7 +11,7 @@ jobs:
           fetch-depth: 0 # so that `git describe` works.
       - name: Set environment variables
         run: |
-          choco install -y dotnetcore-runtime.install --version=3.1.1
+          choco install -y dotnetcore-sdk
           $ENV:CESIUM_UNREAL_VERSION=$(git describe)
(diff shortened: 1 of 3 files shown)


cesium-concierge:

* :grey_exclamation: There was an error checking the CLA! If this is your first contribution, please send in a [Contributor License Agreement](https://github.com/CesiumGS/cesium/blob/master/CONTRIBUTING.md#contributor-license-agreement-cla).
    * Maintainers, this was the error I ran into while attempting to process the CLA check. Please resolve it to continue CLA checking.
    * You'll need to [manually check for a submitted CLA](https://github.com/CesiumGS/cesium/blob/master/Documentation/Contributors/CLAs/README.md)
    ```
    Error: The service is currently unavailable.
    ```

joseph-kaile:
👍

kring:
Our CI builds were really broken, so this PR expanded to fix that, too. Here are the problems and solutions:

* GitHub Actions switched `macos-latest` to start referring to macOS 12 instead of 11, and that broke our UE 5.0 macOS and iOS builds. Solution: use `macos-11` instead of `macos-latest`. UE 5.1 works fine in macOS 12, fwiw.
* Microsoft deprecated .NET Core 3.1, which our UE 5.0 builds use. See [here](https://github.com/CesiumGS/cesium-unreal/issues/1033) for more details and solution.
* While I was making other changes to build images, I also configured them to use Powershell 7 (instead of 5). But that broke the Powershell script used in the UE 5.1 Android build to work around its broken paths. Solution: use absolute paths, because PS 7 resolves relative paths relative to some odd place.

PR #1037, opened 2023-02-10
Add support for rendering points

j9liu:
image

This PR adds support for rendering glTFs with point primitives in Cesium for Unreal. It's mostly done, but it needs tweaks and more thorough testing.

**TODO:**
- [x] Code clean-up and correctness (removing unnecessary functions, setting the correct boolean flags)
- [ ] ~Fix z-fighting bug causing flickering~
- [x] Test for Unreal 5.0 and 5.1

### DIFF
--- a/Source/CesiumRuntime/Private/CesiumGltfPointsComponent.h
+++ b/Source/CesiumRuntime/Private/CesiumGltfPointsComponent.h
@@ -1,0 +1,30 @@
+// Copyright 2020-2021 CesiumGS, Inc. and Contributors
+
+#pragma once
+
+#include "Cesium3DTilesSelection/BoundingVolume.h"
+#include "CesiumEncodedMetadataUtility.h"
+#include "CesiumGltf/MeshPrimitive.h"
+#include "CesiumGltf/Model.h"
+#include "CesiumGltfPrimitiveComponent.h"
+#include "CesiumMetadataPrimitive.h"
+#include "CesiumRasterOverlays.h"
+#include "Components/StaticMeshComponent.h"
+#include "CoreMinimal.h"
+#include <cstdint>
+#include <glm/mat4x4.hpp>
+#include <unordered_map>
+#include "CesiumGltfPointsComponent.generated.h"
+
+UCLASS()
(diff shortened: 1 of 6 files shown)


cesium-concierge:

j9liu:
Update: the flickering is actually being caused by temporal anti-aliasing. The jitter it uses in its algorithm results in artifacts when anti-aliasing thin surfaces, so I'm not sure if there's anything actionable on our end to address this. Users can edit the temporal anti-aliasing settings themselves via console commands, e.g. `r.TemporalAASamples` and `r.TemporalAACurrentFrameWeight`.

j9liu:
After the change I pushed I confirmed this PR works for UE 5.0 and 5.1. Though. whenever I tried to build for 5.1 it kept saying Cesium for Unreal was built for 5.0, so I may be doing something wrong.

Something weird I haven't been able to figure out is why `DrawStaticElements` doesn't actually _draw_ when it's called. I can get the points to show up via `GetDynamicElements`, which works just fine. But the mesh is supposed to be a static mesh, so it seems more proper to mark it for static rendering only. But I can't get it to draw the points this way and I don't know why. :shrug:

Source/CesiumRuntime/Private/CesiumGltfPointsComponent.h
```diff
@@ -0,0 +1,19 @@
+// Copyright 2020-2021 CesiumGS, Inc. and Contributors
```
j9liu, line 1: I just copied this from the `CesiumGltfPrimtiiveComponent` file -- should we update this for 2023?

j9liu:
@kring this is ready for review, whenever you get the chance!

Source/CesiumRuntime/Private/CesiumGltfPointsComponent.h
kring: Haha yeah we should update it everywhere. But that can be a separate PR. It's unfortunate that Epic's marketplace guidelines require us to put this nonsense in every file.
https://www.unrealengine.com/en-US/marketplace-guidelines#262b

ovigbinobaro:
How do you enable collisions on Point clouds? My characters keep falling through the point cloud in Unreal

Issue #327, opened 2021-04-09
Generate navigation mesh on Cesium3DTileset


xuelongmu:
Navigation meshes can be generated in Unreal for movable static meshes. However, navigation meshes currently cannot be generated on the Cesium World Terrain, which is a `Cesium3DTileset` actor. I have experimented with adding a `UNavLinkComponent` to `Cesium3DTileset` in the [navmesh-generation](../tree/navmesh-generation) branch with [Navigation Invokers](https://www.youtube.com/watch?v=QNHInJJ1vUk&list=WL&index=58) and dynamic navmesh generation turned on, but haven't gotten it working yet.

As this is not a priority feature for the upcoming update, I will revisit this issue in the future.

argallegos:
Forum requests for this feature: 
https://community.cesium.com/t/cant-use-nav-mesh-on-cesium-terrain/18465
https://community.cesium.com/t/adjusting-nav-mesh-bound-volume-according-to-the-terrain/17833/4
https://community.cesium.com/t/generating-navmeshes/12998

j9liu:
Closed by #1044.

PR #1044, opened 2023-02-23
Generate navigation mesh on Cesium3DTileset

Project-PLATEAU:
This PR accommodates #327. The latest version(v1.22.0) of Cesium for Unreal can not generate a NavMesh for a Cesium3DTileset
actor. However, this PR make it possible to generate a NavMesh on a Cesium3DTileset actor.

We have confirmed that NavMesh is generated after our modification using the "New York City 3D Buildings" asset provided by
Cesium Ion. The first image is before our modification and the second image is after our modification.

BeforeNavCollision
BeforeNavCollision

AfterNavCollision
AfterNavCollision

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,7 @@
 
 - The `FlyToAltitudeProfileCurve`, `FlyToProgressCurve`, `FlyToMaximumAltitudeCurve`, `FlyToDuration`, and `FlyToGranularityDegrees` properties of `GlobeAwareDefaultPawn`  / `DynamicPawn` may now be read and written from Blueprints.
 - Added an option on `Cesium3DTileset` to ignore the `KHR_materials_unlit` extension entirely and use normal lighting and shadows.
+- Added `CreateNavCollision` property to `Cesium3DTileset`. When enabled, `CreateNavCollision` is called on the static meshes created for tiles.
 
 ##### Fixes :wrench:
 
--- a/Source/CesiumRuntime/Private/Cesium3DTileset.cpp
+++ b/Source/CesiumRuntime/Private/Cesium3DTileset.cpp
@@ -319,6 +319,13 @@ void ACesium3DTileset::SetCreatePhysicsMeshes(bool bCreatePhysicsMeshes) {
   }
 }
 
+void ACesium3DTileset::SetCreateNavCollision(bool bCreateNavCollision) {
+  if (this->CreateNavCollision != bCreateNavCollision) {
+    this->CreateNavCollision = bCreateNavCollision;
+    this->DestroyTileset();
+  }
+}
+
 void ACesium3DTileset::SetAlwaysIncludeTangents(bool bAlwaysIncludeTangents) {
   if (this->AlwaysIncludeTangents != bAlwaysIncludeTangents) {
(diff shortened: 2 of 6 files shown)


cesium-concierge:
Thank you so much for the pull request @Project-PLATEAU! I noticed this is your first pull request and I wanted to say welcome to the Cesium community!

kring:
Thanks for the pull reuest @Project-PLATEAU! Have you looked at the performance impact of this change at all? Is it nearly zero as long as the user isn't using navigation? And if users are using navigation, are the rather large - and also dynamically added/removed! - 3D Tiles meshes very usable with navigation? Does it cause Unreal to do extra work when the tiles are added? If so, does it do that work in the game thread?

Project-PLATEAU:
If users do not use navigation there was no performance impact.
We assume that this fix is not for dynamically deleted meshes, but for 3DTiles meshes, such as buildings, where the character is moved by navigation after all models have been loaded. Therefore, this fix requires that "KeepWorldOriginNearCamera" be set to False. Without the above setting, the 3DTiles will move as the camera moves, which will cause differences in the coordinates of the generated NavMesh.
Navmesh generation is executed in the game thread. However, if Runtime Generation setting of Navmesh is set to Static, Navmesh generation is executed only once and does not affect performance during gameplay.

kring:
> We assume that this fix is not for dynamically deleted meshes, but for 3DTiles meshes, such as buildings, where the character is moved by navigation after all models have been loaded.

This seems like a fairly unlikely use-case to me. It's rare for a 3D Tiles tileset to be loaded completely and then never changed. A large reason to use 3D Tiles at all is to do hierarchical level-of-detail at runtime. Otherwise, you could just import the entire model into the Editor and be done with it, right?

I think we at least need an option to skip generation of nav meshes for 3D Tiles. My worry is that this a) isn't that useful in typical 3D Tiles use-cases, because it's too slow, and b) will break people that are using hand-crafted nav meshes (or nav meshes derived from non-Cesium models, anyway) in levels that incorporate 3D Tiles. Such levels will suddenly start trying to generate nav meshes for large 3D Tiles, thereby breaking a level that used to work fine.

Project-PLATEAU:
Your concerns make sense. However, it does useful when a 3DTileset of a small area is used for simulation, as in our case.
So, we have made a modification to make this feature optional.
Could you confirm the commit we added?

Our case could be found on the following page:
https://www.mlit.go.jp/plateau/use-case/uc22-030/

kring:
Thanks @Project-PLATEAU. Your latest changes are ok with me, and I'm happy to merge that. Quick question though: would it be better to have a per-tileset option to control whether the tileset supports nav meshes, rather than the experimental feature flag? This would be similar to the existing "CreatePhysicsMeshes` option, and allow users to, for example, enable navigation for a smaller local model while disabling it for the background Cesium World Terrain tileset.

kring:
Just a reminder about my question above, @Project-PLATEAU. What do you think?

Project-PLATEAU:
As you commented, we think it would be better to have this setting in each of the 3DTileset actors.
We have removed it from the experimental feature flag and added the flag to the navigation category of the 3DTileset actor.
Could you please confirm this?

kring:
Thanks @Project-PLATEAU!

PR #958, opened 2022-08-31
Minor changes to Cesium for Unreal to work with CesiumGS/cesium-native#554

kring:

PR #1047, opened 2023-02-23
Fix crash when changing project default token

### LINKED ISSUE #1042: Crash when switching tokens with an active tileset in level
**Overview**
Switching project tokens with an active tileset in the level crashes UE5

**Reproduction Steps:**
- Create at least two tokens in Cesium Ion with assets:read access
- Start a new UE project
- Log in to Cesium Ion
- Set one of the existing tokens as project default
- Add Cesium World Terrain + Bing Maps imagery to the level
- Tileset should load correctly
- Open Cesium Token UI and switch to a different existing token
- Click use as project default token
- Unreal will crash

**Versions used**
UE 5.1.1
Cesium for Unreal 1.21.0

joseph-kaile:
Fixes #1042. The problem was that a main thread task is calling Destroy Tileset. The tileset would destroy itself when calling dispatch main thread tasks.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,7 +2,13 @@
 
 ### v1.23.0 - 2023-03-01
 
+##### Additions :tada:
+
 - Added support for rendering point clouds (`pnts`).
+
+##### Fixes :wrench:
+
+- Fixed bug that causes a crash when changing the project default token.
 
 ### v1.22.0 - 2023-02-01
 
--- a/Source/CesiumEditor/Private/SelectCesiumIonToken.cpp
+++ b/Source/CesiumEditor/Private/SelectCesiumIonToken.cpp
@@ -392,11 +392,12 @@ void SelectCesiumIonToken::createRadioButton(
 }
 
 FReply SelectCesiumIonToken::UseOrCreate() {
-  Promise<std::optional<Token>> promise =
-      FCesiumEditorModule::ion()
-          .getAsyncSystem()
-          .createPromise<std::optional<Token>>();
-  SharedFuture<std::optional<Token>> future = promise.getFuture().share();
+  if (!this->_promise || !this->_future) {
+    return FReply::Handled();
+  }
+
+  Promise<std::optional<Token>> promise = std::move(*this->_promise);
+  this->_promise.reset();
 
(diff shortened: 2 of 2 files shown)


cesium-concierge:

* :x: Missing CLA.

kring:
Thanks @joseph-kaile. It took me a minute to realize why this works. The problem is that the `SelectCesiumIonToken` panel does some async work and `thenInMainThread` calls `RefreshTileset` on every tileset that uses the project default token. That main thread task can be - and often is - dispatched from _inside_ the Tileset's `updateView`, so we're effectively destroying an object while we're executing one of its methods.

Your fix avoids this by blocking the button press callback until after the async work completes, guaranteeing that the continuation cannot be invoked from `updateView` because we simply won't invoke `updateView` until after the async work completes. This is obviously far better than the current situation (a crash), but it does mean the UI is unresponsive while we're talking to Cesium ion.

When I originally wrote this panel, we didn't have this problem because the UI used a completely separate `AsyncSystem` from the tilesets. So the `SelectCesiumIonToken` continuation could never be invoked from inside `updateView` because they used completely separate queues. I changed this in #958 as part of an effort to avoid making games freeze completely while destroying tilesets and waiting for async requests to complete. I'm not quite sure if that was essential for some reason, or if it just seemed like a good idea at the time.

So I'm going to merge this because better is better, particularly for releasing tomorrow. But I think longer term there are three ways we can avoid the freeze when changing the token:

1. Make `RefreshTileset` smarter. If it detects its being called from inside an update, it defers the refresh.
2. Instead of directly calling `RefreshTileset` from within a `thenInMainThread`, dispatch a main thread task using Unreal's `AsyncTask(ENamedThreads::GameThread, ...)`. Unreal's tasks are guaranteed to be executed outside `updateView`, or any of our other code, so this will be safe and effective.
3. Change cesium-native so that it doesn't call `dispatchMainThreadTasks` from within `updateView`. Instead, the game engine code would be expected to call it itself at a suitable time. I think this is my favorite because it should eliminate the potential for a whole class of bugs as well as being slightly more performant (because we don't dispatch main thread tasks for every tileset every frame).

I'll write a separate issue for this.

PR #1050, opened 2023-03-01
Update for breaking changes in cesium-native

kring:
I also tried compiling the plugin without the normal unity builds enabled, and found a few missing headers, which I added. See #615 for how that works.

I'm going to self merge this as there is no change in functionality.


cesium-concierge:

PR #1059, opened 2023-03-14
adds the BlueprintReadWrite property specifier to GlobeAwareDefaultPawn FlyTo variables

gvindiol:
Fixes #1052 

https://community.cesium.com/t/transition-between-more-than-one-geo-marker/22612

PR #1060, opened 2023-03-14
Fix errors caused by blueprint-accessible doubles.

kring:
I was a little too quick to merge #1059. It doesn't work in UE4 because properties of type double cannot be accessible from Blueprints. This PR changes the problematic properties, `FlyToDuration` and `FlyToGranularityDegrees`, to floats instead of doubles. This change shouldn't make any difference from a precision perspective, because these properties will have relatively small numbers that don't need to be extraordinarily precise.


cesium-concierge:

PR #1081, opened 2023-04-30
Show credits in Editor

### LINKED ISSUE #1071: Display attribution/credits in the Editor
Currently, tileset and raster overlay credits are only shown when playing (i.e. in Play-in-Editor mode or in a built game). We should show the credits in Editor viewports as well. Most data providers arguably require this.

kring:
Fixes #1071

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,13 @@
 # Change Log
 
+### ? - ?
+
+Starting with this release, Cesium for Unreal requires Unreal Engine v5.0 or later.
+
+##### Additions :tada:
+
+- Tileset and raster overlay credits are now shown in Editor viewports.
+
 ### v1.24.0 - 2023-04-03
 
 This will be the _last_ release that supports Unreal Engine v4.27. Future versions will require Unreal Engine v5.0+.
--- a/Source/CesiumRuntime/Private/CesiumCreditSystem.cpp
+++ b/Source/CesiumRuntime/Private/CesiumCreditSystem.cpp
@@ -11,6 +11,13 @@
 #include <tidybuffio.h>
 #include <vector>
 
+#if WITH_EDITOR
+#include "Editor.h"
+#include "IAssetViewport.h"
+#include "LevelEditor.h"
+#include "Modules/ModuleManager.h"
+#endif
+
 /*static*/ UClass* ACesiumCreditSystem::CesiumCreditSystemBP = nullptr;
 namespace {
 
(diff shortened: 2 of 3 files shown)


cesium-concierge:

j9liu:
This works great on load, but unfortunately when you try to switch to another level, the Unreal editor crashes with this error:

```
[2023.05.01-00.01.27:558][184]LogLoad: Error: Old Package /Game/CesiumSamples/Maps/01/01_CesiumWorld not cleaned up by GC! Object Package /Game/CesiumSamples/Maps/01/01_CesiumWorld is being referenced by SKEL_01_CesiumWorld_C /Game/CesiumSamples/Maps/01/01_CesiumWorld.Default__SKEL_01_CesiumWorld_C:
 (root)  GCObjectReferencer /Engine/Transient.GCObjectReferencer_0
 -> SObjectWidget(ScreenCredits_C_0)::AddReferencedObjects( ScreenCredits_C /Game/CesiumSamples/Maps/01/01_CesiumWorld.01_CesiumWorld:ScreenCredits_C_0)
    ^ UnknownFunction []
```

I tried replicating the `RemoveOverlayWidget` functionality in a destructor `~ACesiumCreditSystem` and in `AActor::Destroyed`, but the error still exists. I'm not sure if I'll be able to find a fix before the release :(

kring:
Ha! I meant to tested that and got distracted, thanks for checking. I think I know how to fix it and can do so before my next flight. 🤞

kring:
Hmm I thought I could subscribe to `FWorldDelegates::LevelRemovedFromWorld` or `FEditorDelegates::NewCurrentLevel`, but neither of those are called when switching levels, at least not before the memory leak assertion is triggered. Very strange. I guess it will need to wait until next release, because it'll be pretty risky even if I get it working at this point.

kring:
This is closer, but still not quite right.

I added some code to remove the credits in the `LevelEditorModule.OnMapChanged` event, which fixes that memory leak assertion. Except sometimes it doesn't. Possibly related to doing a Play-in-Editor before switching levels.

Another problem is that the in-Editor credits sometimes flicker. Specifically, they're disappearing because Cesium for Unreal is not finding _any_ cameras to use that frame, and so the `updateView` isn't happening, and so there are no credits that frame. The reason there are no cameras is because all of the `FEditorViewportClient` instances are returning false for their `IsVisible` method, even though they're clearly visible. I haven't figured out why yet. I don't think it's a new problem, we just never noticed because without in-editor credits, there was no visible symptom of this.

kring:
The "sometimes there's still a leak" problem happened when an object was selected in the Outliner view. During map change, UE would first raise the `OnMapChanged` event, and we'd remove the credits. But then later in the process it would call `SelectNone`, which would trigger the `OnRedrawLevelEditingViewports` event, and we'd re-add the credits. Oops! Solution: switch to the `FEditorSupportDelegates::CleanseEditor` event instead of `LevelEditorModule.OnMapChanged`, which happens much later in the map unload process.

The "credits flickering" problem is caused by my potato GPU, mostly. UE has some questionable logic in `SEditorViewport::IsVisible` where it deems any viewport that hasn't ticked in 250ms to be non-visible. So the Melbourne tileset, in a debug build, on my potato GPU, would sometimes take more than 250ms to render, causing the viewport to be deemed invisible. Cameras associated with invisible editor viewports aren't used, so we didn't have any cameras at all, so we wouldn't run tile selection, so we wouldn't have any credits. Solution: don't have a potato GPU. I wrote an issue (https://github.com/CesiumGS/cesium-native/issues/631) to fix this better, but I think we can consider it low priority.

kring:
This should be ready for another look.

j9liu:
@kring this looks great to me! I found one more bug BUT I don't think it should hold up this PR. The main point is to see the on-screen attributions, and I also feel bad that I didn't find it earlier when testing :cry:

Basically, clicking on "Data Attribution" blurs the viewport, but I don't see any text. 
image

Other than that, this looks good and I'm happy to merge if you agree we should deal with this bug later.

j9liu:
Another thing that can be a follow-up PR: toggling "Show Credits On Screen" on the tileset isn't immediately reflected by the UI. (This was the same in Unity before https://github.com/CesiumGS/cesium-unity/pull/269)

kring:
> Basically, clicking on "Data Attribution" blurs the viewport, but I don't see any text.

This is fixed.

The problem was that the credits popup was too big for the size of the window, so the actual credits were off screen. When playing, the resize behavior is a bit different so it doesn't have that problem even with a small window. I tried to understand the difference in order to fix the in-Editor behavior, but as far as I can tell it's just black magic. 😆

So, I changed the way the popup is sized so that it always occupies 80% of the screen in each direction (and utilized less black magic to do it), and that seems to work fine in both in game and in Editor. It might make for some larger margins than before on small-screen mobile devices, though. I'll test it on my phone once CI pops out a build to make sure it's ok, but I think it will be.

kring:
Looks fine on my Android phone:
Screenshot_20230505-153734

kring:
> Another thing that can be a follow-up PR: toggling "Show Credits On Screen" on the tileset isn't immediately reflected by the UI.

This is fixed. We now simply refresh the tileset when this property changes. It'd be nice to avoid that, but cesium-native unfortunately makes that hard at the moment (the showOnScreen flag is baked into the credits when they're created).

PR #1083, opened 2023-05-01
Remove compilation of PhysX

j9liu:
We're getting warnings in  `\Plugins\CesiumForUnreal\Source\CesiumRuntime\CesiumRuntime.Build.cs(179,13)` that say `'ReadOnlyTargetRules.bCompilePhysX' is obsolete: 'Deprecated in UE5.1 - No longer used as Chaos is always enabled.'` This PR removes that check from the file.

@kring can you check if this is correct?


cesium-concierge:

kring:
I think we need to avoid `Target.bUseChaos` as well, because it is also obsolete. I'll edit this PR accordingly and then merge it once CI passes.

PR #1085, opened 2023-05-02
Add support for point cloud attenuation

j9liu:
This PR adds support for point cloud attenuation to Cesium for Unreal. I can't guarantee the results will look the same between Unity and Unreal because of how differently they handle DPI, but hopefully users will be content adjusting the point cloud shading parameters.

This is Melbourne Point Cloud with SSE 6, default attenuation settings:
| Without Attenuation | With attenuation |
| ----------------------- | ------------------ |
| HighresScreenshot00001 | HighresScreenshot00002 |

And this is Chappes Church with SSE 12, default attenuation settings:

| Without Attenuation | With attenuation |
| ----------------------- | ------------------ |
| HighresScreenshot00005 | HighresScreenshot00004 |

Additionally, because attenuation is implemented via vertex factory, it is compatible with Unreal materials. Here's Melbourne with the `MI_CesiumGridLines` material and attenuation enabled:

HighresScreenshot00003

And here's the church with `MI_CesiumElevationBands` and attenuation:

HighresScreenshot00006

**TODO:**
- Check for compatibility / correctness across all platforms
  - [x] Windows
  - [ ] Android
  - [ ] Linux
  - [x] MacOS
  - [ ] iOS


kring:
@j9liu you will (not) be thrilled to know that this isn't working well on my terrible AMD GPU. It looks fine with attentuation off:
image

But breaks when I turn attentuation on (this is the same camera viewpoint!):

image

I guess you will have no way to debug it, though, so I'll investigate.

kring:
If I switch to using DX12 (which previously fixed the weird behavior on my AMD system in Unity, if you remember), then Unreal Engine crashes when it tries to display a point cloud. Here's the call stack:

```
LoginId:a2fb1be84369824480badbb96c91c8fd
EpicAccountId:e3488bbe74d140fe830edc0291f524e5

Assertion failed: VertexDeclarationRHI [File:D:\build\++UE5\Sync\Engine\Source\Runtime\RHI\Public\RHIResources.h] [Line: 2658]

UnrealEditor_D3D12RHI!FD3D12GraphicsPipelineState::FD3D12GraphicsPipelineState() [D:\build\++UE5\Sync\Engine\Source\Runtime\D3D12RHI\Private\D3D12Pipelinestate.cpp:322]
UnrealEditor_D3D12RHI!FD3D12PipelineStateCacheBase::CreateAndAdd() [D:\build\++UE5\Sync\Engine\Source\Runtime\D3D12RHI\Private\D3D12Pipelinestate.cpp:646]
UnrealEditor_D3D12RHI!FD3D12DynamicRHI::RHICreateGraphicsPipelineState() [D:\build\++UE5\Sync\Engine\Source\Runtime\D3D12RHI\Private\D3D12State.cpp:503]
UnrealEditor_RHI!FCompilePipelineStateTask::DoTask() [D:\build\++UE5\Sync\Engine\Source\Runtime\RHI\Private\PipelineStateCache.cpp:957]
UnrealEditor_RHI!TGraphTask<FCompilePipelineStateTask>::ExecuteTask() [D:\build\++UE5\Sync\Engine\Source\Runtime\Core\Public\Async\TaskGraphInterfaces.h:975]
UnrealEditor_Core!<lambda_b0effac95abe128e68ff2c315eb1e3d1>::operator()() [D:\build\++UE5\Sync\Engine\Source\Runtime\Core\Public\Async\Fundamental\Task.h:350]
UnrealEditor_Core!LowLevelTasks::TTaskDelegate<void __cdecl(bool),48>::TTaskDelegateImpl<<lambda_b0effac95abe128e68ff2c315eb1e3d1>,0>::CallAndMove() [D:\build\++UE5\Sync\Engine\Source\Runtime\Core\Public\Async\Fundamental\TaskDelegate.h:171]
UnrealEditor_Core!LowLevelTasks::FTask::ExecuteTask() [D:\build\++UE5\Sync\Engine\Source\Runtime\Core\Public\Async\Fundamental\Task.h:426]
UnrealEditor_Core!LowLevelTasks::FScheduler::TryExecuteTaskFrom<&LowLevelTasks::TLocalQueueRegistry<1024>::TLocalQueue::DequeueGlobal,0>() [D:\build\++UE5\Sync\Engine\Source\Runtime\Core\Private\Async\Fundamental\Scheduler.cpp:280]
UnrealEditor_Core!LowLevelTasks::FScheduler::WorkerMain() [D:\build\++UE5\Sync\Engine\Source\Runtime\Core\Private\Async\Fundamental\Scheduler.cpp:320]
UnrealEditor_Core!UE::Core::Private::Function::TFunctionRefCaller<<lambda_776551d1e343b6cb7eaf0e5521c77c81>,void __cdecl(void)>::Call() [D:\build\++UE5\Sync\Engine\Source\Runtime\Core\Public\Templates\Function.h:549]
UnrealEditor_Core!FThreadImpl::Run() [D:\build\++UE5\Sync\Engine\Source\Runtime\Core\Private\HAL\Thread.cpp:69]
UnrealEditor_Core!FRunnableThreadWin::Run() [D:\build\++UE5\Sync\Engine\Source\Runtime\Core\Private\Windows\WindowsRunnableThread.cpp:146]
```

kring:
If I switch to Vulkan, UE crashes at startup while compiling shaders:

```
[2023.05.05-06.54.29:084][  0]LogShaderCompilers: Warning: 2 Shader compiler errors compiling WorldGridMaterial for platform SF_VULKAN_SM5:
Shader debug info dumped to: "C:/cesium/cesium-unreal-samples/Saved/ShaderDebugInfo/SF_VULKAN_SM5/WorldGridMaterial/FCesiumPointAttenuationVertexFactory/TShadowDepthVSVertexShadowDepth_VirtualShadowMapPositionOnly/0"
C:/cesium/cesium-unreal-samples/Plugins/cesium-unreal/Shaders/Private/CesiumPointAttenuationVertexFactory.ush(10,26): Shader TShadowDepthVSVertexShadowDepth_VirtualShadowMapPositionOnly, Permutation 0, VF FCesiumPointAttenuationVertexFactory:
	/Plugin/CesiumForUnreal/Private/CesiumPointAttenuationVertexFactory.ush(10,26): error: cannot instantiate StructuredBuffer with given packed alignment; 'VK_EXT_scalar_block_layout' not supported
	StructuredBuffer<float3> PositionBuffer;
	                         ^
C:/Program Files/Epic Games/UE_5.0/Engine/Shaders/Private/VelocityShader.usf(): Shader FVelocityPS, Permutation 0, VF FCesiumPointAttenuationVertexFactory:
	Internal Error!
	WorldGridMaterial/VelocityPipeline/FCesiumPointAttenuationVertexFactory/FVelocityPS/0:/Engine/Private/VelocityShader.usf VF 'FCesiumPointAttenuationVertexFactory' Type 'FVelocityPS' '/Engine/Private/VelocityShader.usf' Entry 'MainPixelShader' Permutation 0 
Fatal error: [File:D:\build\++UE5\Sync\Engine\Source\Runtime\Engine\Private\ShaderCompiler\ShaderCompiler.cpp] [Line: 4409] 
	/Plugin/CesiumForUnreal/Private/CesiumPointAttenuationVertexFactory.ush:10:26: error: cannot instantiate StructuredBuffer with given packed alignment; 'VK_EXT_scalar_block_layout' not supported
	StructuredBuffer<float3> PositionBuffer;
	                         ^
Failed to compile default material /Engine/EngineMaterials/DefaultLightFunctionMaterial.DefaultLightFunctionMaterial!
A breakpoint instruction (__debugbreak() statement or a similar call) was executed in UnrealEditor-Win64-DebugGame.exe.
```

kring:
So in D3D11, my best guess is that this is the same problem that we saw on this GPU in D3D11 in Unity. The positions are coming from a buffer that was created as a vertex buffer, and D3D11 technically doesn't allow a single buffer to be both a VertexBuffer and a StructuredBuffer, even though NVidia GPUs seem to be ok with it. I also noticed in RenderDoc that the `StructureByteStride` of the `FPositionVertexBuffer` is 0, which won't go well when trying to use it as a structured buffer. Probably the same problem, really, because Unreal only sets that property for structured buffers (see `D3D11Buffer.cpp`, `FD3D11DynamicRHI::RHICreateBuffer`).

So... we probably need to copy the vertex buffer to a proper structured buffer, just like we did in Unity. Except it's much less obvious how to do that here. 😬

kring:
I worked out how to copy the vertices to a new StructuredBuffer, and it works. The point cloud renders correctly with my AMD GPU, now. Super hacky PR into this branch coming up.

j9liu:
> So in D3D11, my best guess is that this is the same problem that we saw on this GPU in D3D11 in Unity.

🤦 How did I forget about this! Thanks for coming up with a workaround.

j9liu:
Unfortunately even with the AMD GPU fix, I was still getting errors with DX12. @joseph-kaile also said he was getting errors with Metal. I wondered if I was doing something wrong with the implicit vertex rendering, so I decided to try a different approach for rendering the vertices.

Similar to the Lidar Point Cloud plugin, I create a vertex buffer that duplicates the original vertices 4 times. Then, I explicitly declare a vertex stream, so that the position is received as an attribute in the vertex factory shader. This seems to eliminate the crashes in DX12 and Vulkan on my end. @kring let me know if it still works for you on your AMD GPU.

This _does_ mean that a point cloud with N points will have N * 5 points in memory. I hope the overhead isn't too much of an issue.

j9liu:
@kring based on your suggestions, I tried the dummy vertex buffer that the AR Point Cloud Plugin was using, and it seems to work across all graphics APIs for me. I switched the position buffer from a `StructuredBuffer` to a regular `Buffer` -- let me know if I broke it for AMD GPUs.

joseph-kaile:
> @kring based on your suggestions, I tried the dummy vertex buffer that the AR Point Cloud Plugin was using, and it seems to work across all graphics APIs for me. I switched the position buffer from a `StructuredBuffer` to a regular `Buffer` -- let me know if I broke it for AMD GPUs.

Can confirm it works well in Mac.

kring:
It works on my AMD GPU too! 🎉

kring (commented):
Looks great @j9liu, very clean and easy to follow! I just had a couple of small concerns before merging this.

kring:
I tried this on my Android phone. It works fine when attenuation is off, but when attenuation is on the game crashes the moment the point cloud is rendered. It appears to be a segfault in the RHIThread caused by dereferencing a null pointer. I'm trying to find more details.

kring:
Getting the call stack is a pain. Here's how to do it (for my future reference as much as anything):

1. Run `adb logcat` and find the segfault in all the spew. Copy the crash report, which starts with something like `*** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***`, into a text file, `crashreport.txt`.
2. Find the debug symbol symbols for the game. In my case, I had asked Unreal to build the game to `C:\cesium\cfus-ship-android`, and the symbols were in `C:\cesium\cfus-ship-android\Android_Multi\CesiumForUnrealSamples_Symbols_v1`. But I don't know if Unreal created these, or if Android Studio did, because I was trying to use Android Studio to debug this beforehand. If you need to try debugging with Android Studio, use these instructions: https://docs.unrealengine.com/5.1/en-US/debugging-unreal-engine-projects-for-android-using-android-studio/
3. Run the `ndk-stack` tool that comes with the Android NDK. On my system, that looked like this: ` python C:\Users\kevin\AppData\Local\Android\Sdk\ndk\25.1.8937393\prebuilt\windows-x86_64\bin\ndk-stack.py -sym C:\cesium\cfus-ship-android\Android_Multi\CesiumForUnrealSamples_Symbols_v1 -dump .\crashreport.txt`

```
********** Crash dump: **********
Build fingerprint: 'google/raven/raven:13/TQ2A.230405.003.E1/9802792:user/release-keys'
#00 0x0000000000783c3c /vendor/lib64/egl/libGLES_mali.so (vulkan::descriptor_set::write_descriptor_set(VkWriteDescriptorSet const&)+812) (BuildId: bec4c595930d3102)
#01 0x00000000007830ec /vendor/lib64/egl/libGLES_mali.so (vkUpdateDescriptorSets+124) (BuildId: bec4c595930d3102)
#02 0x0000000003610f34 /data/app/~~n10YGyII3gVexcW1PX5Zpw==/com.YourCompany.CesiumForUnrealSamples-ddHGJ7HmZXn_AGND0mzoHA==/lib/arm64/libUnreal.so (BuildId: 2da05b5707b9b3ed1304430631d9fa9fe6aae285)

                            FVulkanDescriptorSetCache::FCachedPool::CreateDescriptorSets(FVulkanDSetsKey const&, FVulkanDescriptorSetsLayout const&, TArray<FVulkanDescriptorSetWriter, TSizedDefaultAllocator<32> >&, VkDescriptorSet_T**)

                            Module.VulkanRHI.2_of_4.cpp:0:0
#03 0x0000000003610ca4 /data/app/~~n10YGyII3gVexcW1PX5Zpw==/com.YourCompany.CesiumForUnrealSamples-ddHGJ7HmZXn_AGND0mzoHA==/lib/arm64/libUnreal.so (BuildId: 2da05b5707b9b3ed1304430631d9fa9fe6aae285)

                            FVulkanDescriptorSetCache::GetDescriptorSets(FVulkanDSetsKey const&, FVulkanDescriptorSetsLayout const&, TArray<FVulkanDescriptorSetWriter, TSizedDefaultAllocator<32> >&, VkDescriptorSet_T**)

                            Module.VulkanRHI.2_of_4.cpp:0:0
#04 0x000000000361c350 /data/app/~~n10YGyII3gVexcW1PX5Zpw==/com.YourCompany.CesiumForUnrealSamples-ddHGJ7HmZXn_AGND0mzoHA==/lib/arm64/libUnreal.so (BuildId: 2da05b5707b9b3ed1304430631d9fa9fe6aae285)

                            bool FVulkanGraphicsPipelineDescriptorState::InternalUpdateDescriptorSets<true>(FVulkanCommandListContext*, FVulkanCmdBuffer*)

                            Module.VulkanRHI.2_of_4.cpp:0:0
#05 0x000000000360f730 /data/app/~~n10YGyII3gVexcW1PX5Zpw==/com.YourCompany.CesiumForUnrealSamples-ddHGJ7HmZXn_AGND0mzoHA==/lib/arm64/libUnreal.so (BuildId: 2da05b5707b9b3ed1304430631d9fa9fe6aae285)

                            FVulkanPendingGfxState::PrepareForDraw(FVulkanCmdBuffer*)

                            Module.VulkanRHI.2_of_4.cpp:0:0
#06 0x00000000035f6c6c /data/app/~~n10YGyII3gVexcW1PX5Zpw==/com.YourCompany.CesiumForUnrealSamples-ddHGJ7HmZXn_AGND0mzoHA==/lib/arm64/libUnreal.so (BuildId: 2da05b5707b9b3ed1304430631d9fa9fe6aae285)

                            FVulkanCommandListContext::RHIDrawIndexedPrimitive(FRHIBuffer*, int, unsigned int, unsigned int, unsigned int, unsigned int, unsigned int)

                            Module.VulkanRHI.1_of_4.cpp:0:0
#07 0x0000000003e0c3ac /data/app/~~n10YGyII3gVexcW1PX5Zpw==/com.YourCompany.CesiumForUnrealSamples-ddHGJ7HmZXn_AGND0mzoHA==/lib/arm64/libUnreal.so (BuildId: 2da05b5707b9b3ed1304430631d9fa9fe6aae285)

                            UE::Core::Private::Function::TFunctionRefCaller<FRHICommandListImmediate::ExecuteAndReset()::$_34, void ()>::Call(void*)

                            Module.RHI.cpp:0:0
#08 0x00000000023ec22c /data/app/~~n10YGyII3gVexcW1PX5Zpw==/com.YourCompany.CesiumForUnrealSamples-ddHGJ7HmZXn_AGND0mzoHA==/lib/arm64/libUnreal.so (BuildId: 2da05b5707b9b3ed1304430631d9fa9fe6aae285)

                            TGraphTask<TFunctionGraphTaskImpl<void (), (ESubsequentsMode::Type)0> >::ExecuteTask(TArray<FBaseGraphTask*, TSizedDefaultAllocator<32> >&, ENamedThreads::Type, bool)

                            Module.Launch.cpp:0:0
#09 0x0000000003672e14 /data/app/~~n10YGyII3gVexcW1PX5Zpw==/com.YourCompany.CesiumForUnrealSamples-ddHGJ7HmZXn_AGND0mzoHA==/lib/arm64/libUnreal.so (BuildId: 2da05b5707b9b3ed1304430631d9fa9fe6aae285)

                            FNamedTaskThread::ProcessTasksNamedThread(int, bool)

                            Module.Core.1_of_20.cpp:0:0
#10 0x00000000036726e8 /data/app/~~n10YGyII3gVexcW1PX5Zpw==/com.YourCompany.CesiumForUnrealSamples-ddHGJ7HmZXn_AGND0mzoHA==/lib/arm64/libUnreal.so (BuildId: 2da05b5707b9b3ed1304430631d9fa9fe6aae285)

                            FNamedTaskThread::ProcessTasksUntilQuit(int)

                            Module.Core.1_of_20.cpp:0:0
#11 0x0000000003e846a4 /data/app/~~n10YGyII3gVexcW1PX5Zpw==/com.YourCompany.CesiumForUnrealSamples-ddHGJ7HmZXn_AGND0mzoHA==/lib/arm64/libUnreal.so (BuildId: 2da05b5707b9b3ed1304430631d9fa9fe6aae285)

                            FRHIThread::Run()

                            Module.RenderCore.2_of_4.cpp:0:0
#12 0x00000000036c6298 /data/app/~~n10YGyII3gVexcW1PX5Zpw==/com.YourCompany.CesiumForUnrealSamples-ddHGJ7HmZXn_AGND0mzoHA==/lib/arm64/libUnreal.so (BuildId: 2da05b5707b9b3ed1304430631d9fa9fe6aae285)

                            FRunnableThreadPThread::Run()

                            Module.Core.6_of_20.cpp:0:0
#13 0x0000000003670eac /data/app/~~n10YGyII3gVexcW1PX5Zpw==/com.YourCompany.CesiumForUnrealSamples-ddHGJ7HmZXn_AGND0mzoHA==/lib/arm64/libUnreal.so (BuildId: 2da05b5707b9b3ed1304430631d9fa9fe6aae285)

                            FRunnableThreadPThread::_ThreadProc(void*)

                            Module.Core.1_of_20.cpp:0:0
#14 0x00000000000c226c /apex/com.android.runtime/lib64/bionic/libc.so (__pthread_start(void*)+204) (BuildId: dc4001c2ef2dfc23467040797a96840c)
#15 0x0000000000054a30 /apex/com.android.runtime/lib64/bionic/libc.so (__start_thread+64) (BuildId: dc4001c2ef2dfc23467040797a96840c)
```

kring:
The Editor also pops up with a bunch of errors at startup in this branch, all related to the in-editor credits:

```
CreateExport: Failed to load Outer for resource 'WidgetTree': WidgetBlueprint /CesiumForUnreal/ScreenCredits.ScreenCredits
```

I'm guessing the immediate cause is moving the plugin load earlier in the init process. If we need to keep the early init, I can work on fixing the credits.

CesiumForUnreal.uplugin
```diff
@@ -20,7 +20,7 @@
 		{
 			"Name": "CesiumRuntime",
 			"Type": "Runtime",
-			"LoadingPhase": "Default",
+			"LoadingPhase": "PostConfigInit",
```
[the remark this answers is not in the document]
j9liu: Unfortunately, this is mandatory in order to compile the vertex factory. If we don't put it in this loading phase, the attenuation shader will never compile :shrug:

Source/CesiumRuntime/Private/CesiumGltfPointsComponent.cpp
```diff
@@ -39,11 +53,22 @@ class FGltfPointsSceneProxy final : public FPrimitiveSceneProxy {
       FMeshElementCollector& Collector) const override {
     QUICK_SCOPE_CYCLE_COUNTER(STAT_GltfPointsSceneProxy_GetDynamicMeshElements);
 
+    if (!pTilesetActor) {
+      return;
+    }
+
+    const bool hasAttenuation =
+        pTilesetActor->GetPointCloudShading().Attenuation;
```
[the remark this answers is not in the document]
j9liu: Actually I took a closer look at how `ShouldRecreateProxyOnUpdateTransform` is handled. In `FScene` (`RendererScene.cpp`),  there's this code:
```
	// Check if the primitive needs to recreate its proxy for the transform update.
		if (Primitive->ShouldRecreateProxyOnUpdateTransform())
		{
			// Re-add the primitive from scratch to recreate the primitive's proxy.
			RemovePrimitive(Primitive);
			AddPrimitive(Primitive);
		}
```

where `Primitive` is a `UPrimitiveComponent`. So it does seem possible to force a recreation of the proxy, but I'm confused about _where_ to force it. If a tileset's settings change, how is it going to prompt all of the proxies to update?

j9liu:
@kring I just pushed some updates that handle the game thread / render thread data transfer more safely. It's based off the way Lidar Point Cloud does it, but it required a lot of code reorganization. Let me know if you have any questions.

kring:
@j9liu I noticed that if I press "Refresh Tileset" on a point cloud, UE crashes inside FCesiumGltfPointsSceneProxyUpdater dereferencing a null this pointer:

```
>	UnrealEditor-CesiumRuntime-Win64-DebugGame.dll!FCesiumGltfPointsSceneProxyUpdater::RegisterProxy(UCesiumGltfPointsComponent * Component, TWeakPtr<FCesiumGltfPointsSceneProxyWrapper,1> SceneProxyWrapper) Line 53	C++
 	UnrealEditor-CesiumRuntime-Win64-DebugGame.dll!UCesiumGltfPointsComponent::CreateSceneProxy() Line 19	C++
 	UnrealEditor-Renderer.dll!FScene::AddPrimitive(UPrimitiveComponent * Primitive) Line 1324	C++
 	UnrealEditor-Engine.dll!UPrimitiveComponent::CreateRenderState_Concurrent(FRegisterComponentContext * Context) Line 609	C++
 	UnrealEditor-Engine.dll!UStaticMeshComponent::CreateRenderState_Concurrent(FRegisterComponentContext * Context) Line 739	C++
 	UnrealEditor-Engine.dll!UActorComponent::ExecuteRegisterEvents(FRegisterComponentContext * Context) Line 1636	C++
 	UnrealEditor-Engine.dll!UActorComponent::RegisterComponentWithWorld(UWorld * InWorld, FRegisterComponentContext * Context) Line 1283	C++
 	UnrealEditor-CesiumRuntime-Win64-DebugGame.dll!loadPrimitiveGameThreadPart(const CesiumGltf::Model & model, UCesiumGltfComponent * pGltf, LoadGltfResult::LoadPrimitiveResult & loadResult, const glm::mat<4,4,double,0> & cesiumToUnrealTransform, const Cesium3DTilesSelection::Tile & tile, bool createNavCollision, ACesium3DTileset * pTilesetActor) Line 2117	C++
```

But bigger picture, I appreciate that you were following the lead of the lidar point cloud plugin, and that makes sense, but I think this code is a lot more complicated than it needs to be.

For one thing, the `DataLock`. It's only used in one place as far as I can tell, and that one method is only called from the game thread. So why is a lock needed at all?

If I'm understanding correctly, the important operation here is `TransferSettingsToProxies`. That method's goal is to hand over some new settings to each proxy in a thread-safe manner. But I don't think the registration of proxies is necessary to achieve that. TransferSettingsToProxies is called from the main thread, so you can get all the attached UCesiumGltfPointsComponent instances, and get the scene proxy from each. Put them in an array and transfer it over to the render thread for update.

The only concern, I think, is if the render proxy is deleted after this method returns but before the enqueued command is executed on the render thread. I'm not sure if that can actually happen. `UStaticMeshComponent`, for instance, seems to have lots of code that calls `ENQUEUE_RENDER_COMMAND` to invoke a method on the render proxy without doing anything special to ensure it's valid. So there may be a blank system in place in UE to ensure the render proxies don't go away too soon. Perhaps the `DetachFence` in `UPrimitiveComponent`?

kring:
A _little_ progress debugging the crash on Android...

First, I noticed that in the docs for newer versions of UE, it suggests debugging on Android using a Visual Studio extension rather than with Android Studio. So I set that up by following [these instructions](https://docs.unrealengine.com/5.1/en-US/debugging-unreal-engine-projects-for-android-in-visual-studio-with-the-agde-plugin/), and it seems a lot better. Or at least I know how to use it better, which helps.

So with that I was able to step through the `FCesiumGltfPointsSceneProxy` and see what's happening. And it does crash very soon after the first call to `CreateMeshWithAttenuation`. Stepping through there, I noticed that `OriginalVertexFactory.GetPositionsSRV()` is returning `nullptr`, which seems bad. I haven't tracked down why yet, but maybe it will mean something to you?

If not, I'll keep digging tomorrow.

image

j9liu:
> So with that I was able to step through the `FCesiumGltfPointsSceneProxy` and see what's happening. And it does crash very soon after the first call to `CreateMeshWithAttenuation`. Stepping through there, I noticed that `OriginalVertexFactory.GetPositionsSRV()` is returning `nullptr`, which seems bad. I haven't tracked down why yet, but maybe it will mean something to you?

I followed where `PositionComponentSRV` was created and I found this code in `PositionVertexBuffer`:

```
// we have decide to create the SRV based on GMaxRHIShaderPlatform because this is created once and shared between feature levels for editor preview.
bool bSRV = RHISupportsManualVertexFetch(GMaxRHIShaderPlatform) || IsGPUSkinCacheAvailable(GMaxRHIShaderPlatform);

// When bAllowCPUAccess is true, the meshes is likely going to be used for Niagara to spawn particles on mesh surface.
// And it can be the case for CPU *and* GPU access: no differenciation today. That is why we create a SRV in this case.
// This also avoid setting lots of states on all the members of all the different buffers used by meshes. Follow up: https://jira.it.epicgames.net/browse/UE-69376.
bSRV |= (VertexData && VertexData->GetAllowCPUAccess());

if(bSRV)
{
	// When VertexData is null, this buffer hasn't been streamed in yet. We still need to create a FRHIShaderResourceView which will be
	// cached in a vertex factory uniform buffer later. The nullptr tells the RHI that the SRV doesn't view on anything yet.
	PositionComponentSRV = RHICreateShaderResourceView(FShaderResourceViewInitializer(VertexData ? VertexBufferRHI : nullptr, PF_R32_FLOAT));
}
```

So both of the functions for `bSRV` were returning false. I went to the definition of `RHISupportsManualVertexFetch` and this is what it says:
```
/** Whether Manual Vertex Fetch is supported for the specified shader platform.
      Shader Platform must not use the mobile renderer, and for Metal, the shader language must be at least 2. */
```

which.... 😬 I guess I should have looked more carefully at compatibility before I began implementation.

I don't understand why the `ColorComponentsSRV` isn't null there, since the `ColorVertexBuffer` also checks for `RHISupportsManualVertexFetch(GMaxRHIShaderPlatform)`. In any case, though, it seems like Unreal doesn't support this for mobile platforms. But the fallback would be to quadruple the mesh data and use one of the texcoord attributes to denote the corners, right? Is that a reasonable / performant solution on mobile?

kring:
Despite lots of time looking at this, I know almost nothing new. Except one things: `RHISupportsManualVertexFetch` is definitely returning false. So I guess that explains the problem in some sense, because we're definitely trying to use that, right?

I spent a lot of time trying to get debug symbols on Android for UE itself, but got nowhere. And without them, trying to see what's happening inside engine code is just an exercise in frustration. I _might_ be able to get debug symbols by building the engine myself, but that's likely to be a pretty big time investment, so I haven't done it yet.

So taking a step back... we can potentially ship point cloud attenuation without it working on Android. But it needs to not crash. Falling back to non-attenuated would be ok. Printing a warning when doing that would be even better.

It would also be interested to see if a Quest 2 has the same problem as my Pixel 6 Pro. I think it almost certainly will, and that may increase the priority of fixing it. And a Magic Leap?

kring:
I also tried the "Point Cloud support for AR" plugin. But anytime that one is enabled, my Editor crashes at startup while compiling shaders, with these messages:

```
[2023.05.16-09.16.37:016][  0]LogShaderCompilers: Warning: 1 Shader compiler errors compiling WorldGridMaterial for platform PCD3D_SM5:
Shader debug info dumped to: "C:/cesium/cesium-unreal-samples/Saved/ShaderDebugInfo/PCD3D_SM5/WorldGridMaterial/Default/FPointCloudVertexFactory/TShadowDepthVSVertexShadowDepth_OnePassPointLight/0"
C:/Program Files/Epic Games/UE_5.1/Engine/Shaders/Private/ShadowDepthVertexShader.usf(208,40-81): Shader TShadowDepthVSVertexShadowDepth_OnePassPointLight, Permutation 0, VF FPointCloudVertexFactory:
	/Engine/Private/ShadowDepthVertexShader.usf(208,40-81):  error X3004: undeclared identifier 'VertexFactoryGetViewIndex'
Fatal error: [File:D:\build\++UE5\Sync\Engine\Source\Runtime\Engine\Private\ShaderCompiler\ShaderCompiler.cpp] [Line: 4668] 
  /Engine/Private/ShadowDepthVertexShader.usf(208,40-81):  error X3004: undeclared identifier 'VertexFactoryGetViewIndex'
Failed to compile default material /Engine/EngineMaterials/WorldGridMaterial.WorldGridMaterial!
```


It sounds very similar to the "Won't Fix" bug here: https://unreal-engine-issues.herokuapp.com/issue/UE-84895

So no help there, either.

j9liu:
@kring thanks for spending so much time digging. It would definitely be unfortunate if attenuation didn't work in VR, but I think for now it might be good to merge support for attenuation on PC platforms. I can write up an issue to address mobile support separately, summarizing the roadblocks that prevent our current approach from working.

kring:
I can confirm it doesn't crash on Android anymore!

Source/CesiumRuntime/Private/CesiumGltfPointsSceneProxyUpdater.cpp
```diff
@@ -0,0 +1,86 @@
...
+}
+
+void FCesiumGltfPointsSceneProxyUpdater::UpdateSettingsInProxiesIfNeeded() {
+  if (bNeedsUpdate) {
+    UpdateSettingsInProxies();
+  }
+}
+
+void FCesiumGltfPointsSceneProxyUpdater::TransferSettingsToProxies() {
+  TInlineComponentArray<UActorComponent*> ComponentArray(Owner, true);
```
kring, line 50: This constructor calls `Owner->GetComponents()` itself, which is redundant with the call below.

Source/CesiumRuntime/Private/CesiumGltfPointsSceneProxyUpdater.cpp
```diff
@@ -0,0 +1,86 @@
...
+  Owner->GetComponents(
+      UCesiumGltfPointsComponent::StaticClass(),
+      ComponentArray,
+      true);
```
kring, line 54: Are you sure we need components from child actors too? Even in the weird case where one Cesium3DTileset had another one as its parent, I don't think we'd want to find the child point components when operating on the parent.

kring: Also you should use the templatized version of `GetComponents` so that you don't ned to cast to `UCesiumGltfPointsComponent*` later.

Source/CesiumRuntime/Private/CesiumGltfPointsSceneProxyUpdater.cpp
```diff
@@ -0,0 +1,86 @@
...
+  }
+
+  // Update render Data
+  ENQUEUE_RENDER_COMMAND(TransferCesium3DTilesetSettingsToPointsProxies)
+  ([SceneProxies,
+    ProxyTilesetData](FRHICommandListImmediate& RHICmdList) mutable {
+    // Iterate over proxies and, if valid, update their data
+    for (int32 i = 0; i < SceneProxies.Num(); i++) {
+      FCesiumGltfPointsSceneProxy* SceneProxy = SceneProxies[i];
+      if (SceneProxy) {
```
kring, line 81: I don't think this check can ever fail, because you checked it wasn't nullptr before adding it to the list in the first place. Since `SceneProxy` is a pointer type, this `if` can't test anything other than the pointer value. Is there some more thorough validity check that should be done?

Source/CesiumRuntime/Private/LoadGltfResult.h
```diff
@@ -2,6 +2,7 @@
 
 #pragma once
 
+#include "Cesium3DTilesSelection/Tile.h"
```
kring, line 5: Is this include needed? I don't see Tile being used anywhere.

Source/CesiumRuntime/Public/CesiumGltfPointsSceneProxyUpdater.h
```diff
@@ -0,0 +1,53 @@
...
+  void UpdateSettingsInProxiesIfNeeded();
+
+  /** Marks this as needing an update next frame. This is called from
+   * CreateSceneProxy because the proxy won't be assigned to the primitive
+   * component until after that function. */
+  void MarkForUpdateNextFrame();
+
+private:
+  ACesium3DTileset* Owner;
+  bool bNeedsUpdate;
```
kring, line 50: I think you could get rid of this `bNeedsUpdate` field just by updating a new scene proxy immediately after it's created. Scene proxy are created in the game thread, so there shouldn't be any drama.

And then once that field is gone, this class doesn't need to hold any state at all anymore, which means it could just be static methods that take a ACesium3DTileset pointer and a tileset wouldn't need to hold an instance of it.

Source/CesiumRuntime/Public/CesiumGltfPointsSceneProxyUpdater.h
```diff
@@ -0,0 +1,53 @@
+#pragma once
```
kring, line 1: This should probably be in the `Private` directory rather than `Public`, right?

kring (commented):
A few more suggestions, but this is getting really close now.

So I don't forget: we also need to figure out all the credit-related errors at startup. I'll take a look.

kring:
Looks great! I think the credits problem is the last thing remaining. I will take a look today.

kring:
Ok I think I've fixed the credits problem.  The problem was that the constructor for `CesiumCreditSystemBPLoader` was attempting to resolve a Blueprints class. And that constructor was called really early for the CDO. I switched to a different pattern based on `TSoftClassPtr` and it seems to be working now. If I'm perfectly honest, I don't really know what I'm doing here. But I did confirm that credits still work, at least, in both the Editor and in packaged games.

j9liu:
@kring I can confirm that your changes fixed the errors! Thank you :pray:

kring:
Thanks @j9liu!

PR #1087, opened 2023-05-05
Fix for failed to destroy cartographic polygon at runtime

YuchebHu:
Missing calling `Super::BeginPlay()` in `ACartographicPolygon::BeginPlay()` causes we cannot destroy an `ACartographicPolygon` actor at runtime. Because before destroy the actor, we will check the actor `AActor::ActorHasBegunPlay` state. If the value is `EActorBeginPlayState::HasBegunPlay`, the actor will be mark pending killing. But `AActor::ActorHasBegunPlay` will be set to `EActorBeginPlayState::HasBegunPlay` in `AActor::BeginPlay()`.


YuchebHu:
I have signed the CLA.

kring:
Good catch @YuchebHu, and thanks for the PR!

Issue #1049, opened 2023-02-26
Add a way to translate/rotate/scale the entire globe


kring:
The CesiumGeoreference allows users to specify the globe longitude/latitude/height that is at the origin of the Unreal world. But there's no ability from there to translate, rotate, or scale the globe within the world. It's always at the origin, aligned with East-South-Up axes, and scaled by 100 to convert meters to centimeters.

The closest you can come is to set the translation/rotation/scale of Cesium3DTilesets. That works to a point, but it falls apart as soon as you use CesiumGlobeAnchors. The globe anchors anchor to the untransformed globe, not the transformed one we've painstakingly configured on each tileset.

Being able to transform the globe is useful for a variety of applications, from placing the globe in an augmented reality space to scaling down the globe to a smaller size to simulate a planet other than Earth.

We've solved this problem in Cesium for Unity by making the CesiumGeoreference a regular GameObject with a Transform, and so its transform applies to the entire globe. Something like that could make sense in Unreal, too. Or we could add explicit fields to the CesiumGeoreference to allow it to be placed in the world.

Related to #295

PR #1088, opened 2023-05-08
Add Scale property to CesiumGeoreference.

### LINKED ISSUE #295: Add parameter to adjust tileset's UE to ECEF scale
Scaling down tilesets with Unreal's Scale Transform tool can result in precision issues due to the size of the globe. For situations where a 1:1 model of the world is not desired, it may be a good idea to add a parameter that modifies scale further.

This could be a tileset parameter, or a georeference parameter that impacts all tilesets. The most graceful solution may be to create a global scale modifier on the georeference, and an option on individual tilesets to override the global scale modifier.

Additional consideration: Will this scale also automatically impact the maximum screen space error calculations? If not, we should definitely make sure it does.

kring:
Unreal equivalent of CesiumGS/cesium-unity#303. Unlike in Unity, here the scale is a percentage (so 100 is full size). I did that because Unreal has a weird arbitrary limit on how small a number you enter on the details panel (after which it rounds to 0.0). So using a percentage lets us scale down by two more orders of magnitude.

Some problems I know about:

* The CesiumSunSky adjusts the SkyAtmosphere's radius to account for the globe scale. But other SkyAtmosphere parameters, like "Atmosphere Height", need to be adjusted manually for small scales or else the atmosphere will look totally wonky. And this parameter appears to have an arbitrary minimum value of 0.1. So you can't get a globe with a correct atmosphere at a smaller scale than 0.1% (0.001) or so.
* The Water Mask effect fades in based on distance, so at small scales it appears too soon (i.e. while still zoomed out pretty far) and looks weird. There are shader parameters controlling this, and tweaking them will probably work, but users will need to create/copy a material instance to do this.

Fixes #295 
Partially addresses #1049

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,7 @@
 
 ##### Additions :tada:
 
+- Added a `Scale` property to `CesiumGeoreference`. This allows the entire globe to be scaled up or down within the Unreal world.
 - Tileset and raster overlay credits are now shown in Editor viewports.
 
 ##### Fixes :wrench:
--- a/Source/CesiumRuntime/Private/CesiumGeoreference.cpp
+++ b/Source/CesiumRuntime/Private/CesiumGeoreference.cpp
@@ -572,6 +572,12 @@ void ACesiumGeoreference::PostEditChangeProperty(FPropertyChangedEvent& event) {
     this->UpdateGeoreference();
     return;
   } else if (
+      propertyName == GET_MEMBER_NAME_CHECKED(ACesiumGeoreference, Scale)) {
+    if (this->Scale < 1e-6) {
+      this->Scale = 1e-6;
+    }
+    this->UpdateGeoreference();
+  } else if (
       propertyName ==
       GET_MEMBER_NAME_CHECKED(ACesiumGeoreference, CesiumSubLevels)) {
   }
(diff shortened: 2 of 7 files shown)


j9liu (commented):
@kring this just needs a CHANGES.md update, everything else looks fine to me.

Were you planning to address the bullet points in the PR description? Or should we merge this PR and write separate issues for those problems?

kring:
> Were you planning to address the bullet points in the PR description? Or should we merge this PR and write separate issues for those problems?

No, I'll write separate issues for them.

JamieCCH:
Hi there, this issue is still happening in Android Vulkan mode (I am making a native build to Quest 3). When the terrain is scaled lower than 0.001, it shows black but no problem when previewing on PC. Is there any workaround I can do for the Android build? I appreciate any advice. Thank you!

kring:
I'm afraid I don't know @JamieCCH. Unreal has quite a few scale assumptions built into it, and some of them cause problems for globe rendering, particularly at scale extremes. This has been improving every release since the switch to large world coordinates in UE 5.0. You should try our latest v2.13.1 release if you haven't already, because there's a small chance this was fixed by #1599. And using the latest UE version you can may help as well. Beyond that, it could be worth looking at things like near and far planes, because I believe the Android renderer is quite a bit different from the one used on desktop platforms, and may be more sensitive to view parameters. As a last resort, you could consider scaling the rest of your level _up_ instead of scaling the globe _down_.

PR #1092, opened 2023-05-12
Make the credit system's link to the widget transient.

kring:
This avoids an error when attempting to save a sub-level, and that sub-level has a credit system whose widget exists in the persistent level.

As reported here:
https://community.cesium.com/t/latest-update-broke-level-streaming/23967


cesium-concierge:

PR #1095, opened 2023-05-15
Emit warning if camera is too far from the origin

joseph-kaile:


cesium-concierge:

* :x: Missing CLA.

Source/CesiumRuntime/Private/Cesium3DTileset.cpp
```diff
@@ -873,8 +873,26 @@ void ACesium3DTileset::LoadTileset() {
     return;
   }
 
-#if ENGINE_MAJOR_VERSION == 5 && ENGINE_MINOR_VERSION == 0
   AWorldSettings* pWorldSettings = pWorld->GetWorldSettings();
+  if (pWorldSettings && pWorldSettings->bEnableWorldBoundsChecks) {
+    std::vector<FCesiumCamera> cameras = this->GetCameras();
+    if (!cameras.empty()) {
+      for (const FCesiumCamera& camera : cameras) {
+        if (camera.Location.GetAbsMax() > 1048574.0) {
```
kring, line 881: I think this distance check should be removed. Because with it in place, the user will only be warned if the tileset happens to be reloaded while the camera is already far from the origin. Always show the message, and change the text to recommend simply turning off the world bounds check (it's almost never a good idea in Cesium worlds).

kring:
Thanks @joseph-kaile!

PR #1074, opened 2023-04-14
Fix ue5.2 errors

NielsHygum:
Fixed only errors for compiling for UE5.2 but at least it works now. Still some warnings of use of deprecated features that should be fixed.

PR #1096, opened 2023-05-17
Unreal Engine 5.2

### LINKED ISSUE #1073: Compilation fails with Unreal Engine 5.2
The Cesium plugin uses features that was deprecated in UE 5.0 and 5.1. Some of these features are now finally removed in UE 5.2. Also I think they changed some header include order in UE, so some header include statements also needs to be made. I am working on a PR.

kring:
This PR pulls together a bunch of things:

1. The work @NielsHygum started in #1074.
2. @joseph-kaile's ongoing UE 5.2 work in the [fix-ue5.2](https://github.com/CesiumGS/cesium-unreal/tree/fix-ue5.2) branch.
3. My own work to set up CI for UE 5.2.

In addition to supporting UE 5.2 in the way we supported 5.1 and 5.0 before it, this PR brings actual Linux support to Cesium for Unreal for the first time, including in the Editor.

Fixes #1073


cesium-concierge:

kring:
I confirmed that the Linux version of the plugin works on a real Linux system, but in the Editor and in a built game. (well, it didn't at first, but now it does)

j-baker:
Just a flag on the build - on ARM Macs, Unreal by default tries to build for ARM. Right now, CI is only seemingly producing x64 binaries - linking fails complaining that there are no aarch64 symbols in the file. Not sure if this is a problem with Cesium in general (I'm moving from Unreal 4.27 where it's all x86 and Rosetta2'd, found this branch to help unblock me!) or specifically with the new integration, but wanted to flag it to you.

Specifically, the problem is just with the cesium-native artifacts in Source/Thirdparty/lib, the rest works once those are replaced with a hand-built version.

kring:
Thanks for the note @j-baker. I think we'll need to build cesium-native twice, once for each architecture, and then `lipo` the libs together.

.github/workflows/build.yml
```diff
@@ -502,3 +502,269 @@ jobs:
...
+          echo "BUILD_CESIUM_UNREAL_PACKAGE_NAME=${BUILD_CESIUM_UNREAL_PACKAGE_NAME}" >> "$GITHUB_ENV"
+      - name: Install nasm
+        uses: ilammy/setup-nasm@v1.4.0
+      - name: Build cesium-native
+        run: |
+          export UNREAL_ENGINE_DIR="/opt/ue52"
+          export LINUX_MULTIARCH_ROOT="/opt/ue52/Engine/Extras/ThirdPartyNotUE/SDKs/HostLinux/Linux_x64/v21_clang-15.0.1-centos7/"
+          export UNREAL_ENGINE_COMPILER_DIR="${LINUX_MULTIARCH_ROOT}x86_64-unknown-linux-gnu"
+          export UNREAL_ENGINE_LIBCXX_DIR="${UNREAL_ENGINE_DIR}/Engine/Source/ThirdParty/Unix/LibCxx"
+          # TODO: Move this to the image instead
```
j9liu, line 594: Does this TODO need to be addressed before this PR is merged?

.github/workflows/build.yml
```diff
@@ -502,3 +502,269 @@ jobs:
...
+          cmake --build build --config Release --target install -j8
+          cd ..
+          rm -rf extern
+      - name: Build plugin
+        run: |
+          sed -i 's/\"EngineVersion\": \"5.0.0\"/\"EngineVersion\": \"5.2.0\"/g' CesiumForUnreal.uplugin
+          export LINUX_MULTIARCH_ROOT="/opt/ue52/Engine/Extras/ThirdPartyNotUE/SDKs/HostLinux/Linux_x64/v21_clang-15.0.1-centos7/"
+          cd /opt/ue52/Engine/Build/BatchFiles
+          # UnrealBuildTool writes to write to ~/.config and loses its mind if it doesn't exist. So create it.
+          # TODO: move this to the image instead
```
j9liu, line 607: this one too?

Source/CesiumRuntime/Private/CesiumGltfPointsSceneProxy.cpp
```diff
@@ -0,0 +1,222 @@
+#include "CesiumGltfPointsSceneProxy.h"
+#include "CesiumGltfPointsComponent.h"
+#include "Engine/StaticMesh.h"
+#include "RHIResources.h"
+#include "Runtime/Launch/Resources/Version.h"
+#include "SceneInterface.h"
+#include "StaticMeshResources.h"
+
+#if ENGINE_MAJOR_VERSION > 5 || ENGINE_MINOR_VERSION > 1
```
j9liu, line 9: This check seems wrong?

```suggestion
#if ENGINE_MAJOR_VERSION == 5 && ENGINE_MINOR_VERSION > 1
```

Source/CesiumRuntime/Private/CesiumPointAttenuationVertexFactory.cpp
```diff
@@ -6,6 +6,12 @@
 #include "MeshDrawShaderBindings.h"
 #include "MeshMaterialShader.h"
 #include "RenderCommandFence.h"
+#include "Runtime/Launch/Resources/Version.h"
+
+#if ENGINE_MAJOR_VERSION > 5 || ENGINE_MINOR_VERSION > 1
```
j9liu, line 11: Here too:

```suggestion
#if ENGINE_MAJOR_VERSION == 5 && ENGINE_MINOR_VERSION > 1
```

j9liu (commented):
@kring just a few comments from me. So far I confirmed that the plugin works with UE 5.0, but I would like to briefly test it on 5.1 / 5.2 before merging to main.

j9liu:
@kring actually tried this out in UE 5.2 and it works! Don't see any major differences in the sample levels (aside from a material bug I noted in https://github.com/CesiumGS/cesium-unreal-samples/issues/43.)

.github/workflows/build.yml
kring: No, not urgent. It would just make the build a tiny bit faster, at the cost of lots of time building a new image. I'll do it next time I have some other reason to change the image.

Source/CesiumRuntime/Private/CesiumGltfPointsSceneProxy.cpp
kring: It's correct. The idea is to use this code in any release greater than 5.2, including a hypothetical future 6.x.

Technically we should check that the version isn't less than 5, but none of this code would compile then anyway so it doesn't much matter.

j9liu: Ah okay, that makes sense.

j9liu:
Going to merge now. Thanks @kring !

PR #1101, opened 2023-05-19
Fix Credits Crash when creating new scene

### LINKED ISSUE #1100: Crash when creating new scene with CWT + Bing Imagery
Steps to reproduce:

1. Create a blank Level.
2. Quick add a CWT + Bing Imagery.
3. Project will crash in CesiumCreditSystem::createCredit

joseph-kaile:
Fixes #1100 

It looks like the credits worked before because a Credits System was already in the Level and when a new Level was created, the code was not able to load the Blueprint class, perhaps because the class is not static. Loading it as an object instead of a class seems to fix the issue.

### DIFF
--- a/Source/CesiumRuntime/Private/CesiumCreditSystem.cpp
+++ b/Source/CesiumRuntime/Private/CesiumCreditSystem.cpp
@@ -20,7 +20,7 @@
 #include "Modules/ModuleManager.h"
 #endif
 
-/*static*/ UClass* ACesiumCreditSystem::CesiumCreditSystemBP = nullptr;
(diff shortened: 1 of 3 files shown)


cesium-concierge:

* :x: Missing CLA.

kring:
Good catch, @joseph-kaile! I almost certainly broke this in https://github.com/CesiumGS/cesium-unreal/pull/1085/commits/2ab3ae1b5b9d565396b2351ae5d15a868284c63a.

This change looks good. I just changed the `dynamic_cast<>` to a UE `Cast<>` because I think dynamic_cast isn't always guaranteed to be available on all UE platforms.

PR #1102, opened 2023-05-22
Use correct default for glTF occlusion texture.

kring:
The CesiumGlTFFunction material function has an incorrect default for the `occlusionTexture`. As a result, a glTF model without an explicit occlusion texture (which is most of them) will have an `Ambient Occlusion` value of 0.0, rather than 1.0 as expected. Under certain lighting setups, this can make shadows cast onto the model extremely dark (basically pure black).

This PR fixes it by changing the default texture to one that has the value 1.0 in the red channel. Specifically, from this (`NoMetallicRoughnessTexture`):
image

To this (`NoColorTexture`):
image

(Thanks @AlbanBERGERET-Epic for reporting this!)


cesium-concierge:

j9liu:
@kring did you save the material with Unreal Engine 5.1? I'm viewing this in UE 5.0 and I'm getting this message:

image

It's preventing anything in the scene from rendering with the material. Let me see if I can recreate the change in Unreal 5...

kring:
Oh yeah I probably did @j9liu. Sorry about that, and thanks for fixing it!

j9liu:
Thanks @kring !

PR #1105, opened 2023-05-23
Fix CesiumGlTFFunction for UE 5.0

j9liu:
The fix in #1102 was saved with UE 5.1, so the asset could not be opened in UE 5. This replicates the change in UE5.


kring:
Thanks @j9liu!

PR #1106, opened 2023-05-23
Remove degenerate triangles from collision mesh

joseph-kaile:
Using the same condition found in TriangleMeshImplicitObject.cpp, this PR prevents degenerate triangles from being added to the collision mesh. Doesn't seem to have a performance hit.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,10 @@
 - `ACesium3DTileset` now emits a warning if the "Enable World Bounds Checks" option is enabled. That option can make the camera fly toward the origin unexpectedly.
 - Added new settings to the Cesium section of the Project Settings, allowing users to control how many requests to handle before pruning and also how many elements to keep in the cache after pruning.
 
+##### Fixes :wrench:
+
+- Remove degenerate triangles from collision mesh.
+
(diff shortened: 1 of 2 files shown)


j9liu:
@joseph-kaile I had to change `UE_SMALL_NUMBER` to an actual number because it wasn't compiling, but this seems to work well! I can't trigger the debug asserts in `03_CesiumDenver` anymore, and the framerate does not noticeably hitch when I leave the cube.

kring:
@j9liu just to confirm: are you still able to reproduce the hitches in main, but not in this branch?

j9liu:
Yep. I see this lovely view in debug mode:

image

and in release it freezes up, then prints all these degenerate triangle warnings to the console:
image

So this PR is a major improvement.

kring:
Thanks @j9liu! BTW, that lovely view will become much more useful if you install the (enormous) debug symbols for UE. Once you get your new laptop. :)

kring:
@joseph-kaile can you take a look at the build failures on macOS / Linux / Android? Once those are fixed, I think we're ready to merge this.

j9liu:
Thanks @joseph-kaile !

PR #1111, opened 2023-05-31
Fix World Partion crash by removing RF_Public

### LINKED ISSUE #941: World Partition resulting in editor crash with Cesium
Observed here: https://community.cesium.com/t/ue5-crash-when-editing-deleting-saving-cartographic-polygon/19824

User reported that saving a scene while both World Partition and the Cesium for Unreal plugin resulted in a crash. While Cesium for Unreal is not compatible with World Partition, this should probably not be resulting in a crash.

The user first noticed this issue when attempting to save a scene after adding a Cartographic Polygon. They provided this log:
`Fatal error: [File:D:\build++UE5\Sync\Engine\Source\Runtime\CoreUObject\Private\Templates\Casts.cpp] [Line: 10] Cast of CesiumCartographicPolygon /Game/Seattle_Project/Levels/Cesium-BingMapAerial.Cesium-BingMapAerial:PersistentLevel.CesiumCartographicPolygon_UAID_3CECEF77A2E62E2901_1946255554 to MetaData failed`

I am unable to reproduce this on my machine. However, the user has experienced it on multiple machines, in multiple projects, including fresh projects. All the user's machines are using RTX3090 GPUs.

kring:
While playing around with World Partition, I ran into https://github.com/CesiumGS/cesium-unreal/issues/941, where the Editor crashes when attempting to unload a level with World Partition and a Cesium Actor. The crash is caused by Unreal attempting to CastChecked our Actor to UMetaData. Our Actor is not a UMetaData, so this checked cast fails, which is treated as a fatal error.

This cast happens inside [FWorldPartitionActorDesc::Unload](https://github.com/EpicGames/UnrealEngine/blob/5.2.0-release/Engine/Source/Runtime/Engine/Private/WorldPartition/WorldPartitionActorDesc.cpp#L605). Basically if the Actor `IsPackageExternal` (docs helpfully say "returns true if object has a different package than its outer's package"), and also it has the `RF_Public` or `RF_Standalone` flags, then it will be CastChecked to `UMetaData`. Why? I really don't know. A large comment explains some cases where the Actor might _not_ be "package external" (ours is), but those don't seem to apply to us. There's no hint in a comment or otherwise why this cast of some random Actor to `UMetaData` could ever work, nevermind be a good idea. Perhaps you're wondering what `UMetaData` even is? The docs only say "An object that holds a map of key/value pairs", which doesn't really clear it up for me.

But fortunately, I happened to notice that this crash _only_ happens if the Actor is added via the Cesium Quick Add panel. If we instead add a Cesium3DTileset by dragging it in from the Place Actors panel, there's no problem. So I went spelunking through the UE source code, and learned that the Editor UI, when you drag in an Actor, creates it with _only_ the `RF_Transactional` flag, whereas our Quick Add panel also adds `RF_Public`.

So, I removed `RF_Public`. And the problem goes away.

I wish I could tell you what this flag actually does, and why it's there in the first place, but no. I think I added it originally because some of the very early DataSmith glTF import code I was looking at used it, maybe. But that code was mostly importing models as assets, so it was probably not the best example to follow. Or maybe I just thought "public" sounded good? 🤷 The docs say RF_Public means "Object is visible outside its package", which doesn't mean very much to me.

Long story short, removing `RF_Public` makes us consistent with the Place Actors panel, it fixes this bug, and it doesn't have any obvious negative side effects that I've noticed yet. So :shipit:?

Fixes #941

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -14,6 +14,7 @@
 
 - Fixed a bug introduced in v1.26.0 that caused an error when attempting to save a sub-level containing Cesium objects.
 - Fixed a bug in `CesiumGlTFFunction` that caused glTF and 3D Tiles "Ambient Occlusion" value to be 0.0 (instead of the expected 1.0) when the model does not specify an explicit occlusion texture. This could cause some extremely dark shadows.
+- Fixed a bug that could cause a crash when using Cesium Actors with World Partitioning.
 
 ### v1.26.0 - 2023-05-09
 
--- a/Source/CesiumEditor/Private/CesiumEditor.cpp
+++ b/Source/CesiumEditor/Private/CesiumEditor.cpp
@@ -474,7 +474,7 @@ FCesiumEditorModule::CreateTileset(const std::string& name, int64_t assetID) {
       ACesium3DTileset::StaticClass(),
       FTransform(),
       false,
-      RF_Public | RF_Transactional);
+      RF_Transactional);
   ACesium3DTileset* pTilesetActor = Cast<ACesium3DTileset>(pNewActor);
   pTilesetActor->SetActorLabel(UTF8_TO_TCHAR(name.c_str()));
   if (assetID != -1) {
@@ -530,7 +530,7 @@ UCesiumIonRasterOverlay* FCesiumEditorModule::AddOverlay(
   UCesiumIonRasterOverlay* pOverlay = NewObject<UCesiumIonRasterOverlay>(
       pTilesetActor,
       FName(name.c_str()),
-      RF_Public | RF_Transactional);
+      RF_Transactional);
   pOverlay->MaterialLayerKey = overlayKey;
   pOverlay->IonAssetID = assetID;
   pOverlay->SetActive(true);
--- a/Source/CesiumEditor/Private/IonQuickAddPanel.cpp
+++ b/Source/CesiumEditor/Private/IonQuickAddPanel.cpp
@@ -263,7 +263,7 @@ void IonQuickAddPanel::AddCartographicPolygonToLevel() {
       ACesiumCartographicPolygon::StaticClass(),
       FTransform(),
       false,
-      RF_Public | RF_Transactional);
+      RF_Transactional);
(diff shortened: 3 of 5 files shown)


cesium-concierge:

kring:
I wrote a UDN ticket about this, FWIW:
https://udn.unrealengine.com/s/question/0D54z00009FRp5jCAD/crash-with-world-partition-and-an-actor-with-rfpublic-flag

j9liu:
Thanks @kring, I confirmed this fixes the crash! I'll wait for CI to pass before merging.

I noticed another bug while trying to delete the world partition level I used for testing. I created an Open World level and only added CWT. When you switch to another level, then try to delete the open world one in the Editor, it shows this window:

image

Which unfortunately seems to involve the credit system. But then it also spams a ton of other windows that are unrelated to the plugin:

image

And I tried to delete a regular Open World level in the editor, and it still gave me trouble. So it might be a fluke, and just something wrong with Unreal Engine, but I thought I should mention it just in case.

kring:
An engine developer at Epic kindly responded to my UDN ticket. Of the crash, he said:

> We assume that the only object in an actor package should be the meta data, and if that's not the cvase it's an error (that's why I added the check). What you could do is to clear the flag in AActor::PostLoad and fix the code on your side that set the flag and you should be good to go.

We're not currently doing the PostLoad bit, which I understand is needed to clear the RF_Public flag on Actors that were created in previous versions. It's not immediately a problem because existing levels won't have World Partition enabled, so they won't run into this crash. But perhaps it will be necessary in order for the _conversion_ of a level to world partition to succeed. And maybe it's just a good idea in any case. I'll write an issue for it.

Of `RF_Public`, he said:

> RF_Public means that the object is an asset, so it should be only set for worlds, textures, meshes, etc.
> This flags essentially says it's ok to have a reference to public objects from other packages (with the exception of worlds).
> Actors are not assets since they are part of a level which is part of a world, etc. that's why they should not key the public flag.

PR #1116, opened 2023-06-05
Parse feature IDs from `EXT_mesh_features` instead of `EXT_feature_metadata`

j9liu:
This is a PR into a staging branch where we'll make a larger overhaul of Cesium for Unreal's metadata implementation.

In 3D Tiles 1.1, feature IDs are indicated via `EXT_mesh_features` instead of `EXT_feature_metadata`. They're allowed to exist separately from `EXT_structural_metadata`, so this PR attempts to detangle the Unreal implementations similarly. For the sake of incremental change, this still references the old metadata implementation, and a lot of code has been commented out to make sure the new code compiles.

This involves a lot of changes to our current API that I recorded in the changelog -- I'll do a self review first to make sure I didn't miss anything. Please feel free to give feedback on any API / architectural changes I made.


cesium-concierge:

Source/CesiumRuntime/Public/CesiumMetadataPrimitive.h
```diff
@@ -101,7 +71,7 @@ class CESIUMRUNTIME_API UCesiumMetadataPrimitiveBlueprintLibrary
       BlueprintCallable,
       BlueprintPure,
       Category = "Cesium|Metadata|Primitive")
-  static const TArray<FCesiumFeatureIdAttribute>&
+  static const TArray<FCesiumFeatureIDAttribute>&
   GetFeatureIdAttributes(UPARAM(ref)
```
j9liu, line 75: Should I keep this and `GetFeatureIdTextures`? Maybe we'd need to pass `CesiumPrimitiveFeatures` into this to retrieve the original classes from the `FCesiumFeatureIDSet`s.

Source/CesiumRuntime/Public/CesiumMetadataUtilityBlueprintLibrary.h
```diff
@@ -71,35 +68,4 @@ class CESIUMRUNTIME_API UCesiumMetadataUtilityBlueprintLibrary
...
-
-  /**
-   * Gets the feature ID associated with a given face for a feature id
-   * attribute.
-   */
-  UFUNCTION(
-      BlueprintCallable,
-      BlueprintPure,
-      Category = "Cesium|Metadata|Utility")
-  static int64 GetFeatureIDFromFaceID(
```
j9liu, line 83: Again, this tries to act on a `FCesiumMetadataPrimitive` which no longer manages the feature IDs. But this should probably be readded for backwards compatibility.

Source/CesiumRuntime/Private/CesiumMetadataUtilityBlueprintLibrary.cpp
```diff
@@ -59,31 +56,35 @@ UCesiumMetadataUtilityBlueprintLibrary::GetMetadataValuesForFace(
...
+
   // For now, only considers the first feature
   // TODO: expand to arbitrary number of features once testing data is
   // available
   const TMap<FString, FCesiumFeatureTable>& featureTables =
       UCesiumMetadataModelBlueprintLibrary::GetFeatureTables(modelMetadata);
-  const FString& featureTableName =
-      UCesiumFeatureIdAttributeBlueprintLibrary::GetFeatureTableName(
-          featureIdAttributes[0]);
+  const FString& featureTableName = "";
```
j9liu, line 76: I'll fix this properly later. It'd depend on revamping `CesiumFeatureTable`, which is out of the intended scope of this PR.

Source/CesiumRuntime/Private/CesiumFeatureIdSet.cpp
```diff
@@ -0,0 +1,109 @@
...
+}
+
+int64 UCesiumFeatureIDSetBlueprintLibrary::GetFeatureCount(
+    UPARAM(ref) const FCesiumFeatureIDSet& FeatureIDSet) {
+  return FeatureIDSet._featureCount;
+}
+
+int64 UCesiumFeatureIDSetBlueprintLibrary::GetFeatureIDForVertex(
+    UPARAM(ref) const FCesiumFeatureIDSet& FeatureIDSet,
+    int64 VertexIndex) {
```
j9liu, line 87: Validate `VertexIndex` just to be safe?

j9liu (commented):
Leaving comments for myself to get to tomorrow. Mostly just thoughts about how to add backwards compatibility for functions I deleted.

j9liu:
@kring this is ready for a review, if you have the time to review a meaty PR 🍖 

I think CI is complaining about some specs with enum comparisons, so I'll look into them tomorrow.

Source/CesiumRuntime/Public/CesiumFeatureIdAttribute.h
```diff
@@ -2,19 +2,31 @@
...
-struct FeatureTable;
 } // namespace CesiumGltf
 
 /**
- * @brief A blueprint-accessible wrapper for a feature id attribute from a glTF
+ * @brief Reports the status of a FCesiumFeatureIdAttribute. If the feature ID
+ * attribute cannot be accessed, this briefly indicates why.
+ */
+UENUM(BlueprintType)
+enum class ECesiumFeatureIdAttributeStatus : uint8 {
```
j9liu, line 19: Are these "status" enums overkill? I thought it might be helpful to expose to the user if something was wrong, but I know that deviates from CesiumJS's "silently fail" philosophy. I could also consolidate it to a bool that indicates plainly whether the thing is valid.

Source/CesiumRuntime/Private/CesiumFeatureIdAttribute.cpp
```diff
@@ -35,63 +40,79 @@ struct VertexCountFromAccessor {
...
 
 FCesiumFeatureIdAttribute::FCesiumFeatureIdAttribute(
-    const Model& model,
-    const Accessor& featureIDAccessor,
-    int32 attributeIndex,
-    const FString& featureTableName)
-    : _featureTableName(featureTableName), _attributeIndex(attributeIndex) {
-  switch (featureIDAccessor.componentType) {
-  case CesiumGltf::Accessor::ComponentType::BYTE:
+    const Model& InModel,
```
kring, line 43: Out of curiosity, why `InModel` rather than just `Model`? I think it should be ok in most cases to have both a type and parameter named `Model`.

Source/CesiumRuntime/Public/CesiumFeatureIdAttribute.h
kring: This is great! Silently failing (or even failing without explanation) just means more frustrated users and more questions for us on the forums.

Source/CesiumRuntime/Private/Tests/CesiumFeatureIdSpecUtility.h
```diff
@@ -0,0 +1,182 @@
...
+ * accessor in the process.
+ * @returns The index of the accessor.
+ */
+int32_t AddBufferToPrimitive(
+    CesiumGltf::Model& model,
+    CesiumGltf::MeshPrimitive& primitive,
+    const std::string& type,
+    const int32_t componentType,
+    const std::vector<std::byte>&& values) {
+  CesiumGltf::Buffer& buffer = model.buffers.emplace_back();
```
kring, line 19: I think it's worthwhile to move the implementations to a .cpp file. It probably won't really make much difference in the grand scheme of Unreal Engine code, but less code in headers means less code that needs to be compiled when editing source files that use those headers, and that's particularly a good thing for headers meant to be used in tests, because we'd like the iteration time on changing tests to be as quick as possible.

Source/CesiumRuntime/Public/CesiumMetadataPrimitive.h
```diff
@@ -49,23 +45,18 @@ struct CESIUMRUNTIME_API FCesiumMetadataPrimitive {
...
       const CesiumGltf::Model& model,
       const CesiumGltf::MeshPrimitive& primitive,
       const CesiumGltf::ExtensionMeshPrimitiveExtFeatureMetadata& metadata,
-      const CesiumGltf::ExtensionModelExtFeatureMetadata& modelMetadata);
+      const FCesiumPrimitiveFeatures& primitiveFeatures);
 
 private:
-  TArray<FCesiumFeatureIdAttribute> _featureIdAttributes;
-  TArray<FCesiumFeatureIdTexture> _featureIdTextures;
+  const FCesiumPrimitiveFeatures* _pPrimitiveFeatures;
```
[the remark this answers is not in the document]
kring: Looking closer, maybe it's ok because instances of both `FCesiumPrimitiveFeatures` and `FCesiumMetadataPrimitive` are stored side-by-side in `UCesiumGltfPrimitiveComponent`, so they have the same lifetime.

If that will always be the case, you can ignore my suggestion above. But it could be worth a note in the doc for the constructor about this, because it violates C++ norms a bit.

kring (commented):
Looks great @j9liu! I focused primarily on the header files rather than the tests and implementation files, though I did briefly look at all of it. Just a few suggestions below.

Source/CesiumRuntime/Public/CesiumMetadataPrimitive.h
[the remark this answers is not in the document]
j9liu: Yeah, this relationship is awkward. Ideally `FCesiumMetadataPrimitive` wouldn't have to exist (it will be replaced by a new metadata implementation). But to keep backwards compatibility, it needs to be aware of the new implementations and use _their_ functions in order to maintain its functionality. Hence the side-by-side construction and pointers.

I would prefer to keep `FCesiumPrimitiveFeatures` independent of `FCesiumMetadataPrimitive`. The `EXT_mesh_features` extension can exist independently of `EXT_structural_metadata` and I think the implementation should reflect that. So I'll go with the second suggestion, but If there's a way to better do this while adhering to C++ norms, let me know.

Source/CesiumRuntime/Private/CesiumFeatureIdAttribute.cpp
j9liu: I thought Visual Studio complained about that before, but it may have been Intellisense being wonky. I'll rename it to `Model` for consistency with the other variables.

j9liu:
Hm, it seems like Visual Studio disagrees with the clang formatter in CI. When I run "Format Document" on `CesiumPrimitiveFeatures.spec.cpp`, it puts "CesiumPrimitiveFeatures.h" at the top. But CI is complaining that the headers aren't in alphabetical order...

In any case, thanks @kring for the review! I addressed all of your comments, up to you whether you want to give it another pass before you merge.

kring:
> Hm, it seems like Visual Studio disagrees with the clang formatter in CI. 

That's strange. I'm assuming you have clang-format enabled in VS, right?

image

If so, I guess the other possibilities are either that it's not finding the (right?) `.clang-format` file, or (least likely) there's a difference between clang-format versions. There's a way to specify an explicit clang-format.exe for VS to use on that same options panel, so you could try pointing it to the one installed in `node_modules`.

j9liu:
> If so, I guess the other possibilities are either that it's not finding the (right?) `.clang-format` file, or (least likely) there's a difference between clang-format versions. There's a way to specify an explicit clang-format.exe for VS to use on that same options panel, so you could try pointing it to the one installed in `node_modules`.

Manually pointing it to the `node_modules` file worked, thanks!

PR #1126, opened 2023-06-09
Fix Physics Bug caused by Unreal 5.1

joseph-kaile:
Fixes the issue described here: https://community.cesium.com/t/ue-5-1-3d-tileset-collisions-with-physics-simulated-actors-not-working-correctly/22106


cesium-concierge:

* :x: Missing CLA.

kring:
I think the fact that reversing the winding order fixes the physics problems suggests that it is, in fact, related to the negative scale. Perhaps Chaos thinks the object is colliding through the back side of the object, when really it's a negative-scaled front side.

I'm generally ok with reversing the winding order, but I don't think enabling double-sided geometry is a good idea. I think that can lead to things getting stuck inside 3D Tiles models, which is especially problematic because 3D Tiles change level of detail on the fly.

So, do you think we might be able to fix this by eliminating the negative scale and instead inverting the Y coordinate? It might be tricky, I'm really not sure.

Source/CesiumRuntime/Private/CesiumGltfComponent.cpp
```diff
@@ -1210,10 +1216,29 @@ static void loadPrimitive(
   primitiveResult.pModel = &model;
   primitiveResult.pMeshPrimitive = &primitive;
   primitiveResult.RenderData = std::move(RenderData);
-  primitiveResult.transform = transform;
   primitiveResult.pMaterial = &material;
   primitiveResult.pCollisionMesh = nullptr;
 
+  static constexpr glm::dmat4 scaleMatrix = {
```
kring, line 1222: Let's give this a better name and a comment explaining why it exists. It's not scaling, just inverting the Y-axis.

kring (commented):
Very nice @joseph-kaile, it's very clever how you did this! You manually reversed the Y coordinate everywhere, so the mesh itself has the correct winding order, and then modified the transformation chain so that the effect of that is immediately undone. That neatly avoids the order-of-matrix-multiplications problem I was originally worried about when thinking about this.

I also confirmed that this fixes that problems with collision in UE 5.1.

So, just one small comment below and I'll merge this!

kring:
Thanks @joseph-kaile!

PR #1132, opened 2023-06-13
Fix !IsGarbageCollecting() assertion in ACesiumCreditSystem

### LINKED ISSUE #1121: Condition failed (debug) in ACesiumCreditSystem::removeCreditsFromViewports
- In Visual Studio, build a debug version of the Unreal sample (and plugin)
- Press F5 to run
- In Unreal, open 04_MAIN_CesiumSublevels in [cesium-unreal-samples](https://github.com/CesiumGS/cesium-unreal-samples). 
- Start Play-in-Editor mode. 
- Press 4 to fly to Paris.
- Once you land, press 1 to go back to Denver

Almost right away, you'll see an exception thrown in Visual Studio...
(you can continue past it though)

```
Ensure condition failed: !IsGarbageCollecting() [File:D:\build\++UE5\Sync\Engine\Source\Runtime\UMG\Private\Slate\SObjectWidget.cpp] [Line: 43] 
SObjectWidget for 'ScreenCredits_C /Engine/Transient.UnrealEdEngine_0:GameInstance_0.ScreenCredits_C_2' destroyed while collecting garbage.  This can lead to multiple GCs being required to cleanup the object.  Possible causes might be,
1) ReleaseSlateResources not being implemented for the owner of this pointer.
2) You may just be holding onto some slate pointers on an actor that don't get reset until the actor is Garbage Collected.  You should avoid doing this, and instead reset those references when the actor is Destroyed.
[2023.06.08-17.03.54:392][209]LogOutputDevice: Error: Ensure condition failed: !IsGarbageCollecting() [File:D:\build\++UE5\Sync\Engine\Source\Runtime\UMG\Private\Slate\SObjectWidget.cpp] [Line: 43] 
SObjectWidget for 'ScreenCredits_C /Engine/Transient.UnrealEdEngine_0:GameInstance_0.ScreenCredits_C_2' destroyed while collecting garbage.  This can lead to multiple GCs being required to cleanup the object.  Possible causes might be,
1) ReleaseSlateResources not being implemented for the owner of this pointer.
2) You may just be holding onto some slate pointers on an actor that don't get reset until the actor is Garbage Collected.  You should avoid doing this, and instead reset those references when the actor is Destroyed.

[2023.06.08-17.03.54:392][209]LogStats:             FDebug::EnsureFailed -  0.000 s
UnrealEditor-Win64-DebugGame.exe has triggered a breakpoint.


```

csciguy8:
Move `::removeCreditsFromViewports` call out of `::BeginDestroy` and into `::EndPlay`. This prevents more work for the garbage collector being created, while the garbage collector is actively destroying an object.

Fixes #1121

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,6 +7,10 @@
 - Removed the `GetGeoreferencedToEllipsoidCenteredTransform` and `GetEllipsoidCenteredToGeoreferencedTransform` methods from `GeoTransforms`. Because these were transformations between two right-handed coordinate systems, they are not of much use with Unreal's left-handed coordinate system.
 - Deprecated the `FlyToGranularityDegrees` property for `AGlobeAwareDefaultPawn`. Flight interpolation is now computed per-frame, so this property is no longer needed. Any code that refers to `FlyToGranularityDegrees` should be removed or changed to `FlyToGranularityDegrees_DEPRECATED` to still compile.
 
+##### Fixes :wrench:
+
+- Fixed a debug assertion `!IsGarbageCollecting()` that could occur within `ACesiumCreditSystem` when flying to different sublevels.
+
 ### v1.27.0 - 2023-06-1
 
 ##### Additions :tada:
--- a/Source/CesiumRuntime/Private/CesiumCreditSystem.cpp
+++ b/Source/CesiumRuntime/Private/CesiumCreditSystem.cpp
@@ -202,8 +202,6 @@ void ACesiumCreditSystem::OnConstruction(const FTransform& Transform) {
 }
 
 void ACesiumCreditSystem::BeginDestroy() {
-  this->removeCreditsFromViewports();
-
 #if WITH_EDITOR
   FLevelEditorModule* pLevelEditorModule =
       FModuleManager::GetModulePtr<FLevelEditorModule>(LevelEditorName);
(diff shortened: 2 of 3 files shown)


cesium-concierge:

kring:
Thanks @csciguy8! This works well in all the cases I tried.

I did see one assertion failure while testing this, but I haven't been able to recreate it. It's definitely a race condition, but almost certainly not introduced in this PR. 

The assertion was "Unable to create new object: %s %s.%s. Creating UObjects while Collecting Garbage is not allowed!" The `%s` in this case was a texture. Looked like the situation was as follows:

1. A level was being unloaded, so the garbage collector was running.
2. The GC called `IsReadyForFinishDestroy` on an ACesium3DTileset.
3. This method is designed to return true only after all of the tileset's background processing has shut down. So to do that, it ticks the asset accessor and dispatches main thread tasks.
4. Here's where the race condition comes in. In the case where I saw the assertion, ticking the asset accessor just happens to cause an HTTP request _for a credit image_ to complete. Unreal's HttpManager raises the `OnProcessRequestComplete` callback.
5. The credit system attempts to create a texture for the credit image that was just downloaded.
6. Assertion fails because creating a texture during garbage collection is not allowed.

So that's some pretty unlucky timing, but a clear bug. How to fix it, though? Perhaps the credit system could dispatch a separate task to the game thread to create the texture, rather than doing it right on the `OnProcessRequestComplete` callback. That way the stack can unwind out of the GC. Can even call `IsGarbageCollecting()` and only do this if it is. But need to be careful because in some cases that game thread task may end up running after the level that initiated it has been completely unloaded, so we'd just be swapping one race condition for another.

PR #1136, opened 2023-06-14
Check if credit texture load succeeds before proceeding

joseph-kaile:
Exposed by recent AWS outage...

### DIFF
--- a/Source/CesiumRuntime/Private/ScreenCreditsWidget.cpp
+++ b/Source/CesiumRuntime/Private/ScreenCreditsWidget.cpp
@@ -212,10 +212,11 @@ void UScreenCreditsWidget::HandleImageRequest(
     FHttpResponsePtr HttpResponse,
     bool bSucceeded,
     int32 id) {
+  UTexture2D* texture = nullptr;
   if (bSucceeded && HttpResponse.IsValid() &&
-      HttpResponse->GetContentLength() > 0) {
-    UTexture2D* texture =
-        FImageUtils::ImportBufferAsTexture2D(HttpResponse->GetContent());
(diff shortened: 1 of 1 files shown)


Source/CesiumRuntime/Private/ScreenCreditsWidget.cpp
```diff
@@ -212,10 +212,11 @@ void UScreenCreditsWidget::HandleImageRequest(
...
     bool bSucceeded,
     int32 id) {
+  UTexture2D* texture = nullptr;
   if (bSucceeded && HttpResponse.IsValid() &&
-      HttpResponse->GetContentLength() > 0) {
-    UTexture2D* texture =
-        FImageUtils::ImportBufferAsTexture2D(HttpResponse->GetContent());
+      HttpResponse->GetContentLength() > 0 &&
+      (texture = FImageUtils::ImportBufferAsTexture2D(
+           HttpResponse->GetContent())) != nullptr) {
```
csciguy8, line 219: This new check looks ok to me. I double checked that `FImageUtils::ImportBufferAsTexture2D` at least logs something on failure, and it does.
I am a little concerned about the code path that results in `::SetCredits` being called. 
For example, even if an image fails to load, it still decrements `_numImagesLoading`. This means if there are 2 images loading, one sequence of events could fail, succeed, then call `::SetCredits`. And another sequence could succeed, fail, and never call `::SetCredits`.
I assume this isn't what we want?

joseph-kaile: Nice catch, that is definitely a bug.

Issue #1122, opened 2023-06-08
Create tests for AGlobeAwareDefaultPawn


csciguy8:
These tests would be the first iteration and need to test base functionality.

Of particular interest are the FlyToXXXX functions, which are time-based. Ideally, we would be able to run time-based tests, but not wait in real time. (maybe override game time inside the test)

Piggyback on the new testing framework introduced in #1109 

Relates to latest rework in this class in #1120

csciguy8:
Might be some useful testing code in this old draft PR, #548

csciguy8:
@kring , do you think this is good enough to call done from your latest PR? #1149 

(introduction of Source/CesiumRuntime/Private/Tests/GlobeAwareDefaultPawn.spec.cpp)

kring:
I only added one test for the very specific issue I ran into. I think it may be worthwhile to add some more general tests of the functionality.

PR #1149, opened 2023-06-27
Fix rare-ish camera jump when arriving at a flight destination

kring:
In the Sublevels Sample, sometimes, upon arrival in Paris, the flight interpolation percentage would be exactly 1.0. Not just "displays as 1.0", but exactly equal to 1.0. And when that happened, the interpolation of the altitude flight curve would return 1.0, so the flight would put the altitude at its maximum rather than its destination height. This makes absolutely no sense, because the altitude interpolation key points are (0.0, 0.0), (0.5, 1.0), (1.0, 0.0). In other words, it should be 0.0 (minimum height) at the start and end, and 1.0 (maximum height) in the middle, just what you'd expect for altitude on a camera flight. Yet asking for the interpolated value at exactly 1.0 would return that value at 0.5 instead of the value at 1.0.

So stepping into UE's curve interpolation, I found it's doing some fancy cubic interpolation with tangents and whatnot. Ok, fine, but I'm literally asking for one of the keypoints! Well apparently the implementation of this cubic interpolation solves some cubic equation and comes up with three roots. It then picks one that is between 0.0 and 1.0. Three of the roots are well outside that range, but one is very very very slightly larger than 1.0. So Unreal decides none of these solutions are correct and decides to just make up an answer instead, 0.0. The one slightly larger than 1.0 is the correct answer here, but floating point rounding has nudged it to the wrong side of the range.

So, during a camera flight, just as the pawn should be arriving at its destination, it (very briefly, one frame) is given a height equal to the maximum during the flight, before it settles back down to the right place on the next frame.

This PR works around this problem by ensuring that we never interpolate at t=1.0, even if that t=1.0 came from the FlyToProgressCurve, rather than being 1.0 initially. Instead, t=1.0 is simply treated as being at the end of the flight.

I also switched to using `float` instead of `double` for the time variables. The extra precision should be unnecessary, and in theory we could end up with a value that is slightly less than 1.0 as a double, but exactly equal to 1.0 as a float. So using float consistently avoids this (remote) possibility.

### DIFF
--- a/Source/CesiumRuntime/Private/GlobeAwareDefaultPawn.cpp
+++ b/Source/CesiumRuntime/Private/GlobeAwareDefaultPawn.cpp
@@ -129,7 +129,9 @@ void AGlobeAwareDefaultPawn::FlyToLocationECEF(
 
   // The source and destination rotations are expressed in East-South-Up
   // coordinates.
-  this->_flyToSourceRotation = Controller->GetControlRotation().Quaternion();
+  this->_flyToSourceRotation =
+      IsValid(Controller) ? Controller->GetControlRotation().Quaternion()
+                          : FQuat::Identity;
   this->_flyToDestinationRotation =
       FRotator(PitchAtDestination, YawAtDestination, 0).Quaternion();
   this->_flyToECEFDestination = ECEFDestination;
--- a/Source/CesiumRuntime/Private/GlobeAwareDefaultPawn.cpp
+++ b/Source/CesiumRuntime/Private/GlobeAwareDefaultPawn.cpp
@@ -120,6 +120,19 @@ void AGlobeAwareDefaultPawn::FlyToLocationECEF(
     bool CanInterruptByMoving) {
 
   if (this->_bFlyingToLocation) {
+    UE_LOG(
+        LogCesium,
+        Error,
+        TEXT("Cannot start a flight because one is already in progress."));
+    return;
+  }
+
+  if (!IsValid(this->Controller)) {
+    UE_LOG(
+        LogCesium,
+        Error,
+        TEXT(
+            "Cannot start a flight because the pawn does not have a Controller. You probably need to \"possess\" it before attempting to initiate a flight."));
     return;
   }
 
@@ -129,9 +142,7 @@ void AGlobeAwareDefaultPawn::FlyToLocationECEF(
 
   // The source and destination rotations are expressed in East-South-Up
   // coordinates.
-  this->_flyToSourceRotation =
-      IsValid(Controller) ? Controller->GetControlRotation().Quaternion()
-                          : FQuat::Identity;
+  this->_flyToSourceRotation = Controller->GetControlRotation().Quaternion();
   this->_flyToDestinationRotation =
       FRotator(PitchAtDestination, YawAtDestination, 0).Quaternion();
   this->_flyToECEFDestination = ECEFDestination;
--- a/Source/CesiumRuntime/Private/GlobeAwareDefaultPawn.cpp
+++ b/Source/CesiumRuntime/Private/GlobeAwareDefaultPawn.cpp
@@ -81,7 +81,7 @@ FRotator AGlobeAwareDefaultPawn::GetBaseAimRotation() const {
 }
 
 void AGlobeAwareDefaultPawn::_interpolateFlightPosition(
-    double percentage,
+    float percentage,
     glm::dvec3& out) const {
 
   // Rotate our normalized source direction, interpolating with time
@@ -265,14 +265,30 @@ void AGlobeAwareDefaultPawn::_handleFlightStep(float DeltaSeconds) {
     return;
   }
 
-  this->_currentFlyTime += static_cast<double>(DeltaSeconds);
+  this->_currentFlyTime += DeltaSeconds;
 
-  // If we reached the end, set actual destination location and orientation
-  if (this->_currentFlyTime >= static_cast<double>(this->FlyToDuration)) {
+  // In order to accelerate at start and slow down at end, we use a progress
+  // profile curve
+  float flyPercentage;
+  if (this->_currentFlyTime >= this->FlyToDuration) {
+    flyPercentage = 1.0f;
+  } else if (this->FlyToProgressCurve) {
+    flyPercentage = glm::clamp(
+        this->FlyToProgressCurve->GetFloatValue(
+            this->_currentFlyTime / this->FlyToDuration),
+        0.0f,
+        1.0f);
+  } else {
+    flyPercentage = this->_currentFlyTime / this->FlyToDuration;
+  }
+
+  // If we reached the end, set actual destination location and
+  // orientation
+  if (flyPercentage >= 1.0f) {
     this->GlobeAnchor->MoveToECEF(this->_flyToECEFDestination);
     Controller->SetControlRotation(this->_flyToDestinationRotation.Rotator());
     this->_bFlyingToLocation = false;
-    this->_currentFlyTime = 0.0;
+    this->_currentFlyTime = 0.0f;
 
     // Trigger callback accessible from BP
     UE_LOG(LogCesium, Verbose, TEXT("Broadcasting OnFlightComplete"));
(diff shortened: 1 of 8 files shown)


cesium-concierge:

csciguy8:
Mentioning a related testing issue, #1122 . We should probably have time based tests than can exercise end interpolation values like mentioned here.

Source/CesiumRuntime/Private/Tests/GlobeAwareDefaultPawn.spec.cpp
```diff
@@ -0,0 +1,83 @@
...
+              FRequestPlaySessionParams params{};
+              GEditor->RequestPlaySession(params);
+            });
+        BeforeEach(EAsyncExecution::TaskGraphMainThread, [this]() {
+          FEditorDelegates::PostPIEStarted.Remove(subscriptionPostPIEStarted);
+        });
+        AfterEach(EAsyncExecution::TaskGraphMainThread, [this]() {
+          GEditor->RequestEndPlayMap();
+        });
+        It("", [this]() {
```
csciguy8, line 52:
```suggestion
        It(""should not spike altitude when very close to final destination"", [this]() {
```

Source/CesiumRuntime/Private/Tests/GlobeAwareDefaultPawn.spec.cpp
```diff
@@ -0,0 +1,83 @@
...
+    EAutomationTestFlags::ApplicationContextMask |
+        EAutomationTestFlags::ProductFilter)
+
+FDelegateHandle subscriptionPostPIEStarted;
+
+END_DEFINE_SPEC(FGlobeAwareDefaultPawnSpec)
+
+void FGlobeAwareDefaultPawnSpec::Define() {
+  Describe(
+      "should not spike altitude when very close to final destination",
```
csciguy8, line 26:
```suggestion
      "AGlobeAwareDefaultPawn",
```

Source/CesiumRuntime/Private/Tests/GlobeAwareDefaultPawn.spec.cpp
```diff
@@ -0,0 +1,83 @@
...
+                  TSoftObjectPtr<UObject>(FSoftObjectPath(TEXT(
+                      "Class'/CesiumForUnreal/DynamicPawn.DynamicPawn_C'")));
+              AGlobeAwareDefaultPawn* pPawn =
+                  pWorld->SpawnActor<AGlobeAwareDefaultPawn>(
+                      Cast<UClass>(DynamicPawn.LoadSynchronous()));
+
+              subscriptionPostPIEStarted =
+                  FEditorDelegates::PostPIEStarted.AddLambda(
+                      [done](bool isSimulating) { done.Execute(); });
+              FRequestPlaySessionParams params{};
```
csciguy8, line 43: I noticed a bit of audio logging spam to the console and report log with this test. I tried running the PIE without sound, but noticed no change

```
              ULevelEditorPlaySettings* playSettings = NewObject<ULevelEditorPlaySettings>();
              playSettings->EnableGameSound = false;

              FRequestPlaySessionParams params{};
              params.EditorPlaySettings = playSettings;
```

I wonder if we can just disable audio logging for the test completely?

Source/CesiumRuntime/Private/GlobeAwareDefaultPawn.cpp
```diff
@@ -265,14 +267,30 @@ void AGlobeAwareDefaultPawn::_handleFlightStep(float DeltaSeconds) {
...
-  // If we reached the end, set actual destination location and orientation
-  if (this->_currentFlyTime >= static_cast<double>(this->FlyToDuration)) {
+  // In order to accelerate at start and slow down at end, we use a progress
+  // profile curve
+  float flyPercentage;
+  if (this->_currentFlyTime >= this->FlyToDuration) {
+    flyPercentage = 1.0f;
+  } else if (this->FlyToProgressCurve) {
+    flyPercentage = glm::clamp(
+        this->FlyToProgressCurve->GetFloatValue(
```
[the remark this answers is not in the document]
j9liu: I think the idea is that `FlyToProgressCurve` controls how fast you actually go from point A to point B, regardless of altitude. `FlyToAltitudeProfileCurve` controls how it ascends / descends, but `FlyToProgressCurve` controls the overall animation.

So for example, you _can_ have a gradual ascent / descent from two locations with the altitude profile curve, but the lateral movement is still at a constant rate. If you add a sort of S-shaped progress curve, the ascent and descents will be slower, but the zenith of the flight will pass by fast.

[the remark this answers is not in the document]
csciguy8: That makes sense!

Source/CesiumRuntime/Private/GlobeAwareDefaultPawn.cpp
```diff
@@ -129,7 +129,9 @@ void AGlobeAwareDefaultPawn::FlyToLocationECEF(
 
   // The source and destination rotations are expressed in East-South-Up
   // coordinates.
-  this->_flyToSourceRotation = Controller->GetControlRotation().Quaternion();
+  this->_flyToSourceRotation =
```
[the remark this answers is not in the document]
kring: Originally, I tried to run the new test in Edit mode, but it crashed because there was no Controller. So I added this check. But then there were other problems running it in Edit mode (mostly that the flights are completely disabled if there's no game world), so I gave up and made the test use Play-in-Editor mode. I left this change in, though, because not crashing is better than crashing. I believe it's possible to have no Controller even in a game, if the pawn isn't posessed.

Really, camera flights should have nothing to do with Controllers or even Pawns. It should simply be a component that we can add to any Actor, similar to what we've done in Unity.

But I'm not inclined to get into any of that as part of this PR.

Source/CesiumRuntime/Private/Tests/GlobeAwareDefaultPawn.spec.cpp
kring: I intentionally didn't do it this way...

What you've proposed would be fine for right now. But it would break down the moment we add the next test. We can't add another `It` to the existing `Describe`, because then all the same `BeforeEach` and `AfterEach` actions would run for that new test as well. So we'd have to add a new `Describe`. What would we call the new one? `AGlobeAwareDefaultPawn2`? :)

It's kind of silly how the only way (AFAICT) to do a sequence of actions over frames is with these multiple `BeforeEach` actions. But it is what it is, and Epic themselves have run into the same problem, and they adopted the same pattern I did, which is of course where I got the idea. See here for example:
https://github.com/EpicGames/UnrealEngine/blob/5.2.1-release/Engine/Plugins/Tests/AutomationDriverTests/Source/AutomationDriverTests/Private/AutomationDriver.spec.cpp#L287

The test UI in the Editor handles this specially by hiding the anonymous test and making the one test have the label of the Describe instead. So it all works just fine, other than the kind of weird looking anoymous `It`.

Source/CesiumRuntime/Private/Tests/GlobeAwareDefaultPawn.spec.cpp
kring: I'm not sure. I think it's possible to mark certain messages as expected, but I don't think that would be appropriate here.

Source/CesiumRuntime/Private/GlobeAwareDefaultPawn.cpp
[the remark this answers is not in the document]
kring: I cleaned this up a little based on your suggestion. It now aborts earlier when there's no Controller, and logs an error message.

Source/CesiumRuntime/Private/Tests/GlobeAwareDefaultPawn.spec.cpp
csciguy8: I'm not sure either. Maybe just something to keep in mind in the future, if excess test logging starts causing problems.

Source/CesiumRuntime/Private/Tests/GlobeAwareDefaultPawn.spec.cpp
csciguy8: Sounds like a reasonable course of action.

Source/CesiumRuntime/Private/GlobeAwareDefaultPawn.cpp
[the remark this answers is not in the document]
csciguy8: Looks great, thanks.

csciguy8 (approved):

PR #1150, opened 2023-06-27
Support tests for Unreal 5.1 5.2

### LINKED ISSUE #1146: Support CI tests for Unreal 5.1 and 5.2
As of #1117 we have automated tests running under continuous integration, but only for Windows and Unreal 5.0.

Add these tests for Unreal 5.1 and Unreal 5.2.

Consider refactoring duplicate code in the workflow, if possible

csciguy8:
Extend existing automated tests to run on Unreal 5.1 and Unreal 5.2.

Remove some duplicate code by introducing reusable workflows for the Windows jobs.

Fix double zip problem when downloading artifacts from the github UI. When we use `actions/upload-artifact`, remove the added zip extension. This action isn't actually zipping anything. The zip is created in the github UI when we click to download.

Closes #1146


cesium-concierge:

kring:
Looks great @csciguy8! Just one request: please make the test step run in parallel with the combine step. No need to serialize these, or to prevent creation of the combined package just because a test failed.

When you add a package testing step, this should ideally run after the combine so that we're testing the final, combined package. One could argue that the test step should run against the combined package, too. The downside to that is that we don't get any test feedback from CI until after the slowest build (macOS) completes. I think running tests early and packaging late is a reasonable balance of the tradeoffs.

csciguy8:
@kring Thanks!  I agree about a packaging testing step happening after the combine, so that we are testing the final combined package.

But for unit tests, the "Cesium" tests, I would think we would want them like they are, per platform, before the final combine. The rough logic... if any platform unit test fails, the final artifact is invalid, don't bother building it. 

Thoughts? Maybe this is splitting hairs?

csciguy8:
> When you add a package testing step, this should ideally run after the combine so that we're testing the final, combined package. 

Agreed. Makes the most sense.

kring:
Thanks @csciguy8!

PR #1153, opened 2023-06-28
Remove old World Composition based sub-level system, add new one based on level instances

### LINKED ISSUE #729: UE5 Support: Explore alternatives to georeferenced sublevels
In UE4, we used world composition to enable georeferenced sublevels. World composition is typically used to arrange sublevels on a grid so they can be loaded in or out based on where the player is on the grid. But in our case we always spawn the sublevels at the Unreal origin so we didn't need this feature. We only used world composition to explicitly tell the level streaming system that these sublevels are _not_ to be streamed in according to distance, instead our georeference system should control when levels should be loaded from C++.

In UE5, world composition has been removed and replaced with world partition. There is also a level instancing feature which can be used outside of a world partition system as long as a custom streaming management logic is provided, this sounds like exactly what we want ([reference here](https://docs.unrealengine.com/5.0/en-US/WorldFeatures/WorldPartition/LevelInstancing/)). Level instancing also adds the ability for levels to have arbitrary transforms and since this is UE5 I assume these transforms can be in double precision. In some cases, this _may_ remove the need for the georeference origin changing that currently happens when switching levels, but for the foreseeable future it will usually be more convenient to spawn sublevels at the origin and change the georeference. 

In general, we should investigate if it would be possible to have the Unreal coordinate system directly represent ECEF. If so, we should provide that as an alternative option to using the georeference system. This includes providing alternative code paths for Cesium objects to be correctly located when UE coordinates represent ECEF.**

**Technically UE coordinates will be left-handed, Z-up, and in centimeters, so not exactly ECEF, but equivalent up-to a simple conversion transform.

### LINKED ISSUE #1026: Changing sublevel origin does not change the origin of the `CesiumGeoreference`
When you change the `CesiumGeoreference` origin, it changes the origin of the active sublevel. However, this is not true the other way around. Changing the active sublevel's origin should also update the origin of the `CesiumGeoreference`.

georef-and-sublevel

(thanks @kring for the gif!)

kring:
Builds on #1149 so merge that first.

Fixes #729
Fixes #1026

This is a complete overhaul of the sub-level system. It no longer uses World Composition, which is good because that feature has been deprecated in UE. Instead, it uses Level Instance Actors. Old levels based on World Composition are converted to the new system on load.

Creating sub-levels is easy! Let's start with a basic level with Cesium World Terrain, and then add some spiffy cubes to make tumble down onto the terrain.

image

We'll enable "Simulate Physics" on the cubes:
image

So far there are no sub-levels involved. Let's create one. Select all the cubes by shift-clicking, then right-click on one of them and choose Level -> Create Level Instance.

image

Keep the default settings and press Ok. Name the sub-level "CubesInDenver".

Next, click the "CubesInDenver" level instance that was just created and add the "Cesium Sub Level Component" to it. It will disappear (not ideal, I should work on this?). To make it reappear, click the eye icon next to the level instance in the Outliner:

image

If we zoom out a bit, we can see a wireframe sphere indicating the load radius of the sub-level. There's a property on the sub-level to change this.

image

Notice that the cubes aren't really centered within the sub-level sphere. That's because, by default, the sub-level has adopted the origin of the CesiumGeoreference. If you placed your cubes farther from the origin, they might not be inside the load radius at all.

We can fix this by clicking "Place Georeference Origin at Sub Level Origin" button on the sub-level component. This adjusts both the georeference origin and the level instance's location so that the objects don't move, but the georeference origin is at the center of the sub-level. Better:

image

Now if we Play-in-Editor, the cubes will appear when we're inside the load radius, and disappear when we're outside. 

Now let's edit the sub-level. Click the Level Instance Actor, and then click the "Edit" button in the Details panel. UE will grey out everything that isn't in the sub-level:

image

image

Let's add a cone and give it a fancy material (I chose `BasicAsset01`):

image

Click the Level Instance Actor again and choose Commit Changes:

image

You can create additional sub-levels by using the same workflow. Or, you can hide the existing sub-level (by clicking that Eye icon), set the CesiumGeoreference to the new location, and start working there. Be sure to hide the sub-level before modifying the CesiumGeoreference, though. Otherwise, the sub-level will move, too.

.... more to come ...

To Do:

* [x] Place Georeference Origin Here is not undoable, which is bad because pressing it at the wrong time can really mess up your level.
* [x] Should "Place Georeference Origin at Sub Level Origin" also move the Editor camera, so the leve doesn't appear to shift? Probably.
* [x] Don't hide level instances when the CesiumSubLevelComponent is first added.


cesium-concierge:

kring:
All the to dos have been address, so this should be ready for review!

j9liu:
> Notice that the cubes aren't really centered within the sub-level sphere. That's because, by default, the sub-level has adopted the origin of the CesiumGeoreference. If you placed your cubes farther from the origin, they might not be inside the load radius at all.
> 
> We can fix this by clicking "Place Georeference Origin at Sub Level Origin" button on the sub-level component. This adjusts both the georeference origin and the level instance's location so that the objects don't move, but the georeference origin is at the center of the sub-level.

In Unity, the georeference origin is always snapped to a sub-scene origin as long as the sub-scene is active. So as soon as you create a sub-scene in Unity, the georeference origin automatically takes the value of the sub-scene's origin. This is true even in edit mode.

In Unreal, though, it seems like when a Cesium sub level is visible (active?) in edit mode, the georeference doesn't automatically snap to the sub-scene origin. This feels like a confusing difference from Unity, and it doesn't seem like it's tied to any engine-specific limitations. Is there a way to make the behavior between the two engines more consistent?

j9liu:
Also, when I have multiple sub-scenes, the wireframe sphere draws for both sub-scenes even though only one is active. Is that also intentional?

image

kring:
> In Unity, the georeference origin is always snapped to a sub-scene origin as long as the sub-scene is active. So as soon as you create a sub-scene in Unity, the georeference origin automatically takes the value of the sub-scene's origin. This is true even in edit mode.

This is true in Unreal, too. The "Place Georeference Origin at Sub Level Origin" button is doing something subtly different.

When you create a Level Instance from existing Actors, Unreal gives it an origin in the UE world near the center of the objects you've created it from. And then the Actors inside the Level Instance have (small) offsets from the center. Pretty sensible.

When you add a CesiumSubLevelComponent to that Level Instance, it initially adopts the Georeference origin that is already active. If it did otherwise, adding the component would suddenly change the location on the globe of the objects in the Level Instance. So this is also pretty sensible, I think.

However, if you created the Level Instance far from the Georeference origin, then it's possible that the Level Instance itself has quite large coordinate values, and the Actors inside the sub-level are actually nowhere near the sub-level's georeference origin. Which is really not ideal, even though sensible choices got us here.

So the "Place Georeference Origin at Sub Level Origin" button fixes this by changing _both_ the Level Instance's transform (setting its location to (0,0,0) and also updating the sub-levels georeference origin so that the objects in the sub-level don't move at all relative to the globe. I considered doing this automatically when the component is first added to the Level Instance. It's a little tricky to get right, but maybe it's worth it.

Another thing that might help is to consider the "Load Radius" not from the georeference origin, but instead from the Level Instance origin. At least then you won't have the problem that the Load Radius sphere is nowhere near the objects in the sub-level. But I decided against that because it kind of just hides the problem. When the sub-level is far from the georeference origin, you end up with large coordinate values and a wonky orientation of the level. Better to fix it properly by resetting the Level Instance Transform.

But in any case, the georeference origin of the active sub-level is _always_ applied when a sub-level is shown, just like in Unity. If we ever failed to follow that rule, objects would appear in the wrong place on the globe, which would be really confusing and broken-looking for our users.

> Also, when I have multiple sub-scenes, the wireframe sphere draws for both sub-scenes even though only one is active. Is that also intentional?

It is. The old sub-level system did this as well. There are different things we could do here, and I'm not sure what's best:

* Only show the active sub-level's radius.
* Show all the sub-level radii, but highlight the active one in a different color.
* Allow the radius display to be toggled on each sub-level component individually (currently it's a global "show" flag on the CesiumGeoreference).

On my system, at least, those wireframe spheres have a shockingly high impact on frame rate, too. Might be worth trying to do something about that.

Source/CesiumRuntime/Private/Tests/SubLevels.spec.cpp
```diff
@@ -0,0 +1,338 @@
...
+
+  Describe(
+      "does not copy inactive sub-level origin changes to the CesiumGeoreference in the Editor",
+      [this]() {
+        BeforeEach(EAsyncExecution::TaskGraphMainThread, [this]() {
+          pSubLevel1->SetIsTemporarilyHiddenInEditor(false);
+        });
+        BeforeEach(EAsyncExecution::TaskGraphMainThread, [this]() {
+          pSubLevel1->SetIsTemporarilyHiddenInEditor(true);
+        });
```
j9liu, line 152: Is there an async-related reason that these are in separate clauses? I'm probably missing something

Source/CesiumRuntime/Private/Tests/SubLevels.spec.cpp
```diff
@@ -0,0 +1,338 @@
...
+          FRequestPlaySessionParams params{};
+          GEditor->RequestPlaySession(params);
+        });
+    BeforeEach(EAsyncExecution::TaskGraphMainThread, [this]() {
+      FEditorDelegates::PostPIEStarted.Remove(subscriptionPostPIEStarted);
+    });
+    LatentBeforeEach(
+        EAsyncExecution::TaskGraphMainThread,
+        [this](const FDoneDelegate& done) {
+          waitFor(done, GEditor->PlayWorld, 5.0f, [this]() {
```
j9liu, line 218: For me this (and the other play mode test) is failing because `waitFor` times out. I don't know if you can reproduce this on your machine, but if not I can take a look and debug.

j9liu (commented):
@kring looks good, I tried out the new sublevel functionality and can confirm it works! I noticed some remnants of the old implementation, so I pointed that out. Let me know if I'm misunderstanding anything with my comments.

Source/CesiumRuntime/Private/Tests/CesiumTestHelpers.h
```diff
@@ -2,9 +2,168 @@
...
+  }
+}
+
+/// <summary>
+/// Waits for a provided lambda function to become true, ticking through render
+/// frames in the meantime.
+/// </summary>
+/// <typeparam name="T"></typeparam>
+/// <param name="done">The done delegate provided by a LatentIt or
+/// LatentBeforeEach. It will be invoked when the condition is true.</param>
```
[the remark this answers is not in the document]
kring: Yes, it's intentional. Otherwise, the test would hang. Logging an error will trigger the test to fail, though. I'll add some doc to this effect.

kring:
> I definitely think that's sensible. I did the same in Unity to compare, and the baked-in origin immediately takes effect and shifts the entire world 😬 so we probably should add this change on the Unity side too.

Hmm yeah that's not great! Agreed, let's fix it. I wrote an issue for this: https://github.com/CesiumGS/cesium-unity/issues/343

> I misremembered the way that sub-scenes work in Unity because I usually created them with the "Create Sub-Scene Here" button. [...] It would be nice to have an equivalent function in Unreal, but perhaps it's not possible with the level instancing system.

Hmm yeah that's an interesting idea! I think we could probably create the Level Instance and then get the Editor to prompt the user for where to save the new level file. I wrote an issue for this: #1162

Source/CesiumRuntime/Private/Tests/CesiumTestHelpers.h
```diff
@@ -2,9 +2,168 @@
...
+/// <param name="pWorld">The world in which to check the condition.</param>
+/// <param name="timeoutSeconds">The maximum time to wait for the condition to
+/// become true.</param> <param name="condition">A lambda that is invoked each
```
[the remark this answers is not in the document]
kring: Ugh clang-format has no respect for these tags. Thanks for fixing it.

Source/CesiumRuntime/Public/CesiumSubLevelComponent.h
```diff
@@ -0,0 +1,300 @@
...
+ * Only if the "Sub Level Camera" is inside a sub-level's "Load Radius" will
+ * that sub-level be activated. If multiple sub-levels are in range, only the
+ * closest one will be activated.
+ */
+UCLASS(ClassGroup = (Cesium), meta = (BlueprintSpawnableComponent))
+class CESIUMRUNTIME_API UCesiumSubLevelComponent : public UActorComponent {
+  GENERATED_BODY()
+
+public:
+  /** @copydoc UCesiumSubLevelComponent::Enabled */
```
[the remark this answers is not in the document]
kring: These shouldn't have shown up in Blueprints at all. The only reason they did is because I tagged them with `BlueprintPure` and `BlueprintCallable` instead of `BlueprintGetter` and `BlueprintSetter`. Blueprints users should directly access the variables, and it will automatically call through these functions.

So these `@copydoc` tags are really just for C++. Unfortunately they don't work there, either. I mean, they should work if we were to generate documentation with Doxygen, but Intellisense doesn't know how to follow these. I thought maybe if we used Microsoft's tags instead (`<inheritdoc cref="Enabled" />` I believe) it would work, but that doesn't seem to help.

So... as much as it pains me to do so, I copied the doc from the properties to the accessors.

I also noticed while I was doing that that I had somehow connected `OriginLatitude` to `GetOriginLongitude` and `SetOriginLongitude`! 😱

Source/CesiumRuntime/Private/Tests/SubLevels.spec.cpp
kring: It's not strictly necessary, but it's a slighter better test this way. Basically this means:

1. Set the sub-level to be active, and then wait a frame so it actually becomes active.
2. Set the sub-level to be inactive, and wait a frame so it actually becomes inactive.
3. Make sure that the previously-active (but now no longer active) level isn't affected by the georeference origin change.

If we didn't activate it first, we'd instead be testing if a sub-level that was never active is affected by a georeference origin change. And that's somewhat less likely to ever be a problem.

kring: I added some comments to make this clearer.

Source/CesiumRuntime/Private/Tests/SubLevels.spec.cpp
kring: Looks like it's failing only in UE 5.2 maybe? I'll investigate.

kring:
The test failures should be fixed, and I believe I've also addressed all the other feedback. Thanks for the review @j9liu!

Source/CesiumRuntime/Public/CesiumSubLevelComponent.h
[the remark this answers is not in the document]
j9liu: >I also noticed while I was doing that that I had somehow connected OriginLatitude to GetOriginLongitude and SetOriginLongitude! 😱

Ahh I missed that! The words look too similar to each other for their own good 😅

j9liu:
Looks great @kring ! I'll just wait for CI to pass before merging -- wanted to see if the commit in main would fix the Android 5.1 test failure.

brybalicious:
Is this workflow currently usable - in preparation for when v2 milestone is hit?

kring:
Yep. You can try a build of this branch from our CI system, just like you can with any branch. Now that this has been merged into main, look for artifacts from that branch. Or perhaps look at one of the open PRs that build on this. In any case, navigate to the Summary page of the GitHub Actions output and download the appropriate ZIP for your UE version.

We're planning to do a more formal preview before v2.0, too.

brybalicious:
@kring am I misunderstanding? I have confirmed that my plugin version is greater than the release version which the updates have been rolled into. However, I can not find the Cesium Sub Level Component. Are the Cesium plugins automatically updated by Epic Games Launcher not inclusive of these changes? Thanks

kring:
I'm confused about your question. The Epic Games Launcher only has access to final, production versions of Cesium for Unreal. These changes are currently only available in a pre-release form by grabbing a build out of our CI system, as I explained above. You can't infer much from the plugin version, as we only update it for official releases.

brybalicious:
Got it. Downloading an appropriate version from here then - [https://github.com/CesiumGS/cesium-unreal/actions/runs/5466018284](https://github.com/CesiumGS/cesium-unreal/actions/runs/5466018284)

brybalicious:
Looks like this may even be a newer build? - https://github.com/CesiumGS/cesium-unreal/actions/runs/5490194701

PR #1156, opened 2023-07-01
Add packaging tests to CI

csciguy8:
First iteration for packaging tests running after our combined artifacts. Runs for all our Unreal Engine versions: 5.0, 5.1, 5.2. Windows only.

Notes:
* Packaging only compiles the exe and skips the cook + archive steps. This is for speed. The full cook is about 2X slower
* We can turn on full cook by changing the UNREAL_COOK_PARAM var, if needed in the future
* Speed up testWindows job by using sparse checkouts. Saves 1-2 mins per job
* Refresh TestProject to be a skeleton project based on our Cesium Quick Starts (just a Cesium Sun Sky), rather than the Unreal OpenWorld template. In the INI, use the MapsToCook var to explicitly only cook our one map
* Misc cleanup for the buildWindows and testWindows jobs


cesium-concierge:

csciguy8:
@kring, to follow up on speed concerns from an early review... in the end, I solved this by using a `cookonthefly` parameter to skip the cooking step (shader compilation). This delays cooking until after the game has been deployed. [link](https://docs.unrealengine.com/4.27/en-US/SharingAndReleasing/Deployment/BuildOperations/#:~:text=When%20you%20choose%20to%20cook,as%20the%20content%20is%20needed.). 

I originally saw packaging tests times of 45 mins or more, and using this option reduced time by ~50%, depending on the Unreal version. From what I see of the total build times, it seems reasonable. We're parallelizing where we can, but testing the final artifact inevitably has to wait until it gets built. 

I tried various ways to prepopulate caches based on what I saw Unreal doing internally, but eventually gave up. This direction was a lot of guess and check, with 5.0 actually working at some point, but 5.1 and 5.2 inexplicably not. There's no guarantee my mucking about with caches would even be a good long term solution.

Another solution might have been to at least display the final artifacts in the github UI before the all the jobs complete. But alas, this has been a known problem and hasn't been fixed for a few years. [link](https://github.com/actions/upload-artifact/issues/53)

Overall, the packaging tests seem to have value. They don't do everything we want for the first pass, but they can grab our final output artifact, check the existing of a plugin, and compile the code for a project using Cesium.

kring:
Thanks @csciguy8!

PR #1157, opened 2023-07-03
Fix incorrect flat normals.

kring:
After #1126, generated flat normals are very wrong. This can be seen in Denver when set to ignore the unlit extension, for example.


cesium-concierge:

PR #1161, opened 2023-07-05
Remove Android path hack after upgrade to UE 5.1.1.

kring:
Our UE 5.1 build machine image previously used UE 5.1.0, and it has an apparent bug when doing Android builds. Some details here:
https://forums.unrealengine.com/t/5-1-missing-precompiled-manifest-error-when-building-code-plugin-for-android/691983

We had some hacky code in the build process to work around this. But after upgrading the build machine to use UE 5.1.1, the hack no longer works and should also no longer be necessary. So this PR removes it.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,22 +12,22 @@ jobs:
           npm install
           npm run format -- --dry-run -Werror
   Windows50:
-    uses: ./.github/workflows/buildWindows.yml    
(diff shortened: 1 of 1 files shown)


cesium-concierge:

kring:
I'm merging this because it fixes `ue5-main`. It broke as a result of a build server image upgrade, not a code change.

PR #1164, opened 2023-07-07
More improvements to the sub-level system

### LINKED ISSUE #273: Credits missing, duplicate Georeference, when a sub-level needs to have georeferenced objects
Normally, georeferenced objects are put in the PersistentLevel, while local, non-georeferenced objects are placed in sub-levels. But it can sometimes be useful to put georeferenced objects in a sub-level too, so that they're shown/hidden with the sub-level. But when we do that, it doesn't appear to be possible to reference the Georeference or CreditSystem actors in the PersistentLevel, so we end up creating new ones within the sub-level.

This is less than ideal for the Georeference because it usually means we have to make sure they stay in sync, and it's a little weird to have two things doing origin rebasing. But it's more of an immediate problem for the CreditSystem because the UI only attaches to one CreditSystem, meaning some of the credits are not shown.

We should see if there's a way to find/use/reference PersistentLevel objects from sub-levels. And, if not, then the UI should at least be smarter about where it obtains credits.

### LINKED ISSUE #1069: 04_MAIN_CesiumSublevels sample ends up with multiple credit systems
Open `04_MAIN_CesiumSublevels` in [cesium-unreal-samples](https://github.com/CesiumGS/cesium-unreal-samples). Start Play-in-Editor mode. Press 3 to fly to Paris. The credits display in the bottom left corner will look funny:

image

And, depending on where you click the Data Attribution link, the credits for Cesium World Terrain and Bing Maps will be missing:

image

This is almost certainly caused by both the persistent level and the sub-level having a credit system:

image

kring:
See the CHANGES.md update for a summary of what's new in this PR.

Fixes #273
Fixes #1069

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,14 @@
 
 - The "Place Georeference Origin Here" action on CesiumGeoreference is now undoable.
 - Cesium Actors now have the "Is Spatially Loaded" flag disabled by default. When using World Partition, this is essential for some, such as `CesiumGeoreference`.
+- The `CesiumCameraManager` instance to use with a `Cesium3DTileset` can now be specified with a property on the tileset. In addition to offering more flexibility, this avoids the work of finding the camera manager in the level every frame.
+- Cesium Actors created with the Quick Add or Cesium ion panels are now created inside the active sub-level, if there is one.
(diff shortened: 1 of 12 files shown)


cesium-concierge:

j9liu:
Looks great @kring ! One of the jobs in CI was queued for 7 hours, so I restarted it and will merge if / when it passes.

PR #1168, opened 2023-07-12
Account for y-inversion when rendering attenuated points

j9liu:
#1126 broke point cloud shading -- when attenuation is enabled, the point cloud disappears (though the outline is visible in the editor). The y-inversion matrix introduced in that change caused the quads to face the wrong way, so they wouldn't show up with back-face culling. You can confirm this by enabling "Two - Sided" on the Cesium base material; the point cloud does show up, although really dark because we're looking at the "back side" of all the quads.

This PR inverts the Y-offsets of the points in the shader, such that they will be corrected by the y-inversion matrix.


cesium-concierge:

kring:
Thanks @j9liu!

PR #1174, opened 2023-07-19
Allow CesiumGeoreference to be a parent and its transform affects all nested globe objects

### LINKED ISSUE #527: SkyAtmosphere sometimes hides earth's surface

### LINKED ISSUE #1049: Add a way to translate/rotate/scale the entire globe

kring:
From the changelog:

##### Breaking Changes

- `CesiumSunSky` now uses a default `TransmittanceMinLightElevationAngle` value on its `SkyAtmosphere` component of 90.0 degrees instead of -90.0 degrees. This will generally improve lighting when far from the CesiumGeoreference origin, but it is a breaking change because it may change the lighting conditions in existing levels, particularly at sunrise and sunset.
- The `Mobility` property on `ACesium3DTileset` is now obsolete. Instead, use the normal mechanism of setting the root component's mobility.

##### Additions

- `ACesiumGeoreference` can now act as a parent Actor. By adjusting the georeference's transformation, the entire globe can be located, rotated, and scaled within the Unreal Engine world.
- Added `AtmosphereHeight`, `AerialPerspectiveViewDistanceScale`, `RayleighExponentialDistribution`, and `MieExponentialDistribution` properties to `ACesiumSunSky`. These have the same function as the properties of the same name on Unreal's built-in SkyAtmosphere component, except that they automatically respond to the scale of the globe.

##### Fixes

- `ACesiumCartographicPolygon` now operates on the parts of the tileset that are shown in the Editor viewport, even if it is used with a Cesium3DTileset with a non-identity transformation.

Fixes #527 
Fixes #1049


cesium-concierge:

csciguy8:
@kring Was playing with scaling a parent CesiumGeoreference and could trigger an exception. Just set the scale to 0,0,0.

Exception Grab

Source/CesiumRuntime/Private/CesiumSunSky.cpp
```diff
@@ -254,6 +260,39 @@ void ACesiumSunSky::Tick(float DeltaSeconds) {
   if (this->UpdateAtmosphereAtRuntime) {
     this->UpdateAtmosphereRadius();
   }
+
+  if (IsValid(this->SkyAtmosphere)) {
+    double scale = this->_computeScale();
```
csciguy8, line 265: Should we clamp to a minimum sensible value here? 
(Ex. Actor scale gets set to 0, we eventually divide by that in this function)

csciguy8 (commented):
Looks like lots of good changes. The only general suggestion I have would be to add unit tests for checking the nested objects behavior in the typical cases that we expect the parent transform to change.

kring:
> @kring Sounds like a good change, but I still see the Mobility property for Cesium3DTileset in Unreal's Details Window.

You're talking about this?

image

That's UE's standard built-in mobility selector that our custom one has been removed in favor of. Definitely intentional for that one to exist. If there's another one, though, it's a problem (but I don't see it).

Source/CesiumRuntime/Public/CesiumGeoreference.h
```diff
@@ -403,9 +450,9 @@ class CESIUMRUNTIME_API ACesiumGeoreference : public AActor {
...
   /**
    * @brief Computes the normal of the plane tangent to the surface of the
-   * ellipsoid that is used by this instance, at the provided position.
+   * ellipsoid that is used by this instance, at the provided position ECEF.
    *
-   * @param position The cartesian position for which to to determine the
+   * @param position The ECEF position for which to to determine the
    * surface normal.
    * @return The normal.
    */
```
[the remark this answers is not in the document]
kring: I made a whole bunch of changes like the one you suggest - and also removed this particular function entirely - in #1179. So I'm not going to make this change here because it will only create a conflict.

Source/CesiumRuntime/Private/CesiumSunSky.cpp
kring: Hmm not sure where that came from, but it shouldn't be needed. There's a similar check in `Tick`, so no need to do anything in particular in `BeginPlay`. Removed.

Source/CesiumRuntime/Private/CesiumPolygonRasterOverlay.cpp
```diff
@@ -18,6 +19,12 @@ UCesiumPolygonRasterOverlay::UCesiumPolygonRasterOverlay()
 std::unique_ptr<Cesium3DTilesSelection::RasterOverlay>
 UCesiumPolygonRasterOverlay::CreateOverlay(
     const Cesium3DTilesSelection::RasterOverlayOptions& options) {
+  ACesium3DTileset* pTileset = this->GetOwner<ACesium3DTileset>();
+  if (pTileset == nullptr)
```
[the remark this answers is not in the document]
kring: Yeah ok, I like it!

kring:
> The only general suggestion I have would be to add unit tests for checking the nested objects behavior in the typical cases that we expect the parent transform to change.

Yeah I agree it needs more tests. Probably in one of the downstream PRs rather than this one, though. Perhaps we could merge all these PRs into a staging branch like Janine did with the metadata changes?

Source/CesiumRuntime/Public/CesiumGeoreference.h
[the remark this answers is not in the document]
csciguy8: Sounds good

Source/CesiumRuntime/Private/Cesium3DTileset.cpp
```diff
@@ -2024,6 +2024,13 @@ void ACesium3DTileset::Tick(float DeltaTime) {
   glm::dmat4 unrealWorldToCesiumTileset =
       glm::affineInverse(ueTilesetToUeWorld * cesiumTilesetToUeTileset);
 
+  if (glm::isnan(unrealWorldToCesiumTileset[3].x) ||
+      glm::isnan(unrealWorldToCesiumTileset[3].y) ||
+      glm::isnan(unrealWorldToCesiumTileset[3].z)) {
+    // Probably caused by a zero scale.
+    return;
```
csciguy8, line 2031: This takes care of the crash, but we can also validate the scale that causes this a few lines up (line 2020). Maybe even log an error? Might not be obvious to the user why a tileset isn't visible.

csciguy8:
> > The only general suggestion I have would be to add unit tests for checking the nested objects behavior in the typical cases that we expect the parent transform to change.
> 
> Yeah I agree it needs more tests. Probably in one of the downstream PRs rather than this one, though. Perhaps we could merge all these PRs into a staging branch like Janine did with the metadata changes?

Either way works for me. I see the new Georeference test in a downstream PR #1179, thanks.

Source/CesiumRuntime/Private/CesiumSunSky.cpp
csciguy8: Did you make a change to this logic? I saw you removed the blank if statement, but this double on line 265 is still around, and can still be bogus if the actor scale is bogus.

kring: Ah yeah thanks for keeping on me. My comment above was meant to go on the "blank if statement". I somehow skipped over this one.

I think this is ok, though. If we set the scale to zero, the atmosphere height and other things will become zero, but that's arguably correct and it doesn't seem to do any harm. The alternative would be to impose an arbitrary limit on the scale. We've had to do that elsewhere, but it's better not to do that when we can avoid it (as we can here).

Source/CesiumRuntime/Private/Cesium3DTileset.cpp
kring: Logging an error here would spam the output log, because it would happen every frame. We could avoid that, but I'm not sure it's worth the trouble. Hopefully it's not surprising to users that a zero scale means the thing doesn't show up at all?

kring:
Regarding a staging branch: I think it's ok to use ue5-main that way, rather than creating a separate one. Especially since there are already related changes in there. So merge this whenever you're ready!

Source/CesiumRuntime/Private/Cesium3DTileset.cpp
csciguy8: Good point, I didn't think of that. To take it even further, a user might actually be using zero scale to intentionally hide it. If an animation shrinks scale to nothing for example.

Source/CesiumRuntime/Private/CesiumSunSky.cpp
csciguy8: Fair enough, thanks for the explanation.

csciguy8:
Sounds good to me. Nice PRs. Merging this to main..

PR #1188, opened 2023-08-02
Deprecate GeoTransforms class

kring:
This is PR into #1179 so merge that first.
It also depends on CesiumGS/cesium-native#696 so merge that first, too.

Between Unreal itself having double-precision vector and matrix types, and new capabilities in cesium-native (like `LocalHorizontalCoordinateSystem`), the `GeoTransforms` class is no longer needed.

This PR marks it deprecated and switches all of our uses of it to an alternative. I was really happy to see that this ends up deleting more code than it adds.

I didn't switch UCesiumGlobeAnchorComponent to use cesium-native's GlobeAnchor yet, so there's a little duplication because of that at the moment. That will be a separate PR.

### DIFF
--- a/Source/CesiumRuntime/Private/Cesium3DTileset.cpp
+++ b/Source/CesiumRuntime/Private/Cesium3DTileset.cpp
@@ -20,6 +20,7 @@
 #include "CesiumCustomVersion.h"
 #include "CesiumGeospatial/Cartographic.h"
 #include "CesiumGeospatial/Ellipsoid.h"
+#include "CesiumGeospatial/GlobeTransforms.h"
 #include "CesiumGltf/ImageCesium.h"
 #include "CesiumGltf/Ktx2TranscodeTargets.h"
 #include "CesiumGltfComponent.h"
@@ -474,13 +475,10 @@ void ACesium3DTileset::OnFocusEditorViewportOnThis() {
       *this->GetName());
 
   struct CalculateECEFCameraPosition {
-
-    const GeoTransforms& localGeoTransforms;
-
     glm::dvec3 operator()(const CesiumGeometry::BoundingSphere& sphere) {
       const glm::dvec3& center = sphere.getCenter();
       glm::dmat4 ENU =
-          glm::dmat4(localGeoTransforms.ComputeEastNorthUpToEcef(center));
+          CesiumGeospatial::GlobeTransforms::eastNorthUpToFixedFrame(center);
       glm::dvec3 offset =
           sphere.getRadius() *
           glm::normalize(
@@ -493,7 +491,7 @@ void ACesium3DTileset::OnFocusEditorViewportOnThis() {
     operator()(const CesiumGeometry::OrientedBoundingBox& orientedBoundingBox) {
       const glm::dvec3& center = orientedBoundingBox.getCenter();
       glm::dmat4 ENU =
-          glm::dmat4(localGeoTransforms.ComputeEastNorthUpToEcef(center));
+          CesiumGeospatial::GlobeTransforms::eastNorthUpToFixedFrame(center);
       const glm::dmat3& halfAxes = orientedBoundingBox.getHalfAxes();
       glm::dvec3 offset =
           glm::length(halfAxes[0] + halfAxes[1] + halfAxes[2]) *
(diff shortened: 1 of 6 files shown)


cesium-concierge:

Source/CesiumRuntime/Private/CesiumGeoreference.cpp
```diff
@@ -307,65 +323,31 @@ void ACesiumGeoreference::PlaceGeoreferenceOriginHere() {
...
-      FMatrix(
-          FVector(cameraFront.x, cameraFront.y, cameraFront.z),
-          FVector(cameraRight.x, cameraRight.y, cameraRight.z),
-          FVector(cameraUp.x, cameraUp.y, cameraUp.z),
-          FVector::ZeroVector)
+      FMatrix(cameraFront, cameraRight, cameraUp, FVector::ZeroVector)
           .Rotator());
-  pEditorViewportClient->SetViewLocation(
-      FVector(-originLocation.X, -originLocation.Y, -originLocation.Z));
+  pEditorViewportClient->SetViewLocation(FVector::ZeroVector);
```
csciguy8, line 350: This is so much simpler. Tried it out in the editor and works great.

Source/CesiumRuntime/Public/CesiumGeoreference.h
```diff
@@ -707,17 +709,13 @@ class CESIUMRUNTIME_API ACesiumGeoreference : public AActor {
    */
   static FName DEFAULT_GEOREFERENCE_TAG;
 
-  GeoTransforms _geoTransforms;
+  mutable GeoTransforms _geoTransforms;
```
csciguy8, line 712: Any chance we can get rid of this member completely?

Looks like it's created in `GetGeoTransforms`, but isn't updated anywhere else. We could just return a fresh copy every time that function is called.

csciguy8 (commented):
Nice changes. Appreciate seeing more - than + in the diff.

Traced through the refactored code, and loaded some samples in the editor to verify, when appropriate. Everything looks to be behaving correcting.

Just a handful of small comments and questions.

Source/CesiumRuntime/Private/Cesium3DTileset.cpp
```diff
@@ -474,13 +475,10 @@ void ACesium3DTileset::OnFocusEditorViewportOnThis() {
       *this->GetName());
 
   struct CalculateECEFCameraPosition {
```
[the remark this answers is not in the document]
kring: I don't think what you proposed will work with std::visit. I couldn't get it to work, at least. VC++ reports:

```
error C2672: 'std::visit': no matching overloaded function found
note: could be 'unknown-type std::visit(_Callable &&,_Variants &&...)'
note: 'unknown-type std::visit(_Callable &&,_Variants &&...)': could not deduce template argument for '_Callable'
```

I think the problem is that when we write `std::visit(CalculateECEFCameraPosition, boundingVolume)`, the compiler needs to be able to resolve `CalculateECEFCameraPosition` to a particular overload of the function. But that's not possible here because we need it to consider all the overloads when dispatching at runtime, not pick one at the call site.

The cppreference.com [page](https://en.cppreference.com/w/cpp/utility/variant/visit) isn't terribly easy to read, but I think it's presenting at the bottom an `overloaded` template that can be used to resolve this. But it's just a struct again. In C++20 it's apparently provided out of the box, fwiw.

In any case, I don't think there's anything inherently dodgy about a struct with no fields, so I'd rather just leave this as-is.

Source/CesiumRuntime/Public/CesiumGeoreference.h
kring: I think I originally avoided this because it's a breaking change to return GeoTransforms by value instead of const reference. But it's not a very serious one in the context of 2.0, so I did I as you suggested.

csciguy8: Ok, I didn't realize that. But sounds like a minor breaking change, and another nudge to stop using this.

Source/CesiumRuntime/Private/Cesium3DTileset.cpp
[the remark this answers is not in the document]
csciguy8: Thanks for the explanation. I haven't personally implemented the visitor pattern using std variants and visit, so I did brush up on it little [here](https://www.cppstories.com/2018/06/variant/). This overload pattern look legitimate, if not really to my taste, from a language perspective. Looks like there are more instances of it in cesium-native as well.

csciguy8:
Looks great @kring . Will merge once the checks finish.

PR #1190, opened 2023-08-11
Rework `UCesiumEncodedMetadataComponent` -> `UCesiumFeaturesMetadataComponent`

j9liu:
Part of the large metadata system overhaul. This PR does the following:

- Deprecates `UCesiumEncodedMetadataComponent`.
  - Includes backward compatibility, so styling should still work.
  - Removed "Auto Fill" and "Generate Material" functions, to encourage people to upgrade to the new component.
- Adds functions for handling different methods of encoding in `CesiumEncodedMetadataConversions.h`
- Deprecates internal namespace `CesiumEncodedMetadataUtility`. Use `CesiumEncodedFeaturesMetadata` instead.
  - Incorporated ability to parse RGB colors from string from [`metadata-colors`](https://github.com/CesiumGS/cesium-unreal/tree/metadata-colors) (delete that branch after this change?) 
- Adds support for encoding implicit feature IDs (passes them in a texcoord like other feature ID attributes) 
- Adds `UCesiumFeaturesMetadataComponent`.
  - Now displays descriptions of feature ID sets, separate from property tables
  - Property table properties now display their _actual_ type info via `FCesiumMetadataPropertyDetails`. This way, unsupported properties can show up and users can decide if want to try encoding anyways.
  - Encoding info is displayed /  changed via `FCesiumMetadataEncodingDetails`.
    - This allows the user to specify how a property should attempt to be encoded, e.g. "Coerce" or "Parse Color From String".
    - This can be expanded in C++ for custom encoding functionality.
  - "Generate Material" now generates nodes for feature ID sets without requiring them to be connected to property tables
  - "Generate Material" detects when trying to change an asset with unsaved changes, to avoid a crash
- Updates the upgrade guide for most of the changes

Example styling on San Francisco Ferry Building, using "Parse Color from String" as encoding method:

image

This PR does NOT address the following:
- Property textures
- ~Normalized properties (will leave comment)~
- ~Null feature IDs -- will result in undefined property access in materials, e.g. random property colors will show up~
- ~Feature IDs aren't working for point cloud attenuation because the vertex factory isn't passing them in. It seems like the texcoord-related parameters are never binding, but I don't understand why :/~


cesium-concierge:

Source/CesiumRuntime/Private/CesiumEncodedMetadataUtility.cpp
```diff
@@ -71,232 +75,239 @@ EncodedPixelFormat getPixelFormat(
...
+    FTexture2DMipMap* pMip = new FTexture2DMipMap();
+    encodedProperty.pTexture->pTextureData->Mips.Add(pMip);
+    pMip->SizeX = ceilSqrtFeatureCount;
+    pMip->SizeY = ceilSqrtFeatureCount;
+    pMip->BulkData.Lock(LOCK_READ_WRITE);
+
+    void* pTextureData = pMip->BulkData.Realloc(
+        ceilSqrtFeatureCount * ceilSqrtFeatureCount * encodedFormat.pixelSize);
+
+    if (isArray) {
```
j9liu, line 233: Correct me if I'm wrong, but I don't think this code ever accounted for normalization. So if there was a normalized uint property 0-255, it would just get encoded as floats 0.0f-255.0f, and it wouldn't actually denormalize the values.

I think that's the case, and if so, I'll add the correct handling of this in a follow-up PR

Source/CesiumRuntime/Private/CesiumEncodedMetadataConversions.h
```diff
@@ -0,0 +1,109 @@
...
+   * @param propertyDescription The property table property description.
+   * @param property The property table property itself.
+   * @param pTextureData A pointer to the texture data, which will be filled
+   * during encoding.
+   * @param pixelSize The size of a pixel from the given texture, in bytes.
+   */
+  static void encode(
+      const FCesiumPropertyTablePropertyDescription& propertyDescription,
+      const FCesiumPropertyTableProperty& property,
+      void* pTextureData,
```
j9liu, line 74: I wanted this function to encode to the texture directly, without creating an intermediate data structure. But I don't know the best way to do this 😬 please let me know if this is atrocious!

j9liu:
Got feature IDs in point clouds to work! This is what Montreal Point Cloud looks like with the styling borrowed from CesiumJS. This uses implicit feature IDs.

image

j9liu:
Just added support for including null feature IDs in the generated material. The value will appear in the description of the feature ID set.

image

Then, an if statement will be generated with the null feature ID value. It can then be used to do special behavior for "no feature" parts of the model. For example:

image

j9liu:
### 09/14 Update

 I incorporated the changes from https://github.com/CesiumGS/cesium-native/pull/713, which involved refactoring `FCesiumPropertyTableProperty` to handle both normalized and non-normalized property views.

This also allows metadata picking and styling to work with `offset`, `scale`, `noData`, and `default` values from `EXT_structural_metadata`. I'm attaching the tileset I used for testing, which is modified from the one in CesiumJS's spec data.

[TestStructuralMetadata.zip](https://github.com/CesiumGS/cesium-unreal/files/12613418/TestStructuralMetadata.zip)

In this tileset, there is a `temperatureCelsius` property that is normalized, and has `scale` defined in the class property. This is overridden by the `scale` defined in the property table property, which also defines an `offset`. So this results in the following material nodes:

image

The offset of 18 and scale of 6 means values are in the range 18-24. If you remap them to a color gradient you can see something like this:

image

The `employeeCount` property has a `noData` value specified, which results in these nodes. It's a bit hard to read visually, but essentially the value is being compared to the `noData` value. The user must define what happens when it equals the `noData` value, since no default value is defined for the property.

image

I also incorporated the changes from https://github.com/CesiumGS/cesium-native/pull/726, which allows omitted properties to be treated as special valid properties, such that their default value can be accessed. In the same example tileset, the `year` property has a `default` value specified, and it has absolutely no data in the property table. The property will still be visible while metadata picking, and can also be passed to the material as a parameter.

Set the encoded details on the property to be "Scalar", "Float", and "Coerce". This results in these nodes:

image

If the property actually encodes a texture with values, they will be used. Otherwise, this falls back on the default value parameter.

@kring This is ready for a look when you get the chance! My changes made the PR bigger, sorry 😅

Source/CesiumRuntime/Private/CesiumEncodedMetadataConversions.h
kring: Is this the raw pointer you were asking about in our call today? If so, I think the only thing I might suggest improving about it is to be clearer about the required size of the `pTextureData` buffer. So documentation, mostly. But it might be better yet to pass a `gsl::span<std::byte>` rather than a `void*`. Then you can throw if the buffer isn't big enough. That will terminate the process if exceptions are disabled, but that's much better than corrupting memory.

kring:
Just dumping some thoughts / experiences as I run into them...

## Material Layer Names

I had a lot of trouble getting styling to work at first. I eventually figured out the problem: I forgot to rename the material layer from `Metadata` to `FeaturesMetadata`. 🤦 I don't think there's any action there, it just reinforces the fact that good documentation - perhaps with a troubleshooting section - will be essential!

## Offset / Scale

I noticed that when "Has Offset" and "Has Scale" are ticked, extra nodes are generated in the material to handle the offset and scale. The offset and scale values themselves presumably come from the ClassProperty, which makes sense. But I expected I would be able to tick those boxes and _still_ have a working material even if my ClassProperty had no offset and scale. It would just use 0 and 1, respectively. But that doesn't seem to be the case; the material stops working. Is it possible / reasonable to to make this work? It _sounds_ like it might be easy, and it would help for cases where some tiles have an offset/scale and others do not. But if it's difficult it probably isn't a big deal. Same with "Has No Data Value" and "Has Default Value"? I guess you'd need to use NaN or something for those, though, so maybe it's a bad idea.

## A Feature ID Set is required

When I read this in the upgrade guide:

> Everything that is listed in the descriptions will be encoded, and anything not listed will be skipped. As such, only include feature ID sets and metadata that will be actually used by the application. Otherwise, the unused properties may affect memory and performance.

I initially thought that if I didn't need to use the feature ID directly, I could remove the feature ID set. But that breaks the generated material and makes it impossible to access the property table data. It makes sense that a feature ID is required, so maybe just a doc tweak needed here to make it more obvious?

kring:
## Undo

I can get a crash like this:

1. Open a version of the `06_CesiumMetadata` level that I have previously converted to use the new system.
2. Click on the CesiumFeaturesMetadata component attached to the New York City 3D Buildings tileset.
3. Press Generate Material. The material editor opens.
4. Without closing the material editor, click the Auto Fill button.
5. Press Ctrl-Z to undo the Auto Fill. The Editor crashes.

image

```
	UnrealEditor-MaterialEditor.dll!FMatExpressionPreview::FMatExpressionPreview(UMaterialExpression * InExpression) Line 70	C++
 	UnrealEditor-MaterialEditor.dll!FMaterialEditor::GetExpressionPreview(UMaterialExpression * MaterialExpression, bool & bNewlyCreated) Line 6306	C++
 	UnrealEditor-MaterialEditor.dll!FMaterialEditor::RefreshExpressionPreviews(bool bForceRefreshAll) Line 6244	C++
 	UnrealEditor-MaterialEditor.dll!FMaterialEditor::PostUndo(bool bSuccess) Line 6039	C++
 	UnrealEditor-UnrealEd.dll!UEditorEngine::BroadcastPostUndoRedo(const FTransactionContext & UndoContext, bool bWasUndo) Line 5392	C++
 	UnrealEditor-UnrealEd.dll!UEditorEngine::HandleTransactorRedoUndo(const FTransactionContext & TransactionContext, bool Succeeded, bool WasUndo) Line 1209	C++
 	[Inline Frame] UnrealEditor-UnrealEd.dll!Invoke(void(UEditorEngine::*)(const FTransactionContext &, bool)) Line 66	C++
 	[Inline Frame] UnrealEditor-UnrealEd.dll!UE::Core::Private::Tuple::TTupleBase<TIntegerSequence<unsigned int>>::ApplyAfter(void(UEditorEngine::*)(const FTransactionContext &, bool) &) Line 324	C++
 	UnrealEditor-UnrealEd.dll!TBaseUObjectMethodDelegateInstance<0,UEditorEngine,void __cdecl(FTransactionContext const &,bool),FDefaultDelegateUserPolicy>::ExecuteIfSafe(const FTransactionContext & <Params_0>, bool <Params_1>) Line 611	C++
 	[Inline Frame] UnrealEditor-UnrealEd.dll!TMulticastDelegate<void __cdecl(FTransactionContext const &,bool),FDefaultDelegateUserPolicy>::Broadcast(const FTransactionContext &) Line 967	C++
 	UnrealEditor-UnrealEd.dll!UTransBuffer::Undo(bool bCanRedo) Line 1550	C++
 	UnrealEditor-UnrealEd.dll!UEditorEngine::UndoTransaction(bool bCanRedo) Line 1426	C++
 	UnrealEditor-UnrealEd.dll!UEditorEngine::Exec_Transaction(const wchar_t * Str, FOutputDevice & Ar) Line 5376	C++
 	UnrealEditor-UnrealEd.dll!UEditorEngine::Exec(UWorld * InWorld, const wchar_t * Stream, FOutputDevice & Ar) Line 5770	C++
 	UnrealEditor-UnrealEd.dll!UUnrealEdEngine::Exec(UWorld * InWorld, const wchar_t * Stream, FOutputDevice & Ar) Line 659	C++
 	UnrealEditor-MainFrame.dll!FMainFrameActionCallbacks::ExecuteExecCommand(FString Command) Line 547	C++
 	[Inline Frame] UnrealEditor-MainFrame.dll!Invoke(void(*)(FString) &) Line 47	C++
 	[Inline Frame] UnrealEditor-MainFrame.dll!UE::Core::Private::Tuple::TTupleBase<TIntegerSequence<unsigned int,0>,FString>::ApplyAfter(void(*)(FString) &) Line 324	C++
 	UnrealEditor-MainFrame.dll!TBaseStaticDelegateInstance<void __cdecl(void),FDefaultDelegateUserPolicy,FString>::ExecuteIfSafe() Line 731	C++
 	UnrealEditor-Slate.dll!FUICommandList::ConditionalProcessCommandBindings(const FKey Key, bool bCtrl, bool bAlt, bool bShift, bool bCmd, bool bRepeat) Line 217	C++
 	UnrealEditor-Slate.dll!FUICommandList::ProcessCommandBindings(const FKeyEvent & InKeyEvent) Line 161	C++
 	UnrealEditor-MainFrame.dll!FMainFrameActionCallbacks::OnUnhandledKeyDownEvent(const FKeyEvent & InKeyEvent) Line 256	C++
 	[Inline Frame] UnrealEditor-MainFrame.dll!Invoke(FReply(*)(const FKeyEvent &) &) Line 47	C++
 	[Inline Frame] UnrealEditor-MainFrame.dll!UE::Core::Private::Tuple::TTupleBase<TIntegerSequence<unsigned int>>::ApplyAfter(FReply(*)(const FKeyEvent &) &) Line 324	C++
 	UnrealEditor-MainFrame.dll!TBaseStaticDelegateInstance<FReply __cdecl(FKeyEvent const &),FDefaultDelegateUserPolicy>::Execute(const FKeyEvent & <Params_0>) Line 721	C++
 	[Inline Frame] UnrealEditor-Slate.dll!TDelegate<FReply __cdecl(FKeyEvent const &),FDefaultDelegateUserPolicy>::Execute(const FKeyEvent &) Line 588	C++
 	UnrealEditor-Slate.dll!FSlateApplication::ProcessKeyDownEvent(const FKeyEvent & InKeyEvent) Line 4473	C++
 	UnrealEditor-Slate.dll!FSlateApplication::OnKeyDown(const int KeyCode, const unsigned int CharacterCode, const bool IsRepeat) Line 4360	C++
```

kring:
Assertion failure here. Probably lucky it wasn't a crash?

image

```
Assertion failed: (Index >= 0) & (Index < ArrayNum) [File:C:\Program Files\Epic Games\UE_5.0\Engine\Source\Runtime\Core\Public\Containers\Array.h] [Line: 691] 
Array index out of bounds: 0 from an array of size 0
A breakpoint instruction (__debugbreak() statement or a similar call) was executed in UnrealEditor-Win64-DebugGame.exe.
```

```
	UnrealEditor-CesiumRuntime-Win64-DebugGame.dll!UCesiumPrimitiveFeaturesBlueprintLibrary::GetFeatureIDFromFace(const FCesiumPrimitiveFeatures & PrimitiveFeatures, __int64 FaceIndex, __int64 FeatureIDSetIndex) Line 138	C++
 	UnrealEditor-CesiumRuntime-Win64-DebugGame.dll!UCesiumPrimitiveFeaturesBlueprintLibrary::execGetFeatureIDFromFace(UObject * Context, FFrame & Stack, void * const Z_Param__Result) Line 82	C++
 	UnrealEditor-CoreUObject.dll!UObject::execCallMathFunction(UObject * Context, FFrame & Stack, void * const Z_Param__Result) Line 945	C++
 	[Inline Frame] UnrealEditor-CoreUObject.dll!FFrame::Step(UObject *) Line 422	C++
 	UnrealEditor-CoreUObject.dll!UObject::execLet(UObject * Context, FFrame & Stack, void * const Z_Param__Result) Line 2656	C++
 	[Inline Frame] UnrealEditor-CoreUObject.dll!FFrame::Step(UObject *) Line 422	C++
```

My blueprint looked like this:

image

And the crash happened when my line trace missed the New York tileset and hit Cesium World Terrain instead.

kring:
I tried to use the fancy new "Parse Color from String" to color Cesium OSM Buildings, but I couldn't quite get it to work. I'm probably missing something small. The closest I got was this:

image

I set up the property encoding like this:

image

The material layer is pretty simple:

image

And it's incorporate into this material instance:

image

There are a bunch of message like this in the log:

> LogCesium: Warning: The value type of the metadata property cesium#color does not match the type specified by the metadata description. It will still attempt to be encoded, but may result in empty or unexpected values.

Which is expected. Not all buildings have a color. Ideally I'd be able to specify a default value when the string can't be parsed as a color. Is this possible?

j9liu:
Thanks @kring for the thorough testing! I've gone through your comments and will address them incrementally.

> I expected I would be able to tick those boxes and still have a working material even if my ClassProperty had no offset and scale. It would just use 0 and 1, respectively. But that doesn't seem to be the case; the material stops working. Is it possible / reasonable to to make this work?

Ah yes -- this was unintentional, and I'm adding a fix for it in my next commit.

> Same with "Has No Data Value" and "Has Default Value"? I guess you'd need to use NaN or something for those, though, so maybe it's a bad idea.

Yeah... the "no data" value and default value are a little trickier, and there's (understandably) no fallback values provided by the spec. My initial thought was just pick the value 0 for both, so the value would be unchanged if both "Has No Data Value" and "Has Default Value" were true. But if only "Has No Data Value" is true, then we have just picked some arbitrary value from the data to be null, which isn't too great. 

Following that, I then thought might be useful to specify custom "no data" or default values under the "Encoding Details". The value could be used as a fallback if the class property didn't provide such values -- or as an override, if desired. This could also be used to address this comment:

> Ideally I'd be able to specify a default value when the string can't be parsed as a color. Is this possible?

But I worry the UI for this may be convoluted. The type of the values is determined by the type of the property -- at least, the type that it is trying to coerce to in the "Encoding Details". So the method of displaying / editing values in the UI would have to change based on that type. I'm not familiar with how to build custom editor UIs in Unreal, so I have no idea how feasible this is. Do you have any insights from the work you incorporated from #646 ?

If a custom UI isn't feasible, one workaround is to have users specify as a string, and have it attempt to parse a scalar / vector value from there. We could also simply document that it is a limitation of the metadata system that users should be aware of if the class definitions differ across a tileset. Any feelings about either option?

j9liu:
The commit I recently pushed does the following:

- Use `gsl::span` instead of `void*` when encoding
- Remove unintentional requirement of Feature ID sets for property tables
- Fix typo that allowed a crash in `GetFeatureIDFromFace`

We talked offline about letting users specify their own "no data" / default values. Ultimately, we'll document the issue as a known limitation for now. While it's not impossible to implement, it's a bit out of the way for now.

>Open a version of the 06_CesiumMetadata level that I have previously converted to use the new system.
Click on the CesiumFeaturesMetadata component attached to the New York City 3D Buildings tileset.
Press Generate Material. The material editor opens.
Without closing the material editor, click the Auto Fill button.
Press Ctrl-Z to undo the Auto Fill. The Editor crashes.

Auto Fill wasn't recording the changes to itself, so I've fixed that. But the crash actually happens whenever you're trying to do undo the "Generate Material" step. This keeps crashing in the "MaterialEditor::PostUndo" step where:

1. if there is a user-added node in the material, it crashes at `GetExpressionPreview(UMaterialExpression * MaterialExpression, bool & bNewlyCreated)` because `InExpression->Material` is null
2. if there is no user-added node in the material, it crashes at `GetTabManager()->TryInvokeTab(FMaterialEditorTabs::GraphEditor);` 

It should be noted that with this window closed, undoing does not crash the application. So there's something with the Editor window that isn't able to handle our material changes via code, but not sure why. I'm trying to compare our code to other Editor code, but it's not obvious what we're missing.

For now, I think I'll focus on finishing the "Upgrade to 2.0" guide and then requesting another review. It's definitely important to fix the crash but given time constraints, maybe we can just list it as a known issue for now?

kring:
> It's definitely important to fix the crash but given time constraints, maybe we can just list it as a known issue for now?

Yep that makes sense to me.

j9liu:
So I tried playing around with `cesium#color` on Cesium OSM buildings today. It looks like this in CesiumJS



> (this is mainly for my personal reference)

In @kring 's screenshots, many buildings are bright white is because the colors are sent to the material as-is, in the range 0-255. Since Unreal colors are meant to be from 0-1, this results in really bright colors. The only reason the other non-white colors are showing up are because they don't use all three color components, so they're able to look red / yellow / green.

So we'd have to divide by 255 in the material before passing the color into the Base Color field. This is *better*, but it still results in really bright colors...



I thought perhaps this was a colorspace issue, and adding an "sRGB to Linear" node definitely desaturates all of the colors. But it looks rather grayscale...



But if you *multiply* this grayscale by the normalized color, it looks better!



This is the material graph that resulted in the above:



Something I also tried at first was dividing by 2550 instead of 255, without the sRGB to Linear / multiply nodes. It looks closer to the CesiumJS colors, but OSM Buildings in CesiumJS seem generally "darker" than Unreal, so I'm not sure if this is necessarily more correct.



In brief, I think that the data is passing accurately to the Unreal material, but there needs to be extra steps to convert it to a linear color. And it's possible, but I'm not sure if what I'm doing is correct.

j9liu:
Also, I tried adding an If node to catch the pure black (0, 0, 0) colors. And if so, use a default color of white. It makes the buildings look a nicer together, but weirdly, it changes a red building on the right to blue!

| No default color | With default color |
| ---------------- | ------------------ |
| image | image |

So I don't understand what's going on there 🤔 This is the material graph for the second screenshot:

image

j9liu:
@kring Now that CI has passed, this is ready for another review!

kring:
I got Cesium OSM Buildings working well!
image

Here's unlit so you can see the colors are correct:
image

It's still doing that thing where the one red building becomes blue. No idea! But the rest look correct.

To get this to work, though, I had to add this weird hack where I add 1 to the feature IDs. It seems they're all off by 1? Unless I'm doing something stupid (entirely possible!) that seems like a bug.

image

The other key bit is that Unreal's sRGBToLinear function only operates on a single float, not a whole vector. If you try to use it on a whole vector, Unreal helpfully grabs the red channel, converts it to linear, and then copies that to all the components. Hence the grayscale. The mind boggles. But anyway, easy fix. Here's the key part of the material:

image

kring:
Adding 1 to the feature ID seems to make it better overall, but there are a few cases where it's correct without the +1, and it becomes wrong with it. So I guess it's more subtle than a simple off-by-one.

kring:
Finally figured it out. Floating point precision. It's always floating point precision.

The old metadata styling system had a custom node with code like this:

```
uint propertyIndex = round(PropertyIndexUV.r);
uint width;
uint height;
Year_array.GetDimensions(width, height);
uint pixelX = propertyIndex % width;
uint pixelY = propertyIndex / width;
```

The new one looks like this:

```
uint _czm_width;
uint _czm_height;
cesium_color_DATA.GetDimensions(_czm_width, _czm_height);
uint _czm_pixelX = FeatureID % _czm_width;
uint _czm_pixelY = FeatureID / _czm_width;
cesium_color = asuint(cesium_color_DATA.Load(int3(_czm_pixelX, _czm_pixelY, 0)).rgb);
return FeatureID;
```

FeatureID is always rounded before it's passed into this function, but it's still a float (even though the value is an integer). When we use `/` and `%` on a float, we're not guaranteed to get integers. And then when we clamp them to integers, we'll sometimes end up with the next lowest integer instead.

I changed it to this, and all is well:

```
uint _czm_width;
uint _czm_height;
cesium_color_DATA.GetDimensions(_czm_width, _czm_height);
uint index = round(FeatureID);
uint _czm_pixelX = index % _czm_width;
uint _czm_pixelY = index / _czm_width;
cesium_color = asuint(cesium_color_DATA.Load(int3(_czm_pixelX, _czm_pixelY, 0)).rgb);
return FeatureID;
```

I also went through and changed all the metadata-related texture creation code to be sure to set the `sRGB` flag to false, because we don't want Unreal to mess with our texture values when it loads them. 

And the red building turning cyan? Yeah that's caused by the `If` node doing a component-wise comparison. It treats the X,Y,Z channels are completely separate and does the comparison on each. So if we have this:

image

Then any component that is zero will be changed to 192. It doesn't require that all of the components be zero. 🙄

kring:
Replace the `if` with a custom node and I think it's finally all working!

```
return all(color == float3(0.0, 0.0, 0.0)) ? float3(192.0, 192.0, 192.0) : color;
```

image

kring:
Well, I certainly can't claim to have reviewed every line of code here, but I've played around with it quite a bit, and I think it looks really, really good! I think you've been extremely thorough (This PR single-handedly increases our test count 10x!). So I'm going to merge this into the staging branch. You should probably take a look at the changes I've made @j9liu to make sure I haven't done anything terrible, and then feel free to merge it to ue5-main whenever you're ready! 🎉🚀:shipit:

j9liu:
Thanks for the thorough reviewing @kring! Don't worry about not poring over the code, there's around 20K changed lines and only so much time :')

Issue #1184, opened 2023-07-31
Add ability to exclude tiles by implementing a C++ or Blueprints interface


kring:
This is the Unreal equivalent of CesiumGS/cesium-unity#248.

In Unreal C++ code, users can implement the cesium-native `ITileExcluder` interface directly. So the first step is to simply allow users to use C++ to specify an `ITileExcluder` that will be passed through to cesium-native.

The next step is to allow this to be done from Blueprints. This will look a bit like the C# wrapper in Cesium for Unity, involving a Blueprints-accesible wrapper for the cesium-native Tile class that provides access to a minimal set of properties, like the tile's bounding box. I suspect that calling into Blueprints for every visited tile on every frame is likely to cause a major performance problem, though. If so, we can solve this with caching. Maintain a map of Tile pointer to exclusion state, and only ask Blueprints for the exclusive state once. Provide a way to invalidate the cache, so that the per-tile exclusion state can be changed when needed.

kring:
Fixed in #1194.

PR #1194, opened 2023-08-24
Implement Unreal Tile Excluder

joseph-kaile:
This pull request fixes issue #1184. It is based on a similar pull request in Cesium For Unity and implements the Unreal equivalent of the Unity version. The implementation includes CesiumTile, CesiumTileExcluder, and CesiumTileExcluderAdapter classes. An adapter was needed because an Unreal UObject/UClass could not inherit from the ITileExcluder, so an outside class was required.

Performance testing showed that caching the results was not necessary for blueprints, so it was not implemented. However, this may be subject to change. The Unreal plugin does not allow RTTI, so some changes were necessary to work around the lack of dynamic casting.

One area for improvement is intersection testing, which currently uses the FSphereBoxBounds class. It would be better to use the IsComponentOverlapping function and convert the tile bounds to an FCollisionShape object to improve accuracy and compatibility with different shapes.


cesium-concierge:

* :x: Missing CLA.

kring:
I enabled Stat -> Engine -> Unit and looked at the "Game" time with a blueprint excluder that always returns false compared to no excluder at all. Adding a do-nothing excluder seems to add less than 0.5 ms to the game time per thread with the Melbourne tileset (it goes from about 6.0ms to about 6.5ms on my system). That's really really good, considering it's invoking a blueprint callback over 300 times per frame! So no need for anything fancy here I'd say, this is working well.

Source/CesiumRuntime/Public/CesiumTileExcluder.h
```diff
@@ -0,0 +1,59 @@
+#pragma once
+
+#include "Cesium3DTileset.h"
```
kring, line 3: I don't think this include is needed.

Source/CesiumRuntime/Private/CesiumTileExcluderAdapter.h
```diff
@@ -0,0 +1,24 @@
+#pragma once
+#include "Cesium3DTileset.h"
+#include "CesiumTile.h"
+#include "CesiumTileExcluder.h"
+#include <Cesium3DTilesSelection/ITileExcluder.h>
+
+class CesiumTileExcluderAdapter : public Cesium3DTilesSelection::ITileExcluder {
```
kring, line 7: I think it should be possible to make `CesiumTileExcluder` implement `ITileExcluder`, rather than using an Adapter. The `shouldInclude` method on `ITileExcluder` obviously won't be callable from Blueprints (and in fact it should be private). But I think moving the implementation of the cesium-native interface into CesiumTileExcluder itself will simply things a bit. For example, the raw points in this class make me nervous, because they'll be left dangling if the corresponding UE objects get garbage collected. I'm not certain that can happen, but I'm also not certain that it _can't_. If we only reference UObjects via UPROPERTYs on `CesiumTileExcluder`, we pretty much don't have to worry about it.

Source/CesiumRuntime/Private/CesiumTile.cpp
```diff
@@ -0,0 +1,81 @@
+#include "CesiumTile.h"
+#include "Components/PrimitiveComponent.h"
+#include "VecMath.h"
+
+namespace {
+struct GetBoundsOperation {
```
kring, line 6: This looks copied from `CalcBounds.cpp`. Can we use that instead of copying it? Perhaps with minor changes.

Source/CesiumRuntime/Private/CesiumTile.cpp
```diff
@@ -0,0 +1,81 @@
...
+};
+} // namespace
+
+FBoxSphereBounds UCesiumTile::GetBounds() const {
+  return std::visit(
+      GetBoundsOperation{_transform},
+      _pTile->getBoundingVolume());
+}
+
+bool UCesiumTile::IsIntersecting(UPrimitiveComponent* Other) const {
```
kring, line 73: I think this function is a little misleading, because it promises to check for intersection with a primitive but then it only checks against the primitives FBoxSphereBounds. I think it would be better to make this function just take the FBoxSphereBounds directly so that it's more honest about what it's doing.

kring (commented):
This looks really good @joseph-kaile! Just a few comments below. I also look forward to seeing the new Samples level to demonstrate it.

kring:
To test this out, I started with the `02_CesiumMelbourne` level in the Samples project.

1. Adde a new Static Mesh Component to the "Melbourne Photogrammetry" actor, named "ExclusiveVolume". Set the Static Mesh to a cube and make it big (set the scale) and move it so it intersects part of Melbourne. Add a tag to it called `ExclusionVolume`.
2. Add a "New Blueprint Script Component" and chose CesiumTileExcluder as the parent class. It should open in the Blueprint Editor.
3. Click the Override button next to "Functions" on the left and choose the `ShouldExclude` function.
4. Add this Blueprint to the Should Exclude function to exclude tiles that do not intersect with the exclusion volume created in step 1.
image
5. Click the Refresh Tileset on Melbourne, because it doesn't seem to pick it up automatically. (can it somehow?)

Source/CesiumRuntime/Private/CesiumTile.cpp
joseph-kaile: I changed it to use OverlapComponent. The reason why I made it input a UPrimitiveComponent is because there doesn't seem to be a way to access FBoxSphereBounds in Blueprints.

Also, looks like OverlapComponent can describe an OrientedBoundingBox, but I wasnt able to figure that out.

joseph-kaile:
@kring , I think this is ready for a second review.

kring:
This is pretty cool:
excluder2

kring:
@joseph-kaile this looks great and is really really close, but I want to take the time to give it a proper review and there will probably be a couple of tweaks from that. So I think there's just not quite enough time before tomorrow's release to do it justice. Plus it's built on the ue5-main branch, which means it will have to be disentangled a bit from some of the 2.0 features if we want to ship it this month. So rather than rushing it out the door tomorrow, I think we should ship it as part of the 2.0 release, including a preview as soon as we can. Does that sound ok?

joseph-kaile:
I finished the sample level using Tile Excluders here https://github.com/CesiumGS/cesium-unreal-samples/pull/47

Source/CesiumRuntime/Public/CesiumTile.h
```diff
@@ -0,0 +1,51 @@
...
+  UFUNCTION(BlueprintCallable, Category = "Cesium")
+  FBoxSphereBounds GetBounds() const;
+
+  /**
+   * Tests whether a primitive component intersects with this tile.
+   * This function provides a convenient way to test for intersection between a
+   * primitive component and this tile.
+   */
+  UFUNCTION(BlueprintCallable, Category = "Cesium")
+  bool IsIntersecting(const UPrimitiveComponent* Other) const;
```
kring, line 40: I still think `IsIntersecting` and `IsContainedBy` are misleading. They imply they're testing the tile against the primitive, but really they're testing a very rough AABB or sphere approximation of the tile against the primitive. Improving the documentation to make this very clear would help, but making it clear from the names of these functions would be even better.

Here's what I'm thinking...

There's already a `GetBounds` function, and it is pretty clear about what it does. So let's make these two functions reference that.

For `IsIntersecting`:

* [ ] Rename it to `TileBoundsOverlapsPrimitive`.
* [ ] Its implementation can actually use GetBounds rather than computing it separately. The trick is to notice that there is an actual overlap only if _both_ the sphere and box overlap with the primitive.
* [ ] Update the docs to make clear that we're testing the tile's bounding box and bounding sphere against the primitive using `OverlapComponent`.

For `IsContainedBy`:

* [ ] Rename it to `PrimitiveBoxFullyContainsTileBounds`.
* [ ] Again we can implement it in terms of GetBounds, but here the tile is fully contained if _either_ the box or sphere is fully contained in the primitive's box.
* [ ] Update the docs to make clear that we're testing the tile's bounding box and bounding sphere against the primitive's axis-aligned bounding box.

This is a lot better because we're at least being clear - both in the documentation and in the naming - about exactly what is being tested. But the `PrimitiveBoxFullyContainsTileBounds` function is still a little strange. We'd really like a function that proves the tile bounds is fully inside the primitive itself, not an AABB around the primitive (which could be a lot bigger and therefore include tiles that aren't really inside the primitive at all). Is there a better way to implement it? I couldn't find one with a quick look around, unfortunately.

kring (commented):
I check out the new Samples level, and it's really good. Nice job! Some review comments below.

joseph-kaile:
This is ready for another review.

Source/CesiumRuntime/Public/CesiumTileExcluder.h
```diff
@@ -0,0 +1,75 @@
...
+    Blueprintable,
+    Abstract)
+class CESIUMRUNTIME_API UCesiumTileExcluder : public UActorComponent {
+  GENERATED_BODY()
+private:
+  CesiumTileExcluderAdapter* pExcluderAdapter;
+
+protected:
+  UPROPERTY()
+  UCesiumTile* CesiumTile;
```
kring, line 28: Seems like this could be private, right? I think a Blueprint client can't use it anyway, because it's not marked BlueprintReadable or BlueprintReadWrite. And a C++ derived class probably has no reason to use it.

Source/CesiumRuntime/Private/CesiumTile.cpp
```diff
@@ -0,0 +1,34 @@
...
+#include "Components/PrimitiveComponent.h"
+#include "VecMath.h"
+
+bool UCesiumTile::TileBoundsOverlapsPrimitive(
+    const UPrimitiveComponent* Other) const {
+  if (IsValid(Other)) {
+    return Bounds.GetBox().Intersect(Other->Bounds.GetBox()) &&
+           Bounds.GetSphere().Intersects(Other->Bounds.GetSphere());
+  } else {
+    return true;
```
kring, line 13: It's more intuitive to me if this returns false, because a tile can't overlap with something that doesn't exist anymore.

Source/CesiumRuntime/Public/CesiumTile.h
```diff
@@ -0,0 +1,45 @@
...
+   * and box comparison. This function provides a convenient way to test for
+   * intersection between a primitive component and this tile.
+   */
+  UFUNCTION(BlueprintCallable, Category = "Cesium")
+  bool TileBoundsOverlapsPrimitive(const UPrimitiveComponent* Other) const;
+
+  /**
+   * Checks if this tile is fully inside the given primitive component using a
+   * sphere and box comparison. It uses the FBox::IsInside function to compare
+   * the the FBox of the component and the tile's bounds.
```
kring, line 35:
```suggestion
   * the FBox of the component and the tile's bounds.
```

kring (commented):
Just a couple more tweaks on this!

kring:
Thanks @joseph-kaile!

PR #1200, opened 2023-08-31
Tweaks to performance testing framework.

kring:
This is a PR into #1198 with some changes I made while debugging CesiumGS/cesium-native#717. I don't think you'll want to merge it as-is, which is why I've marked it draft. More for your consideration @csciguy8.

I'll add notes inline explaining these changes.


cesium-concierge:

Source/CesiumRuntime/CesiumRuntime.Build.cs
```diff
@@ -81,6 +81,8 @@ public CesiumRuntime(ReadOnlyTargetRules Target) : base(Target)
         string[] libs = new string[]
         {
             "async++",
+            "Cesium3DTiles",
+            "Cesium3DTilesReader",
```
kring, line 85: Needed after the cesium-native upgrade that fixes CesiumGS/cesium-native#717.

Source/CesiumRuntime/Private/Cesium3DTileset.cpp
```diff
@@ -890,6 +890,12 @@ void ACesium3DTileset::UpdateLoadStatus() {
 
   this->LoadProgress = this->_pTileset->computeLoadProgress();
 
+  // If we have tiles to hide next frame, we haven't completely finished loading
+  // yet. We need to tick once more
+  if (!this->_tilesToHideNextFrame.empty()) {
+    this->LoadProgress = glm::min(this->LoadProgress, 99.9999f);
+  }
```
kring, line 897: Because tile hides are always delayed by one frame, it was possible before for the performance test to end (and suspend further updates) with tiles of multiple LODs visible simultaneously. This change makes it so that loading isn't 100% complete until these tiles have been hidden.

Source/CesiumRuntime/Private/Cesium3DTileset.cpp
```diff
@@ -2064,6 +2069,8 @@ void ACesium3DTileset::Tick(float DeltaTime) {
       updateTileFade(pTile, false);
     }
   }
+
+  this->UpdateLoadStatus();
```
kring, line 2073: This needs to be moved later, after `_tilesToHideNextFrame` has been set.

Source/CesiumRuntime/Private/Tests/CesiumLoadTest.cpp
```diff
@@ -227,7 +233,9 @@ void createCommonWorldObjects(LoadTestContext& context) {
   context.pawn->AutoPossessPlayer = EAutoReceiveInput::Player0;
 }
 
-bool RunLoadTest(const size_t locationIndex) {
+bool RunLoadTest(
+    std::function<void(LoadTestContext&)> locationSetup,
```
kring, line 237: I thought that passing in a setup function, rather than an index indicating which setup function to use, was much cleaner and less error prone.

Source/CesiumRuntime/Private/Tests/CesiumLoadTest.cpp
```diff
@@ -299,15 +298,93 @@ bool RunLoadTest(const size_t locationIndex) {
         loadElapsedTime);
   }
 
+  if (afterTest) {
+    afterTest(context);
+  }
```
kring, line 303: This `afterTest` thing allows doing additional checks after the performance test is complete. For example, to make sure the test did what you expected. Or in my case, to do an additional test after zooming out.

In retrospect, though, I think maybe it would be better to restructure things further so that these std::functions aren't needed at all. For example, the Denver test could instead look like this:

```cpp
LoadTestContext context;
createCommonWorldObjects(context);
setupForDenver(context);
RunLoadTest(contex);
```

Where `RunLoadTest` starts with "Halt tileset updates and reset them", skipping the context and object creation.

Source/CesiumRuntime/Private/Tests/CesiumLoadTest.cpp
```diff
@@ -299,15 +298,93 @@ bool RunLoadTest(const size_t locationIndex) {
...
+  auto after = [this](LoadTestContext& context) {
+    // Zoom way out
+    FCesiumCamera zoomedOut;
+    zoomedOut.ViewportSize = FVector2D(1024, 768);
+    zoomedOut.Location = FVector(0, 0, 7240000.0);
+    zoomedOut.Rotation = FRotator(-90.0, 0.0, 0.0);
+    zoomedOut.FieldOfViewDegrees = 90;
+    context.setCamera(zoomedOut);
+
+    context.pawn->SetActorLocation(zoomedOut.Location);
```
kring, line 356: It's weird that we need to `setCamera` and also set the DynamicPawn's location. Why do we need the custom camera at all?

Source/CesiumRuntime/Private/Tests/CesiumLoadTest.cpp
```diff
@@ -299,15 +298,93 @@ bool RunLoadTest(const size_t locationIndex) {
...
+        }
+
+        UCesiumGltfComponent* Gltf = static_cast<UCesiumGltfComponent*>(
+            pRenderContent->getRenderResources());
+        if (Gltf && Gltf->IsVisible()) {
+          ++visibleTiles;
+        }
+      });
+
+      TestEqual("visibleTiles", visibleTiles, 1);
```
kring, line 383: This unfortunately fails, and it calls attention to a bit of a problem in the performance testing framework. The problem is that the performance test is executed in an Editor world, which means that tiles are selected using any active editor viewports. That's a bit problem because different editor views will affect the performance, and in this particular case it also causes my test to fail because the editor viewport is zoomed in close and causes additional tiles to be selected. To see what I mean, switch your default Editor viewport to an orthographic view (click Perspective, then Top) and note that the test now passes.

In general, I worry that the way the performance test is executed here is just a little bit artificial, and that will affect our results in unknown ways. I think it would be better to use Play-in-Editor mode and let the Engine itself manage the ticking, by following the model for this that's used in the sublevel tests.

kring:
Some of the comments above are suggestions for improvements, and some are commentary on the original PR. Sorry if that's a bit confusing, let me know if you have any questions @csciguy8.

Source/CesiumRuntime/Private/Tests/CesiumLoadTest.cpp
csciguy8: Agreed. The `RunLoadTest ` part is really the test code we care about. The `setupForXXX` should be clearly separate and reusable. What if we want this setup code for other tests?

Source/CesiumRuntime/Private/Tests/CesiumLoadTest.cpp
csciguy8: Good catch, we don't need it. Setting the pawn's location / rotation should be enough to run the test.

This was leftover setup code when I was playing with creating a world outside of the editor. It was needed this because `ACesium3DTileset` wasn't picking up any cameras in `::GetCameras()`.

Source/CesiumRuntime/Private/Tests/CesiumLoadTest.cpp
csciguy8: Agreed. I never really liked how it just used an active Editor viewport. This needs to be more rigid and predictable.

Committed some changes to start / end PIE mode, which simultaneously
- Uses the additional camera (and viewport) that each location configures
- Fixes your test so that it passes
- Is less artificial. Calls typical in-game functions like ::BeginPlay ::BeginDestroy

Source/CesiumRuntime/Private/Tests/CesiumLoadTest.cpp
csciguy8: Actually, we do need it now. Just committed changes to start / end PIE mode, and in this mode, it does use this additional camera rather than the active viewport camera.

Source/CesiumRuntime/Private/Tests/CesiumLoadTest.cpp
csciguy8: > I think it would be better to use Play-in-Editor mode and let the Engine itself manage the ticking, by following the model for this that's used in the **sublevel tests**.

Looks like the sublevel test uses two mechanisms for hooking into the game loop:

1) Executing tasks at the end of frame
image

2) A special tick call in the main tick loop
image

I haven't hooked into those yet for these reasons
1) I haven't got that far yet. The original test was made as quickly and simply as possible
2) I like the control of determining how often the tick loop executes, what tick times are used, and what is getting ticked
3) The inevitable next steps are to set up repeatable benchmarks and profile problems. If the way we tick doesn't impact either of these, do we need to change it?

But, it might be the case in the future that we MUST hook into the main tick loop for one reason for another. For example, what if we discover we need a test to render to a texture, but nothing is showing up because of this hacky tick logic?

csciguy8: I'm still thinking on this btw. Stay tuned...

Source/CesiumRuntime/Private/Tests/CesiumLoadTest.cpp
```diff
@@ -299,15 +304,93 @@ bool RunLoadTest(const size_t locationIndex) {
...
+
+    context.pawn->SetActorLocation(zoomedOut.Location);
+
+    context.setSuspendUpdate(false);
+    tickWorldUntil(context, 10, breakWhenTilesetsLoaded);
+    context.setSuspendUpdate(true);
+
+    Cesium3DTilesSelection::Tileset* pTileset =
+        context.tilesets[0]->GetTileset();
+    if (TestNotNull("Tileset", pTileset)) {
```
csciguy8, line 370: So were you going the direction of combining a performance test with tests for correctness?
I viewed these as somewhat separate, thus the Cesium.Performance.X namespace.
But even if we make them separate, we should reuse as much code as possible.
(I like this camera reposition test btw)

kring: Well.... yeah, but it wasn't a very carefully considered choice. I was just trying to chase down a bug, and the performance framework gave me an easy way to do that, even if I was abusing it.

I think assertions of correctness _do_ have a place in performance tests, though. Mostly as sanity checks: is that performance test fast because we optimized something awesomely? Or is it fast because we broke something and produces incorrect results (quickly)?

Source/CesiumRuntime/Private/Tests/CesiumLoadTest.cpp
csciguy8: Ok, committed some changes that do a more 'normal' tick loop, using latent commands. Kind of a blend of what Unreal recommends [here](https://docs.unrealengine.com/4.27/en-US/TestingAndOptimization/Automation/TechnicalGuide/), and the test helper functions you use for the sublevels test.

csciguy8:
Caught an interesting bug today. The test seems to have completed, but there was obviously more loading happening after the end mark. This means the reported time is faster than it actually is.

This suggests that `LoadProgress` was calculated as 100, but wasn't really done yet. Was testing Google Tiles, Unreal Insights capture below...

image

csciguy8:
Merging this into the base. It doesn't make sense to review both as separate. Performance tests are a new effort this branch is the better one.

PR #1204, opened 2023-08-31
Add workaround for Linux intermediates on the UE Marketplace.

kring:
The Unreal Marketplace build process for our plugin seems to a) delete our Linux intermediate files, and b) fail to produce new ones itself. As a workaround, the Marketplace team has told us to put our Linux intermediate files into a separate directory so that they can pass through Epic's build process unscathed. 🤷 So that's what this PR does. Users will also need to manually copy the files to the correct place, though, in order for Linux builds to work.


cesium-concierge:

PR #1207, opened 2023-09-07
Make the plugin compile on Unreal Engine 5.3

kring:


cesium-concierge:

kring:
This currently works on Windows, Linux, and Android. I'll add support for Apple platforms when I can, but it can be a separate PR. So this is ready for review.

kring:
macOS and iOS are building now, too. @joseph-kaile would you mind trying out the built 5.3 package from this branch on those platforms and confirm that it works ok?

.github/workflows/build.yml
```diff
@@ -771,3 +771,272 @@ jobs:
...
+          rm -rf extern
+      - name: Build plugin
+        run: |
+          sed -i '' 's/\"EngineVersion\": \"5.0.0\"/\"EngineVersion\": \"5.3.0\"/g' CesiumForUnreal.uplugin
+          export UNREAL_ENGINE_DIR=$HOME/UE_5.3
+          cd $UNREAL_ENGINE_DIR/Engine/Build/BatchFiles
+          ./RunUAT.sh BuildPlugin -Plugin="$GITHUB_WORKSPACE/CesiumForUnreal.uplugin" -Package="$GITHUB_WORKSPACE/packages/CesiumForUnreal" -CreateSubFolder -TargetPlatforms=Mac
+      - name: Publish plugin package artifact
+        if: ${{ success() }}
+        uses: actions/upload-artifact@v3
```
csciguy8, line 927: I checked the size of the macOS artifact, and it's 497MB. The other versions are ~275MB-ish.

You know of any reason for the oddly large 5.3 macOS zip?

[Run link](https://github.com/CesiumGS/cesium-unreal/actions/runs/6152360237)
`[CesiumForUnreal-53-macos-v1.30.1-18-ge6c88d7a`

Config/Engine.ini
```diff
@@ -1,3 +1,10 @@
+# UE 5.3+ defaults this to 11, which is not really high enough.
+# 100 is probably high enough for reasonable values of Cesium3DTileset's 'MaximumSimultaneousTileLoads property,
+# but if you get log messages like "Warning: Reached threaded request limit", and you're sure you want to use
+# such a high value, you may need to increase this.
+[HTTP.HttpThread]
+RunningThreadedRequestLimit=100
+
```
csciguy8, line 7: 100 simultaneous HTTP requests sounds really high to me. Is this something that users really need?

There are some discussions about web browser parallel connection [here](https://stackoverflow.com/questions/985431/max-parallel-http-connections-in-a-browser), and typical values are less than 10.

In my personal experience with writing client code that does parallel fetches, I typically settled with 4. Values that were higher, like 8, 12, 16 showed diminishing returns with improving load time.

Source/CesiumRuntime/Private/CesiumPointAttenuationVertexFactory.cpp
```diff
@@ -13,7 +13,12 @@
 #include "MaterialDomain.h"
 #endif
 
+#if ENGINE_MAJOR_VERSION > 5 || ENGINE_MINOR_VERSION >= 3
```
csciguy8, line 16: I think this #if works, but my brain had to pause a bit and work out the logic.

What do you think of this?
```
#define ENGINE_VERSION_5_3_OR_HIGHER   (ENGINE_MAJOR_VERSION > 5 || ENGINE_MINOR_VERSION >= 3)
...
#if ENGINE_VERSION_5_3_OR_HIGHER   
  ... some code
```

csciguy8 (commented):
Looks good. Just a few comments / questions.

Tested on a windows machine with just UE 5.3 and Visual Studio 2022 community. Seems to work fine. Generated a solution from unreal-samples. Built debug / release. Tested out all the samples levels. Ran unit tests.

Source/CesiumRuntime/Private/CesiumPointAttenuationVertexFactory.cpp
j9liu: I also was confused by this when I reviewed a PR with this kind of check. But ultimately we have a lot of these checks across our files, and if we did this here, we would want to do it across the plugin for consistency. And you'd need a little manage all the `#defines` so that they don't get redefined between files. It seems a little messy 😅

csciguy8: For sure. A centralized place for a #define such as this is would be ideal.

.github/workflows/build.yml
kring: I suspect it's because the macOS distribution contains binaries for both Intel and Apple CPUs.

Config/Engine.ini
kring: The number of simultaneous requests is a setting in the UI (maximum simultaneous tile loads), and it defaults to 20. More simultaneous tile loads makes a _huge_ difference in some cases. For example, from here in Australia we have good bandwidth but very high latency, so downloading lots of tiles in parallel is a big win. In any case, this value in the config file needs to be higher than the value a user could _reasonably_ try to use, or else they'll get a warning. It doesn't imply we'll actually _do_ 100 simultaneous requests.

Source/CesiumRuntime/Private/CesiumPointAttenuationVertexFactory.cpp
kring: I don't think we need to mix this in with this PR. If you want to open a separate one to do this plugin-wide, I'm all for it.

csciguy8: That works. New issue is here, #1214

.github/workflows/build.yml
csciguy8: Yes indeed. Here's a diff of the 52-macos artifact on left, the new 53-macos artifact on right.

The dylib binaries are about twice the size, suggesting that yes, the Apple CPU version is in there. 

Were you expecting this as output? And if so, is this something we need to note?

image

kring: Hmm yeah ok that's weird. UE 5.2 should have binaries for both processors, too. So I can't explain why UnrealEditor-CesiumRuntime.dylib is twice as large in 5.2 versus 5.3.

kring: Hmm it seems I'm wrong. I confirmed with the `lipo` tool that our 5.2 build only includes Intel binaries, no Apple ones. We do include fat binaries of the cesium-native libs, though. The build process we're using is identical between 5.2 and 5.3, so this seems to be something Epic fixed/improved? I guess I'd say this is a bug in our 5.2 build process on CI, but it's largely gone unnoticed because the binaries on the Unreal Marketplace are built by Epic, and those _do_ support both CPU architectures. The sizes of the Marketplace binaries are larger, just like our CI-built 5.3 binaries.

TL;DR: the larger binaries in our 5.3 build are a feature, not a bug.

csciguy8: Ok, for the scope of this PR then (UE 5.3), things are working great.

Do we expect our 5.1 and 5.0 builds to also include Apple binaries? Because based on the sizes I'm seeing, they probably do not.

kring: No, I expect that 5.0 and 5.1 won't include Apple Silicon binaries. We don't even include Apple binaries for cesium-native for those versions.

joseph-kaile:
> macOS and iOS are building now, too. @joseph-kaile would you mind trying out the built 5.3 package from this branch on those platforms and confirm that it works ok?

Using the CI binaries, macOS works well, the performance is similar to a scene without Cesium. I still need to look into a way to fix the Reset Orientation Node. I was able to run it on iOS (after changing it from Desktop to Mobile). The tiles loaded but the material was black, which I'll look into.

csciguy8:
> I was able to run it on iOS (after changing it from Desktop to Mobile). The tiles loaded but the material was black, which I'll look into.

Sounds like macOS runs fine, and iOS does, with a material issue. 

I'm going to merge this PR, as overall, it's doing what it promises.

@joseph-kaile , if the black material is a legit bug for with UE 5.3 on iOS, let's write a new issue for it.

joseph-kaile:
> > macOS and iOS are building now, too. @joseph-kaile would you mind trying out the built 5.3 package from this branch on those platforms and confirm that it works ok?
> 
> Using the CI binaries, macOS works well, the performance is similar to a scene without Cesium. I still need to look into a way to fix the Reset Orientation Node. I was able to run it on iOS (after changing it from Desktop to Mobile). The tiles loaded but the material was black, which I'll look into.

Update, it looks like the IOS renderer doesn't like MLB_CesiumAlphaBlendBaseColor. After removing it, the textures come through. Also, the performance Is very smooth.

The tiles around the edge of the view are getting culled though.

IMG_0345 copy

kring:
@joseph-kaile can you please check if these are new problems in 5.3? I'm guessing so, but please confirm. I don't know what would cause either of those problems.

Issue #1114, opened 2023-06-01
Clear RF_Public flag on our Actors in PostLoad


kring:
Prior to #1111, the Cesium UI created Actors with the `RF_Public` flag set. This was not the right thing to do. As explained by an engine developer at Epic:

> RF_Public means that the object is an asset, so it should be only set for worlds, textures, meshes, etc.
This flags essentially says it's ok to have a reference to public objects from other packages (with the exception of worlds).
Actors are not assets since they are part of a level which is part of a world, etc. that's why they should not key the public flag.

#1111 changed the code to no longer set the flag. However, Cesium Actors that were created in previous versions will still have this flag. This _may_ cause problems when converting an existing level to World Partition, or perhaps cause other subtle issues that we haven't realized yet. The same engine developer suggested that we clear the flag from old Actors in `PostLoad`.

PR #1216, opened 2023-09-14
Clear RF_Public flag on our Actors in PostLoad

csciguy8:
Closes issue with same name, #1114 .

Based on the changes in #1111 , the only actors that could have previously had the `RF_Public` flag set would be `ACesium3DTileset`, `ACesiumCartographicPolygon`, and `UCesiumIonRasterOverlay` (actor component).

Hook into these classes' `PostLoad` and clear this flag if it set.

Tested using cesium-unreal-samples, 02_CesiumMelbourne and 08_CesiumClipping

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -27,6 +27,7 @@
 - `ACesiumGeoreference`, `ACesiumCameraManager`, and `ACesiumCreditSystem` are now created in the Persistent Level, even if the object that triggered their automatic creation (such as `ACesium3DTileset`) exists in a sub-level. It is very rarely useful to have instances of these objects within a sub-level.
 - An instance of `ACesiumCreditSystem` in a sub-level will no longer cause overlapping and broken-looking credits. However, we still recommend deleting credit system instances from sub-levels.
 - `ACesiumCartographicPolygon` now operates on the parts of the tileset that are shown in the Editor viewport, even if it is used with a Cesium3DTileset with a non-identity transformation.
+- Fixed bug where older scenes that used Cesium UI created actors could have their `RF_Public` flag set. This could cause problems when converting an existing level to World Partition, or perhaps cause other subtle issues that we haven't realized yet.
 
 ### v1.30.1 - 2023-09-03
 
--- a/Source/CesiumRuntime/Private/Cesium3DTileset.cpp
+++ b/Source/CesiumRuntime/Private/Cesium3DTileset.cpp
@@ -2089,7 +2089,13 @@ void ACesium3DTileset::PostLoad() {
 
   Super::PostLoad();
 
-  CesiumActors::validateActorFlags(this);
+#if WITH_EDITOR
+  // Only do this in the editor, when not in play mode
+  // We want to exclude fixing up any objects that aren't in this level
+  if (!GEditor->IsPlaySessionInProgress()) {
+    CesiumActors::validateActorFlags(this);
+  }
+#endif
 }
 
 void ACesium3DTileset::Serialize(FArchive& Ar) {
--- a/Source/CesiumRuntime/Private/Cesium3DTileset.cpp
+++ b/Source/CesiumRuntime/Private/Cesium3DTileset.cpp
@@ -12,6 +12,7 @@
 #include "Cesium3DTilesSelection/TilesetOptions.h"
 #include "Cesium3DTilesetLoadFailureDetails.h"
 #include "Cesium3DTilesetRoot.h"
+#include "CesiumActors.h"
 #include "CesiumAsync/IAssetResponse.h"
 #include "CesiumBoundingVolumeComponent.h"
 #include "CesiumCamera.h"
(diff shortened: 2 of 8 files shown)


cesium-concierge:

kring:
Thanks @csciguy8, this looks good. I just have one concern, and it might be unfounded. This code will, of course, remove the RF_Public flag from Cesium3DTileset and other objects. For the most part that is the correct thing to do. But assets are supposed to have the RF_Public flag, and will probably break if it's removed. So the question is: is it possible for a Cesium3DTileset to ever be an asset instead of an object in a level? Perhaps a Blueprint derived from Cesium3DTileset? Would you mind spending a little time investigating if something like this is possible and will cause a problem?

Also, please update CHANGES.md.

csciguy8:
> ..is it possible for a Cesium3DTileset to ever be an asset instead of an object in a level? Perhaps a Blueprint derived from Cesium3DTileset? Would you mind spending a little time investigating if something like this is possible and will cause a problem?
 
Changes updated.

I did quick test for both your questions. Yes, it is possible to create a Blueprint derived from a Cesium3DTileset, and yes, I could save this as a .uasset that shows up in the Content Browser.

Do these flags present a problem? For my test case of dragging this new asset into a blank level, no, it doesn't seem to be a problem. I stuck a breakpoint in `ACesium3DTileset::ACesium3DTileset` and traced how the actor was created. I saw this in `UActorFactory::PlaceAsset`

```
SpawnParams.ObjectFlags = InPlacementOptions.bIsCreatingPreviewElements ? EObjectFlags::RF_Transient : EObjectFlags::RF_Transactional;
AActor* NewActor = CreateActor(InPlacementInfo.AssetToPlace.GetAsset(), InPlacementInfo.PreferredLevel.Get(), InPlacementInfo.FinalizedTransform, SpawnParams);
```

This suggests that the object flags in a new asset don't really matter, as they are created fresh when you drag them in.

csciguy8:
Ran another test where I created a level blueprint, a  "Load Asset Blocking" node on play start, and tried to load that asset I mentioned previously. 

Again, had a breakpoint in `ACesium3DTileset::ACesium3DTileset` and traced how the actor was created. Bad news here, it gets created, and `ACesium3DTileset::PostLoad` is indeed getting called, running the code that strips out the RF_Public flag. In my test case there was no RF_Public flag present, but it is running this code on assets outside of the level. This is doing more than we originally intended.

csciguy8:
Committed changes to only do the fixup in the editor, when not in play mode. 

The flag was originally introduced via the quick add panel in the editor, during edit mode. It seems fitting that we'd fix it up in the same context.

kring:
Only removing RF_Public in the Editor sounds fine to me, but I think it might not totally solve the problem. For example, if `PostLoad` is ever called in the Editor for Assets, then the flag would still be erroneously removed. I wouldn't be surprised to see `PostLoad` called either when the project is opened, or when the user opens the asset in the Editor by double-clicking it in the content browser.

csciguy8:
Agreed. The changes are 'better', but hard to call them complete without having a good idea of all the cases.

Opening an asset in the editor is something I'll test next.

csciguy8:
Committed changes to forego the fixup if loading a .uasset in the editor. Thankfully there are useful object flags to know the difference between this and a level.

@kring Any other fixup scenarios come to mind?

csciguy8:
@kring Another thought about this issue. Would it be valid to consider not integrating this change at all?
What if the problems that it's trying to fix might actually be smaller than the problems that it could potentially cause in the future?

kring:
> Would it be valid to consider not integrating this change at all?

A very good question! Hard to answer, because we don't actually know with certainty the effect of not fixing it. Nor do we really understand the effect of fixing it!

kring:
After looking at the current code, I think it's very unlikely that it is going to cause problems for anyone. Basically the only way it could cause a problem if someone somewhere has some object that:

1. Is a `Cesium3DTileset`, `CesiumCartographicPolygon`, or `CesiumIonRasterOverlay`,
2. Has `RF_Public` and needs to keep it for some reason, and
3. Is not also marked with `RF_ClassDefaultObject` or `RF_ArchetypeObject`.

Can that happen? Maybe, but we'll just have to trust that someone will let us know if it does. Meanwhile, an existing actor with `RF_Public` will almost certainly problems (a crash) when the object is used with World Partition. So on balance I think this is likely to be a good change.

The only thing I can think of that we could do to improve the safety slightly is to make it only remove the RF_Public flag for objects saved with _previous_ versions of Cesium for Unreal. Which would at least give users an "out" by recreating an object that should have RF_Public in the new version, even if Cesium code insists on removing it from old objects. But I don't think it's really worth the trouble.

So... I'm going to merge this. Thanks @csciguy8!

csciguy8:
> The only thing I can think of that we could do to improve the safety slightly is to make it only remove the RF_Public flag for objects saved with previous versions of Cesium for Unreal. Which would at least give users an "out" by recreating an object that should have RF_Public in the new version, even if Cesium code insists on removing it from old objects. But I don't think it's really worth the trouble.

I like that thought. Kind of a file format versioning, but just for a plugin. 

Agreed though, could be overkill if our cesium objects don't really change very often or drastically.

kring:
> I like that thought. Kind of a file format versioning, but just for a plugin.

Yep, Unreal has a nice built-in system for doing this. Search for `Ar.UsingCustomVersion` in the plugin code and you'll find several places we do it.

No action necessary here, though.

PR #1218, opened 2023-09-15
Fix UE 5.3 unit tests (disable sublevel test)

csciguy8:
Turn off the Sublevels test "ensures only one sub-level instance is visible in the editor at a time".

Disable it only when running on UE 5.3, and only when running headless (CI).

From my debugging, I can see that `UCesiumSubLevelSwitcherComponent` isn't getting a chance to tick in between when `SetIsTemporarilyHiddenInEditor` is called and when the result is checked using `IsTemporarilyHiddenInEditor`. Works fine in the other versions.

Could be a bug introduced in 5.3, or they changed the timing for the automation framework intentionally.


cesium-concierge:

csciguy8:
Closing this PR as it is not required anymore. Issue has been fixed with #1219

PR #1219, opened 2023-09-18
Make sure editor sub-levels are hidden even when not ticking.

kring:
Fixes the test that was disabled in #1218 because it failed in UE 5.3.

The problem is that in UE 5.3, when running headless, ActorComponents never have their TickComponent called. Why? Because there are no editor viewports, and so the engine only does a `ELevelTick::LEVELTICK_TimeOnly` type of tick, which explicitly does not tick actor or their components. Without the `UCesiumSubLevelSwitcherComponent` tick, there's nothing that disables other sub-levels when a new one is activated.

I don't know if this is a UE 5.3 bug. The closest thing I can find in the changelog is this:

> Disabled real-time viewports when the editor process is in the background and bThrottleCPUWhenNotForeground.

But in any case I've worked around it in this PR by making `CesiumEditorSubLevelMutex` now explicitly schedule the necessary operation on the game thread. This way it will still happen even if the engine doesn't tick. It's harmless and has almost no cost for it to happen multiple times.

### DIFF
--- a/Source/CesiumEditor/Private/CesiumEditorSubLevelMutex.cpp
+++ b/Source/CesiumEditor/Private/CesiumEditorSubLevelMutex.cpp
@@ -1,6 +1,7 @@
 // Copyright 2020-2023 CesiumGS, Inc. and Contributors
 
 #include "CesiumEditorSubLevelMutex.h"
+#include "Async/Async.h"
 #include "CesiumGeoreference.h"
 #include "CesiumSubLevelComponent.h"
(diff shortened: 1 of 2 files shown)


cesium-concierge:

csciguy8 (approved):
Thanks @kring for fixing this odd one.

I'm not sure if this is a bug or not with UE 5.3 either. I wonder if this hints at a implied design where we shouldn't wait until a ::Tick for updating critical game objects? Or maybe it is a bug and we'll see an update soon.

PR #1229, opened 2023-09-28
Usability improvements for CesiumGeoreference parentage

kring:
This is a PR into #1227.

There are two main changes here:

1. The Quick Add and Cesium ion Assets UIs now add the objects they create as children of a CesiumGeoreference.
2. The logic used to find the default CesiumGeoreference for an object that doesn't have one specific explicitly has been improved to search its parent attachment chain first. If it doesn't find a CesiumGeoreference there, then it follows the same logic as it did previously. It also does an additional search at the end (if a CesiumGeoreference hasn't been found already) for _any_ CesiumGeoreference in the PersistentLevel.

These changes make it more convenient (automatic, even) to nest Cesium objects inside a CesiumGeoreference so that the globe can easily be transformed in the world. We can also easily do fancy things like have two globes with different georeferencing in the same Unreal world.


cesium-concierge:

Source/CesiumRuntime/Private/CesiumGeoreference.cpp
```diff
@@ -57,83 +57,132 @@ createCoordinateSystem(const FVector& center, double scale) {
...
+
+  // Spawn georeference in the persistent level
+  FActorSpawnParameters spawnParameters;
+  spawnParameters.SpawnCollisionHandlingOverride =
+      ESpawnActorCollisionHandlingMethod::AlwaysSpawn;
+  spawnParameters.OverrideLevel = World->PersistentLevel;
+
+  ACesiumGeoreference* Georeference =
+      World->SpawnActor<ACesiumGeoreference>(spawnParameters);
+  Georeference->Tags.Add(Tag);
```
csciguy8, line 137: Looks like you removed the null check against the returned pointer from SpawnActor.

That's fine, but if you're convinced this should always be non-null, might add an assert for that.

Source/CesiumRuntime/Private/CesiumGeoreference.cpp
```diff
@@ -57,83 +57,132 @@ createCoordinateSystem(const FVector& center, double scale) {
...
+/*static*/ ACesiumGeoreference*
+ACesiumGeoreference::GetDefaultGeoreferenceForActor(AActor* Actor) {
+  if (!IsValid(Actor))
+    return nullptr;
+
+  ACesiumGeoreference* Georeference = FindGeoreferenceAncestor(Actor);
+  if (IsValid(Georeference))
+    return Georeference;
+
+  Georeference = FindGeoreferenceWithTag(Actor, DEFAULT_GEOREFERENCE_TAG);
```
csciguy8, line 175: Lines 175 - 185 look like the same logic in `::GetDefaultGeoreference`. Any change to just call that function instead? Would reduce duplication, but could also make it clearer what's happening... find an ancestor, if didn't find it, find it in the world

Source/CesiumRuntime/Private/CesiumGlobeAnchorComponent.cpp
```diff
@@ -233,22 +233,38 @@ void UCesiumGlobeAnchorComponent::Sync() {
...
 
-  if (this->ResolvedGeoreference) {
-    this->ResolvedGeoreference->OnGeoreferenceUpdated.AddUniqueDynamic(
-        this,
-        &UCesiumGlobeAnchorComponent::_onGeoreferenceChanged);
+  if (Previous != Next) {
+    if (IsValid(Previous)) {
+      // If we previously had a valid georeference, first synchronize using the
+      // old one so that the ECEF and Actor transforms are both up-to-date.
+      this->Sync();
```
csciguy8, line 253: I like the addition of `::Sync`, here.

Because this exists, we could remove the other `::Sync` calls from line 82, and 513

csciguy8 (approved):
Looks good! All tested well and just a few non-critical suggestions.

I like the usability improvements. As someone new to cesium-unreal, all the changes feel intuitive.

kring:
Thanks for the review @csciguy8! I've addressed all your comments.

kring:
I'm merging this for the v2.0 preview. Further review comments still welcome.

PR #1233, opened 2023-09-29
Support `EXT_mesh_features` and `EXT_structural_metadata` picking / styling

j9liu:
This upgrades the metadata picking / styling APIs to handle the 3D Tiles 1.1 extensions. This involved an extensive amount of changes that are noted in detail in `upgrade-to-2.0-guide`.

Many functions / classes were renamed to match to the new spec. Some have been deprecated entirely. Core redirects were added when applicable, but this will definitely break any projects based on our plugin in C++. The nature of these changes means this breaking changes can't really be helped.

In #1190 we discovered a crash when pressing Undo after **GenerateMaterial** was called. I posted on UDN about it, and Stephen suggested I add `GEngine()->EndTransaction()` before we open the asset editor. It does prevent the crash, but it doesn't look like the changes from **GenerateMaterial** are able to be undone. Still, it's much better than crashing on undo, so I'm adding it to this PR.

Note: In light of this change, `EXT_feature_metadata` will no longer be supported. Batch tables will still work since they can be upgraded to `EXT_structural_metadata`, but glTFs with the old extension will not work with the metadata API.


cesium-concierge:

kring:
🚀

PR #1239, opened 2023-10-02
Improve build error handling on Continuous Integration (CI)

### LINKED ISSUE #1235: Continuous Integration (CI) reporting successful build, when it shouldn't
This run shows "Build CesiumForUnreal plugin" as succeeded (check), yet it didn't really.

Click on the arrow to see the log and you'll see many "fatal error C10883" and "error 2838" lines.

I would expect an X to show and the subsequent steps to not execute.

https://github.com/CesiumGS/cesium-unreal/actions/runs/6368512241/job/17287933466

csciguy8:
Fix bug where CI wouldn't show an error on the build step. 

This would not only lead to a false positive, but would also upload artifacts even when the build wasn't valid.

Applies to Windows builds only.

Closes #1235

### DIFF
--- a/.github/workflows/buildWindows.yml
+++ b/.github/workflows/buildWindows.yml
@@ -73,9 +73,9 @@ jobs:
         run: |
           echo "Temp directory: $env:TEMP"
           cd "$env:UNREAL_BATCH_FILES_PATH"
-          $exitCode = ./RunUAT.bat BuildPlugin -Plugin="$env:GITHUB_WORKSPACE/CesiumForUnreal.uplugin" -Package="$env:BUILD_CESIUM_UNREAL_PACKAGE_PATH" -CreateSubFolder -TargetPlatforms=Win64
-          echo ExitCode=$exitCode
-          if ($exitCode -eq 0) {
+          $consoleLog = ./RunUAT.bat BuildPlugin -Plugin="$env:GITHUB_WORKSPACE/CesiumForUnreal.uplugin" -Package="$env:BUILD_CESIUM_UNREAL_PACKAGE_PATH" -CreateSubFolder -TargetPlatforms=Win64
+          echo LastExitCode=$LASTEXITCODE
+          if ($LASTEXITCODE -eq 0) {
             exit 0
           }
           else {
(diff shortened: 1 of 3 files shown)


cesium-concierge:

csciguy8:
For reference, using these changes, here is a CI build showing an intentional compile error causing failure...
https://github.com/CesiumGS/cesium-unreal/actions/runs/6398299040

kring:
@csciguy8 is there any way to continue to get build output while the build is still running? It's not ideal to have no build output until the end of a process that can sometimes take over an hour. Also, I have in the past seen CI systems kill long-running processes that don't output anything for a long time. Certainly on Travis; I'm not 100% sure whether that could be a problem on GH Actions, too.

csciguy8:
> @csciguy8 is there any way to continue to get build output while the build is still running?

Sure! I disabled this, and moved the log display to a separate step just for organizational convenience. Personally, I like to see logs separate from the execution of build steps.

No worries though. I'll put it back to the way it was.

.github/workflows/buildWindows.yml
```diff
@@ -37,7 +37,8 @@ jobs:
           $env:CESIUM_UNREAL_VERSION=$(git describe)
           $env:BUILD_CESIUM_UNREAL_PACKAGE_NAME="${{ inputs.upload-package-base-name }}-${env:CESIUM_UNREAL_VERSION}"
           $env:BUILD_CESIUM_UNREAL_PACKAGE_PATH="$env:GITHUB_WORKSPACE/packages/CesiumForUnreal"
-          $env:UNREAL_BATCH_FILES_PATH="${{ inputs.unreal-batch-files-path }}"
+          $env:UNREAL_BATCH_FILES_PATH="C:/Program Files/Epic Games/${{ inputs.unreal-program-name }}/Engine/Build/BatchFiles"
+          $env:UNREAL_BUILD_LOG_PATH="C:/Users/ec2-user/AppData/Roaming/Unreal Engine/AutomationTool/Logs/C+Program+Files+Epic+Games+${{ inputs.unreal-program-name }}"
```
kring, line 41: I don't think this is used anymore.

csciguy8: You are correct. Removed it.

csciguy8:
@kring Ready for another look

PR #1241, opened 2023-10-04
Don't use `std::variant` in `FCesiumPropertyTableProperty`

j9liu:
`PropertyTablePropertyViewType` was a `typedef` for a large `std::variant` defined in Cesium Native. It was so large that all of the normalized property versions had to be put in their own `std::variant`, and the two were then grouped under their own `std::variant` in Cesium for Unreal. While Visual Studio's compiler was able to handle this well, Clang would experience incredibly slow compilation times and often would not be able to complete CI.

This replaces the `std::variant` with a `std::any`. The original property type is stored in the `FCesiumMetadataValueType`, and the struct can be used to re-find the C++ type to `std::any_cast` to. This allows Clang + CI to finally compile successfully, and to _somewhat_ reduce the size of our object files.


cesium-concierge:

PR #1244, opened 2023-10-06
Fix GLTF primitive component positioning

joseph-kaile:
This pull request fixes a bug in the GLTF primitive component positioning. Previously, the GLTF primitive component were using absolute coordinates to set their location:

```
Gltf->SetUsingAbsoluteLocation(true);
```

but now its using relative. 

Because the GLTF primitive component is attached to the Cesium3DTilesetRoot component, the relative transform is already applied, and it was being applied twice.


cesium-concierge:

* :x: Missing CLA.

kring:
Thanks @joseph-kaile, this change makes sense to me. Any idea what's happening with the UE 5.3 test failure, though?

kring:
The test failure went away upon re-rerunning.

PR #1251, opened 2023-10-18
Add support for file:/// URLs across platforms and UE versions

### LINKED ISSUE #378: Can't load from file:// URLs on Mac and Linux (or on any platform in UE 5.3!) because they are disabled in Unreal's libcurl
According to a [question about loading tilesets from files under Linux](https://community.cesium.com/t/loading-tilesets-from-files-under-linux/13313) from the forum, there might be an issue with the current approach of loading tilesets using the standard `UnrealAssetAccessor`.

The `UnrealAssetAccessor` uses the `IHttpRequest` to resolve the requests. But under linux, it might be necessary to use an `IPlatformFile` for local file requests, because on Linux, the libcurl library that is used in Unreal Engine is compiled with the `file` protocol being disabled.

This behavior has to be verified. If this turns out to be a problem, then it might be easy to fix, by just checking the protocol of the URL, and either using an `IHttpRequest` or an `IPlatformFile`, roughly at https://github.com/CesiumGS/cesium-unreal/blob/3daac114185202a8ee28b058d1131b401bad37dd/Source/CesiumRuntime/Private/UnrealAssetAccessor.cpp#L131 

---

Edit: I think that it would be better in many ways to have the option to load local tilesets in `cesium-native` itself. This would essentially mean to move the https://github.com/CesiumGS/cesium-native/blob/main/Cesium3DTiles/test/SimpleAssetAccessor.h or the https://github.com/CesiumGS/cesium-native/pull/127/files#diff-43c4884466567e2f908defe62c8ba7921b5310694a2dcf3ec0501cbc5e520f9d from the "test" package into the actual `cesium-native` core. This would allow to create an `IAssetAccessor` that "dispatches" either to the current `UnrealAssetAccessor` or the `FileAssetAccessor`, depending on the protocol.

kring:
Fixes #378 

I was updating the [Adding Datasets](https://cesium.com/learn/unreal/unreal-datasets/) tutorial for v2.0, and it seemed easier (or at least more rewarding) to fix file URLs rather than document all the caveats surrounding their use.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,10 @@
 
 ### ? - ?
 
+##### Additions :tada:
+
+- Added support for `file:///` URLs across all platforms and Unreal Engine versions.
+
 ##### Fixes :wrench:
 
 - Fixed a bug that could cause tiles in a Cesium3DTileset to have an incorrect transformation.
--- a/Source/CesiumRuntime/CesiumRuntime.Build.cs
+++ b/Source/CesiumRuntime/CesiumRuntime.Build.cs
@@ -176,7 +176,8 @@ public class CesiumRuntime : ModuleRules
                 "GLM_FORCE_XYZW_ONLY",
                 "GLM_FORCE_EXPLICIT_CTOR",
                 "GLM_FORCE_SIZE_T_LENGTH",
-                "TIDY_STATIC"
+                "TIDY_STATIC",
+                "URI_STATIC_BUILD"
             }
         );
 
--- a/Source/CesiumRuntime/Private/Tests/UnrealAssetAccessor.spec.cpp
+++ b/Source/CesiumRuntime/Private/Tests/UnrealAssetAccessor.spec.cpp
@@ -1,9 +1,10 @@
+#include "UnrealAssetAccessor.h"
 #include "CesiumAsync/IAssetResponse.h"
 #include "CesiumRuntime.h"
 #include "HAL/PlatformFileManager.h"
 #include "Misc/AutomationTest.h"
 #include "Misc/FileHelper.h"
-#include "UnrealAssetAccessor.h"
+#include "Misc/Paths.h"
 
 BEGIN_DEFINE_SPEC(
     FUnrealAssetAccessorSpec,
--- a/Source/CesiumRuntime/Private/Tests/UnrealAssetAccessor.spec.cpp
+++ b/Source/CesiumRuntime/Private/Tests/UnrealAssetAccessor.spec.cpp
@@ -1,9 +1,9 @@
-#include "UnrealAssetAccessor.h"
 #include "CesiumAsync/IAssetResponse.h"
 #include "CesiumRuntime.h"
 #include "HAL/PlatformFileManager.h"
 #include "Misc/AutomationTest.h"
 #include "Misc/FileHelper.h"
+#include "UnrealAssetAccessor.h"
 
 BEGIN_DEFINE_SPEC(
     FUnrealAssetAccessorSpec,
--- a/Source/CesiumRuntime/Private/Tests/UnrealAssetAccessor.spec.cpp
+++ b/Source/CesiumRuntime/Private/Tests/UnrealAssetAccessor.spec.cpp
@@ -1,3 +1,4 @@
+#include "Async/Async.h"
 #include "CesiumAsync/IAssetResponse.h"
 #include "CesiumRuntime.h"
 #include "HAL/PlatformFileManager.h"
--- a/Source/CesiumRuntime/Private/Tests/UnrealAssetAccessor.spec.cpp
+++ b/Source/CesiumRuntime/Private/Tests/UnrealAssetAccessor.spec.cpp
@@ -1,0 +1,65 @@
+#include "UnrealAssetAccessor.h"
+#include "CesiumAsync/IAssetResponse.h"
+#include "CesiumRuntime.h"
+#include "HAL/PlatformFileManager.h"
+#include "Misc/AutomationTest.h"
+#include "Misc/FileHelper.h"
+
+BEGIN_DEFINE_SPEC(
+    FUnrealAssetAccessorSpec,
+    "Cesium.Unit.UnrealAssetAccessor",
+    EAutomationTestFlags::ApplicationContextMask |
+        EAutomationTestFlags::ProductFilter)
+END_DEFINE_SPEC(FUnrealAssetAccessorSpec)
+
+void FUnrealAssetAccessorSpec::Define() {
+  It("Can access file:/// URLs", [this]() {
+    const char text[] = "Some random text.";
+
+    FString Filename = FPaths::ConvertRelativePathToFull(
+        FPaths::CreateTempFilename(*FPaths::ProjectSavedDir()));
+
+    IPlatformFile& FileManager = FPlatformFileManager::Get().GetPlatformFile();
+    FFileHelper::SaveStringToFile(
+        UTF8_TO_TCHAR(text),
+        *Filename,
+        FFileHelper::EEncodingOptions::ForceUTF8WithoutBOM);
+
+    try {
+      FString Uri = TEXT("file:///") + Filename;
+      Uri.ReplaceCharInline('\\', '/');
+      Uri.ReplaceInline(TEXT(" "), TEXT("%20"));
+
+      bool done = false;
+
+      UnrealAssetAccessor accessor{};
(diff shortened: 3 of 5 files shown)


cesium-concierge:

csciguy8 (changes requested):
Thanks for fixing this. Agreed, feels better to have something just work, then document limitations.

Verified fixed in UE 5.3.

Noticed these things:
* I could get file loading working by using this url locally 
`file:///C:/Tileset/tileset.json`
* I could NOT get file loading working by using these URLs with the same data
`file:///C:/Space%20In%20Name/Tileset/tileset.json`
`file:///C:/Space In Name/Tileset/tileset.json`
* The console error messages are not very helpful when loading fails

Can we get a simple unit test loading a tileset from file? This should help us detect this problem in the future (and across versions)

And also, for the future, should we have a "From File" option to complement "From URL" and "From Cesium Ion"? This might make the user experience a little clearer, and automatically do steps like "Replace spaces with %20".

kring:
This now depends on https://github.com/CesiumGS/cesium-native/pull/742, so merge that first.

New since the review:
* Added support for `%20` and other URI shenanigans by using the uriparser library to convert file URIs to filenames.
* Simplified the file reading by not using Unreal's insane async file reading API. It wasn't actually doing non-blocking I/O anyway, it was just doing blocking I/O in an I/O thread pool. Now it does that same thing but in a much more direct way. And it actually closes the file after it reads it, which I literally couldn't figure out how to do with the original API. 🤯
* The new, blocking, higher-level `FFileHelper::LoadFileToArray` that is used to do the actual reading also logs a warning with some more details when the read fails, so the console messages are hopefully more useful now.
* Added a test of using the UnrealAssetAccessor to read from a file URL.

This should be ready for another look.

csciguy8 (approved):

csciguy8:
Thanks @kring .

I've:
* Reviewed and merged the cesium-native changes
* Reviewed the code changes in this PR (looks good!)
* Tested the "path with spaces" case again, is fixed now
* Added a new failing test case to complement your success case
* Tried to fix some stubborn CI include path problems while I was there

If it looks good to you at this point, and CI is happy, I'd say let's merge.

kring:
Your changes look good to me @csciguy8. I just added one more small change, to move the deletion of the temp file to an `AfterEach` rather than doing it in each test. Merging!

csciguy8:
@lolleko  Can you post your tileset here? I can take a look.
(if it's of reasonable size and doesn't contain any sensitive data)

kring:
@lolleko not sure if this is the entire problem, but your first problem is that you're using backslashes. You have to use forward slashes in a file URL.

kring:
By the way, scraping tilesets from the web is against most providers terms of use.

lolleko:
> @lolleko Can you post your tileset here? I can take a look. (if it's of reasonable size and doesn't contain any sensitive data)

Unfortunately, it's around 900GB so i don't think I will be able to share it.

> @lolleko not sure if this is the entire problem, but your first problem is that you're using backslashes. You have to use forward slashes in a file URL.

Using forward slashes, results in a different error:

```
LogCesium: Loading tileset from URL file:///E:/tileset/tileset.json
LogCesium: Loading tileset from URL file:///E:/tileset/tileset.json done
LogStreaming: Warning: Failed to read file 'E:\tileset\7523\tileset.json?id=7523' error.
LogStreaming: Warning: Failed to read file 'E:\tileset\7526\tileset.json?id=7526' error.
LogCesium: Error: [2023-10-29 07:20:04.257] [error] [TilesetJsonLoader.cpp:886] Received status code 404 for tile content file:///E:/tileset/7523/tileset.json?id=7523
LogStreaming: Warning: Failed to read file 'E:\tileset\7524\tileset.json?id=7524' error.
LogStreaming: Warning: Failed to read file 'E:\tileset\7529\tileset.json?id=7529' error.
LogStreaming: Warning: Failed to read file 'E:\tileset\7525\tileset.json?id=7525' error.
LogStreaming: Warning: Failed to read file 'E:\tileset\7528\tileset.json?id=7528' error.
LogStreaming: Warning: Failed to read file 'E:\tileset\mesh_3230448.b3dm?id=3230448&x=3775235.2217615079135&y=5042296.3882614867762&z=912665.3464751446154' error.
LogCesium: Error: [2023-10-29 07:20:04.257] [error] [TilesetJsonLoader.cpp:886] Received status code 404 for tile content file:///E:/tileset/7526/tileset.json?id=7526
LogCesium: Error: [2023-10-29 07:20:04.258] [error] [TilesetJsonLoader.cpp:886] Received status code 404 for tile content file:///E:/tileset/7529/tileset.json?id=7529
LogCesium: Error: [2023-10-29 07:20:04.257] [error] [TilesetJsonLoader.cpp:886] Received status code 404 for tile content file:///E:/tileset/7524/tileset.json?id=7524
LogStreaming: Warning: Failed to read file 'E:\tileset\mesh_3230249.b3dm?id=3230249&x=3795252.5100151738152&y=5031587.0992207946256&z=888791.1531713148579' error.
LogStreaming: Warning: Failed to read file 'E:\tileset\mesh_3230579.b3dm?id=3230579&x=3779733.7439543749206&y=5037214.811092376709&z=922154.69383614265826' error.
LogStreaming: Warning: Failed to read file 'E:\tileset\mesh_3229310.b3dm?id=3229310&x=3770740.6180102843791&y=5046424.8120765183121&z=908422.0489208198851' error.
LogStreaming: Warning: Failed to read file 'E:\tileset\mesh_3230494.b3dm?id=3230494&x=3781387.0469807512127&y=5037313.612601429224&z=914845.94649977982044' error.
LogCesium: Error: [2023-10-29 07:20:04.258] [error] [TilesetJsonLoader.cpp:886] Received status code 404 for tile content file:///E:/tileset/7528/tileset.json?id=7528
LogCesium: Error: [2023-10-29 07:20:04.258] [error] [TilesetJsonLoader.cpp:886] Received status code 404 for tile content file:///E:/tileset/mesh_3230448.b3dm?id=3230448&x=3775235.2217615079135&y=5042296.3882614867762&z=912665.3464751446154
LogCesium: Error: [2023-10-29 07:20:04.258] [error] [TilesetJsonLoader.cpp:886] Received status code 404 for tile content file:///E:/tileset/7525/tileset.json?id=7525
LogCesium: Error: [2023-10-29 07:20:04.258] [error] [TilesetJsonLoader.cpp:886] Received status code 404 for tile content file:///E:/tileset/mesh_3230579.b3dm?id=3230579&x=3779733.7439543749206&y=5037214.811092376709&z=922154.69383614265826
LogCesium: Error: [2023-10-29 07:20:04.258] [error] [TilesetJsonLoader.cpp:886] Received status code 404 for tile content file:///E:/tileset/mesh_3229310.b3dm?id=3229310&x=3770740.6180102843791&y=5046424.8120765183121&z=908422.0489208198851
LogCesium: Error: [2023-10-29 07:20:04.258] [error] [TilesetJsonLoader.cpp:886] Received status code 404 for tile content file:///E:/tileset/mesh_3230494.b3dm?id=3230494&x=3781387.0469807512127&y=5037313.612601429224&z=914845.94649977982044
LogCesium: Error: [2023-10-29 07:20:04.258] [error] [TilesetJsonLoader.cpp:886] Received status code 404 for tile content file:///E:/tileset/mesh_3230249.b3dm?id=3230249&x=3795252.5100151738152&y=5031587.0992207946256&z=888791.1531713148579
```

> By the way, scraping tilesets from the web is against most providers terms of use.

I know, but this one is provided by the city itself for Public & Free use. While they don't specifically allow scraping I couldn't find  anything that says its forbidden.

lolleko:
Ahh and I forgot to mention, this tilset worked perfectly fine before UE5.3 dropped support for `file:///` in CURL requests.

kring:
Are the filenames it's trying to read, such as `E:\tileset\7524\tileset.json?id=7524`, correct?

lolleko:
> Are the filenames it's trying to read, such as `E:\tileset\7524\tileset.json?id=7524`, correct?

Good call... they are not. On disk they are stored without the additional URL params. In the corresponding tileset.json they are stored with URL params. I never noticed that before because it just used to work before 5.3. I guess CURL strips params when requesting files from disk.
I will have to write a script to fix up my tileset.json files. Hopefully it will work afterwards.

Thanks for the help. Assuming it works with the fixed json files there shouldn't be an issue with the new file based loading. Unless you want to match CURL behaviour, in that case you could consider supporting forward & backward slashes and maybe look into URL param handling.

kring:
ok, thanks, I wrote #1287.

PR #1254, opened 2023-10-19
Add support for property texture picking with `EXT_structural_metadata`

### LINKED ISSUE #1247: CoreRedirects error while packaging in v2.0 preview
Reported here:
https://community.cesium.com/t/cesium-for-unreal-v2-0-preview-release/27181/3

Seems it might just be a whitespace issue in `Engine.ini`?

j9liu:
This PR expands the property texture API to allow for picking, even when "Support UV Hit Info" is not enabled for the project. Here's an example of the metadata picking level with the [US Muscatatuck dataset](https://drive.google.com/file/d/1Ge9en-y4IFJ7EBba0esqhz3RuFd7gS2G/view?usp=share_link):

image

Each property texture property can specify it's own UV set, which complicated the metadata picking API. Multiple functions were added that take an input `FHitResult`, which lets them access the `Hit.Component` more readily. This is used to retrieve the barycentric coordinates of the point on the face, then compute the UV at that point.

This abstracts almost everything away from the user, and it is enough to simply connect a line trace hit to this node. However, there are still other functions if people want more granular access to the model and primitive metadata.

Unfortunately, this PR does **not** cover metadata styling with property textures. I made a lot of changes to try and get it to work, but ultimately abandoned the effort due to time. I'm happy to pick it up again after 2.0 is out. This should be noted as a regression in the release notes.

Upgrade guide is still WIP for the picking section, but I don't want the technical review to be delayed by the needed changes. Also I included a change that will fix #1247.


cesium-concierge:

kring:
I got this working on Linux CI by reducing the build parallelism. The CI build machine type we are usually running on has 8 cores and 16 GB of RAM. Unreal sees this and runs 8 simultaneous clang processes, which runs out of memory. Running out of memory is fatal because the Linux VM has swap disabled. Telling Unreal to only run two simultaneous clang processes avoids the issue.

kring:
Thanks @j9liu!

PR #1255, opened 2023-10-19
Don't spam the log when a globe anchor does not have a world.

kring:
I noticed that when opening the DynamicPawn in the Blueprint Editor, Cesium for Unreal spams endless messages to the log about the pawn's GlobeAnchor not being registered:

```
LogCesium: Error: Cannot get the rotation from CesiumGlobeAnchorComponent GlobeAnchor because the component is not yet registered or does not have a valid CesiumGeoreference.
LogCesium: Warning: CesiumGlobeAnchorComponent GlobeAnchor globe position is invalid because the component is not yet registered.
LogCesium: Warning: CesiumGlobeAnchorComponent GlobeAnchor globe position is invalid because the component is not yet registered.
```

This is normal and expected in this scenario, so I changed the code to avoid the spam.

It doesn't make sense to set the globe anchor position/orientation parameters in this scenario, and the current UI implementation (both before and after this PR) hides the fields, which I guess is fine:
image

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,7 @@
 
 - Fixed a bug that could cause tiles in a `Cesium3DTileset` to have an incorrect transformation.
 - Fixed a crash that occurred when a `LevelSequenceActor` in the level did not have a `LevelSequencePlayer` assigned.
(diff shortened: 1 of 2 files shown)


cesium-concierge:

csciguy8 (approved):
Looks sensible and verified the log is no longer spammed in this scenario. Thanks @kring !

PR #1263, opened 2023-10-24
Add Google to the Quick Add panel

kring:


cesium-concierge:

csciguy8 (approved):
Code changes look good. I can see the quick add entry in the editor and it appears to behave correctly. 
My account doesn't have the new asset so I can't verify.

PR #1267, opened 2023-10-27
Cesium sub-level improvements

### LINKED ISSUE #1162: Add a "Create Sub-Level Here" button
Similar to the one on the CesiumGeoreference in Unity.

It should create a new Level Instance, add a CesiumSubLevelComponent to it, and set the component's origin to the current camera location. Ideally, it should also prompt the user for where to save a new level and assign that level to the Level Instance.

### LINKED ISSUE #1261: CesiumSubLevel's "Place Georeference Origin at Sub Level Origin" button doesn't work for tilesets
Add a Cesium3DTileset to a sub-level, then use the "Place Georeference Origin at Sub Level Origin" button. The tileset will be given a transformation, just as normal objects are, which will offset the tileset from its correct location.

kring:
Numerous sub-level workflow improvements:

* New Cesium Actors are now created with the same world transform as the CesiumGeoreference. This is particularly important when Actors are created in sub-levels, because the previous behavior (an identity world transform) would end up with them misaligned with other Cesium globe things. Outside sub-levels, it should behave the same as before, but it's hopefully a bit clearer. Previously we created the new Tileset (for example) with an identity FTransform, and then moved it under the CesiumGeoreference while preserving the (identity) _relative_ transform. That would cause it to adopt the same world transform as the CesiumGeoreference. Now we create it explicitly with the same world transform as the CesiumGeoreference, and preserve the world transform when moving it under the CesiumGeoreference.
* Added `Create Sub Level Here` button to the `CesiumGeoreference`. Fixes #1162.
* Fixed a bug in the `Place Georeference Origin Here` button where it failed to take into account the CesiumGeoreference's transform.
* Made `PlaceGeoreferenceOriginAtSubLevelOrigin` on `CesiumSubLevelComponent` work correctly when the sub-level contained tilesets. Previously the tilesets would end up misaligned with the rest of the globe. Fixes #1261 
* Added `Place Georeference Origin Here` button on `CesiumSubLevelComponent`. It moves the georeference origin to the current camera position while ensuring the sub-level content doesn't move relative to the globe. It does this by setting the level instance's transform.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,12 +5,15 @@
 ##### Additions :tada:
 
 - Added support for `file:///` URLs across all platforms and Unreal Engine versions.
+- Added "Create Sub Level Here" button on `CesiumGeoreference`.
+- Added "Please Georeference Origin Here" button to `CesiumSubLevelComponent`.
 
 ##### Fixes :wrench:
 
 - Fixed a bug that could cause tiles in a `Cesium3DTileset` to have an incorrect transformation.
 - Fixed a crash that occurred when a `LevelSequenceActor` in the level did not have a `LevelSequencePlayer` assigned.
 - Fixed a bug that would spam Georeference-related messages to the log when editing a globe anchor component that is not embedded in a world. For example, when editing a Blueprint asset with a globe anchor.
+- Fixed several problems that could cause tilesets in sub-levels to be misaligned with the rest of the globe.
 
 ### v2.0.0 Preview 1 - 2023-10-02
 
--- a/Source/CesiumEditor/Private/CesiumEditor.cpp
+++ b/Source/CesiumEditor/Private/CesiumEditor.cpp
@@ -54,6 +54,9 @@ TSharedPtr<FSlateStyleSet> FCesiumEditorModule::StyleSet = nullptr;
 FCesiumEditorModule* FCesiumEditorModule::_pModule = nullptr;
 
 namespace {
+
+AActor* SpawnActorWithClass(UClass* actorClass);
+
 /**
  * Register an icon in the StyleSet, using the given property
  * name and relative resource path.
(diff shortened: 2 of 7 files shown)


cesium-concierge:

csciguy8:
Discovered an odd bug:

1) Open the samples map 01_CesiumWorld
2) Click on the CesiumGeoreference, click on the new "Create Sub Level Here" button. Click save when prompted to save the new map
3) Navigate the camera somewhere else
4) Click on the new "Create Sub Level Here" button again. Click save when prompted to save the new map

Notice your first sublevel coordinates have changed and now match your new sublevel. I assume this isn't desired?

Oddly, repeating for a 3rd and 4th sublevel does not show this problem.

kring:
Yeah definitely not desired. I'll check it out.

csciguy8:
Another crash bug. Admittedly, I'm doing something not great.

1) Open the samples map 01_CesiumWorld
2) Click on the CesiumGeoreference, click on the new "Create Sub Level Here" button. Click save when prompted to save the new map
3) In the outliner, move your new LevelInstance actor so it's not a child of the CesiumGeoreference actor
4) Delete the CesiumGeoreference actor. Click Yes to confirm breaking references.
5) Right click on the sublevel actor, Edit XXX
6) In the Details window, go the CesiumSubLevelComponent_0 
7) Click "Place GeoreferenceOriginAtSubLevelOrigin"

Observe this crash...
image

csciguy8 (approved):
Thanks @kring. Tested the new functionality described in this PR and it looks like some nice improvements. 

Code looks good with just those two bugs that I noticed along the way.

For the future, it would be nice to have unit tests validate these UI workflows and resulting transforms in some way.

kring:
Thanks for the review @csciguy8. I've fixed the problems you mentioned above.

csciguy8:
Verified both issues are fixed, thanks.

Ready to merge once CI finishes.

kring:
I'm going to merge this because the only change since the last green tick is a merge from main.

PR #1269, opened 2023-10-30
Merge v1.31.2 into main

j9liu:
This merges the v1.31.2 changes into main for the upcoming v2.0 release.

The changelog is a bit convoluted because of the v1.31.2 release happening in parallel with v2.0 development. Maybe we can condense the 2.0.0 and 2.0.0 Preview 1 notes? Let me know your thoughts.


cesium-concierge:

kring:
Thanks @j9liu! I re-added v1.31.2 to the changelog, because we don't want to lose the history of that version. I also added the one "Addition" from v1.31.2 to the v2.0.0 section. It's a little awkward to have these two different versions listed in a single changelog, but I don't think it's worth the trouble to separate them. So in the absence of that, the idea is just 1) list all the versions that we've released, 2) put them in chronological order, 3) the changes listed are relative to the previous version in that "line".

PR #1274, opened 2023-10-31
Fix inaccurate and "snappy" flights when using FlyTo component

### LINKED ISSUE #1236: `UCesiumFlyToComponent` flight sometimes inaccurate and "snaps" at end of flight
This can be reproduced with `04_CesiumSublevels`, `11_CesiumPointClouds`, and  `12_CesiumGoogleMapsTiles`. Hit play, then move the camera slightly away from the starting position. Then press **1**.

The expected behavior is that you smoothly fly to that same place you started. However, it is sometimes strangely inaccurate. It flies to a completely different place, then "snaps" to the intended location. For example:

ezgif-2-956185d57b

(The sudden teleporting at the end is exactly what I'm seeing in the PIE view.)

I initially thought this might be a problem with the sublevels specifically, but it happens both with and without sub-levels. This also happens **regardless** of the origin-shifting mode.

csciguy8:
Closes #1236 

Revert logic that calculates flight rotation axis and angle to use `glm::` functions, like we did a few versions ago. 

The Unreal counterparts have hidden epsilons that were returning (1,0,0) for the rotation axis when small flight distances were needed.

Easy to reproduce in cesium-unreal-samples. Open 04_CesiumSublevels. Enter Play. Move somewhere else on the roof top, then press '1'.

### DIFF
--- a/Source/CesiumRuntime/Private/CesiumFlyToComponent.cpp
+++ b/Source/CesiumRuntime/Private/CesiumFlyToComponent.cpp
@@ -7,6 +7,8 @@
 #include "GameFramework/Pawn.h"
 #include "UObject/ConstructorHelpers.h"
 
+#include <glm/gtx/quaternion.hpp>
+
 UCesiumFlyToComponent::UCesiumFlyToComponent() {
   // Structure to hold one-time initialization
   struct FConstructorStatics {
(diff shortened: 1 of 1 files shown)


cesium-concierge:

kring:
Code looks good to me. I'm going to merge this before CI finishes in order to hopefully have fewer CI cycles before we can release v2.0.

PR #1275, opened 2023-10-31
Add missing includes, fix a warning

kring:
This fixes some errors caused by missing includes in a non-unity build. Also fixed a warning caused by the compiler being unable to see that a variable would actually be initialized.


cesium-concierge:

csciguy8:
By chance are you using non-unity builds for UE 5.3.1? 
Because I couldn't get that to work last I checked.

kring:
No, I'm not. I had the same problem with 5.3. This is with 5.1.

kring:
Merging this, too, in order to minimize CI cycles.

PR #1277, opened 2023-10-31
Remove deprecated exclusion zones features.

kring:
Removed the old "exclusion zones" feature, which has been deprecated since v1.11.0. Cesium for Unreal has `CesiumCartographicPolygon` and `CesiumTileExcluder` now, which are better.


cesium-concierge:

PR #1278, opened 2023-11-01
Fix the syntax of a couple of the core redirects.

### LINKED ISSUE #1262: Problems packaging on Android in Cesium for Unreal v2.0
As reported on the forum:
https://community.cesium.com/t/cesium-2-0-preview-building-for-android-on-ue-5-3-app-doesn-t-start/27295

It's apparently working well in v1.31.1, but not in v2.0-preview1. It's not obvious to me what might have changed, but we should see if we can reproduce this and investigate.

kring:
Fixes #1262


cesium-concierge:

Issue #1237, opened 2023-10-02
"FlyToLocation" yaw / pitch values result in different views for `UCesiumFlyToComponent`


j9liu:
This was most noticeable in the `12_CesiumGoogleMapsTiles` map with the the following cities:

- Sydney (4 key)
- London (6 key)
- Las Vegas (8 key)

First, check out the views with the deprecated "Fly To Location" Blueprints. Then, replace the blueprint with a "Fly To Location" on the DynamicPawn's `CesiumFlyToComponent`, with the same pitch and yaw. The views end up being different from the original view.

This happens even if origin shifting is disabled. It's totally possible that this new behavior is actually correct, but we should verify just in case it isn't.

kring:
I modified the "FlyToSydney" part of the Blueprint to look like this:
<img width="849" alt="image" src="https://github.com/CesiumGS/cesium-unreal/assets/924374/50591926-452c-42ba-b261-5dde566e20fa">

And the view at the end is identical whether I use the deprecated Fly to function (top) or the non-deprecated one (bottom). Anything I might have missed here?

kring:
Closing this because it appears to be fixed!

PR #1279, opened 2023-11-01
DynamicPawn should use ControlRotationInEastSouthUp RotationToUse.

kring:
I realized while updating the "Transition Between Locations" tutorial that the DynamicPawn's `CesiumFlyToComponent` has the wrong default value for the `RotationToUse`. As a result, flights wouldn't set the DynamicPawn's orientation correctly (it would set the Actor's orientation, rather than the Pawn's ControlRotation). It's possible that this is what @j9liu saw in #1237.


cesium-concierge:

PR #1280, opened 2023-11-01
Cleanup engine version defines

### LINKED ISSUE #1214: Cleanup usage of engine version #defines
Let's cleanup the preprocessor logic that deals with `ENGINE_MAJOR_VERSION` and `ENGINE_MINOR_VERSION` in the cesium-unreal codebase.

Ex.
```
  if (pMesh) {
#if ENGINE_MAJOR_VERSION >= 5 ||                                               \
    (ENGINE_MAJOR_VERSION == 4 && ENGINE_MINOR_VERSION >= 27)
    pMesh->SetRenderData(nullptr);
#else
    pMesh->RenderData.Reset();
#endif
  }
```

In this example it would be nice to have a single `IS_VERSION_4_27_OR_HIGHER` define. And if we do introduce this, it should be defined in one place only, plugin wide.

The goal is to have code that is easier to read, easier to verify correct, and less error prone.

A few more instances coming in from PR #1207

csciguy8:
Closes #1214.

Use new helper macros when defining Unreal Engine specific code.

Remove support for UE versions 5.0 and below.


cesium-concierge:

Source/CesiumRuntime/Private/CesiumSubLevelSwitcherComponent.cpp
```diff
@@ -355,7 +355,7 @@ UCesiumSubLevelSwitcherComponent::_getLevelStreamingForSubLevel(
             if (!pInstanceStreaming)
               return false;
 
-#if ENGINE_MAJOR_VERSION > 5 || ENGINE_MINOR_VERSION > 0
+#if ENGINE_VERSION_5_1_OR_HIGHER
             return pInstanceStreaming->GetLevelInstance() == SubLevel;
 #else
             return pInstanceStreaming->GetLevelInstanceActor() == SubLevel;
```
j9liu, line 361: Just a drive-by comment, looks like this if statement can be removed

csciguy8: I guess we can just assume that everything is 5.1 or higher. Good catch!

csciguy8: Fixed

j9liu (commented):
Looks good @csciguy8 ! I just had one comment about some other code we could also clean up as a result of these changes.

Source/CesiumRuntime/Public/CesiumCommon.h
```diff
@@ -4,11 +4,12 @@
 
 #include "Runtime/Launch/Resources/Version.h"
 
-#if ENGINE_MAJOR_VERSION >= 5
+#define ENGINE_VERSION_5_3_OR_HIGHER                                           \
+  (ENGINE_MAJOR_VERSION > 5 || ENGINE_MINOR_VERSION >= 3)
+#define ENGINE_VERSION_5_2_OR_HIGHER                                           \
+  (ENGINE_MAJOR_VERSION > 5 || ENGINE_MINOR_VERSION >= 2)
+
 #define CESIUM_UNREAL_ENGINE_DOUBLE 1
```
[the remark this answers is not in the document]
csciguy8: Another good catch. Removed both CESIUM_UNREAL_ENGINE_DOUBLE and CesiumReal

csciguy8:
@j9liu Ready for another look

j9liu:
Looks good @csciguy8 ! I'll wait for CI to pass before I merge.

PR #1281, opened 2023-11-01
Cesium3DTileset custom category ordering

### LINKED ISSUE #1260: On Cesium3DTileset, not-very-useful "Rendering" section appears above "Cesium"
We should see if we can move this section lower.

image

csciguy8:
Closes #1260 .

Move Cesium category so it's right after Transform and before Rendering.

These changes allow us to do future customization as well.

image


cesium-concierge:

csciguy8:
@kring If you're inspired to streamline this even more, feel free. 

There's actually quite a bit of customization we can do here,
https://docs.unrealengine.com/5.3/en-US/reordering-and-hiding-properties-in-details-panel-customizations-in-unreal-engine/

kring:
Thanks @csciguy8!

PR #1296, opened 2023-11-16
Improve network performance by fixing libcurl thread

csciguy8:
Decrease the value of `HttpThreadActiveFrameTimeInSeconds`, from 0.005 to 0.001. This controls how often Unreal Engine polls libcurl. Now we do it more often.

Overall, all test locales seem to benefit. 

In the most demanding tests (Google Tiles), load times are **~2.5x faster.** Previously, a user who viewed the NYC Chrysler building would wait ~14 seconds for it to load. Now it only takes 6 seconds.

Tested these things:
* 4 different performance tests, see below - All passed, and are faster
* Flying around Unreal editor with Google Tiles - Good framerate and responsiveness
* Flying around a packaged game - Good framerate and responsiveness
* Unreal Insights - Everything looked like before, except the events were more tightly packed (which is expected)

Performance Tests Summary (cold cache)
-----------------------------------------
```
SampleLocaleDenver - 13% load time reduction
SampleLocaleMelbourne - 16% load time reduction
GoogleTiles.LocaleDeathValley - 61% load time reduction
GoogleTiles.LocaleChrysler - 57% load time reduction
```

Complete Testing Data
-----------------------------------------
```
HttpThreadActiveFrameTimeInSeconds set to 0.005 (existing default)
--------------------------------------------------------------------------
SampleLocaleDenver (27 MB)
	Cold cache - 3.32, 3.03, 2.94 - 3.09 avg
	Warm cache - 0.71, 0.79, 0.75 - 0.75 avg
SampleLocaleMelbourne (89 MB)
	Cold cache - 4.68, 4.19, 4.24 - 4.37 avg
	Warm cache - 1.16, 1.15, 1.14 - 1.15 avg
GoogleTiles.LocaleDeathValley (27 MB)
	Cold cache - 9.95, 10.02, 10.20 - 10.05 avg
	Warm cache - 1.37, 1.33, 1.41 - 1.37 avg
GoogleTiles.LocaleChrysler (55 MB)
	Cold cache - 13.89, 15.32, 14.38 - 14.53 avg
	Warm cache - 2.28, 2.28, 2.34 - 2.30 avg

HttpThreadActiveFrameTimeInSeconds set to 0.001
--------------------------------------------------------------------------
SampleLocaleDenver (27 MB)
	Cold cache - 2.78, 2.68, 2.62 - 2.69 avg
	Warm cache - 0.68, 0.71, 0.79 - 0.72 avg
SampleLocaleMelbourne (89 MB)
	Cold cache - 4.07, 3.48, 3.45 - 3.66 avg
	Warm cache - 1.20, 1.17, 1.18 - 1.18 avg
GoogleTiles.LocaleDeathValley (27 MB)
	Cold cache - 3.98, 3.77, 3.88 - 3.87 avg
	Warm cache - 1.32, 1.32, 1.33 - 1.32 avg
GoogleTiles.LocaleChrysler (55 MB)
	Cold cache - 6.29, 6.17, 6.13 - 6.19 avg
	Warm cache - 2.27, 2.24, 2.31 - 2.27 avg

```


cesium-concierge:

csciguy8:
Some more technical points about _why_ this change is needed...

* In `FLegacyHttpThread::Run` a loop runs in which every iteration eventually calls `curl_multi_perform`
* This loop throttles itself so it doesn't consume 100% of the CPU. It uses `::Sleep`, or if set to 0, `::SwitchToThread`
* Some users have experienced the same bandwidth throttling as we have ([link](https://stackoverflow.com/questions/44389909/idling-thread-between-curl-multi-perform))
* From curl documentation, these type of loops should be using [curl_multi_poll](https://curl.se/libcurl/c/curl_multi_poll.html)

Overall, this PRs change accomplishes increased download speed, but the solution is not optimal. 

Ideally `FLegacyHttpThread::Run()` needs to be written to use `curl_multi_poll` instead of `Sleep`. This would allow sleeping as needed, based on the amount of activity detected, rather than polling at a fixed 1000hz.

Unfortunately, this is inside Unreal Engine's source and is not configurable from an INI file.

kring:
Here in Oz, the load time for the Chrysler locale goes from ~18s to ~12s with this PR. Not as impressive as your numbers, but still a very nice improvement! I don't see any downside to this, and a huge upside, so I'm merging it.

PR #1299, opened 2023-11-21
Add missing headers and forward declarations

kring:
I ran into these because I'm building without "unity" mode, but they're (small) legitimate problems.


cesium-concierge:

PR #1302, opened 2023-11-21
Check and remove query params from file URLs

### LINKED ISSUE #1287: New file:/// URL support is confused by query parameters
Originally mentioned here:
https://github.com/CesiumGS/cesium-unreal/pull/1251#issuecomment-1785966880

It seems curl strips out query parameters from file URLs before resolving the file on disk, whereas uriparser, somewhat surprisingly, does not. It seems unlikely someone would ever want/expect us to include the query parameters in the filename we attempt to read from disk, so we should fix this.

joseph-kaile:
Fixes #1287


cesium-concierge:

* :x: Missing CLA.

csciguy8 (approved):
Thanks @joseph-kaile , look like good changes.

The only suggestion I would make would be to make a separate test for a file with query params. This way we can verify a test without query params still passes.

No worries though, I committed this change.

PR #1306, opened 2023-11-27
Fix "Select New Token" dialog when not connected to ion

### LINKED ISSUE #1187: Crash when attempting to specify a Cesium ion token while not logged in
If you're not logged in, the "Select a Cesium ion Token" panel looks like this:

image

This is kind of broken in a few of ways:
1. There's only one option - to specify a token - but the button is mislabeled "Create New Project Default Token"
2. The radio button next to the Specify a token box is invisible. It also should be selected by default because it's the only option, but it's not.
3. Clicking that "Create New Project Default Token" causes Cesium for Unreal to crash.

If you actually select "Specify a token" first, then the panel looks like this:

image

And now you can click "Use as project default token" without crashing.

joseph-kaile:
Fixes #1187

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,6 +7,10 @@
  - Added support for styling with property textures in `EXT_structural_metadata`.
  - Significantly improved tile download performance by adding `HttpThreadActiveFrameTimeInSeconds=0.001` to `Engine.ini`. This results in a major performance improvement for all tilesets, particularly Google Photorealistic 3D Tiles.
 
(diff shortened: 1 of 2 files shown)


cesium-concierge:

* :x: Missing CLA.

csciguy8 (approved):
Ran through various scenarios and could no longer reproduce any of the 3 symptoms cited in the original issue.

Thanks @joseph-kaile !

PR #1307, opened 2023-11-27
Fix tiles on edges of frustum disappearing (UE.5.3)

### LINKED ISSUE #1303: Tiles on edges of wide frustum don't render
On Windows 10, Unreal Engine 5.3.1, with a Nvidia Quadro K2200.

1) Load cesium-unreal-samples, level 05_CesiumWorldPostProcessing. 
2) Push the play button to go into Play in Editor mode
3) Move the camera side to side

Granted, this is an old video card, and Unreal Engine does complain to me on startup (see popup). Still, I wouldn't expect tiles on the edges to be missing simply because of D3D 12 compatibility. 

Also odd, I don't see this behavior when not in Play in Editor mode.

TilesOnEdges

image

csciguy8:
Closes #1303.

Explicitly set the value of `AspectRatioAxisConstraint` to `AspectRatio_MaintainXFOV`. Our code in `ACesium3DTileset::CreateViewStateFromViewParameters` assumes horizontal field of view is constant, and vertical field of view is calculated from this.

This fixes a bug where the editor viewport did not match the viewport you would see when entering Play-in-Editor mode or when running a packaged game. It's noticeable whenever the aspect ratio isn't 1.0, although in extreme cases, tile might not show up around the edges of the view.

In Unreal Engine v5.2, the default was `AspectRatio_MaintainXFOV`. In Unreal Engine v5.3, the default is now `AspectRatio_MaintainYFOV`. This change caused the issue.

image

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,10 @@
 
 ### ? - ?
 
+##### Fixes :wrench:
+
+- Fixed a bug where viewports could appear wider than configured in the Dynamic Pawn's Camera Field of View. Noticed in Unreal Engine v5.3 when in Play-In-Editor mode, or a packaged game. In extreme cases, tiles would be missing near the edges of the view
+
 ##### Additions :tada:
 
  - Added support for styling with property textures in `EXT_structural_metadata`.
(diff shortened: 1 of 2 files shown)


cesium-concierge:

kring:
As discussed offline, I think it would be better to add support for the new setting, rather than reverting it. #1310 covers that. In the meantime, though, this PR is a big improvement over the incorrect tile selection that we get otherwise. Thanks @csciguy8!

PR #1312, opened 2023-11-30
Multiple cesium ions review

joseph-kaile:
Fixes two bugs:

1. When resuming, the connection could still be connected even when the API url is incorrect and so the UI appears that the user is logged in. Now, the Login Panel shows instead.

2. When the API url is being fetched, the ConnectionStatus Widget would show the URL to be blank and the Cancel button wouldn't work. Now, the ConnectionStatus Widget will appear only after the Server has acquired the API url and tries to authorize.


cesium-concierge:

* :x: Missing CLA.

PR #1315, opened 2023-12-06
Apply texture transformation in UMaterial instead of Native

joseph-kaile:
Same as https://github.com/CesiumGS/cesium-unity/pull/385 but in Unreal.


kring:
@joseph-kaile can you please open a PR for the cesium-native changes as well? It looks like there will be two, one into main and one into v0.27.x.

kring:
Joseph's code in this PR is solid, but it only applies the texture transform to the "baseColorTexture". I extended it to work with all the other textures, too. And added support for texture rotation, so that I could use @emackey's awesome [TextureTransformMultiTest](https://github.com/KhronosGroup/glTF-Sample-Assets/blob/main/Models/TextureTransformMultiTest/README.md) glTF to validate it. It seems to be working well:

image

The occlusion test passes, too, it's just a bit hard to see:
image

(Note that I had to turn off "Static Lighting" in the Project Settings in order to make Unreal render the occlusion texture at all.)

Now that I've worked on this, someone else review and merge it. It depends on CesiumGS/cesium-native#799 so merge that first.

j9liu:
This looks great to me @kring. I'm pushing a few commits to fix an unrelated test failure, and also some cleanup in the material graph. (No nodes added or removed, just reorganized for visual clarity).

I have one concern about the image you posted: I'm seeing a discrepancy between the "Sample" normal textures and the transformed ones.

image

It almost looks like the normals are completely reversed from the sample ones, even though they're supposed to be the same. But I wasn't able to verify it myself because I didn't know how to get the glTF into the scene. In the end it does seem like the texture coordinate _transformations_ were correct, and maybe the output is somehow being interpreted wrong, but I'm not 100% sure.

emackey:
Sharp eyes @j9liu.  I'm seeing some similar results in Blender, but not in BabylonJS and others.  My initial hunch is that there's disagreement as to whether rotating a normal map also rotates the TBN matrix, or if rotating it just influences the UV coordinates only.  This probably needs a bit more investigation and possibly a new sample model to test it.  But it seems likely unrelated to this PR.

CHANGES.md
```diff
@@ -16,6 +16,7 @@
 - Fixed several building warnings when packing in Unreal Engine 5.3.
 - Readded backwards compatibility for feature textures from `EXT_feature_metadata`, which was mistakenly removed.
 - Fixed a bug that caused nav mesh creation to be slow due to creating duplicate physics meshes.
+- Apply `KHR_texture_transform` texture coordinate transformation in UMaterial instead of in native.
```
j9liu, line 19: I don't want to retrigger a build, but when this merges in main I think this should say:

> `KHR_texture_transform` is now applied in `UMaterial`, instead of being baked directly into texture coordinates.

PR #1317, opened 2023-12-07
Move nav mesh creation after setting physics mesh

### LINKED ISSUE #1264: When `CreateNavCollision` is enabled, Unreal unnecessarily creates physics meshes on the game thread
As reported here:
https://community.cesium.com/t/line-trace-passing-through-cesium3dtileset-on-high-lod/27485/7

CesiumGltfComponent calls `CreateNavCollision` before it sets up the physics meshes. Because the nav collision (sometimes? always?) requires physics meshes, Unreal generates them at that point, which is inside the game thread. That's wasteful because we've already generated them in a worker thread, and will be applying them just a few lines later. It also causes errors to be written to log in UE 5.3, at least. And it's _possible_ that it also breaks the line traces entirely.

This looks like an easy fix that should definitely improve performance when NavCollision is enabled, and may fix other problems as well.

joseph-kaile:
Fixes #1264


csciguy8 (commented):
Changes looks reasonable. 

I tested it with a tileset by checking "Create Nav Collision" and witnessing the console spam. With these changes, the spam is gone, and everything appears to execute normally.

How did you test if Nav Collision still works after this? I'm not sure how to. I didn't find a unit test for this either.

kring:
Thanks @joseph-kaile!

PR #1322, opened 2023-12-14
Upgrade to v4 of download/upload-artifact

kring:
The biggest improvement here is that it should allow us to access artifacts of intermediate build steps before the entire workflow is complete.


kring:
This looks like a clear improvement. Artifacts are now available after each step, rather than waiting until the entire build pipeline is complete. Other than that, it works the same as before. So I'm going to merge this.

Issue #1311, opened 2023-11-29
DitherFade material layer breaks glTF alphaMask rendering


stephLVCim:
Hello,

I have some glb objects with alphaMode set to "MASK" using png with alpha channel texture, but the texture appear totally opaque in the rendering in Unreal Editor:

![image](https://github.com/CesiumGS/cesium-unreal/assets/125025785/732b2652-e23e-4a24-a65b-f581f972f6d8)

I tried to change alpha cutoff but without more success.

is alphamode "MASK" supported with 3D tiles 1.1 ?

thanks.

stephLVCim:
Thanks for your answer,

I think the problem is that the alpha channel of the BaseColor texture needs to be linked to the "Opacity Mask" slot of the material : 
https://docs.unrealengine.com/5.3/en-US/material-blend-modes-in-unreal-engine/#masked 

but I don't see how to do this in code.

Found this : https://forums.unrealengine.com/t/how-to-hook-up-alpha-channel-of-a-texture-in-c/453636 but did not manage to apply it in the cesium-unreal code (and it's perhaps some editor only code ? not runtime).

kring:
The BaseColor is correctly linked to the Opacity Mask in your default materials. The problem is the `DitherFade` layer in our default materials. The `MLB_DitherFade` replaces the opacity mask with the one from the dither. It should multiply instead.
<img width="547" alt="image" src="https://github.com/CesiumGS/cesium-unreal/assets/924374/9ef47a43-87a9-4281-a939-7d11bd043d20">

As a workaround, you can open `MLB_DitherFade` included in the Cesium for Unreal plugin and made the changes in the Blueprint above. Or you can open `MI_CesiumThreeOverlaysAndClipping` and remove the DitherFade layer entirely.

stephLVCim:
Thanks Kevin,

     modifying the MLB_DitherFade works fine.

     Do you plan to merge this directly in the repository ?

         Stéphane

________________________________
De : Kevin Ring ***@***.***>
Envoyé : dimanche 10 décembre 2023 22:39
À : CesiumGS/cesium-unreal ***@***.***>
Cc : Stéphane Capo ***@***.***>; Author ***@***.***>
Objet : Re: [CesiumGS/cesium-unreal] gltf objects with alphaMode MASK and transparent png (Issue #1311)


[...]

PR #1324, opened 2023-12-20
 DitherFade material layer breaks glTF alphaMask rendering #1311 

stephLVCim:
Changed MLB_DitherFade like this : 

image

So that the gltf objects with a "MASK" blending mode are correctly displayed, after a discussion with @kring .


kring:
Thanks for the PR @stephLVCim! I don't think we have a Contributor License Agreemnt (CLA) from you yet. Details here:
https://github.com/CesiumGS/cesium/blob/main/CONTRIBUTING.md#contributor-license-agreement-cla

kring:
Thanks for signing the CLA @stephLVCim. I've pushed your changes to https://github.com/CesiumGS/cesium-unreal/tree/alpha-mask (that link will stop working once the branch is merged and deleted). I've also replaced the "Min" function with "Multiply", which I think is more correct, and updated CHANGES.md. If CI is happy, I'll merge this. Thanks again for the PR!

kring:
Merged! Thanks @stephLVCim.

Issue #1327, opened 2024-01-05
Unnecessary tile pop in when viewing a previously loaded area (memory caching bug)


csciguy8:
1. Open the `cesium-unreal-samples` project

2. Open level `12_GoogleMapsTiles`

3. In the editor perspective window, you should see the Googleplex and surrounding area load

4. Move the camera 180 degrees so you are looking behind you

5. Move the camera back to the starting position

You will see the original tiles load again, faster, because they are cached on disk. This isn't the desired behavior though.

It is expected that you should **not** see any tiles load or pop in at all. The Cesium3DTileset's `Maximum Cached Bytes` is configured for 256MB, and all of these tiles only take up ~40MB in the disk cache (`cesium-request-cache.sqlite`). The tiles is in the original view should stay loaded in this scenario.

Thanks @kring for mentioning this last month.

kring:
I noticed that changing the `Maximum Cached Bytes` property from 268435456 (256MiB) to 26843545600 (25600MiB) makes the problem go away. So the problem does seem to be that tiles are being unloaded when you look away due to the cache size being exceeded, but I'm not sure why.

kring:
Turns out I was wrong about how the tile data size is measured for cache purposes. It's actually formed by adding up the number of bytes in all of the glTF buffers (`cesium.data.size()`) and images (`cesium.pixelData.size()`). That is _after_ things like Draco and JPG decompression, so it can be a lot larger than the size of the data downloaded over the network (even after it is gunzipped). A single 256x256 JPG ends up being 341KiB.

kring:
Right, for that Googleplex scene, the actual size of the traversed tiles is about 300 MiB. I'm still trying to be 100% certain it's correct, but I think it is. In which case the default 256MiB max cache for Google Photorealistic 3D Tiles is simply too small.

csciguy8:
Ok, if that's true, then my expectations were just wrong.

But this problem could lead to more work with this feature. For example, a new "auto" setting that determines a reasonable size based on your system memory or some other metrics. We could still have a fixed memory size for users that know exactly their device constraints.

kring:
[...]

PR #1329, opened 2024-01-11
Reduce texture memory usage on non-Windows systems

kring:
This is a simple change that significantly reduces the amount of (CPU) memory used by textures on non-Windows systems.

We create textures in most situations and on most platforms by populating Unreal's `FBulkData` class with the raw pixel data from the glTF or raster overlay. Then, sometime later, Unreal's uploads it to the GPU. I noticed while spelunking through the Unreal source code while working on #1327 that `FBulkData` has a flag: `BULKDATA_SingleUse`. When this is set, Unreal directly grabs the buffer from the FBulkData as the source of the GPU upload, and frees it afterward. It's single use because we can't use that buffer ourselves after Unreal has grabbed it. When it's _not_ set (which is the default), Unreal copies the bulk data again, uses the copy as the source of the upload, and leaves our original buffer in FBulkData intact.

We don't need the buffer in FBulkData - we already have our own copy of this data in the glTF - so `BULKDATA_SingleUse` is perfect for our use, and this PR adds it. Specifying this flags avoids a game-thread copy of the texture data. It also significantly reduces memory usage by texture data because the copy in the FBulkData doesn't sit around for the entire lifetime of the texture anymore.

This affects:

1. Every texture on non-Windows platforms (D3D11 and D3D12 use a custom texture creation path that is not affected by this change)
2. Feature metadata-related textures on all platforms (even Windows, because these textures don't use the custom texture creation path on Windows)

To measure the memory reduction, I hacked the Windows implementation to use the standard texture creation instead of the custom one. If you want to do this yourself, look for `GRHISupportsAsyncTextureCreation` in `CesiumTextureUtility.cpp` and replace it with `false`.

Then I opened up level 1 of the Samples project in the Editor and let it load. Before this change, the Editor process used 1.669 GB. Afterward, it used 1.538. Then I loaded the Google Photorealistic 3D Tiles sample level, and let that load. On that level, the Editor process went from 2.182 GB down to 1.918 GB with this change. These don't sound like huge changes percentage-wise, but the Editor uses a lot of memory and only some of it is our textures. Textures are our biggest use of memory in many cases, so eliminating a copy of the texture data is a big deal, especially on on devices with limited memory, like mobile and AR/VR devices.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,10 @@
 
 - Fixed a bug in `MLB_DitherFade` that made glTF materials with an `alphaMode` of `MASK` incorrectly appear as fully opaque.
 
+##### Additions :tada:
+
+- Significantly reduced CPU memory usage by textures on non-Windows systems.
+
 ### v2.2.0 - 2023-12-14
 
 ##### Breaking Changes :mega:
--- a/Source/CesiumRuntime/Private/CesiumEncodedFeaturesMetadata.cpp
+++ b/Source/CesiumRuntime/Private/CesiumEncodedFeaturesMetadata.cpp
@@ -171,6 +171,7 @@ std::optional<EncodedFeatureIdSet> encodeFeatureIdTexture(
         pFeatureIdImage->pixelData.size());
 
     pMip->BulkData.Unlock();
+    pMip->BulkData.SetBulkDataFlags(BULKDATA_SingleUse);
   }
 
   return result;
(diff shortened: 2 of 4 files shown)


azrogers:
Testing both levels, I see a 14.3% reduction on 01_CesiumWorld with this change (from 2.24GB to 1.92GB) and a 15.6% reduction in 12_CesiumGoogleMapsTiles (from 3.53GB to 2.98GB). Great change!

PR #1330, opened 2024-01-11
Further reduce texture memory usage by clearing pixelData in glTF images after creating Unreal textures.

kring:
This depends on CesiumGS/cesium-native#804 and CesiumGS/cesium-native#806 so merge them first.

After we create an Unreal texture from a glTF `ImageCesium`, we "mostly" don't need the `pixelData` anymore. Clearing it saves a lot of memory - an entire copy of the raw, uncompressed pixel data. This is an improvement comparable to #1329 _again_, and it applies even to Windows (unlike that PR).

So that's what this PR set out to do, but it kind of got away from me and ended up being a pretty substantial overall of how textures get created in Unreal. There are a bunch of benefits:

1. In all cases except metadata textures, we now free the CPU-side copy of the texture data once we've handed it off to Unreal. This was the original goal, and is an enormous reduction in texture memory usage. It would be easy to do this for metadata textures, too, but the assumption is that we (sometimes? always?) need access to that data on the CPU.
2. Metadata-related textures are now created using the same code path as all the other textures. Previously they used a separate, slower path. (note: I don't expect any major surprises here, but I'm not sure I was able to test every variation of a metadata texture. Bugs are possible, so some attention on this in review would be much appreciated!)
3. We avoid an extra copy of the texture data in a worker thread on non-Windows platforms. Previously, every time we created a texture on non-D3D platforms, we made one copy of the pixel data in a worker thread, and then copied it again in the renderer thread. Now we only copy it once, in the renderer thread. On D3D, we also make one copy, but in that case that one copy is made only in the worker thread, which is better (this didn't change in this PR).
4. When multiple glTF `texture` instances reference a single `image`, we now only create one copy of the texture data on the GPU. Previously we'd upload the image multiple times. This isn't super common, but it's a big win when it happens. Probably the most notable place it happens is when creating upsampled tiles for raster overlays. Previously the upsampled children would each get a full copy of the parent's textures. Now, they all share the parent's textures.

Some other notes:

1. Metadata textures are now created in `TEXTUREGROUP_8BitData` instead of `TEXTUREGROUP_World`. This is a little bit arbitrary, but I think it's an improvement.
2. Previously, the size of the `pixelData` was used to determine the size of texture data for caching purposes. Now that's pretty much always going to be zero, so the new `ImageCesium::sizeBytes` field (see CesiumGS/cesium-native#804) is used to track the size instead.


kring:
I'm using [TextureSettingsTest](https://github.com/KhronosGroup/glTF-Sample-Assets/tree/main/Models/TextureSettingsTest) to test support for multiple textures with different sampler settings referencing a single image. I'm using this tileset.json to display it:

```json
{
  "asset": { "version": "1.0" },
  "root": {
    "boundingVolume": {
      "sphere": [
        0,0,0,10
      ]
    },
    "children": [],
    "content": { "uri": "TextureSettingsTest.glb" },
    "geometricError": 0.0,
    "refine": "REPLACE"
  }
}
```

This is as close as we get to working, because double-sided materials are not supported in Cesium for Unreal:
image

It's doing a GPU copy of the texture data. There are edge cases, though. What if the first texture didn't have mipmaps, but a later one does? Ugh.

csciguy8:
@j9liu , if you have the time, can you help with a sanity check / test on anything that changed related to metadata?

I've already reviewed the dependent cesium-native changes, tested the unreal samples, and am looking at rest of the code the now.

csciguy8:
@kring any thoughts on how to test this properly?

I see all the unit tests passed, which is good, but I'm not seeing any unit tests that target texture loading / texture memory specifically.

I've also run through all the Unreal samples, in Windows, but don't have a non-windows device to check against.

As for code review, looks like a lot of "overhaul" type code, like you stated. Can definitely look for obvious issues, but there's a chance that anything subtle and non-obvious can slip through here.

csciguy8:
@kring One more thing, do you have any numbers of showing how much texture memory was saved?

Ex. In cesium-unreal-samples, I saw memory usage go from X to X, a X% reduction.

kring:
> I see all the unit tests passed, which is good, but I'm not seeing any unit tests that target texture loading / texture memory specifically.

It'll be extremely difficult to write tests for. We'd have somehow confirm that the GPU texture has been populated correctly. I'm open to ideas, but it doesn't seem worth the effort when it's very obvious very quickly just by looking if textures aren't being created successfully.

One thing to watch for is memory leaks. I've put a lot of effort into convincing myself there are none, but that's probably the biggest risk. Well, that and the risk that I've broken something related to metadata textures, because breaking those wouldn't be so obvious at a glance.

> One more thing, do you have any numbers of showing how much texture memory was saved?

It eliminates a complete copy of the texture in CPU memory. After this PR, there should be none left (as far as I know!), so the CPU texture memory usage goes from X to 0, for an infinite improvement! 😁

Counting CPU and GPU memory together, this will cut the texture memory usage in half.

kring:
Well I looked into it, and it's actually not too difficult to read back a texture in Unreal. So I added some tests. I also cleaned up some TODOs. Sadly this PR no longer removes more code than it adds, but it's probably worth it. 😁

kring:
I unfortunately had to mark the new tests `NonNullRHI`, meaning they effectively won't run on CI, because the null RHI doesn't allow to create a texture and then read back its pixels, so all the new tests fail. It's also not possible to run with a non-null RHI on the non-D3D11-capable GH Actions runners.

j9liu:
I ran into a crash earlier that I can't reproduce 🤔 It was because my 06_CesiumMetadata had been changed for #1360, and when I opened it, it crashed some class that was "SharedThread____" ? But after I `git restore`'d the level, I didn't see it again, and I have no idea what I could have done for my PR to induce the crash!

Unfortunately, I took a screenshot and pasted it here to post later, but I accidentally navigated to a different tab and now I can't reproduce it. I'm documenting this anyway in case I run into it again.

Source/CesiumRuntime/Private/CesiumEncodedFeaturesMetadata.cpp
```diff
@@ -130,48 +130,23 @@ std::optional<EncodedFeatureIdSet> encodeFeatureIdTexture(
...
-    encodedFeatureIdTexture.pTexture->sRGB = false;
-    // TODO: upgrade to new texture creation path
-    encodedFeatureIdTexture.pTexture->textureSource = LegacyTextureSource{};
+    // Copy the image, so that we can keep a copy of it in the glTF.
+    CesiumGltf::ImageCesium imageCopy(*pFeatureIdImage);
+    encodedFeatureIdTexture.pTexture =
+        MakeShared<LoadedTextureResult>(std::move(*loadTextureAnyThreadPart(
+            imageCopy,
+            TextureAddress::TA_Clamp,
+            TextureAddress::TA_Clamp,
```
j9liu, line 139: I just realized we didn't change this for https://github.com/CesiumGS/cesium-native/pull/793. But this can look exactly the same as the code in `encodePropertyTexture`, where we find the matching `TextureAddress`

Source/CesiumRuntime/Private/CesiumEncodedFeaturesMetadata.cpp
```diff
@@ -716,67 +672,49 @@ EncodedPropertyTexture encodePropertyTextureAnyThreadPart(
...
+          addressX = TextureAddress::TA_Wrap;
           break;
         case CesiumGltf::Sampler::WrapS::MIRRORED_REPEAT:
-          encodedProperty.pTexture->addressX = TextureAddress::TA_Mirror;
+          addressX = TextureAddress::TA_Mirror;
         case CesiumGltf::Sampler::WrapS::CLAMP_TO_EDGE:
         default:
-          encodedProperty.pTexture->addressX = TextureAddress::TA_Clamp;
+          addressX = TextureAddress::TA_Clamp;
         }
```
j9liu, line 687: Can we make this a helper function, since it comes up so often? Could put it in `CesiumTextureUtility`

Source/CesiumRuntime/Private/CesiumEncodedMetadataConversions.cpp
```diff
@@ -133,7 +133,7 @@ template <typename T>
 void coerceAndEncodeArrays(
     const FCesiumPropertyTablePropertyDescription& propertyDescription,
     const FCesiumPropertyTableProperty& property,
-    gsl::span<std::byte>& textureData,
+    const gsl::span<std::byte>& textureData,
```
j9liu, line 136: Not a suggestion, I'm just surprised we can add `const` here considering that we're overriding the contents of `textureData` 😮

Source/CesiumRuntime/Private/CesiumGltfComponent.cpp
```diff
@@ -78,33 +78,9 @@ using TMeshVector4 = FVector4f;
 static uint32_t nextMaterialId = 0;
 
 namespace {
-void destroyHalfLoadedTexture(
-    TUniquePtr<CesiumTextureUtility::LoadedTextureResult>& pHalfLoadedTexture) {
-  if (pHalfLoadedTexture) {
-    CesiumTextureUtility::destroyHalfLoadedTexture(*pHalfLoadedTexture.Get());
-  }
-}
 class HalfConstructedReal : public UCesiumGltfComponent::HalfConstructed {
```
j9liu, line 81: I'm confused on if we _need_ this `HalfConstructedReal` class anymore? Could it be merged with its base class `UCesiumGltfComponent::HalfConstructed`?

j9liu (commented):
@kring looks great! My comments are mainly tweaks + questions. We should probably resolve / merge those cesium-native PRs before we merge this, just to keep things tidy

Source/CesiumRuntime/Private/CesiumEncodedMetadataConversions.cpp
csciguy8: The `const `keyword is denoting that the `gsl::span` object is not modifiable. Internally it's just keeping track of a a pointer and size.

If you additionally changed this to `const gsl::span<const std::byte>`, this would declare that the byte contents are constant, and you'll get the compile error you would have originally expected.

Source/CesiumRuntime/Private/Cesium3DTileset.cpp
```diff
@@ -794,23 +809,18 @@ class UnrealResourcePreparer
...
+            pLoadedTexture.Get());
     if (!pTexture) {
       return nullptr;
     }
 
-    pTexture->AddToRoot();
-    return pTexture;
+    // Don't let this ReferenceCountedUnrealTexture be destroyed when the
+    // intrusive pointer goes out of scope.
+    pTexture->addReference();
```
csciguy8, line 822: For my sanity, the corresponding `::Release` to this `::addReference` is a couple lines up, in `::loadTextureGameThreadPart` correct?

csciguy8 (commented):
Did a pass over the code, just one question, but nothing jumps out as wrong or has issues. 

I like the new `ReferenceCountedUnrealTexture` class and wrangling lifetime control away from Unreal's GC.

csciguy8:
@kring back to you. There's a few comments left here, and also some comments in [cesium-native 806](https://github.com/CesiumGS/cesium-native/pull/806).

Appreciate the new texture unit tests, thanks. I think one of your big concerns was memory leaks, and indeed that is a general concern for the plugin right now, so I wrote a new issue to create memory leak tests.

I finished code review and didn't find anything major. It's a lot of code that's changed though, so it's very easy for subtle bugs to slip through. Hopefully unit testing will help. I'll also manual test a bit more through next week.

Source/CesiumRuntime/Private/CesiumEncodedMetadataConversions.cpp
kring: Yeah what Brian said, but I'll also point out that the `gsl::span` behavior is different from `std::vector`. If you have a const vector, the things in it are treated as const as well. This maybe makes sense because a vector owns its contents while a span just points to something owned elsewhere.

Source/CesiumRuntime/Private/CesiumGltfComponent.cpp
kring: It's a compiler firewall. `CreateOffGameThread` needs to be able to return a thing so that it can be passed to `CreateOnGameThread` and so that it can be freed later. But the actual contents of that thing are an implementation detail. If we were to put the actual contents of `HalfConstructedReal` into `HalfConstructed`, then any source file including CesiumGltfComponent.h would (by necessity) have a full declaration of all the contents of that class, which would require including a whole bunch more headers. So this approach makes compilation a bit faster; we have to recompile fewer source files and less often.

Source/CesiumRuntime/Private/Cesium3DTileset.cpp
kring: No, it's a couple lines down in `freeRaster`.

kring:
Thanks for the reviews @csciguy8 and @j9liu. I believe I've addressed all your comments.

Source/CesiumRuntime/Private/CesiumGltfComponent.cpp
j9liu: That makes sense, thanks! :pray:

j9liu:
Thanks @kring !

PR #1333, opened 2024-01-12
Fix jump at end of flights.

### LINKED ISSUE #1332: CesiumFlyToComponent "jumps" at the end due to spherical math that doesn't quite work on an ellipsoid
Here's a diagram to try to illustrate the problem:
image

The blue oval is a highly exaggerated ellipsoidal model of the Earth. To do a flight between the start and end points marked, `CesiumFlyToComponent` first computes a rotation axis and angle between those two points. At each step, it interpolates the angle and rotates the initial direction by that angle to produce the direction of the path at this step. It then scales this direction to the ellipsoid surface and adds a separately-computed height to it to get the actual position for the flight.

The problem, as shown in the diagram, is that scaling the final direction to the ellipsoid and then adding the final position's height does not actually result in the final position. It would on a sphere, but not on an ellipsoid. It's our old friend geocentric versus geodetic angles.

As a result, flights often appear to jump to a new position just as they complete, because the final position along the curve is quite different from the real final position. This can be seen quite obviously by opening the Google Photorealistic 3D Tiles level in Cesium for Unreal Samples, hitting Play-In-Editor, pressing 3 to fly to Tokyo, then moving the camera a good distance _up_ and away from Tokyo before pressing 3 to fly toward it again. When the intial position of the camera is fairly high, the jump at the end should be very obvious.

It's probably easy to fix this, but I'm not sure it's worth the trouble. I'm inclined to just port the camera flight code from CesiumJS rather than trying to fix this approach.

kring:
Fixes #1332

I wrote in the linked issue that I didn't think it was worth fixing the problem, versus porting the flight code from CesiumJS. But then I realized the fix is really easy: just scale the positions to the ellipsoid before computing the start and end directions. So here it is.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,7 @@
 ##### Fixes :wrench:
 
 - Fixed a bug in `MLB_DitherFade` that made glTF materials with an `alphaMode` of `MASK` incorrectly appear as fully opaque.
+- Fixed a bug in `CesiumFlyToComponent` that could cause the position of the object to shift suddenly at the very end of the flight.
 
 ### v2.2.0 - 2023-12-14
 
--- a/Source/CesiumRuntime/Private/CesiumFlyToComponent.cpp
+++ b/Source/CesiumRuntime/Private/CesiumFlyToComponent.cpp
@@ -6,6 +6,7 @@
 #include "GameFramework/Controller.h"
 #include "GameFramework/Pawn.h"
 #include "UObject/ConstructorHelpers.h"
+#include "VecMath.h"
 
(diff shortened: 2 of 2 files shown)


csciguy8 (approved):
Changes look good.

From the related issue and looking closer at this function, looks like our Start was good (`_sourceDirection`), but the total amount needed to rotate was incorrect (`_totalAngle`). This fixes that nicely.

csciguy8:
Thanks @kring !

Reviewed and tested out in 2 cesium-unreal-samples levels (sublevels, google tile).

Merging...

PR #1337, opened 2024-01-16
Track texture size stats in custom D3D texture loading path.

kring:
This PR makes the "Texture Memory Used" stat of the "Memory" group:

image

And the "TextureGroup" stats:

image

show the memory used by our textures even on D3D11 and D3D12. Previously, our custom texture creation path side-stepped Unreal's counter logic, so the memory used by our textures was conspiciously missing from these debug displays.

This was a pain because some things necessary to make it work are inexplicably private to the engine. Specifically `FTextureResource::TextureGroupStatFNames` is a public static field in a class declared in a public header file. But trying to use it results in linker errors on Windows because Epic didn't (forgot to?) mark it as DLL exported. So I had to duplicate the engine code that declares a stat group for each texture group.


azrogers:
Looks great to me!

PR #1338, opened 2024-01-17
Remove unnecessary parameters from glTF normals methods.

### LINKED ISSUE #1290: Probably incorrect vertex index in setUniformNormals function
In setUniformNormals we are iterating through vertices, but use size of indices array. So we can go out of boundaries in vertices array.
https://github.com/CesiumGS/cesium-unreal/blob/84ba3a121b45d8497ac2c9a9fd7fefebe32fe987/Source/CesiumRuntime/Private/CesiumGltfComponent.cpp#L287C16-L287C16

azrogers:
Fixes #1290.

As pointed out in the issue, this doesn't actually constitute a bug as these methods will only be called when `indices` and `vertices` are the same length. However, it's a bit confusing and the `indices` parameter is unnecessary regardless. This change removes the parameter from both methods and changes the loops to use the length of the vertices array instead.

### DIFF
--- a/Source/CesiumRuntime/Private/CesiumGltfComponent.cpp
+++ b/Source/CesiumRuntime/Private/CesiumGltfComponent.cpp
@@ -280,10 +280,9 @@ static void computeTangentSpace(TArray<FStaticMeshBuildVertex>& vertices) {
 }
 
 static void setUniformNormals(
-    const TArray<uint32_t>& indices,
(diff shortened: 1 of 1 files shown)


j9liu:
Looks great @azrogers ! Happy to merge once CI passes.

j9liu:
Merging now. Congrats on your first PR @azrogers !

PR #1340, opened 2024-01-17
Fix crash when enabling Ignore `KHR_materials_unlit` on point clouds

### LINKED ISSUE #1268: Ignore `KHR_materials_unlit` causes crash for point clouds
`computeFlatNormals` expects a triangle mesh, so it crashes when run on a point cloud. There should be a catch for this so the Editor doesn't crash.

azrogers:
Fixes #1268.

Enabling the setting to Ignore `KHR_materials_unlit` forces the glTF implementation to compute flat normals for the primitive, as per the glTF spec. However, we can't compute flat normals for points or line primitives, causing a crash if this setting was enabled on a point cloud. This change adds a check that ignores this setting if specified on a non-triangle primitive.

### DIFF
--- a/Source/CesiumRuntime/Private/CesiumGltfComponent.cpp
+++ b/Source/CesiumRuntime/Private/CesiumGltfComponent.cpp
@@ -1026,6 +1026,13 @@ static void loadPrimitive(
       !options.pMeshOptions->pNodeOptions->pModelOptions
            ->ignoreKhrMaterialsUnlit;
 
+  // We can't calculate flat normals for points or lines, so we have to force
+  // them to be unlit if no normals are specified. Otherwise this causes a
+  // crash when attempting to calculate flat normals.
+  if (primitive.mode < MeshPrimitive::Mode::TRIANGLES && !hasNormals) {
+    primitiveResult.isUnlit = true;
+  }
+
   const MaterialPBRMetallicRoughness& pbrMetallicRoughness =
       material.pbrMetallicRoughness ? material.pbrMetallicRoughness.value()
                                     : defaultPbrMetallicRoughness;
--- a/Source/CesiumRuntime/Private/CesiumGltfComponent.cpp
+++ b/Source/CesiumRuntime/Private/CesiumGltfComponent.cpp
@@ -1029,7 +1029,11 @@ static void loadPrimitive(
   // We can't calculate flat normals for points or lines, so we have to force
   // them to be unlit if no normals are specified. Otherwise this causes a
   // crash when attempting to calculate flat normals.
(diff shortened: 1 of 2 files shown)


Source/CesiumRuntime/Private/CesiumGltfComponent.cpp
```diff
@@ -1026,6 +1026,13 @@ static void loadPrimitive(
       !options.pMeshOptions->pNodeOptions->pModelOptions
            ->ignoreKhrMaterialsUnlit;
 
+  // We can't calculate flat normals for points or lines, so we have to force
+  // them to be unlit if no normals are specified. Otherwise this causes a
+  // crash when attempting to calculate flat normals.
+  if (primitive.mode < MeshPrimitive::Mode::TRIANGLES && !hasNormals) {
```
j9liu, line 1032: @kring @csciguy8 Purely style question, is it okay to hardcode checks like this where the values of spec constants are assumed? Or is it better to be more explicit, e.g.:

```
bool isTriangles = primitive.mode == MeshPrimitive::Mode::TRIANGLES ||
   primitive.mode == MeshPrimitive::Mode::TRIANGLE_FAN || 
   primitive.mode == MeshPrimitive::Mode::TRIANGLE_STRIP;
   
if (!isTriangles && !hasNormals) {
```

j9liu (commented):
Thanks @azrogers ! I confirmed that this fixes the crash. I just had a concern about code style but I'll defer to our teammates for their opinion :')

Also, can you add a line about this to CHANGES.md?

Source/CesiumRuntime/Private/CesiumGltfComponent.cpp
csciguy8: If I had to chose, I think I would prefer explicit.

With a < operator and enums, you're implicitly assuming a conversion to an integer, and in addition a specific order in the enum. Explicit avoids this.

Also, if this enum ever switched to the more modern scoped enum (enum class X), which doesn't as easily allow casts to an integer, explicit becomes a more pragmatic choice..

kring: Given that these constants come from a formal spec, they're extremely unlikely to change. So the original code should be pretty safe, and a tiny bit more performant. But it's a little clearer if it's explicit. You could get the best of both worlds (at the cost of more typing) by using a `static_assert` to prove the values are what you expect. And this would also serve as a bit of documentation. Probably overkill in this case, though. Probably just bike shedding here, I'm basically ok with any of the approaches.

j9liu:
Thanks @azrogers !

PR #1343, opened 2024-01-26
Update cesium-native, fix metadata test failures.

### LINKED ISSUE #1342: New metadata unit test failures caused by texture wrapping modes
There are a number of Cesium for Unreal test failures when we upgrade to the latest version of cesium-native. I think the cause is CesiumGS/cesium-native#793.

In the one I looked at in detail (Unit -> FeatureIdSet -> GetFeatureIDFromHit -> returns correct value for texture set), the problem is that the default wrapping mode is REPEAT, and the test doesn't change that. And as a result, now that we're honoring the wrap mode, the texture coordinates (1,0) get wrapped to (0,0). Arguably, wrapping a value that is not actually outside the 0-1 range is wrong? But using the REPEAT mode in this test is also wrong, I'd say. So I'm not sure if this is a bug in the test or the production code (or both).

kring:
Fixes #1342 

This PR just changes the tests, setting the wrapping mode to CLAMP_TO_EDGE where necessary. As I mentioned in the issue above, it's possible there's something we want to change outside of the tests as well, so that a texture coordinate of 1.0 doesn't get wrapped to 0.0.

### DIFF
--- a/Source/CesiumRuntime/Private/Tests/CesiumFeatureIdTexture.spec.cpp
+++ b/Source/CesiumRuntime/Private/Tests/CesiumFeatureIdTexture.spec.cpp
@@ -136,8 +136,13 @@ void FCesiumFeatureIdTextureSpec::Define() {
          image.cesium.channels = 1;
          image.cesium.pixelData.push_back(std::byte(42));
 
+         Sampler& sampler = model.samplers.emplace_back();
+         sampler.wrapS = Sampler::WrapS::CLAMP_TO_EDGE;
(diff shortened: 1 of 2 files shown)


j9liu:
Thanks @kring! I wrote these tests while `FeatureIdTextureView` didn't care about the texture's `Sampler`. Now that it does care, it's exposing the invalid test setup. 😅 

There's apparently an unrelated test failing in UE 5.3, so I'll see if I can reproduce + fix that before merging.

PR #1346, opened 2024-01-29
Replace `CesiumMetadataConversions` with `CesiumGltf::MetadataConversions`

j9liu:
Many of the type conversions in `CesiumMetadataConversions.h` were moved to `cesium-native` so they could be made available in Unity. This PR removes the duplicate functionality in Cesium for Unreal, replacing it with calls to `CesiumGltf::MetadataConversions` instead.

Some conversions were Unreal-reliant, e.g. converting to `FString` or `FVector3f`. So `CesiumMetadataConversions` has actually been renamed to `UnrealMetadataConversions`, and some relevant functions were added to convert from the `cesium-native` to the Unreal type.

Also, I realized that in `CesiumEncodedMetadataUtility.cpp` I had left some code commented out. So I uncommented the code + updated it for backwards compatibility.


kring:
This all looks good to me. Is there anything that needs to be mentioned in CHANGES.md? The newly-uncommented code, maybe?

j9liu:
@kring Good point, I added a note about re-adding the backwards compatibility for feature textures.

kring:
Thanks @j9liu!

PR #1347, opened 2024-01-29
Use `AccessorUtility` in cesium-native instead of `GltfAccessors`

j9liu:
While working on features / metadata in Cesium for Unity, I ported a lot of the `GltfAccessors.h` code from Unreal to cesium-native. The code is exactly the same, so there's no need to keep the duplicate code in Unreal. This PR uses the cesium-native functions / `typedef`s and deletes the Unreal-specific files.


kring:
This all looks good to me. Can you please take a look at the build failures on Linux and Android, though?

PR #1348, opened 2024-01-30
Let `AGlobeAwareDefaultPawn` use `SimplePlanarEllipsoidCurve`, add more tests

### LINKED ISSUE #1122: Create tests for AGlobeAwareDefaultPawn

azrogers:
Fixes #1122.

Currently, `AGlobeAwareDefaultPawn` only has a single test for a bug previously encountered. This change adds more tests that check its fly-to functionality.


csciguy8:
These tests look a lot like the tests in https://github.com/CesiumGS/cesium-native/pull/797.

What do you think about not bothering with the new Unreal tests, but just integrating with the native implementation, thus relying on the native tests instead?

(that said, Unreal side tests do have value, but I have a feeling most of the corner cases are in path calculation)

j9liu:
My fault for not reading the issue closely when I passed this to @azrogers. @csciguy8 was there anything else you were thinking of testing when you wrote up the issue? One thing I could think of is testing that it adjusts its orientation as it moves across the Earth.

Regarding the fly-to behavior, though, perhaps we can simplify the tests just to "make sure it works" without going to deeply into the math?

csciguy8:
> adjusts its orientation as it moves across the Earth

I like that, a test for  orientation sounds reasonable. And maybe a test to make sure the flight path doesn't fly through the earth? (height stays >0). 

As far as what goes into the Unreal side, maybe we can just check that we can start at X and arrive at Y when we expect? (verifying the integration piece). All the subtleties in the middle can be a cesium-native test. 

Open to discussion though. Just trying to avoid too much duplication and test coverage that won't really help us in the end.

csciguy8:
Modified the title a bit. We're doing a bit more than just adding tests :)

csciguy8 (approved):
Looks great. Thanks @azrogers .

Reviewed the code with the new SimplePlanarEllipsoidCurve integration and tested.

Will merge once CI finishes...

Issue #831, opened 2022-04-27
Crash when adding CesiumSunSky while the viewport is hidden


yuanlei88:
Add Cesium SunSky  error

Unhandled Exception: EXCEPTION_ACCESS_VIOLATION reading address 0x0000000000000030

UnrealEditor_CesiumRuntime!ACesiumSunSky::UpdateAtmosphereRadius() [D:\build\U5M-Marketplace\Sync\LocalBuilds\PluginTemp\HostProject\Plugins\CesiumForUnreal\Source\CesiumRuntime\Private\CesiumSunSky.cpp:439]
UnrealEditor_Engine!AActor::TickActor() [D:\build\++UE5\Sync\Engine\Source\Runtime\Engine\Private\Actor.cpp:1360]
UnrealEditor_Engine!FActorTickFunction::ExecuteTick() [D:\build\++UE5\Sync\Engine\Source\Runtime\Engine\Private\Actor.cpp:197]
UnrealEditor_Engine!FTickFunctionTask::DoTask() [D:\build\++UE5\Sync\Engine\Source\Runtime\Engine\Private\TickTaskManager.cpp:284]
UnrealEditor_Engine!TGraphTask<FTickFunctionTask>::ExecuteTask() [D:\build\++UE5\Sync\Engine\Source\Runtime\Core\Public\Async\TaskGraphInterfaces.h:975]
UnrealEditor_Core!FNamedTaskThread::ProcessTasksNamedThread() [D:\build\++UE5\Sync\Engine\Source\Runtime\Core\Private\Async\TaskGraph.cpp:753]
UnrealEditor_Core!FNamedTaskThread::ProcessTasksUntilQuit() [D:\build\++UE5\Sync\Engine\Source\Runtime\Core\Private\Async\TaskGraph.cpp:642]
UnrealEditor_Core!FTaskGraphCompatibilityImplementation::WaitUntilTasksComplete() [D:\build\++UE5\Sync\Engine\Source\Runtime\Core\Private\Async\TaskGraph.cpp:2169]
UnrealEditor_Engine!FTickTaskSequencer::ReleaseTickGroup() [D:\build\++UE5\Sync\Engine\Source\Runtime\Engine\Private\TickTaskManager.cpp:565]
UnrealEditor_Engine!FTickTaskManager::RunTickGroup() [D:\build\++UE5\Sync\Engine\Source\Runtime\Engine\Private\TickTaskManager.cpp:1592]
[...]

PR #1349, opened 2024-02-01
Add checks to CesiumSunSky to prevent possible crashes

azrogers:
There's been a few reports in our forums ([here](https://community.cesium.com/t/cesium-causes-access-violation-crash-in-unreal-5-3-2/28696) and [here](https://community.cesium.com/t/my-project-cannot-be-opened-hello-i-used-cesuim-to-make-the-project-for-a-month-suddenly-cannot-be-opened/29660)) that show some users are experiencing a crash on the line `FVector location = transform.TransformPosition(getViewLocation(this->GetWorld()));` in `CesiumSunSky.cpp`. While none of us are able to reproduce this bug, it seems likely that some sort of race condition is resulting in `this->GetWorld()` returning nullptr. Even if this isn't the source of these crashes, it's a good check to add anyways. I've also gone through `CesiumSunSky` to add checks to other locations where an unexpected nullptr might be possible and would cause a crash.

### DIFF
--- a/.github/workflows/buildWindows.yml
+++ b/.github/workflows/buildWindows.yml
@@ -66,7 +66,7 @@ jobs:
         run: |
           ((Get-Content -path CesiumForUnreal.uplugin -Raw) -replace '"EngineVersion": "5.0.0"','"EngineVersion": "${{ inputs.unreal-engine-version }}"') | Set-Content -Path CesiumForUnreal.uplugin
       - name: Build CesiumForUnreal plugin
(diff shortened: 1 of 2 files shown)


j9liu:
Wondering if this old issue is also related + can be fixed by this: https://github.com/CesiumGS/cesium-unreal/issues/831

j9liu:
Merging now that CI passed; thanks @azrogers !

PR #1352, opened 2024-02-05
Update ion token for performance tests

csciguy8:
They can't run without a valid ion token. Looks like we missed updating it on the last release.

Also update the release documentation that mentions a separate google key. This is no longer needed.
https://github.com/CesiumGS/cesium-unreal/blob/Update-ion-token-for-performance-tests/Documentation/release-process.md#verify-the-code


kring:
Thanks @csciguy8! Sorry I forgot to update this during the release process.

PR #1353, opened 2024-02-06
Update release documentation for UE5

j9liu:
While looking at #1352 I noticed that our release documentation was rather out-of-date, so I did a pass to clean it up and remove unnecessary steps (e.g. separate releases for UE4 and UE5).


csciguy8 (approved):
Look like good revisions.

Not sure if we have to mention the `04_MAIN_CesiumSublevels` scene and testing procedure anymore. There are unit tests that cover some of this. That said, I still use that level a lot, so it's an important one.

PR #1355, opened 2024-02-16
Update `Build.cs` for zlib-ng

j9liu:
This PR updates `CesiumRuntime.Build.cs` to refer to the correct zlib library. There were some linker errors due to https://github.com/CesiumGS/cesium-native/pull/798:

linker errors

PR #1360, opened 2024-02-21
Support `KHR_texture_transform` for feature ID textures

j9liu:
Depends on #1355 (which I still have to get working, sorry :sob:)

This adds support for `KHR_texture_transform` for feature ID textures, both with picking and styling. To make sure this actually worked, I used the sample model that CesiumJS uses for its specs: https://github.com/CesiumGS/cesium/tree/main/Specs/Data/Models/glTF-2.0/FeatureIdTextureWithTextureTransform/glTF.

**HOWEVER**, be sure to modify the glTF so that the feature ID set has `"featureCount": 64,`. This is required by the spec, so it's an error on the model's part.

To test **picking**:

- Open the 06_CesiumMetadata sample level
- Load a tileset with the aforementioned glTF in it.
- Modify the Level Blueprint to print the feature ID like so:
image
- Press play. If correct, the very top-left pixel should return the value "54". The bottom-right pixel will return 153.

To test **styling**:
- Add a CesiumFeaturesMetadata component to the tileset
- Click Auto Fill. You should see a feature ID texture with the "Has Khr Texture Transform" flag = true
- Click Generate Material. There should be nodes for the scale / offset / rotation from `KHR_texture_transform`:
image
- The base color texture is also used for feature IDs, so if the texture transform works, the square should appear _exactly_ the same. Configure the material graph to look like so.
image
- Create a new material instance with this as the layer (and **remember to name it** "FeaturesMetadata"!). Then assign it to the tileset. It should look _exactly_ the same. You can test this by clicking the "revert" button next to the material and checking if the appearance changes.
image

Bonus Changes:
- Reorganized the way Material Functions were passed in `CesiumFeaturesMetadataComponent` because I initially was using other material functions. But I've removed them and am just keeping the new organization because it seems cleaner.
- Vector parameters were hooked incorrectly for Feature ID textures, and I re-wrote the code to connect them properly. I will sweep through the other parameters and check for correctness in a following PR.


kring:
Thanks @j9liu!

javagl:
@j9liu 

> HOWEVER, be sure to modify the glTF so that the feature ID set has "featureCount": 64,. This is required by the spec, so it's an error on the model's part.

Just drop me a note when you notice something like this - I added this file without validating, so that's my fault.

Fix is in https://github.com/CesiumGS/cesium/pull/11863

j9liu:
Thanks for noticing @javagl -- I intended to create a fix, but got completely distracted. 😓

PR #1367, opened 2024-02-28
Support `KHR_texture_transform` in property texture properties

j9liu:
Depends on https://github.com/CesiumGS/cesium-native/pull/822, so merge that first.

Similar to #1360 . Use  https://github.com/CesiumGS/cesium/tree/main/Specs/Data/Models/glTF-2.0/PropertyTextureWithTextureTransform/glTF for this one.

### Picking

- Open the 06_CesiumMetadata sample level
- Load a tileset with the aforementioned glTF in it.
- Modify the Level Blueprint to print the feature ID like so:
image
- Press play. If correct, the very top-left pixel should return the value 54 / 255 = 0.21176470588. The bottom-right pixel should return 135 / 255 = 0.5294117647

### Styling

- Add a CesiumFeaturesMetadata component to the tileset
- Click Auto Fill. You should see a feature ID texture with the "Has Khr Texture Transform" flag = true
- Click Generate Material. There should be nodes for the scale / offset / rotation from `KHR_texture_transform`:
image
- The base color texture is also used for feature IDs, so if the texture transform works, the square should appear exactly the same. Configure the material graph to look like so.
image
- Create a new material instance with this as the layer (and remember to name it "FeaturesMetadata"!). Then assign it to the tileset. It should look exactly the same. You can test this by clicking the "revert" button next to the material and checking if the appearance changes.

### Bonus Changes

#1330 unfortunately made things complicated for the sample models used, because the base color texture _and_ the features / metadata texture used are the same. I feel like this isn't a robust fix, but what I've done as a band-aid for the release:

- Opened https://github.com/CesiumGS/cesium-native/pull/822.
- Before `FCesiumModelMetadata` is created, iterate over all of the primitive's materials, and find if any of the images are used by metadata. If so, we insert a `makeImageCopy = true` variable in the `extras`.
- Before `FCesiumPrimitiveFeatures` is created (and before material textures are loaded), do the same thing as above.
- If this extra is present, then the `TextureView` will be created with a CPU copy, which avoids picking errors.

@kring I know you suggested the approach of duplicating textures in the glTF itself, but I'm hesitant to modify the glTF images itself? (maybe unjustifiably so). But let me know if what I've done is atrocious.


Source/CesiumRuntime/Private/CesiumGltfComponent.cpp
```diff
@@ -471,26 +471,65 @@ static void applyWaterMask(
...
-  const ExtensionMeshPrimitiveExtStructuralMetadata* pMetadata =
-      primitive.getExtension<ExtensionMeshPrimitiveExtStructuralMetadata>();
-  if (!pMetadata) {
-    return FCesiumPrimitiveMetadata();
+static bool hasMaterialTextureConflicts(
+    const CesiumGltf::Model& model,
+    const CesiumGltf::Material& material,
+    int32_t imageIndex) {
+  if (material.pbrMetallicRoughness) {
+    auto maybeBaseColorTexture =
```
kring, line 491:
```suggestion
    auto& maybeBaseColorTexture =
```

Source/CesiumRuntime/Private/CesiumGltfComponent.cpp
```diff
@@ -471,26 +471,65 @@ static void applyWaterMask(
...
+        material.pbrMetallicRoughness->baseColorTexture;
+    if (maybeBaseColorTexture && textureUsesSpecifiedImage(
+                                     model,
+                                     maybeBaseColorTexture->index,
+                                     imageIndex)) {
+      return true;
+    }
+
+    auto maybeMetallicRoughnessTexture =
```
kring, line 500:
```suggestion
    auto& maybeMetallicRoughnessTexture =
```

Source/CesiumRuntime/Private/CesiumGltfComponent.cpp
```diff
@@ -850,6 +889,139 @@ static void updateTextureCoordinatesForMetadata_DEPRECATED(
...
+          featureId.texture->extras.insert({"makeImageCopy", true});
+        }
+      }
+    }
+  }
+
+  const ExtensionMeshPrimitiveExtStructuralMetadata* pMetadata =
+      primitive.getExtension<ExtensionMeshPrimitiveExtStructuralMetadata>();
+
+  auto pModelOptions = options.pMeshOptions->pNodeOptions->pModelOptions;
```
kring, line 945: In general I think we should avoid using auto like this, because it makes it harder to read the code because it's not clear at a glance what the type is. Especially in a review on GitHub where Intellisense can't help. And that in turn makes it harder to tell if the code is correct. Is this an object that is going to be _copied_ if it's assigned to an auto (and so changing it to `auto&` would be preferable)? Or is it just a raw pointer so it doesn't matter? The `p` prefix makes it possible to make a pretty good guess in this case, of course.

In general, I prefer to only use auto in cases where the type is obvious (`auto p = std::make_unique<Whatever>()` is fine) or impossible to express (such as a lambda).

Source/CesiumRuntime/Private/CesiumPropertyTexture.cpp
```diff
@@ -26,12 +26,33 @@ FCesiumPropertyTexture::FCesiumPropertyTexture(
...
   }
 
-  propertyTextureView.forEachProperty([&properties = _properties](
-                                          const std::string& propertyName,
-                                          auto propertyValue) mutable {
-    FString key(UTF8_TO_TCHAR(propertyName.data()));
-    properties.Add(key, FCesiumPropertyTextureProperty(propertyValue));
-  });
+  const CesiumGltf::Class* pClass = propertyTextureView.getClass();
+  for (auto classPropertyIt : pClass->properties) {
```
kring, line 30:
```suggestion
  for (const auto& classPropertyPair : pClass->properties) {
```

kring (commented):
Looks good, just minor comments.

kring:
@j9liu when you've addressed the minor comments here and in https://github.com/CesiumGS/cesium-native/pull/822, please go ahead and merge them yourself. That way I won't need to wait for CI before I can start the release process at the end of your day today.

j9liu:
Thanks @kring for the review! Merging with your approval.

PR #1368, opened 2024-02-29
`apt update` before `apt remove`

kring:
Otherwise an outdated package source seems to break the removal process and cause CI builds to fail. Which is infurating.


kring:
Merging this to unbreak CI.

PR #1371, opened 2024-03-07
Add support for EXT_mesh_gpu_instancing

timoore:
Create UInstancedStaticMeshComponent objects for meshes that use the EXT_mesh_gpu_instancing extension. Nothing is done yet about metadata.

I used multiple inheritance to share the Cesium data that was in UCesiumGltfPrimitiveComponent with the new UCesiumGltfInstancedComponent class, but this is kind of awkward in the context of Unreal's automatically generated reflection support et al. I'd like to get feedback on whether it would be better to use a composition-based approach instead.


csciguy8:
> I'd like to get feedback on whether it would be better to use a composition-based approach instead.

While multiple inheritance doesn't seem to be explicitly disallowed in [UE's coding standard](https://docs.unrealengine.com/4.26/en-US/ProductionPipelines/DevelopmentSetup/CodingStandard/), from reading discussions on UE forums, it seems that multiply inheriting from interfaces is the general accepted pattern.

I think I understand your approach and seems like it could work. If it was up to my preferences, I'd probably lean toward a single inheritance approach here
* `UCesiumGltfPrimitiveComponent` derives from `UStaticMeshComponent` only
* `UCesiumGltfPrimitiveComponent` contains an instance of `CesiumGltfPrimitiveBase`
* Pros - More UE coding paradigm friendly, no multi-inheritance means no multi-inheritance complexities (problems)
* Cons - Maybe more duplicated code than you want, but maybe not that bad in the end?

timoore:
> > I'd like to get feedback on whether it would be better to use a composition-based approach instead.
> 
> While multiple inheritance doesn't seem to be explicitly disallowed in [UE's coding standard](https://docs.unrealengine.com/4.26/en-US/ProductionPipelines/DevelopmentSetup/CodingStandard/), from reading discussions on UE forums, it seems that multiply inheriting from interfaces is the general accepted pattern.
> 
> I think I understand your approach and seems like it could work. If it was up to my preferences, I'd probably lean toward a single inheritance approach here
> 
> * `UCesiumGltfPrimitiveComponent` derives from `UStaticMeshComponent` only
> * `UCesiumGltfPrimitiveComponent` contains an instance of `CesiumGltfPrimitiveBase`
> * `UCesiumGltfInstancedComponent` derives from `UInstancedStaticMeshComponent` only
> * `UCesiumGltfInstancedComponent` contains an instance of `CesiumGltfPrimitiveBase`
> * Pros - More UE coding paradigm friendly, no multi-inheritance means no multi-inheritance complexities (problems)
> * Cons - Maybe more duplicated code than you want, but maybe not that bad in the end?

Thanks for the feedback. I'm starting to look at this in detail. The one issue I have with moving to a single inheritance approach -- where UCesiumGltfPrimitiveComponent and UCesiumGltfInstancedComponent don't share a common Cesium base class -- is how to implement UpdateTransformFromCesium(). While it's nice that it's virtual in my current implementation, I suppose a combination of Cast and a template function could get around it. I'm going to continue to play it and try to figure out if multiple inheritance is really that bad here.

Source/CesiumRuntime/Private/CesiumGltfPrimitiveComponent.h
```diff
@@ -22,14 +28,9 @@ struct Model;
 struct MeshPrimitive;
 } // namespace CesiumGltf
 
-UCLASS()
-class UCesiumGltfPrimitiveComponent : public UStaticMeshComponent {
-  GENERATED_BODY()
-
+class CesiumGltfPrimitiveBase {
```
kring, line 31: Yeah sorry Tim I'm not a huge fan of this multiple inheritance approach. My main complaints are:

1. It requires some implementation to move to the .h file, and the corresponding headers to be included. I'm pretty big on keeping as much as we can out of the headers because it increases compile time and object file size (we have to ship object files for every platform, so it adds up).
2. The `_uobject` bit. It's hacky. Especially since it's public.
3. Multiple implementation inheritance is widely considered something to avoid.

I mean, these things aren't huge, so if we were getting some significant advantage from it, I'd be willing to let it go. But I don't think we are. The `UpdateTransformFromCesium` can easily be handled by delegating to a helper function in the two classes. It probably doesn't even need to be a template; I think everything that method uses is available on `UPrimitiveComponent` which is the common base class.

As for all those public fields in `CesiumGltfPrimitiveBase`... well, they're a bit of a mess anyway, aren't they? Ideally I'd like to see thoughtful, Blueprint-accessible accessors for the ones that make sense, and the rest made private. But that's a lot of scope creep for this PR, so I'd be happy enough if you just copied them into the `UCesiumGltfInstancedComponent` (a template function in CesiumGltfCompnent.cpp can be used to set them). Or, better yet, moved them to a common object that is aggregated in both classes (this is a breaking change, but no one is likely to care. We just need to document it). Or... better still... do we even need most of them? A lot of them are probably only relevant for things that aren't supported yet, like metadata. So just leaving them out of the new class for now is an excellent option.

kring (changes requested):
Looks great @timoore! I tried it out and it's working well, including when moving the goereference origin. Most of my comments are fairly nitpicky, though I would like to see the multiple inheritance based approach change (sorry!), which will be bigger. Let me know if you have any questions or disagree with anything.

Source/CesiumRuntime/Private/CesiumGltfComponent.cpp
```diff
@@ -1817,6 +1820,151 @@ static void loadMesh(
...
+  // operation is correct.
+  std::vector<glm::dmat4> instanceTransforms(count);
+  for (auto& mat : instanceTransforms) {
+    mat = glm::dmat4(1.0);
+  }
+  // Note: the glm functions translate() and scale() post-multiply the matrix
+  // argument by the new transform. E.g., translate() does *not* translate the
+  // matrix.
+  if (translations) {
+    // XXX Need to verify float scalar3.
```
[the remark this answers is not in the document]
timoore: It's a note to verify that the format of the instance data conforms to the extension specification, but it's garbled. Translation and scale data should be VEC3<float>; I will check that using the AccessorView status.

Source/CesiumRuntime/Private/CesiumGltfPrimitiveComponent.h
timoore: > Yeah sorry Tim I'm not a huge fan of this multiple inheritance approach. My main complaints are:
> 
> 1. It requires some implementation to move to the .h file, and the corresponding headers to be included. I'm pretty big on keeping as much as we can out of the headers because it increases compile time and object file size (we have to ship object files for every platform, so it adds up).
> 2. The `_uobject` bit. It's hacky. Especially since it's public.
> 3. Multiple implementation inheritance is widely considered something to avoid.

While I don't agree with your points in general, I have to agree in the context of an Unreal plugin. _uobject is only needed because regular dynamic_cast isn't available; Unreal turns off rtti. Otherwise, I think multiple inheritance is a fine way to add a common facet of data and behavior to an inheritance chain that has a very obvious "single inheritance" axis. Again, downcasting is very awkward in the Unreal universe.

I've committed the change. I think the the results a messier and more verbose than using multiple inheritance, but then there's no accounting for taste 😄 
> 
> 
> As for all those public fields in `CesiumGltfPrimitiveBase`... well, they're a bit of a mess anyway, aren't they?

Not really for me to say.

 >Ideally I'd like to see thoughtful, Blueprint-accessible accessors for the ones that make sense, and the rest made private. But that's a lot of scope creep for this PR, so I'd be happy enough if you just copied them into the `UCesiumGltfInstancedComponent` (a template function in CesiumGltfCompnent.cpp can be used to set them). Or, better yet, moved them to a common object that is aggregated in both classes (this is a breaking change, but no one is likely to care. We just need to document it). Or... better still... do we even need most of them? A lot of them are probably only relevant for things that aren't supported yet, like metadata. So just leaving them out of the new class for now is an excellent option.

I'm hoping to use them soon for instances. Maybe we can have another go-around then.

kring: > _uobject is only needed because regular dynamic_cast isn't available; Unreal turns off rtti.

I haven't looked at your changes yet, but a quick note here... Unreal has its own RTTI system for UObjects. You can downcast using the `Cast<T>` function. It works much like `dynamic_cast<T>`, e.g.: 

```cpp
UObject* pWhatever = ...;
UCesiumGltfPrimitiveComponent* pPrimitive = Cast<UCesiumGltfPrimitiveComponent>(pWhatever);
if (pPrimitive) {
  // ...
}
```

Source/CesiumRuntime/Private/CesiumGltfPrimitiveComponent.h
```diff
@@ -105,16 +106,118 @@ class UCesiumGltfPrimitiveComponent : public UStaticMeshComponent {
...
 
-  /**
-   * Updates this component's transform from a new double-precision
-   * transformation from the Cesium world to the Unreal Engine world, as well as
-   * the current HighPrecisionNodeTransform.
-   *
-   * @param CesiumToUnrealTransform The new transformation.
-   */
-  void UpdateTransformFromCesium(const glm::dmat4& CesiumToUnrealTransform);
+  void BeginDestroyPrimitive();
```
kring, line 109: This could use a better name. The overridden method on the actual UObjects has to be called `BeginDestroy` to hook into Unreal, but here the `Begin` part doesn't make much sense. Something like `Destroy` or `Clear` would be fine IMO.

Source/CesiumRuntime/Private/CesiumGltfPrimitiveComponent.h
```diff
@@ -105,16 +106,118 @@ class UCesiumGltfPrimitiveComponent : public UStaticMeshComponent {
...
+
+  friend CesiumGltfPrimitiveBase* getPrimitiveBase(USceneComponent* Component);
+  friend const CesiumGltfPrimitiveBase*
+  getPrimitiveBase(const USceneComponent* Component);
+
+protected:
+  CesiumGltfPrimitiveBase _cesiumBase;
 };
+
+CesiumGltfPrimitiveBase* getPrimitiveBase(USceneComponent* pComponent);
```
kring, line 221: This should be a virtual method on an interface (perhaps `ICesiumTile`) that is implemented by both primitive types, rather than a global function that uses a dynamic cast. See here for how to create interfaces in UE:
https://dev.epicgames.com/documentation/en-us/unreal-engine/interfaces-in-unreal-engine?application_version=5.2

Source/CesiumRuntime/Private/CesiumGltfPrimitiveComponent.h
```diff
@@ -105,16 +106,118 @@ class UCesiumGltfPrimitiveComponent : public UStaticMeshComponent {
...
-   * the current HighPrecisionNodeTransform.
-   *
-   * @param CesiumToUnrealTransform The new transformation.
-   */
-  void UpdateTransformFromCesium(const glm::dmat4& CesiumToUnrealTransform);
+  void BeginDestroyPrimitive();
+};
+
+template <typename UnrealType>
+void UpdateTransformFromCesium(
```
kring, line 113: This function and `BeginDestroyPrimitive` below should be moved to the .cpp file and the anonymous namespace. If they need to be called externally, they should be added to the `ICesiumTile` interface (see above), and the virtual methods can be implemented by calling these templatized functions.

Source/CesiumRuntime/Private/CesiumGltfPrimitiveComponent.h
```diff
@@ -105,16 +106,118 @@ class UCesiumGltfPrimitiveComponent : public UStaticMeshComponent {
...
+    if (pBodySetup) {
+      CesiumLifetime::destroy(pBodySetup);
+    }
+
+    CesiumLifetime::destroy(pMesh);
+  }
+}
+
+template <typename T>
+class CesiumGltfPrimitive : public CesiumGltfPrimitiveBase {
```
kring, line 170: Is this still used? Doesn't look like it.

kring (changes requested):
Thanks for getting rid of the multiple inheritance. A few comments on the new approach, though...

Source/CesiumRuntime/Private/CesiumGltfPrimitiveComponent.h
```diff
@@ -22,14 +28,9 @@ struct Model;
...
-UCLASS()
-class UCesiumGltfPrimitiveComponent : public UStaticMeshComponent {
-  GENERATED_BODY()
-
+class CesiumGltfPrimitiveBase {
 public:
-  // Sets default values for this component's properties
-  UCesiumGltfPrimitiveComponent();
-  virtual ~UCesiumGltfPrimitiveComponent();
+  CesiumGltfPrimitiveBase();
```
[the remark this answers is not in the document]
timoore: I finally chose CesiumPrimitiveData. CesiumTileData doesn't seem write as the data for each primitive, not the whole tile.

Source/CesiumRuntime/Private/CesiumGltfPrimitiveComponent.h
timoore: Thinks for that link, it's very helpful. I would note that its example shows inheriting from the interface via multiple inheritance, but that would be unproductive 😎

kring: >  I could note that its example shows inheriting from the interface via multiple inheritance

Sure, that's the only way to implement interfaces in C++. Multiple inheritance is widely considered acceptable when it's inheritance of an interface only (no implementation, no fields). Inheriting from multiple classes with fields has extra gotchas.

This goes back to Effective C++, item 40 if my quick google search is to be believed (I don't have the book handy).

As always with these sorts of rules of thumb ("don't use multiple inheritance, unless all but one is an interface"), there are exceptions. But this isn't one of them.

Source/CesiumRuntime/Private/CesiumGltfPrimitiveComponent.h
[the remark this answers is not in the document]
kring: Oh yes, good point. `CesiumPrimitiveData` works for me.

kring:
I noticed a crash while loading an i3dm tileset from the CesiumJS account on Cesium ion, asset ID 3877. Looks like it's dereferencing an empty optional.

image

After a little investigation, I think this is a glTF 1.0 tileset, which cesium-native doesn't support at all, so I don't expect it to work. But it shouldn't crash. It should warn about an unsupported version, like it does for the same situation in b3dm.

Source/CesiumRuntime/Private/CesiumGltfComponent.cpp
```diff
@@ -1817,6 +1820,151 @@ static void loadMesh(
...
+          translationAccessor[i].value[2]);
+      instanceTransforms[i] = glm::translate(instanceTransforms[i], transVec);
+    }
+  }
+  if (rotations) {
+    createAccessorView(model, *rotations, [&](auto&& quatView) -> void {
+      using QuatType = decltype(quatView[0]);
+      using ValueType = std::decay_t<QuatType>;
+      if constexpr (is_float_quat_v<ValueType>) {
+        for (int i = 0; i < count; ++i) {
```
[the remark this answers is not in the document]
kring: You marked this resolved but didn't resolve it.

Source/CesiumRuntime/Private/CesiumGltfPrimitiveComponent.cpp
```diff
@@ -100,17 +54,82 @@ void UCesiumGltfPrimitiveComponent::BeginDestroy() {
...
+  }
+
+  return std::visit(
+      CalcBoundsOperation{
+          LocalToWorld,
+          getPrimitiveData()->HighPrecisionNodeTransform},
+      *getPrimitiveData()->boundingVolume);
+}
+
+FBoxSphereBounds UCesiumGltfInstancedComponent::CalcBounds(
```
kring, line 82: This function and the one above are nearly identical. I'm not too fussed about it, but it might be nice to eliminate it since it's pretty easy to do. Something like this (not tested):

```cpp
namespace {

std::optional<FBoxSphereBounds> calcBounds(ICesiumPrimitive& primitive, const FTransform& LocalToWorld) {
  if (!primitive.boundingVolume) {
    return std::nullopt;
  }

  return std::visit(
      CalcBoundsOperation{
          LocalToWorld,
          primitive.HighPrecisionNodeTransform},
      *primitive.boundingVolume);
}

}

FBoxSphereBounds UCesiumGltfPrimitiveComponent::CalcBounds(
    const FTransform& LocalToWorld) const {
  std::optional<FBoxSphereBounds> maybeBounds = calcBounds(getPrimitiveData(), LocalToWorld);
  return maybeBounds ? *maybeBounds : Super::CalcBounds(LocalToWorld);
}

```

Source/CesiumRuntime/Private/CesiumGltfPrimitiveComponent.cpp
```diff
@@ -20,77 +19,32 @@ PRAGMA_DISABLE_DEPRECATION_WARNINGS
...
-    this->SetRelativeTransform_Direct(transform);
-    this->UpdateComponentToWorld();
-    this->MarkRenderTransformDirty();
-    this->SendPhysicsTransform(ETeleportType::ResetPhysics);
-  }
-}
-
-void UCesiumGltfPrimitiveComponent::BeginDestroy() {
+namespace {
+void destroyCesiumPrimitive(UStaticMeshComponent* uobject) {
```
kring, line 35: We're a little inconsistent about naming in Cesium for Unreal. Sometimes we use the cesium-native conventions (local variable and parameter names start with a lowercase letter, pointers prefixed with `p`), and sometimes we use the Unreal Engine conventions (capital first letter, no `p`). The latter is important for the bits that are public, because following Unreal's standard makes it easier for people to use our plugin. The former I guess we just do where it doesn't matter out of habit from working in cesium-native. But `uobject` is kind of a third way. At the risk of nitpicking, I think I'd rename this to either `pComponent` or `Component`.

Source/CesiumRuntime/Private/CesiumGltfPrimitiveComponent.h
```diff
@@ -3,118 +3,56 @@
...
-#include "CesiumRasterOverlays.h"
+#include "CesiumPrimitive.h"
+#include "Components/InstancedStaticMeshComponent.h"
 #include "Components/StaticMeshComponent.h"
 #include "CoreMinimal.h"
-#include <CesiumGltf/AccessorUtility.h>
-#include <cstdint>
-#include <glm/mat4x4.hpp>
-#include <unordered_map>
+// #include "Engine/StaticMesh.h"
```
kring, line 10: Remove this line.

kring (commented):
A few last details, but this is looking really close! If you're able to address these comments Thursday or Friday, we should be able to include it in the release on Monday.

Source/CesiumRuntime/Private/CesiumPrimitive.h
```diff
@@ -0,0 +1,136 @@
...
+    return &_cesiumData;
+  }
+  /*
+   * @brief Reset the component's physics transform. The Unreal function to do
+   * this, SendPhysicsTransform(), is protected.
+   */
+  virtual void resetPhysicsTransform() = 0;
+
+private:
+  CesiumPrimitiveData _cesiumData;
```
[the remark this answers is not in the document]
timoore: I'll make your proposed fix, but I think that you're fixating a bit too much on "Interface." I agree that ICesiumPrimitive isn't a real interface class, but I'm forced to pretend it is one to fit in the Unreal RTTI scheme with generated bodies, Cast<>, etc. CesiumPrimitive is a common inchoate part of the Cesium mesh types, and as such has common data; it has common behavior too, but we can't really model that due to Unreal's class hierarchy. The C++ Core Guidelines make it clear that multiple inheritance is a legit strategy for implementing such classes.

I  personally don't think it is more elegant to duplicate the _cesiumData definitions in the derived classes, just to keep data definitions out of an interface class... which isn't actually an interface 😎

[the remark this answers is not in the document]
kring: IMO if you're using a UInterface for something that isn't an interface, you're abusing the UInterface mechanism, and you're asking for trouble. Sometimes that can be justified, but here the cost of playing by the rules is really low.

Source/CesiumRuntime/Private/CesiumGltfPrimitiveComponent.cpp
timoore: I did change this, but I'm curious: does the "pComponent" convention extend to things that behave like pointers, like std::shared_ptr or std::unique_ptr?

kring: > I'm curious: does the "pComponent" convention extend to things that behave like pointers, like std::shared_ptr or std::unique_ptr?

Yes, smart pointers count as pointers. An explanation of why we use the `p` prefix is in the [style guide](https://github.com/CesiumGS/cesium-native/blob/main/doc/style-guide.md):

> DON'T use Hungarian notation, except that it's useful to prefix pointer-like variables with a p, e.g. pThing. This is justified because it's not meant to convey type information (not so useful), but rather as an easily-understood abbreviation for the descriptive name of what the variable actually is (very useful). The variable isn't a thing, it's a pointer to a thing, but pointerToThing would get annoying quickly. (see [NL.5](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#nl5-avoid-encoding-type-information-in-names))

Although not called out in the style guide explicitly, this extends to other similar cases as well. For example, we often prefix the name of variables of std::optional with `maybe`. Once again, this isn't Hungarian notation, it's a description of what the variable actually is. If an `int32_t` is an index, a `std::optional<int32t_t>` could be an `optionalIndex` or `maybeAnIndex` or, our usual convention, `maybeIndex`.

kring:
Your changes look good, thanks @timoore! It looks like you may have missed my comment about the crash [above](https://github.com/CesiumGS/cesium-unreal/pull/1371#issuecomment-2138534687), though. I'll check if that's still an issue and fix it before the release if so. Merging!

PR #1386, opened 2024-04-04
Improve DynamicPawn movement after flight is interrupted

### LINKED ISSUE #1238: Improve DynamicPawn movement after `UCesiumFlyToComponent` flight is interrupted
In `04_CesiumSublevels`, try interrupting any of the flights when the pawn is far above the Earth. Sometimes you can still move, but sometimes you are stuck there until you try another flight and move closer to the Earth.

I've seen this behavior sometimes even in the old implementation, so I don't think this is related to the 2.0 changes. It's still worth revisiting the DynamicPawn logic so we can prevent the camera from being stuck high above the Earth.

csciguy8:
Closes #1238 .

Fix getting stuck in space when interrupting a flight by removing block that rejects large speed increases. For cases where a tile is unloaded, the previous "height trace return 0" test covers this...

Removal Block Capture

While we're here, fix case where you can interrupt flight in space, fly back down to the ground, and your speed is gigantic. Do this by updating our speed when our camera location changes, not just when there's input.

Added "Did movement happen this tick?" block (previously used `Get Last Movement Input Vector`)

Addition Block Capture


j9liu:
@csciguy8 Just letting you know that I'm not ignoring these PRs, but I'll probably wait until after the release this week to review / merge them.

j9liu:
Thanks @csciguy8 ! It seems like this works well, I can't seem to get the camera stuck despite my numerous attempts :smile:

Will merge when CI passes

csciguy8:
@j9liu Thanks! Just one caveat with this chain of PRs.

I did find a bug, and fixed it, but snuck the fix to the[ last PR in this chain](https://github.com/CesiumGS/cesium-unreal/pull/1390).

Here's the commit details...
https://github.com/CesiumGS/cesium-unreal/pull/1390/commits/110e77896c0b1323073f75b7765214acf780de6b

To repro, just start cesium-unreal-samples, level 13 (google tiles), and enter PIE. You'll get stuck.

PR #1387, opened 2024-04-04
Smooth DynamicPawn speed changes

csciguy8:
Builds on #1386 , so merge that first (especially since these are binary assets).

Fix abrupt and jarring speed changes when navigating the DynamicPawn. 

This is visible in examples like cesium-unreal-samples, `04_CesiumSublevels`. Start PIE mode, press A to move left. You'll quickly leave the rooftop and see your speed increase very abruptly. Press D to move back and you'll see your speed decrease abruptly as your position changes back to the rooftop.

The problem is that pawn speed is being set instantly from height queries. This PR changes this behavior to set a target speed, and continuously interpolate to it. This smooths sudden changes, but as a side effect, creates a lag before the final target is reached.

Feedback welcome. Getting the feel just right may take some iterations.

image


csciguy8:
> I do wish it was slightly faster when leaving the top of a building, but that might be specific to this use case

The logic could probably be modified to lerp at the current speed when slowing down, but lerp faster when speeding up.

When now it just has a "x 3" fudge factor. I don't think there's any rule here as to what feels right...

image

j9liu:
> When now it just has a "x 3" fudge factor. I don't think there's any rule here as to what feels right...

Yeah I saw that, but I'm worried about the implications of having a skewed acceleration  / deceleration rate. For example, that could be catastrophic for the far away from Earth -> ground case, where not decelerating fast enough has made us go underground.

csciguy8:
> When we're stuck underground, it's difficult to get back up (especially because there's no longer anything underneath us to raycast). I'm not sure off the bat what modification you should make to catch this :/

I'm kind of wondering how you flew though the ground to start with. Just because you're going fast shouldn't mean that collisions with terrain get disabled.

It may be possible to work around this though. Maybe some logic like "If we're moving mostly in the direction of our ray cast (towards center of the earth), respond to our speed changes quickly. Else if we're moving somewhat perpendicular, smooth it out."

csciguy8:
Ok @j9liu , I think I have a fix for flying through the world.

I just disabled speed smoothing if over a certain speed. Let me know how it feels!

image

csciguy8:
One more set of changes @j9liu . (all done, I promise!)

I decided I did want to fix the "fly through the ground" problem by turning off speed interpolation, but decided to do it with a height threshold, rather than a speed threshold. This seemed a bit more intuitive.

image

Also fixed movement getting stuck sometimes at PIE start (added check for movement input vector)...

image

j9liu:
Thanks @csciguy8 ! This is much better.

I found that when I use the scroll wheel to add speed multiplier, I'm still able to phase through the ground. However, this happens even on `main`, so I'll write an issue for it so it doesn't hold up this PR.

PR #1396, opened 2024-04-17
Add support for Unreal Engine 5.4

csciguy8:
Draft PR until official UE 5.4 is available


CHANGES.md
```diff
@@ -4,6 +4,7 @@
 
 ##### Fixes :wrench:
 
+- Fixed any C++ and shader errors that occur when building under Unreal Engine 5.4
```
j9liu, line 7: Usually we count support for new UE versions as an "addition" instead of a fix. You can see the changelog [here](https://github.com/CesiumGS/cesium-unreal/blob/main/CHANGES.md#v1310---2023-09-20) for what we said for 5.3.

Source/CesiumRuntime/CesiumRuntime.Build.cs
```diff
@@ -211,7 +211,12 @@ public CesiumRuntime(ReadOnlyTargetRules Target) : base(Target)
 
         PCHUsage = PCHUsageMode.UseExplicitOrSharedPCHs;
         PrivatePCHHeaderFile = "Private/PCH.h";
+
+#if UE_5_4_OR_LATER
+        CppStandard = CppStandardVersion.Cpp20;
```
j9liu, line 216: Just checking, is C++ 20 now the min version in 5.4?

Source/CesiumRuntime/Private/CesiumCreditSystem.cpp
```diff
@@ -41,20 +42,25 @@ ACesiumCreditSystem* findValidDefaultCreditSystem(ULevel* Level) {
         TEXT("No valid level for findValidDefaultCreditSystem"));
     return nullptr;
   }
+#if ENGINE_VERSION_5_4_OR_HIGHER
+  TArray<TObjectPtr<AActor>>& Actors = Level->Actors;
+  TObjectPtr<AActor>* DefaultCreditSystemPtr;
```
j9liu, line 47: Might be good form to set it explicitly to nullptr here?

```suggestion
  TObjectPtr<AActor>* DefaultCreditSystemPtr = nullptr;
```

Source/CesiumRuntime/Private/CesiumCreditSystem.cpp
```diff
@@ -41,20 +42,25 @@ ACesiumCreditSystem* findValidDefaultCreditSystem(ULevel* Level) {
...
-        }
-        if (!InItem->IsA(ACesiumCreditSystem::StaticClass())) {
-          return false;
-        }
-        if (!InItem->GetName().StartsWith("CesiumCreditSystemDefault")) {
-          return false;
-        }
-        return true;
-      });
+  AActor** DefaultCreditSystemPtr;
```
j9liu, line 50: Same here:
```suggestion
  AActor** DefaultCreditSystemPtr = nullptr;
```

j9liu (commented):
Thanks @csciguy8 ! I just had a few comments.

I'll test compilation with each engine for compatibility, just to be sure it works across the board. Will update my checklist as I go:

- [x] 5.1
- [x] 5.2
- [ ] 5.3
- [ ] 5.4

CHANGES.md
csciguy8: Makes sense! I was actually torn as to which section, when I put that in.

Source/CesiumRuntime/CesiumRuntime.Build.cs
csciguy8: Good question. The plugin didn't build right away with 17, so I bumped it to 20.

The [Unreal coding standard](https://dev.epicgames.com/documentation/en-us/unreal-engine/epic-cplusplus-coding-standard-for-unreal-engine) says 17 is the minimum, but wouldn't be surprised if it was out of date.
image


We could invest some time into figuring out how to get 5.4 to compile using C++17... or just bump it, which I did instead.

CHANGES.md
csciguy8: Change committed

Source/CesiumRuntime/Private/CesiumCreditSystem.cpp
csciguy8: Yeah, avoiding having a pointer in an uninitialized state sounds cleaner.
Checked in some changes that avoids this completely (put an alias declaration in the #if instead)

Source/CesiumRuntime/CesiumRuntime.Build.cs
j9liu: Hmm, do you remember any of the errors that you saw with C++17? That will probably affect whether it's realistic to try and fix it in this timeframe.

@kring do you have any strong feelings about compiling with C++20? Just wanted to check if this had any consequences / implications we're not considering.

csciguy8: Easy enough to check, just set it back and build again with 5.4.

These are the first errors you would see... 

```
2>------ Building 15 action(s) started ------
2>[1/15] Compile [x64] PCH.CesiumRuntime.cpp
2>[2/15] Compile [x64] Module.CesiumRuntime.2.cpp
2>[3/15] Compile [x64] Module.CesiumRuntime.1.cpp
2>C:\Program Files\Epic Games\UE_5.4\Engine\Source\Runtime\Engine\Public\NaniteSceneProxy.h(513): error C7582: 'bHasRayTracingInstances': default member initializers for bit-fields requires at least '/std:c++20'
2>C:\Program Files\Epic Games\UE_5.4\Engine\Source\Runtime\Engine\Public\NaniteSceneProxy.h(514): error C7582: 'bNeedsDynamicRayTracingGeometries': default member initializers for bit-fields requires at least '/std:c++20'
2>C:\Program Files\Epic Games\UE_5.4\Engine\Source\Runtime\Renderer\Private\ScenePrivate.h(2938): error C7589: defaulting comparison operators requires at least '/std:c++20'
```

csciguy8: Oddly enough, those are the only errors I see when compiling with C++17. 
I just see them repeated over and over again.

j9liu: Ah well. If they require C++ 20, then that's what we need to compile with too :')

csciguy8: Total guess, but it might have slipped through the cracks for the UE 5.4 release. 
Feels like so little code that requires C++20.

j9liu:
@csciguy8 this looks great to me. Only thing is I think we should set up CI for 5.4 before merging into main. As much as I don't want it to hold up this PR, I'm concerned we might find compilation errors with Linux or MacOS.

j9liu:
Based on offline discussion, I'll merge this now. CI changes to come later.

PR #1398, opened 2024-04-19
Update copyright across all files

j9liu:
Updates the copyright at the top of all files for the range 2020-2024. Also adds it in the files where it was missing.

I just opened this PR because the inconsistency kept frustrating me. 🙃


csciguy8:
Agreed, it is a bit frustrating when copyrights go out of date. I've worked on some projects where I solved this by just not including them anymore.

Tagging self for review and starting the arduous 236 file diff ...

kring:
> Only recommendation I would have would be write an issue to consider automating this (or removing the copyright altogether?)

The reason we have a copyright header in cesium-unreal and nowhere else is because it's required by Epic's Marketplace guidelines. Or at least it was, I haven't checked lately.

PR #1402, opened 2024-04-25
Fix intermittent shutdown crash (static class GC'd)

### LINKED ISSUE #1363: Crash on exit
I frequently see a crash on exit. This is the call stack:

```
>	[Inline Frame] UnrealEditor-CesiumEditor.dll!FCesiumGeoreferenceCustomization::Unregister(FPropertyEditorModule &) Line 20	C++
 	[Inline Frame] UnrealEditor-CesiumEditor.dll!?A0x82102606::unregisterDetailCustomization() Line 130	C++
 	UnrealEditor-CesiumEditor.dll!FCesiumEditorModule::ShutdownModule() Line 391	C++
 	[Inline Frame] UnrealEditor-Core.dll!FModuleManager::UnloadModule(const FName) Line 686	C++
 	UnrealEditor-Core.dll!FModuleManager::UnloadModulesAtShutdown() Line 816	C++
 	UnrealEditor.exe!FEngineLoop::Exit() Line 4650	C++
 	[Inline Frame] UnrealEditor.exe!EngineExit() Line 77	C++
 	[Inline Frame] UnrealEditor.exe!GuardedMain::__l2::EngineLoopCleanupGuard::{dtor}() Line 131	C++
 	UnrealEditor.exe!GuardedMain(const wchar_t * CmdLine) Line 216	C++
 	UnrealEditor.exe!LaunchWindowsStartup(HINSTANCE__ * hInInstance, HINSTANCE__ * hPrevInstance, char * __formal, int nCmdShow, const wchar_t * CmdLine) Line 233	C++
 	UnrealEditor.exe!WinMain(HINSTANCE__ * hInInstance, HINSTANCE__ * hPrevInstance, char * pCmdLine, int nCmdShow) Line 282	C++
 	[Inline Frame] UnrealEditor.exe!invoke_main() Line 102	C++
 	UnrealEditor.exe!__scrt_common_main_seh() Line 288	C++
 	kernel32.dll!00007ffc17c2257d()	Unknown
 	ntdll.dll!00007ffc1900aa58()	Unknown
```

It's an access violation that looks like this:

```
Exception thrown: read access violation.
**__imp_ACesiumGeoreference::GetPrivateStaticClass**(...) returned 0x21364955F50.
```

image

So we're unloading the CesiumEditor module, and apparently `ACesiumGeoreference::StaticClass()` is returning an invalid pointer. Based on the Unreal log, it doesn't look like the CesiumRuntime module was unloaded first, so I don't really know why that pointer would be invalid.

csciguy8:
Fixes #1363 .

For all of our classes deriving from `IDetailCustomization`, change the call to `::UnregisterCustomClassLayout` to use a saved version of the class name that we called `::RegisterCustomClassLayout` with.

We don't want to access that class with `::StaticClass()` during shutdown, it's already been sent to the garbage collector to be destroyed. Sometimes the GC wouldn't have gotten to it yet (and you don't crash). But other times the static class data is garbage and we'll get an access violation.

I could reproduce this 100% of the time in UE 5.4 in DebugGame, with the following UI layout...
image

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,12 @@
 # Change Log
 
+### ? - ?
+
+##### Fixes :wrench:
+
+- Fixed an issue where tilesets would render completely black on Quest headsets and some iOS devices.
+- Fixed a crash on Unreal Editor shutdown that would happen occassionally.
+
 ### v2.4.1 - 2024-04-01
 
 This release updates [cesium-native](https://github.com/CesiumGS/cesium-native) from v0.33.0 to v0.34.0. See the [changelog](https://github.com/CesiumGS/cesium-native/blob/main/CHANGES.md) for a complete list of changes in cesium-native.
--- a/Source/CesiumEditor/Private/Cesium3DTilesetCustomization.cpp
+++ b/Source/CesiumEditor/Private/Cesium3DTilesetCustomization.cpp
@@ -6,18 +6,22 @@
 #include "DetailCategoryBuilder.h"
 #include "DetailLayoutBuilder.h"
 
+FName FCesium3DTilesetCustomization::RegisteredLayoutName;
+
 void FCesium3DTilesetCustomization::Register(
     FPropertyEditorModule& PropertyEditorModule) {
+
+  RegisteredLayoutName = ACesium3DTileset::StaticClass()->GetFName();
+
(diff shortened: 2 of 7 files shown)


j9liu:
Looks good to me, thanks @csciguy8 !

Issue #871, opened 2022-06-07
libcurl can't support the SSL 


lishu2006ll:
hi,
I am using unreal 4.27.2, but I found a problem, that is, I can't read the map image of SSL. It seems that the SSL certificate doesn't support it. Is there a solution, or can other versions support it? thank you!
the console tips： logHttp: Warning: xxxxxx: request failed libcurl error: 35( SSL connect error)

kring:
Unreal Engine's libcurl definitely supports SSL/TLS, but if your server's certificate is signed by a custom or unusual certificate authority you may run into problems. From what I can tell, the trusted certificate authority list used by Unreal Engine can be found here:
C:\Program Files\Epic Games\UE_4.27\Engine\Content\Certificates\ThirdParty\cacert.pem

So you can likely solve the problem by upgrading or editing this file.

I'm going to close this issue because I don't think there's any action on our side. Please post to the [community forum](https://community.cesium.com/c/cesium-for-unreal/10) if you would like to discuss further.

PR #1403, opened 2024-04-30
Add CI for Unreal Engine 5.4

### LINKED ISSUE #1400: Add CI support for Unreal Engine 5.4
For our github action builds, we need new jobs for Unreal Engine 5.4.

- Support all the existing platforms `Windows, Android, Linux, macOS, iOS`.
- Generate new artifacts, ex. `CesiumForUnreal-54-<platform>-<branch>`
- Potentially remove support for 5.1, as we traditionally only support 3 versions at a time.

The cesium-unreal plugin should build successfully as of [this PR](https://github.com/CesiumGS/cesium-unreal/pull/1396)

kring:
This builds on #1385 because main is broken.

Closes #1400

### DIFF
--- a/.github/actions/install-unreal-linux/action.yml
+++ b/.github/actions/install-unreal-linux/action.yml
@@ -27,5 +27,6 @@ runs:
       shell: bash
       run: |
         mkdir -p /mnt/work/${{ inputs.unreal-program-name }}
-        unzip -q ./ue.zip -d /mnt/work/${{ inputs.unreal-program-name }}
+        # Selectively extract files in order to save disk space
+        unzip -q ./ue.zip "Engine/*" -d /mnt/work/${{ inputs.unreal-program-name }} -x "Engine/Extras/*" "Engine/Binaries/Android/*" "Engine/Intermediate/Build/Android/*"
         rm ./ue.zip
--- a/.github/actions/install-unreal-linux/action.yml
+++ b/.github/actions/install-unreal-linux/action.yml
@@ -28,5 +28,5 @@ runs:
       run: |
         mkdir -p /mnt/work/${{ inputs.unreal-program-name }}
         # Selectively extract files in order to save disk space
-        unzip -q ./ue.zip "Engine/*" -d /mnt/work/${{ inputs.unreal-program-name }} -x "Engine/Extras/*" "Engine/Binaries/Android/*" "Engine/Intermediate/Build/Android/*"
+        unzip -q ./ue.zip "Engine/*" -d /mnt/work/${{ inputs.unreal-program-name }} -x "Engine/Binaries/Android/*" "Engine/Binaries/LinuxArm64/*" "Engine/Intermediate/Build/Android/*" "Engine/Intermediate/Build/LinuxArm64/*"
         rm ./ue.zip
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -416,7 +416,7 @@ jobs:
       cmake-toolchain: "version=14.38"
       cmake-platform: "x64,version=10.0.18362.0"
       visual-studio-version: "2022"
-      visual-studio-components: "Microsoft.VisualStudio.Component.Windows10SDK.18362"
+      visual-studio-components: "Microsoft.VisualStudio.Component.VC.14.38.17.8.x86.x64,Microsoft.VisualStudio.Component.Windows10SDK.18362"
   TestWindows54:
     needs: [Windows54]
     uses: ./.github/workflows/testWindows.yml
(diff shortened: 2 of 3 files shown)


.github/actions/install-unreal-linux/action.yml
```diff
@@ -27,5 +27,6 @@ runs:
       shell: bash
       run: |
         mkdir -p /mnt/work/${{ inputs.unreal-program-name }}
-        unzip -q ./ue.zip -d /mnt/work/${{ inputs.unreal-program-name }}
+        # Selectively extract files in order to save disk space
```
csciguy8, line 30: Looks like a straight forward optimization..

... but if there are unneeded files in the linux ue zip, could just make our own .zip on s3 (like the others) that already have the unnecessary files pruned out. Would save on download bandwidth and the extra unzip complexity.

csciguy8 (commented):
Thanks @kring!

Good changes to get CI up and running. It looks like things are building and artifacts are being generated, but I did notice a couple of issues that may / may not be solvable...

1) Build times for the new 5.4 version seem to be slower. The Windows 5.4 build is ~2x slower than the 5.3 build. Android54, linux54, iOS54 seem to be slight slower too. macOS54 is about the same

2) Artifact size for 5.4 seems to be ~2x larger than 5.3. This is a bit concerning.

csciguy8:
> 2\. Artifact size for 5.4 seems to be ~2x larger than 5.3. This is a bit concerning.

Good news! I diff'd the artifacts and discovered that the problem seems to be in the `Intermediate` folder. Usually we see `Module.CesiumRuntime.X.cpp.o` files in there. But with the new 5.4 build, we're seeing additional .o files (non-unity build).

image

Bad news, not sure what we did to cause this. From the looks of it, we build 5.4 just like we build 5.3. Worst case we just clean up these extra files before creating the artifact?

csciguy8:
Also noticed that 5.4 packaging failed...

```
[11/13] Compile [x64] Module.TestsProject.cpp
[12/13] Link [x64] TestsProject.exe
CesiumGltfReader.lib(GltfReader.obj) : error LNK2005: stbir_resize already defined in Module.ImageCore.cpp.obj
CesiumGltfReader.lib(GltfReader.obj) : error LNK2005: stbir_resize_uint8_srgb already defined in Module.ImageCore.cpp.obj
   Creating library D:\a\cesium-unreal\cesium-unreal\TestsProject\Binaries\Win64\TestsProject.lib and object D:\a\cesium-unreal\cesium-unreal\TestsProject\Binaries\Win64\TestsProject.exp
D:\a\cesium-unreal\cesium-unreal\TestsProject\Binaries\Win64\TestsProject.exe : fatal error LNK1169: one or more multiply defined symbols found
```

https://github.com/CesiumGS/cesium-unreal/actions/runs/8893486883/job/24432801834

.github/actions/install-unreal-linux/action.yml
kring: This isn't so much an optimization as a desperate attempt to get it to work at all. The GH Actions Linux runner doesn't have enough disk space to extract the whole thing. Yes, we could optimize the ZIP from here.

csciguy8:
> [11/13] Compile [x64] Module.TestsProject.cpp
> [12/13] Link [x64] TestsProject.exe
> CesiumGltfReader.lib(GltfReader.obj) : error LNK2005: stbir_resize already defined in Module.ImageCore.cpp.obj
> CesiumGltfReader.lib(GltfReader.obj) : error LNK2005: stbir_resize_uint8_srgb already defined in Module.ImageCore.cpp.obj
>    Creating library D:\a\cesium-unreal\cesium-unreal\TestsProject\Binaries\Win64\TestsProject.lib and object D:\a\cesium-unreal\cesium-unreal\TestsProject\Binaries\Win64\TestsProject.exp
> D:\a\cesium-unreal\cesium-unreal\TestsProject\Binaries\Win64\TestsProject.exe : fatal error LNK1169: one or more multiply defined symbols found

`Module.ImageCore.cpp.obj` is an obj file in the 5.4 engine source... which hints that stb symbols were newly defined here in 5.4, and we need to work around this somehow
`UE_5.4\Engine\Intermediate\Build\Win64\x64\UnrealGame\Development\ImageCore`

.

kring:
> Module.ImageCore.cpp.obj is an obj file in the 5.4 engine source... which hints that stb symbols were newly defined here in 5.4, and we need to work around this somehow

Yep. We'll probably have to modify the cesium-native build to optionally not include those `sbir` functions. We may also have to add ImageCore or similar to CesiumRuntime's .build.cs so that non-packaged builds don't break.

csciguy8:
I think I have a fix for the packaging test failure (stb linker error). It involved a change to cesium-native so I committed that.

In theory, both cesium-native and cesium-unreal should now be able to build (crossing fingers).

kring:
If your fix works (it looks promising), then perhaps we can ship UE 5.4 support today after all. I see you cancelled the build, though? Was that because you didn't think it would work, or you were just trying not to use build time during the release?

csciguy8:
> Was that because you didn't think it would work, or you were just trying not to use build time during the release?

It was because I didn't think it would work. I saw an error, then fixed it, then reupdated cesium-native in this PR.

Should be good to go now. Just waiting for CI results for this PR.

We now depend on [cesium-native PR #871](https://github.com/CesiumGS/cesium-native/pull/871)

PR #1408, opened 2024-05-01
Fixes for Unreal Engine 5.4 problems

### LINKED ISSUE #1405: UE 5.4 spams "LogHttp: Warning: Reached threaded request limit (20)" warnings to the console
It doesn't seem to cause any immediate harm, but we should investigate. We may be missing out on some performance, and if nothing else, the messages are distracting.

### LINKED ISSUE #1407: Collision / tracing / picking not working in UE 5.4
In the `06_CesiumMetadata` sample level, clicking makes the red line trace lines appear, but does not show any metadata.

kring:
Fixes #1407 
In UE 5.4, `UBodySetup::ChaosTriMeshes` has been "deprecated" and replaced with `TriMeshGeometries`. I put deprecated in quotes because the old property still exists, so code that uses it compiles, it just doesn't work at all anymore. Which is not my favorite kind of deprecation, if I'm being honest.

Fixes #1405 
In UE 5.3 we got the `RunningThreadedRequestLimit` configuration property. In UE 5.4 we now have its friend, `RunningThreadedRequestLimitEditor` that needs to be set, too. Both put an unnecessarily low limit on the number of simultaneous network requests that are allowed, so we increase the limits using our plugin's config.

### DIFF
--- a/Config/Editor.ini
+++ b/Config/Editor.ini
@@ -1,0 +1,6 @@
+# UE 5.3+ defaults this to 11, which is not really high enough.
+# 100 is probably high enough for reasonable values of Cesium3DTileset's 'MaximumSimultaneousTileLoads property,
+# but if you get log messages like "Warning: Reached threaded request limit", and you're sure you want to use
+# such a high value, you may need to increase this.
+[HTTP.HttpThread]
+RunningThreadedRequestLimitEditor=100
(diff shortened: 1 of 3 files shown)


j9liu:
Thanks @kring ! I'll trust that you've already tested this, so I'll go ahead and merge to speed up the release.

csciguy8:
Tested with latest. Verified these changes are good!

* Thread request limit console spam no longer occurs in editor
* The `06_CesiumMetadata` sample level picking looks like it works fine
* Dynamic Pawn speed control looks fine

PR #1409, opened 2024-05-01
Update release-process.md

j9liu:
Updates the release process guide for how to replace the default project token.


kring:
Thanks @j9liu! It's also really important to use the oldest supported version of UE when updating the token, so I added some words to that effect.

PR #1411, opened 2024-05-02
Force "None" source code provider on CI.

### LINKED ISSUE #1410: Plugin artifacts for UE 5.4 are huge (too many objects files)
Artifacts for Unreal 5.4 seem to be twice as big as the last version...

Ex.
CesiumForUnreal-54-main -> 2.35 GB
CesiumForUnreal-53-main -> 1.12 GB

Refer to [this CI build](https://github.com/CesiumGS/cesium-unreal/actions/runs/8910908675)

This was originally [noticed during code review](https://github.com/CesiumGS/cesium-unreal/pull/1403#issuecomment-2085862636), and looks to be that UE 5.4 is generating way more object files than we expect.

kring:
This avoids UE 5.4 considering our non-read-only source files to be part of the working set and leaving them out of the unity build.

Fixes #1410

### DIFF
--- a/.github/workflows/buildAndroid.yml
+++ b/.github/workflows/buildAndroid.yml
@@ -71,6 +71,18 @@ jobs:
       - name: Overwrite plugin engine version
         run: |
           ((Get-Content -path CesiumForUnreal.uplugin -Raw) -replace '"EngineVersion": "5.1.0"','"EngineVersion": "${{ inputs.unreal-engine-version }}"') | Set-Content -Path CesiumForUnreal.uplugin
+      - name: Customize BuildConfiguration.xml
+        run: |
+          mkdir -p "$env:USERPROFILE\AppData\Roaming\Unreal Engine\UnrealBuildTool"
+          # - We limit parallel actions because our builds use a lot more memory than UBT thinks they will.
+          # - We set the source code control Provider to None so UBT includes all files in the unity build.
+          Set-Content -Path "$env:USERPROFILE\AppData\Roaming\Unreal Engine\UnrealBuildTool\BuildConfiguration.xml" -Value '<?xml version="1.0" encoding="utf-8" ?>
+            <Configuration xmlns="https://www.unrealengine.com/BuildConfiguration">
+              <BuildConfiguration>
+                <MaxParallelActions>2</MaxParallelActions>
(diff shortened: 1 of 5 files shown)


kring:
This seems to have worked well. The 5.4 binaries are a similar size to earlier versions now.

csciguy8 (approved):
Thanks @kring ! Nice find. 

Artifacts are now back to normal size (about the same as 5.3 and earlier). 

I was hoping 5.4 build performance would also be a bit faster, but it still seems a little slower than 5.3 builds, and about as fast as the CI build for the official 2.5.0 release. Anyway, I checked the build log and verified we didn't have any weird excluded files or files compiling in addition to the unity build.

.github/workflows/buildAndroid.yml
```diff
@@ -71,6 +71,18 @@ jobs:
...
           ((Get-Content -path CesiumForUnreal.uplugin -Raw) -replace '"EngineVersion": "5.1.0"','"EngineVersion": "${{ inputs.unreal-engine-version }}"') | Set-Content -Path CesiumForUnreal.uplugin
+      - name: Customize BuildConfiguration.xml
+        run: |
+          mkdir -p "$env:USERPROFILE\AppData\Roaming\Unreal Engine\UnrealBuildTool"
+          # - We limit parallel actions because our builds use a lot more memory than UBT thinks they will.
+          # - We set the source code control Provider to None so UBT includes all files in the unity build.
+          Set-Content -Path "$env:USERPROFILE\AppData\Roaming\Unreal Engine\UnrealBuildTool\BuildConfiguration.xml" -Value '<?xml version="1.0" encoding="utf-8" ?>
+            <Configuration xmlns="https://www.unrealengine.com/BuildConfiguration">
+              <BuildConfiguration>
+                <MaxParallelActions>2</MaxParallelActions>
```
csciguy8, line 82: Just curious, is setting `MaxParallelActions` just a safety measure?

It wasn't obvious to me that this had any impact, at least on performance

kring: Some of the platforms need that in order to avoid exhausting their memory and running really slow (by swapping) or crashing. But I probably got overzealous copying it everywhere.

Issue #1359, opened 2024-02-21
Visual noise in Cesium3DTiles


ftrx:
I've noticed today that there is a visual noise on the Cesium3DTiles (its very hard to see, kind of white noise/ pixel artifacts):

https://github.com/CesiumGS/cesium-unreal/assets/1744939/8fe1da6b-05bf-40d4-9dfa-b53facb4df43

it is even more visible when the mesh/3dTiles is selected (so its something in the mesh?):

https://github.com/CesiumGS/cesium-unreal/assets/1744939/051c5b6c-1125-442e-a564-56c7368d2627

Its quite subtle visual noise and far better visible when the 3dTileset is selected. It's mostly noticeable in darker areas and is in VR quite noticeable (also in builds)

#### It seems like its a Material issue (?) with the standard masked material:

If I change the 3dTileset material to the `M_DEPRECATED_CesiumDefaultMasked` the issue is gone.
Also if disconnect/replace the Alpha Channel of the Vertex Color node from the Opacity/ OpacityMask in the `CesiumGlTFFunction MaterialFunction`, the issue/noise is gone (so it seems vertex color related?)

If I have time, I'll try to dive into it a bit, but maybe you guys already know the issue/culprit

### My Specs
- UE 5.3
- Cesium Plugin 2.3.0 but affects probably all version with the newer material

### Steps to reproduce:
1. Open Cesium for Unreal Examples Project
2. Open any map (for example 04_MAIN_CesiumSublevels)
3. Go to a darker area (dark textured building or so)

csciguy8:
I see it too. Even turning off antialiasing (Project Settings->Anti-Aliasing Method = None), the problem persists.

This does stop the noise from animating at least.

ftrx:
Some more information from our side:
It seems to have something to do with the Material Layers,
as the `CesiumGltFFunction` is also used in the old deprecated material directly and there it works without issue.

Also if I use a new Material Instance of the `M_CesiumBaseMaterial` with the same parameters as the `MI_CesiumThreeOverlaysAndClipping` Material Instance the issue seems to be gone...

So maybe a bad asset/ old parameters? but i cannot exclude yet if its cache/DDC related from our side.

Issue #1104, opened 2023-05-23
DitherFade material layer causes serious artifact with forward rendering / MSAA


kring:
1. Enable forward shading by check the Forward Shading box under Edit -> Project Settings -> Engine - Rendering -> Forward Renderer.
2. Change "Anti-Aliasing Method" to "Multisample Anti-Aliasing (MSAA)" under Edit -> Project Settings -> Engine - Rendering -> DEfault Settings)

You'll see serious dithering artifacts in all Cesium objects, while objects that user other materials look fine:

<img width="565" alt="image" src="https://github.com/CesiumGS/cesium-unreal/assets/924374/11da609e-46ef-4c37-b7a7-6953f90de5b7">

Deleting the `DitherFade` layer from our default material fixes the problem. We're using an engine-provided material node that is meant for temporal AA, but it's not clear why it fails so badly with MSAA.

j9liu:
#1414 doesn't address this. I think that it has to do with the fact that the "DitherTemporal AA" node that we use, doesn't have good defaults when Temporal AA isn't enabled. So it's stuck on a permanent dither pattern. You can actually kind of see that pattern in the material function preview: 
![image](https://github.com/CesiumGS/cesium-unreal/assets/32226860/5d75c76b-0fb7-4bb7-bdc8-1b9695edaac3)

We could possibly account for the current AA mode by querying it, and passing a flag into the material indicating whether or not to use dithering. But because the user could possibly change it, we'd need to constantly check for updates to that setting. So something like this (taken from `RendererScene.cpp`):

```
static auto* CVarAntiAliasingMethod = IConsoleManager::Get().FindTConsoleVariableDataInt(TEXT("r.AntiAliasingMethod"));
static int32 CachedAntiAliasingMethod = CVarAntiAliasingMethod->GetValueOnGameThread();
const int32 AntiAliasingMethod = CVarAntiAliasingMethod->GetValueOnGameThread();
if (AntiAliasingMethod != CachedAntiAliasingMethod) {
     // ... 
}
```

But I'm concerned about doing it this way because it seems like things can commonly override the project's default AA method? From `SceneView.cpp`:

```
// see EAntiAliasingMethod
static TAutoConsoleVariable<int32> CVarDefaultAntiAliasing(
	TEXT("r.AntiAliasingMethod"),
	4,
	TEXT("Engine default (project setting) for AntiAliasingMethod is (postprocess volume/camera/game setting still can override)\n")
[...]

PR #1414, opened 2024-05-06
Fix LOD dithering artifacts

### LINKED ISSUE #1388: White pixel flickering artifacts with "DitherFade" material layer
This may be related to #1104, but that issue is about seeing temporal dithering artifacts with forward rendering + AA. There are actually dithering artifacts that happen _with_ temporal rendering.

Reported on the forum [here](https://community.cesium.com/t/light-white-pixel-flickering-on-tiles/31203/). The user sees bright white pixels flickering throughout the tileset, which are especially visible when the lighting is dark. This is with:

- Unreal Engine 5.3
- Cesium for Unreal 2.4.1
- Google Photorealistic 3D Tiles
- Temporal Super Resolution enabled in project settings (this is how it's set in the samples)
- **Enable LOD Transitions** on the tileset set to false

To reproduce the view, set the georeference to
**Latitude**: 59.908135
**Longitude**: 10.721472
**Height**: 99.238093

image
( Zoom in to see more clearly. From a distance it will look like you have dust on your computer. :') )

The workaround was to delete the "DitherFade" material layer from the default tileset material. 

One action item is to test if the plugin version matters, since the user says it worsened with time. Engine version may matter too.

j9liu:
Fixes #1388, #1359 

This PR modifies the ML_DitherFade material layer to prevent some flickering pixels that could be seen on tilesets. These were most visible on Google Photorealistic 3D Tiles, and especially when the tileset was selected. You can see it in `12_CesiumGoogleMapsTiles` behind the intro billboard, or check out #1388 for an example.

### Before

image

image

### After

image

image

I will also test if this addresses #1104, TBD.


Source/CesiumRuntime/Private/CesiumGltfComponent.cpp
```diff
@@ -3582,10 +3580,12 @@ void UCesiumGltfComponent::UpdateFade(float fadePercentage, bool fadingIn) {
     return;
   }
 
+  fadePercentage = glm::clamp(fadePercentage, 0.0f, 1.0f);
```
csciguy8, line 3583: Did moving this line fix anything? Or was it just code cleanup?

j9liu: Just code cleanup from when I was debugging the fades (which was ultimately not necessary)

j9liu:
Unfortunately this does not address #1104. ☹️

kring:
Code looks good, and this is working well, no matter if LOD transitions are enabled or disabled. Merging.

PR #1415, opened 2024-05-06
Add soak test for memory usage with external tilesets

azrogers:
To test out CesiumGS/cesium-native#876, I've created a long-running [soak test](https://en.wikipedia.org/wiki/Soak_testing) to evaluate the memory usage of cesium-native over a long period. The test consists of loading the Google P3DT and flying to random locations around the globe over and over again for twelve straight hours. There's a few issues with this test as it stands:

1. Because of CesiumGS/cesium-native#749, the test can only actually run for three hours total, since after that the Google P3DT token will need to be refreshed and we don't currently handle this possibility. I can't think of a better dataset than P3DT to run this test with, so do we fix this bug instead?
2. After each flight it chooses a new position in the opposite quadrant of the earth to fly to (if it's in the northern and western hemispheres, it will choose a point in the southern and eastern hemispheres, for example). This means most of the points it chooses are in the middle of nowhere (fields, ocean) which aren't exactly a great test. It would be great to choose random locations that have lots of buildings and therefore content to load instead, but I can't think of a way to do this other than working from a set list of locations.


kring:
Let me try to describe the problem and possible solution, as I see it. Most of this probably isn't new to you at this point, but hopefully it helps to have it in one place.

Ok, so first things first, what are we actually trying to do? We're concerned with tiles which are the root of an external tileset. In other words, tiles for which `tile.getContent().isExternalContent()` returns true. Currently, we simply never unload the content for such tiles (`TilesetContentManager::unloadTileContent` [exits early](https://github.com/CesiumGS/cesium-native/blob/43187540a5199d3b5fa8693fca90a1b81bf73c80/Cesium3DTilesSelection/src/TilesetContentManager.cpp#L1042) without doing anything when it sees external content). We'd like to be able to unload them.

But what does it mean to unload the content for such a tile? Usually, when we talk about unloading tile content, we're talking about clearing its [_content](https://github.com/CesiumGS/cesium-native/blob/43187540a5199d3b5fa8693fca90a1b81bf73c80/Cesium3DTilesSelection/include/Cesium3DTilesSelection/Tile.h#L529) field. For renderable tiles, this contains the tile's whole glTF, including geometry and textures, so it's pretty big. But for a tile with external content, it's actually [pretty tiny](https://github.com/CesiumGS/cesium-native/blob/43187540a5199d3b5fa8693fca90a1b81bf73c80/Cesium3DTilesSelection/include/Cesium3DTilesSelection/TileContent.h#L52). That `TilesetMetadata` instance is mostly empty in most cases, so it doesn't take up much memory at all.

However, when we load an external tileset, the parsed tileset.json is turned into a tree of `Tile` instances. Sometimes thousands of them. These tiles don't have content (we sometimes call them "skeletons"), but they still add up when we have a lot of them (which we very often do). So, in the same way we can _load_ extra tiles from an external tileset.json, we'd like to later be able to _unload_ (destroy!) these same tiles. This will save us a lot of memory, particularly with tilesets that use external tilesets heavily like Google Photorealistic 3D Tiles.

So to get started, we can simply allow `TilesetContentManager::unloadTileContent` to operate on tiles with external content, and we can implement the unload of external content to _also_ clear the tile's `_children` vector. 

(Note that we should _never_ clear the `_children` vector in any other circumstances - that would just result in losing information that we couldn't easily recover. This is only safe to do for tiles where the children came from an external tileset because we can get them back just by loading that external tileset again.)

The problem with this simple approach is that clearing the `_children` vector invalidates pointers/references to every `Tile` in the vector, plus every `Tile` in those tiles' `_children` vectors, and so on. So before we can do that safely, we have to make sure that _no one_ is holding pointers to any of those tiles.

So then the question becomes: where might we hold `Tile` pointers/references, and how can we detect this?

Here are the places I know about. I'm not sure if it's exhaustive, but hopefully it's close:

1. The `ViewUpdateResult`, specifically `tilesToRenderThisFrame` and `tilesFadingOut`. We generally shouldn't need to worry about `tilesToRenderThisFrame` because any tile rendered this frame won't be unloading anyway, according to existing unload rules. In fact, if a particular tile deep down in the tree is rendered, all its ancestors will be prevented from unloading, too. So generally we don't need to worry about this one. `tilesFadingOut`, on the other hand, can happen over multiple frames and after the tile subtree isn't traversed at all. So we definitely need to check that one. Specifically, if any Tile in the external tileset is in this list, we can't unload the external tileset yet.
2. The load pipeline, initiated at `TilesetContentManager::loadTileContent`. A Tile reference is put into `TileContentLoadInfo`. So if any Tile in the external tileset is loading, we can't unload the external tileset yet. We can determine this by looking at Tile [states](https://github.com/CesiumGS/cesium-native/blob/43187540a5199d3b5fa8693fca90a1b81bf73c80/Cesium3DTilesSelection/include/Cesium3DTilesSelection/Tile.h#L30). The `Unloaded`, `ContentLoaded`, `Done`, and `Failed` states are safe. All others are not.
3. A Tile that is "upsampling" from a parent Tile holds a pointer to the parent tile, so the parent tile (and its external tileset) can't be unloaded. We can determine if a Tile is upsampling from this Tile with the logic [here](https://github.com/CesiumGS/cesium-native/blob/43187540a5199d3b5fa8693fca90a1b81bf73c80/Cesium3DTilesSelection/src/TilesetContentManager.cpp#L1068).

Checking all of these things can be expensive, especially because we need to check every tile in an external tileset in order to determine if we can unload it. That can easily be thousands of tiles. So one idea is keep track of unload safety all along rather than computing it when we need it. Here's how I think I'd approach that (warning: this may not be a fully baked idea):

Add `doNotUnloadCount` to Tile. Every time we enter a state that forbids unloading (1-3 above), we increment this count. Then we _also_ increment the same counter on all of the Tile's parent Tiles, all the way to the root. When we leave the state that forbids unloading, we decrement the counter (and also do the same for all ancestors). A single Tile can be in multiple unload-forbidding states simultaneously, such as if it's in `tilesFadingOut` and is also an upsampling source. If that's done correctly, checking if we can unload an external tileset is a simple matter of checking whether the root Tile of the external tileset has a `doNotUnloadCount` of zero.

Making sure these reference counts are prefectly balanced can be tricky and error prone, so a debug mode that tracks not just the count but also the source of the count (perhaps as a descriptive string) can be helpful.

I haven't thought of the flaw with this plan yet (I'll update if I do), so hopefully it'll work well, and I don't think it'll be too difficult to implement, either. There are probably other reasonable solutions as well.

azrogers:
Closing in favor of #1615.

PR #1416, opened 2024-05-07
Add check to disable LOD transitions in material layer

### LINKED ISSUE #1104: DitherFade material layer causes serious artifact with forward rendering / MSAA

j9liu:
Follow up to #1414. Fixes #1104.

This PR does the following:

- Renames `UseLodTransitions` to `EnableLodTransitions`, to keep consistent with the rest of the naming in `Cesium3DTileset`.
- Adds a `EnableLodTransitions` parameter to the `ML_DitherFade` material layer, to control whether or not the dithering is applied.
- Clarifies that `EnableLodTransitions` is only compatible with temporal AA modes in the comments.

This material parameter corresponds exactly with the `Cesium3DTileset`'s parameter. When false, dithering won't be applied at all. This helps for projects that use non-temporal AA modes. As long as they don't enable LOD transitions (which would be user error), they don't need to worry about these dithering artifacts on their tilesets.

### Before

image

### After

image

image


j9liu:
Hm, the artifacts don't appear when Forward Rendering is disabled, even for non-temporal AA modes. So I'll adjust the documentation

kring:
@j9liu I see pretty significant artifacts in this branch when "Enable LOD Transitions" is enabled. It kind of looks like the old tile is being removed completely before the new one comes in, so I see big holes in the surface. This is with Google Photorealistic 3D Tiles:
image

I see the same thing with CWT+Bing, though.

j9liu:
I'm going to close this since this is really old, and I won't get around to it anytime soon.

PR #1432, opened 2024-05-24
Add support for custom ellipsoids

### LINKED ISSUE #366: Allow users to choose a custom ellipsoid
We should allow users to set a custom ellipsoid on the **CesiumGeoreference** instead of the default WGS84. I think the cesium-native calculations are agnostic of what ellipsoid is used, so this should be relatively easy. 

If someone from the community wants to give this a shot, look for all hard-coded references to `CesiumGeospatial::Ellipsoid::WGS84` in `CesiumGeoreference.cpp`. Each of these needs to be replaced with a reference to a custom `CesiumGeospatial::Ellipsoid` that should be configurable through a C++ setter. It should still default to using the `WGS84` ellipsoid if no other ellipsoid is set. You can have a look at the cesium-native source code ("CesiumGeospatial/Ellipsoid") to determine how to create the custom ellipsoid.

azrogers:
Currently, Cesium for Unreal is hardcoded to use the WGS84 ellipsoid. This change adds support for specifying other ellipsoids by creating assets for them and assigning them to georeferences. 

Closes #366


csciguy8:
> One main issue that needs to be fixed before this can be merged in is that changing the ellipsoid of a georeference should cause all of its associated tilesets to be reloaded.

@azrogers , just double checking, is this still an issue?

(you've marked this PR as ready, yet I saw this comment in the description)

csciguy8:
Found a crash, I _think_ it's valid?

Was just playing around with the Unreal Editor Details window and ellipsoid options. Debug build, UE 5.4.1

image

csciguy8:
Found another crash, but in UE 5.2 (still debug build).

Opened up the 01 sample level, then in the Outliner window, find "Edit DynamicPawn", then click it.

Obviously not directly related to ellipsoids, but triggers a related UI refresh I'm guessing...

image

csciguy8 (commented):
This looks like a nice added feature. Can we get mention of it in CHANGES.md?

Also, it's not obvious to me what the typical workflow is to make a new custom ellipsoid, maybe that can be mentioned as well?

Content/WGS84.uasset
csciguy8, line 1: Looks good and verified this works on UE 5.2

Source/CesiumRuntime/Public/CesiumGeoreference.h
```diff
@@ -84,9 +98,30 @@ class CESIUMRUNTIME_API ACesiumGeoreference : public AActor {
   UPROPERTY(BlueprintAssignable, Category = "Cesium")
   FGeoreferenceUpdated OnGeoreferenceUpdated;
 
+  /**
+   * An event that will be called whenever the georeference's ellipsoid has
+   * been modified.
+   */
+  UPROPERTY(BlueprintAssignable, Category = "Cesium")
+  FGeoreferenceEllipsoidChanged OnEllipsoidChanged;
```
csciguy8, line 106: Adding a new event when the ellipsoid is changed makes sense, but this makes me think that `OnGeoreferenceUpdated` isn't really pulling its weight.

Have you considered modifying `OnGeoreferenceUpdated` to pass the old / new ellipsoid to the callers? In theory, we can just have the one event, and it can serve its existing callers, but also support what `ACesium3DTileset` needs (`RefreshTileset` when the ellipsoid changes)

Source/CesiumRuntime/Public/Cesium3DTileset.h
```diff
@@ -1127,6 +1128,14 @@ class CESIUMRUNTIME_API ACesium3DTileset : public AActor {
   void UpdateTransformFromCesium();
 
 private:
+  /**
+   * The event handler for ACesiumGeoreference::OnEllipsoidChanged.
+   */
+  UFUNCTION()
+  void _onGeoreferenceEllipsoidChanged(
```
csciguy8, line 1135: The naming on this function looks stylistically inconsistent.

It is a handler for an event named `OnEllipsoidChanged`. For reference, one existing example of a callback that adds itself to the `OnGeoreferenceUpdated` event is `HandleGeoreferenceUpdated`.

Maybe `HandleEllipsoidChanged` would be better?

Source/CesiumRuntime/Private/Cesium3DTileset.cpp
```diff
@@ -1052,6 +1075,16 @@ void ACesium3DTileset::LoadTileset() {
...
 
+  TSharedPtr<CesiumGeospatial::Ellipsoid> pNativeEllipsoid =
+      this->ResolveGeoreference()->GetEllipsoid()->GetNativeEllipsoid();
+
+  // Since native will be working with an STL shared_ptr, not an Unreal one,
+  // we'll copy the ellipsoid here. It's not much data to copy and saves us some
+  // headaches.
+  std::shared_ptr<CesiumGeospatial::Ellipsoid> pEllipsoid =
+      std::make_shared<CesiumGeospatial::Ellipsoid>(
+          pNativeEllipsoid->getRadii());
```
csciguy8, line 1086: If my memory serves me, cesium-native no longer uses shared pointers for its internal use of ellipsoid.

So we shouldn't need this logic, and `GetNativeEllipsoid` can probably just return a `CesiumGeospatial::Ellipsoid` rather than a share ptr to one.

Source/CesiumRuntime/Private/CesiumEllipsoid.cpp
```diff
@@ -0,0 +1,97 @@
...
+FVector UCesiumEllipsoid::GetRadii() { return this->Radii; }
+
+void UCesiumEllipsoid::SetRadii(FVector NewRadii) { this->Radii = NewRadii; }
+
+double UCesiumEllipsoid::GetMaximumRadius() { return this->Radii.X; }
+
+double UCesiumEllipsoid::GetMinimumRadius() { return this->Radii.Z; }
+
+FVector UCesiumEllipsoid::ScaleToGeodeticSurface(
+    const FVector& EllipsoidCenteredEllipsoidFixedPosition) {
```
csciguy8, line 24: This parameter is so wordy, yet I have no suggestion to make it better

Source/CesiumRuntime/Private/CesiumEllipsoid.cpp
```diff
@@ -0,0 +1,97 @@
...
+  }
+}
+
+TSharedPtr<Ellipsoid> UCesiumEllipsoid::GetNativeEllipsoid() {
+  const double MinRadiiValue = TMathUtilConstants<double>::Epsilon;
+
+  if (!this->NativeEllipsoid.IsValid()) {
+    // Radii of zero will throw Infs and NaNs into our calculations which will
+    // cause Unreal to crash when the values reach a transform.
+    if (this->Radii.X < MinRadiiValue || this->Radii.Y < MinRadiiValue ||
```
csciguy8, line 81: Looks like a good check. 

Optionally could be moved to `::SetRadii` where we could log the bad input, like we do here, and even reject making the change to the member variables.

Source/CesiumRuntime/Private/CesiumEllipsoidFunctions.h
```diff
@@ -0,0 +1,63 @@
...
+
+namespace CesiumGeospatial {
+class Ellipsoid;
+};
+
+/**
+ * A collection of methods for working with {@link CesiumGeospatial::Ellipsoid}
+ * objects from Unreal.
+ */
+class CesiumEllipsoidFunctions {
```
csciguy8, line 15: This a good way to separate shared functionality, but for code style, I'm don't see any other examples of `XxxFunctions` in the cesium-unreal code base.

There's always a first though.

Another optional way to do it is to just make these static functions in `CesiumEllipsoid`. `UCesiumWgs84Ellipsoid` should still be able to access that.

package-lock.json
```diff
@@ -1,12 +1,12 @@
...
   "name": "cesium-unreal",
-  "version": "2.4.0",
+  "version": "2.6.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "cesium-unreal",
-      "version": "2.4.0",
+      "version": "2.6.0",
```
csciguy8, line 9: Did you mean to change to this file? 
(it's easy to accidentally commit it)

csciguy8 (changes requested):
Thanks for the PR @azrogers!

It surprised me how much code has to change to support doing something as seemingly simple as changing the ellipsoid. Lots of good work.

Requesting changes for the crashes I found. 

Also made lots of code comments. Some may result in more changes, like the ones related to `GetNativeEllipsoid`.

Source/CesiumRuntime/Public/CesiumGeoreference.h
azrogers: The reason there's two events is that there's two different things that need to happen. One is to update a globe anchor's position when the georeference changes in a way that would affect its computations, the other is to unload and reload a tileset when the ellipsoid has changed. We only need to do one or the other - because GlobeAnchor stores its positions in ECEF coordinates, its position isn't going to change when the ellipsoid is changed. And we definitely don't want to unload and reload the tileset whenever the georeference is updated. So either we have one event with an enum or bool value that specifies which of the two operations to perform - or we have two distinct events.

Source/CesiumRuntime/Public/Cesium3DTileset.h
azrogers: It's named this way to match `UCesiumGlobeAnchorComponent::_onGeoreferenceChanged` and `_onActorChanged`. But neither of these match Unreal's naming conventions, so I'm fine changing it.

Source/CesiumRuntime/Private/CesiumEllipsoid.cpp
azrogers: The reason it's here is because, when you create a new Ellipsoid asset, the Radii field is initialized to (0, 0, 0) by default. If you pass this asset to a UCesiumGeoreference without changing the values, it'll crash Unreal (and if you saved the scene, it'll crash Unreal every time you open the scene). Could initialize `UCesiumEllipsoid::Radii` to something other than (0, 0, 0), but I think it's better to keep the check here.

Source/CesiumRuntime/Private/CesiumEllipsoidFunctions.h
azrogers: It's used by `CesiumWgs84Ellipsoid` - since the existing code expects `UCesiumWgs84Ellipsoid` to have static methods for dealing with a WGS84 ellipsoid, our options are either 1. putting a bunch of static methods on `UCesiumEllipsoid` that accept an ellipsoid parameter, which both `UCesiumEllipsoid` and `UCesiumWgs84Ellipsoid` will use, 2. storing a static instance of a WGS84 `UCesiumEllipsoid` in `UCesiumWgs84Ellipsoid` which all the static methods use, or 3. creating static methods that accept an ellipsoid parameter and using them for both `UCesiumEllipsoid` and `UCesiumWgs84Ellipsoid` (which is what I went with here).

package-lock.json
azrogers: It's automatically changed by npm when running npm install - it seemed fine to commit it, since the version should probably stay up to date anyways, but I agree it's not really related to this PR. I can remove it.

azrogers:
@csciguy8 Addressed your review comments. Fixed crashes related to the `UCesiumGlobeAnchorComponent` details editor and added a bunch more checks so that you should get errors logged instead of crashes.

Source/CesiumRuntime/Public/CesiumGeoreference.h
csciguy8: > So either we have one event with an enum or bool value that specifies which of the two operations to perform

Right, that's what I was implying that we could do instead. Both approaches seem valid though.

Source/CesiumRuntime/Public/Cesium3DTileset.h
csciguy8: Thanks. I _think_ this is more consistent now?

csciguy8:
Thanks @azrogers for the changes.

Two comments remain that got lost in the shuffle...

1) Is the issue mentioned in the description still an issue that needs to be addressed?
(the ellipsoid of a georeference should cause all of its associated tilesets to be reloaded)

2) `GetNativeEllipsoid` is still returning a `TSharedPtr` even though we don't use shared pointer functionality.

azrogers:
In response to the point in my original post, that should be fixed. The tileset should be reloaded when the ellipsoid is changed.

I've removed the usage of TSharedPtr. Let me know if there's any final things you need me to take care of @csciguy8

azrogers:
Oh, I should also mention at some point that this PR will close #366.

Source/CesiumRuntime/Public/CesiumEllipsoid.h
```diff
@@ -0,0 +1,142 @@
...
+   * Returns a new {@link CesiumGeospatial::LocalHorizontalCoordinateSystem}
+   * with the given scale, center, and ellipsoid.
+   */
+  CesiumGeospatial::LocalHorizontalCoordinateSystem
+  CreateCoordinateSystem(const FVector& Center, double Scale);
+
+  /**
+   * Returns the underlying {@link CesiumGeospatial::Ellipsoid}
+   */
+  CesiumGeospatial::Ellipsoid& GetNativeEllipsoid();
```
csciguy8, line 125: One last catch! (I swear)

We're returning a reference to an ellipsoid here, implying that the caller is free to change the value of it, which would change the value of our protected member too. Pretty sure we don't want this. 

I'm guessing we'd instead want to return a const ref, or just return by value.

Source/CesiumRuntime/Public/CesiumEllipsoid.h
```diff
@@ -0,0 +1,142 @@
...
+   * the Z coordinate should be the radius of the smallest axis.
+   */
+  UPROPERTY(
+      EditAnywhere,
+      Category = "Cesium|Ellipsoid",
+      meta = (DisplayName = "Radii"))
+  FVector Radii;
+
+private:
+  TUniquePtr<CesiumGeospatial::Ellipsoid> NativeEllipsoid = nullptr;
```
csciguy8, line 141: Optionally, could change this to not even use `TUniquePtr`...

Could be simply,
`CesiumGeospatial::Ellipsoid NativeEllipsoid;`

csciguy8:
Did some testing and good news, no crashes! 

Looks pretty much done, but I did find 2 issues with the tilesets refreshing...

1) When changing which ellipsoid is used through the "Details" window, I noticed that the globe shape updated, but the tiles didn't look to reload. If I hit the "Refresh Button" on the tileset manually, things do what I expect. I even set a breakpoint in `ACesium3DTileset::HandleOnGeoreferenceEllipsoidChanged` and noticed it wasn't getting trigged.

2) Similarly, if I make a new ellipsoid, edit it in its new Details window, then save, I'm not seeing the tileset refresh.

azrogers:
@csciguy8 Made `GetNativeEllipsoid` return a const ref, made `UCesiumEllipsoid` store a `TOptional` instead of a `TUniquePtr`, fixed the OnEllipsoidChanged event (apparently the handler needs to be marked with "CallInEditor"), and added a method that reloads any georeferences using an ellipsoid asset when that ellipsoid asset is saved. Should be good to go!

csciguy8 (approved):
Looks great @azrogers!

Will merge once CI passes (almost done).

Kudos for catching the event when the ellipsoid data changes. `UCesiumEllipsoid::PostSaveRoot` does not look like an easy or obvious way to handle this

PR #1439, opened 2024-05-31
Fix experimental occlusion culling

### LINKED ISSUE #1429: Occlusion culling is stuck on waiting for results?
Hi there,

I decided to look into occlusion culling to help performance in my application as mentioned in [this blog article](https://cesium.com/blog/2022/08/18/occlusion-culling-cesium-for-unreal/) but I am having some issues with it. Having occlusion culling enabled on my 3d tileset causes no tiles to load. It appears to be stuck waiting for the occlusion results as the latest entry in my output log shows:
```
LogCesium: Display: Cesium3DTileset_0: 156 ms, Visited 2, Culled Visited 0, Rendered 1, Culled 0, Occluded 0, Waiting For Occlusion Results 1, Max Depth Visited: 1, Loading-Worker 0, Loading-Main 0, Loaded tiles 0%
```

I have enabled the `Enable Experimental Occlusion Culling Feature` in the plugin settings inside the project settings and if I disable occlusion culling on the 3D tile, it renders all my tiles (without culling of course).

I have done some investigation into what might be causing this and have looked at some unreal engine visualizations. I have disabled occlusion culling to render the tiles for these screenshots:

image
*Bounds appear to be set correctly for the google 3d tiles*

image
*I have a platform with some normal actors and have enabled `r.VisualizeOccludedPrimitives 1` in the console to see that normal occlusion culling appears to work*

image
*Interestingly it appears that the depth buffer has this staggered effect from the camera, I am visualizing this by Buffer Visualisation -> Scene Depth. I would assume that this is not correct and could be the cause of my occlusion culling to not work?*

image
*Here is a screenshot of my 3d tile settings incase this might have caused some issues, however a lot of them are defaults.*

Some more information on my versions:
 - CesiumForUnreal v2.3.0 downloaded from the release page and placed in the plugins folder of the project
 - Unreal Engine 5.3.2
 - Google 3D tiles using googles API (tile.googleapis.com)

Thanks for any help with this and let me know if you'd like any more details!

kring:
Fixes #1429 

This was broken by a change in UE 5.1: the signature of `PostRenderViewFamily_RenderThread` method of `ISceneViewExtension` changed. Had that been all, our code would have failed to compile when we upgraded to UE 5.1, and we would have fixed it. But for some reason, Epic kept the old signature around, they just made it do nothing. It's marked with a deprecating macro:

```cpp
UE_DEPRECATED(5.1, "Use PostRenderViewFamily_RenderThread with an RDG builder instead. RHI commands can be enqueued directly using GraphBuilder.RHICmdList or queued onto the graph.")
virtual void PostRenderViewFamily_RenderThread(FRHICommandListImmediate& RHICmdList, FSceneViewFamily& InViewFamily) {}
```

But apparently C++ compilers don't flag a warning when you override a deprecated method, which is really unfortunate. So we had no help from the compiler regarding the breaking change, and being an experimental feature we don't use it often enough to have noticed it ourselves.

Once I updated the signature, occlusion culling started working again, _except_ when the tileset was selected in the Outliner. The problem here was that Unreal skips all occlusion checks for primitives selected in the Editor. Because the occlusion check was skipped, and because `UCesiumBoundingVolumeComponent` doesn't actually render anything (it's only used for occlusion checks), Unreal was deeming the primitive "not visible". This might be a change from previous UE versions? I'm not sure. In any case, our code then assumed this meant the primitive was frustum culled (which is what "not visible" usually means at this stage), and so the associated tile wasn't refined. End result: low detail or entirely missing tiles when the tileset was selected in the Outliner.

This PR fixes this second problem by using the same check that UE itself uses when deciding to skip the occlusion culling step: it checks if the primitive is selected in the Editor. If it is, it's treated as unoccluded.


kring:
Self merging this in order to include it in today's release. The changes are isolated to a code path that is only activated when an experimental flag is enabled, and that path is completely broken without these changes, so the risk here is minimal.

PR #1444, opened 2024-06-03
Add Editor.ini to FilterPlugin.ini

### LINKED ISSUE #1443: "Reached threaded request limit" warning in UE 5.4
When running in the Editor in UE 5.4 with a released version of the plugin, the log is spammed with this message:

> LogHttp: Warning: Reached threaded request limit (20)

I theoretically fixed this in #1408, and I did for local development builds. But it doesn't work for released versions of the plugin because I forgot to include the newly-added `Editor.ini` file in `FilterPlugin.ini`, so the change isn't included in the shipped version of the plugin.

kring:
Fixes #1443

### DIFF
--- a/Config/FilterPlugin.ini
+++ b/Config/FilterPlugin.ini
@@ -1,6 +1,7 @@
 [FilterPlugin]
 /Config/FilterPlugin.ini
 /Config/Engine.ini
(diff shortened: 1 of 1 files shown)


j9liu:
Thanks @kring !

PR #1447, opened 2024-06-06
Use the vcpkg-enabled version of cesium-native.

kring:
This depends on CesiumGS/cesium-native#820.

To do:

* [x] Create an [overlay triplet](https://learn.microsoft.com/en-us/vcpkg/users/examples/overlay-triplets-linux-dynamic) for Unreal so the above steps aren't necessary.
* [x] Get all the platforms building on CI.
* [x] Test the Editor on all platforms:
  * [x] Windows
  * [x] macOS
  * [x] Linux
* [x] Test packaging for all platforms
  * [x] Windows
  * [x] macOS
  * [x] Linux
  * [x] Android
  * [x] iOS
* [x] Create [overlay ports](https://learn.microsoft.com/en-us/vcpkg/concepts/overlay-ports) for abseil and sqlite that rename our symbols to avoid conflicts with the same libraries included in other Unreal plugins.
* [x] Investigate why the distributions are bigger. Probably because of [this](https://github.com/microsoft/vcpkg/issues/9084).

CC @jherico (thanks for all your work on this on the cesium-native side!)


extern/CMakeLists.txt
```diff
@@ -37,26 +37,18 @@ set(PRIVATE_CESIUM_SQLITE ON)
...
-set(CMAKE_MINSIZEREL_POSTFIX ${CESIUM_RELEASE_POSTFIX})
-set(CMAKE_RELWITHDEBINFO_POSTFIX ${CESIUM_RELEASE_POSTFIX})
-
 # On Mac and Linux, Unreal uses -fvisibility-ms-compat.
 # On Android, it uses -fvisibility=hidden
 if (${CMAKE_SYSTEM_NAME} STREQUAL "Linux" OR ${CMAKE_SYSTEM_NAME} STREQUAL "Darwin")
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fvisibility-ms-compat -fvisibility-inlines-hidden")
 elseif (${CMAKE_SYSTEM_NAME} STREQUAL "Android")
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fvisibility=hidden -fvisibility-inlines-hidden")
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fvisibility=hidde\n -fvisibility-inlines-hidden")
```
jherico, line 51: Uh... what?

kring: 😆 I don't know where that came from. I must have fat-fingered it. Thanks for flagging it!

.github/actions/vcpkg-cache/action.yml
```diff
@@ -0,0 +1,15 @@
...
+    required: true
+    type: string
+runs:
+  using: composite
+  steps:
+    - name: Cache vcpkg artifacts
+      uses: actions/cache@v4
+      with:
+        path: ${{ runner.os == 'Windows' && '/.ezvcpkg' || '~/.ezvcpkg' }}
+        key: vcpkg-${{ inputs.id }}-${{ hashFiles('extern/vcpkg-overlays/**/vcpkg.json') }}-${{ hashFiles('extern/cesium-native/CMakeLists.txt', 'extern/CMakeLists.txt') }}
```
jherico, line 13: does the caching mechanism automatically segment by OS, or is the intent to have all the different platforms / architectures in one cache blob?  The latter is going to be very tricky because as far as I know the cache upload won't merge the local packages with the overall package list.

Also, I don't know what the max cache entry size is, but trying to put all the binaries together is going to be more likely to overrun it.

jherico: Bleh, never mind... I get that the inputs.id will do the segemenation now.

extern/vcpkg-overlays/ktx/0001-Use-vcpkg-zstd.patch
```diff
@@ -0,0 +1,122 @@
...
+     lib/basisu/transcoder/basisu.h
+-    lib/basisu/zstd/zstd.c
+     lib/checkheader.c
+     lib/dfdutils/createdfd.c
+     lib/dfdutils/colourspaces.c
+@@ -532,7 +531,6 @@ macro(common_libktx_settings target enable_write library_type)
+         $<INSTALL_INTERFACE:lib/basisu/transcoder>
+ 
+         $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/lib/basisu/zstd>
+-        $<INSTALL_INTERFACE:lib/basisu/zstd>
```
jherico, line 17: _ptsd about basisu intensifies_

kring:
I finally got around to trying to use the CI-produced Windows binaries on my system, and of course they immediately crash on startup. The immediate problem is the lovely new MSVC 14.40 compiler that GitHub recently rolled out on the windows-2022 runner, which apparently generates code that crashes the first time your code tries to use std::mutex if your MSVCRT runtime is an older version. The real problem, though, is that vcpkg is not building with MSVC 14.34, which is the version UE 5.2 uses and therefore is also the version that cesium-native and all of its dependencies must use. Should be fixable by setting the [VCPKG_PLATFORM_TOOLSET_VERSION](https://learn.microsoft.com/en-us/vcpkg/users/triplets#vcpkg_platform_toolset_version) variable in the x64-windows-unreal.cmake triplet file.

kring:
Just dropping a reminder for myself here that this PR also removes the debug section compression on Linux. The compressed debug sections cause linker errors when cross-compiling for Linux from Windows:

> PackagingResults: Error: C:/Program Files/Epic Games/UE_5.2/Engine/Plugins/Marketplace/CesiumForUnreal/Intermediate/Build/Linux/x64/UnrealGame/Development/CesiumRuntime/Module.CesiumRuntime.2_of_11.cpp.o:(.debug_gnu_pubtypes) is compressed with ELFCOMPRESS_ZLIB, but lld is not built with zlib support

Now that the binaries aren't so enormous, compressing the debug sections isn't as essential.

kring:
This is working well and I believe it's finally ready for review!

kring:
Final testing across all platforms:

* Editor
  * [x] Windows x64
  * [ ] Linux x64
  * [x] macOS x64
  * [ ] macOS ARM64
* Build Game from Windows
  * [x] Windows x64
  * [x] Android ARM64
* Build Game from macOS
  * [x] macOS x64
  * [ ] macOS ARM64
  * [ ] iOS ARM64
* Build Game from Linux
  * [ ] Linux x64
* Run Game
  * [x] Windows x64
  * [ ] Linux x64
  * [x] macOS x64
  * [ ] macOS ARM64
  * [ ] iOS ARM64
  * [x] Android ARM64 (tested on Pixel 6 Pro)

j9liu (commented):
@kring Looks good to me! Just a few questions and a typo, but it doesn't have to hold up this PR.

Let me know when you're done with testing so I can hit merge!

kring:
Reminder for myself: I need to update CHANGES.md.

Source/CesiumEditor/Private/PCH.h
```diff
@@ -1,29 +0,0 @@
-// Copyright 2020-2024 CesiumGS, Inc. and Contributors
```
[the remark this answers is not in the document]
kring: PCH stands for "precompiled header". It's a header file that is included by all (or at least most) of the source files with the intention of making builds go faster by only processing big, complicated, and very commonly used headers once per project instead of once per source file. The downside is that it makes every object file bigger if the PCH includes headers that some source files otherwise wouldn't need. And also if any of the headers included by the precompiled header change, you get to recompile everything. I removed it because it wasn't really doing us any good.

Source/CesiumEditor/CesiumEditor.Build.cs
```diff
@@ -26,78 +24,68 @@ public CesiumEditor(ReadOnlyTargetRules Target) : base(Target)
...
             }
         }
 
-        PublicAdditionalLibraries.AddRange(libs.Select(lib => Path.Combine(libPath, libPrefix + lib + postfix + libPostfix)));
+        string libPath = useDebug ? libPathDebug : libPathRelease;
+
+        string[] allLibs = Directory.GetFiles(libPath, libSearchPattern);
+
+        // Remove libssl and libcrypto because these come from Unreal Engine itself
+        // TODO: Exclude these from the cesium-native side instead of excluding them here.
```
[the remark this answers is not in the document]
kring: I implemented what's suggested by this TODO and removed the TODO.

kring:
I've addressed the review feedback and tested every platform that I'm currently able to test. Let's not wait for the testing on the remaining platforms and instead get this into main as soon.

j9liu:
Thanks @kring ! Merging now 💪

PR #1458, opened 2024-06-18
Add functions to get the approximate location of a tileset during runtime

arbertrary:
As discussed in the following Cesium community forum thread, there is no API-level support to get the world location of a tileset during runtime from UE blueprints.

https://community.cesium.com/t/get-coordinates-of-cesium3dtileset-spawned-at-unknown-location/33087/4

As suggested by @kring I had a look at the `ACesium3DTileset::OnFocusEditorViewportOnThis()` function. 
With this pull request I added two functions that return an approximate location (Unreal and ECEF) of the tileset calculated using the rootTile.

**Disclaimer:**

- The code might not be in the most suitable locations, I simply lack the overview of the bigger picture of the Cesium for Unreal source code and where to place which kinds of functions
- I did not use the center of the bounding Box or Sphere to calculate the location. This is because for one of my own tilesets that I used to test this code with, this would always place me far above the tileset (thousands of meters)
- I think calling it "approximate" is necessary because otherwise this would probably fuel expectations that this code cannot fulfill 😄 
- I was only able to test these functions anecdotally

### DIFF
--- a/Source/CesiumRuntime/Private/Cesium3DTileset.cpp
+++ b/Source/CesiumRuntime/Private/Cesium3DTileset.cpp
@@ -165,6 +165,31 @@ void ACesium3DTileset::InvalidateResolvedGeoreference() {
   this->ResolvedGeoreference = nullptr;
 }
 
+FVector ACesium3DTileset::GetApproximateTilesetLocationUnreal() {
+    ACesiumGeoreference* pGeoreference = this->ResolveGeoreference();
+
+    FVector approximateRootTileLocationECEF = this->GetApproximateTilesetLocationECEF();
+    FVector approximateRootTileLocationUnreal = pGeoreference->TransformEarthCenteredEarthFixedPositionToUnreal(approximateRootTileLocationECEF);
+
+    return approximateRootTileLocationUnreal;
+}
+
+
+FVector ACesium3DTileset::GetApproximateTilesetLocationECEF() {
+    const Cesium3DTilesSelection::Tile* pRootTile =
+        this->_pTileset->getRootTile();
+    if (!pRootTile) {
+        return FVector::ZeroVector;
+    }
+
+    glm::dmat4x4 rootTileTransform = pRootTile->getTransform();
+    glm::dvec3 translation(rootTileTransform[3][0], rootTileTransform[3][1], rootTileTransform[3][2]);
+
+    FVector approximateRootTileLocationECEF = VecMath::createVector(translation);
+
+    return approximateRootTileLocationECEF;
+}
+
 TSoftObjectPtr<ACesiumCreditSystem> ACesium3DTileset::GetCreditSystem() const {
   return this->CreditSystem;
 }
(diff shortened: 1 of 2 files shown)


arbertrary:
If there is no root tile returned by `this->_pTileset->getRootTile()` this currently returns the Zerovector without warning which results in obviously incorrect coordinates.

I'm not sure what causes a tileset to not have a root tile but it seems to be the case for some tilesets. 

`ACesium3DTileset::OnFocusEditorViewportOnThis()` also simply returns if it encounters a nonexisting root tile so I'm not sure how to best handle this here 🤔

kring:
Thanks for the PR @arbertrary!

The root tile is part of the tileset.json, which is download when the tileset is first loaded. Until that download / parse completes, we don't know anything about the tileset, not even where it's located.

There's nothing for it other than to give the function a way to return "I don't know yet" as the answer.

arbertrary:
I have added a warning and a boolean out value to the function so that this gets propagated and can be checked in blueprints.

Unfortunately this wasn't actually the problem. For some tilesets the function returns 0,0,0 even if the root tile is available. The `!prootTile` condition was actually never entered.

Tilesets where that was the case:

- Nearmap Boston Photogrammetry https://ion.cesium.com/assetdepot/354759
- New York City 3D Buildings https://ion.cesium.com/assetdepot/75343

Tilesets where the code does work for me as expected are for example
 
- Melbourne Photogrammetry https://ion.cesium.com/assetdepot/69380
- Montreal Pointcloud https://ion.cesium.com/assetdepot/28945

arbertrary:
> Unfortunately this wasn't actually the problem. For some tilesets the function returns 0,0,0 even if the root tile is available. The `!prootTile` condition was actually never entered.

Checking whether the initial load of the tileset is complete by confirming that `LoadProgress` equals 100.0 before calling `getApproximateLocation` doesn't change this as well.

dyllan-aeroai:
Hello,
Thank you for generating this solution! I implemented your solution to my Cesium3DTilset C++ class exactly as you did, but I am unable to see those functions in the unreal editor. Did you also run into this issue / do you have any advice for exposing your functions in the unreal editor if it is not being shown?

kring:
@dyllan-aeroai if you're making changes to the plugin, be aware that Unreal Engine will not automatically recompile source files installed with the engine (i.e., in `C:\Program Files\Epic Games\UE_5.2\Engine\Plugins\Marketplace\CesiumForUnreal` or similar). You'll need to put the plugin into your project instead.

qichaoguo:
我尝试了您给出的解决方法，是可行的，太棒了，非常感谢！！！

dyllan-aeroai:
> @dyllan-aeroai if you're making changes to the plugin, be aware that Unreal Engine will not automatically recompile source files installed with the engine (i.e., in `C:\Program Files\Epic Games\UE_5.2\Engine\Plugins\Marketplace\CesiumForUnreal` or similar). You'll need to put the plugin into your project instead.

Thank you so much! This solution is now exposed in the blueprint editor!

arbertrary:
> Sorry to bother again, but did you ever figure out how to fix the tileset cases in which the function returned 0,0,0. For my program, certain tilesets are not loading such as the ones you pointed out (Nearmap Boston Photogrammetry and New York City 3D Buildings).

No. I haven't been able to figure out what the essential difference is between the tilesets for which this approach works and for those where it doesn't. 

Let me know if you do, though 😄

PR #1462, opened 2024-06-25
Add Doxygen step to CI

azrogers:
Bringing cesium-unreal in line with cesium-unity and cesium-native, this change adds a step to CI that builds the Doxygen documentation for us.


kring:
Thanks @azrogers!

PR #1464, opened 2024-06-27
Turn on debug assertions

csciguy8:
Assertions now work again for cesium-unreal debug builds.

Uses functionality from [cesium-native PR 874](https://github.com/CesiumGS/cesium-native/pull/874).

Also updates cesium-native to the latest


kring:
Thanks @csciguy8! I tweaked the cmake to use a generator expression so that it works with multi-config generators, too (CMAKE_BUILD_TYPE won't always be defined in this case). Merging.

Issue #1084, opened 2023-05-02
Degenerate triangle crashes in debug mode when using character controller and physics


j9liu:
Run 03_CesiumDenver in the Cesium for Unreal samples. As soon as you leave the cube that the character initially stands on, Unreal Engine crashes:

```
[2023.05.02-14.49.06:878][500]LogChaos: Warning: Degenerate triangle 2449: (26.909554 1580.875610 -317.959564) (26.919434 1580.875610 -317.987000) (26.900770 1580.875610 -317.987000)
[2023.05.02-14.49.06:881][500]LogOutputDevice: Warning: 

Script Stack (0 frames):

Ensure condition failed: false [File:D:\build\++UE5\Sync\Engine\Source\Runtime\Experimental\Chaos\Private\Chaos\TriangleMeshImplicitObject.cpp] [Line: 1285] 

[2023.05.02-14.49.06:882][500]LogOutputDevice: Error: Ensure condition failed: false [File:D:\build\++UE5\Sync\Engine\Source\Runtime\Experimental\Chaos\Private\Chaos\TriangleMeshImplicitObject.cpp] [Line: 1285] 


[2023.05.02-14.49.06:883][500]LogStats:             FDebug::EnsureFailed -  0.003 s
A breakpoint instruction (__debugbreak() statement or a similar call) was executed in UnrealEditor.exe.

The program '[38628] UnrealEditor.exe' has exited with code 0 (0x0).
```

The line number corresponds to the `ensure` in this part of the engine code:
```
if(Normal.SafeNormalize() < SMALL_NUMBER)
{
   UE_LOG(LogChaos, Warning, TEXT("Degenerate triangle %d: (%f %f %f) (%f %f %f) (%f %f %f)"), FaceIdx, A.X, A.Y, A.Z, B.X, B.Y, B.Z, 
   C.X, C.Y, C.Z);
   ensure(false);
   return FVec3(0, 0, 1);
}
```

Note that in a non-development build, the framerate stalls when you leave the cube, but the program _does_ resume and you are able to run across the tileset. I don't know if this is something we can reliably catch, but I'm noting it for future reference.

kring:
We should modify `BuildChaosTriangleMeshes` (in CesiumGltfComponent.cpp) to filter out degenerate triangles, using the same definition of degeneracy that Chaos itself uses. See the code in `TriangleMeshImplicitObject.cpp` mentioned in the log above.

[...]

PR #1465, opened 2024-06-30
Fix Chaos physics epsilon silliness by scaling meshes up and then down.

### LINKED ISSUE #1250: Line trace sometimes passes through tileset geometry
While testing metadata picking with 3D Tiles 1.1, I found that the San Francisco Ferry Building had some places where line traces would pass through the tileset geometry. See below:

image

This is just the debug view of line trace, so I don't think it's related to any of the features / metadata changes I'm making. I haven't tried it with other tilesets, so maybe it's just a quirk of this tileset? You can see this pass-through behavior most prominently in the middle window and around the pillar areas. 

image

[Tileset Google Drive link](https://drive.google.com/file/d/1giI-mBn156BLXHMkbHRqSVuzCgYbxjzy/view?usp=drive_link)

kring:
Fixes #1250

Lending further support to my theory that "anytime a software developer is typing the word 'epsilon', they're typing a bug"...
(See my previous rant about epsilons in CesiumGS/cesium-native#914)

We've had trouble with Chaos physics ever since Unreal Engine switched to it by default. The first problem was that Chaos would frequently crash in debug builds, or hang and spam the log in release builds, when doing collisions against many of our meshes (#1084). It was caused by checking for (and logging) "degenerate triangles". So, we decided to filter out degenerate triangles from our meshes using the same test that Chaos itself uses (#1106).

That worked well in that it stopped the crashing and freezing. But we still occasionally heard reports (#1250) of failed lined traces against our meshes. We assumed this was because the triangles it was supposed to hit were quite small, and so Chaos is (incorrectly) flagging them as degenerate.

But with the San Francisco Ferry Building model, the triangles being missed really aren't all that small. What is going on here? I finally spent a chunk of time looking at it today.

My diagnosis is that the degenerate triangle test that Chaos is using is astoundingly bad. It looks like this:

```cpp
const ParticleVecType& A = MParticles.X(Elements[FaceIdx][0]);
const ParticleVecType& B = MParticles.X(Elements[FaceIdx][1]);
const ParticleVecType& C = MParticles.X(Elements[FaceIdx][2]);

const ParticleVecType AB = B - A;
const ParticleVecType AC = C - A;
ParticleVecType Normal = ParticleVecType::CrossProduct(AB, AC);
			
if(Normal.SafeNormalize() < UE_SMALL_NUMBER)
{
	UE_LOG(LogChaos, Warning, TEXT("Degenerate triangle %d: (%f %f %f) (%f %f %f) (%f %f %f)"), FaceIdx, A.X, A.Y, A.Z, B.X, B.Y, B.Z, C.X, C.Y, C.Z);
	CHAOS_ENSURE(false);
	return FVec3(0, 0, 1);
}
```

A, B, and C are the three vertex positions and `UE_SMALL_NUMBER` is 1e-8. And if you dig into that `SafeNormalize`, you'll see there's yet another epsilon test. If the squared magnitude of the vector is less then 1e-4, then SafeNormalize will return 0.0. Zero is definitely less than 1e-8, so this warning will be triggered.

This computation is done in model coordinates. Which means.... wait for it... the quantities here do not have any known units (it depends on how the model is scaled). So here we go again applying an absolute epsilon to quantities of unknown magnitude. 🤦 Which is guaranteed to end in heartbreak.

Let's say our model coordinates are meters, and we have a simple right triangle that is 9cm on a side. The `Normal` will have a magnitude of 0.09 * 0.09 = 0.0081 meters. So the magnitude squared of that vector will be 0.00006561 m^2. That is clearly less than 1e-4, therefore SafeNormalize will return 0.0 and Chaos will log a "degenerate triangle" warning about this triangle! For a 9cm per side right triangle!

Maybe the author of this code didn't realize that `SafeNormalize` has a 1e-4 epsilon check on the magnitude squared? Probably not, because I don't think the UE_SMALL_NUMBER check will _ever_ make sense with that in place. I mean, to be fair, it's _absolutely insane_ that SafeNormalize - a function working with double-precision values! - returns zero when the magnitude squared is less then 1e-4.

I guess maybe UE can get away with this when model coordinates match world coordinates, and so the units are centimeters instead of meters. A 9cm triangle being degenerate is a big problem, but a 0.09cm triangle being degenerate is I guess kinda sorta more acceptable.

But it's worth noting that there are a bunch of other peope complaining about this problem; it's not just us:
https://forums.unrealengine.com/t/degenerate-triangle-and-tvector-ensure-crashes/544465

Ok, so what do we do about it? Short of changing UE itself, I can only think of one solution: change the scale of our mesh. So that's what this PR does. It multiplies every vertex position in our meshes by 1024.0 (chosen so that only the exponent of the floating point number is affected, not the mantissa... I think I did that right?). And then adjusts the UCesiumGltfPrimitiveComponent matrix to scale by 1 / 1024.0 to bring the model back down to the correct size. With the mesh scaled up by 1024, the silly absolute epsilon becomes less silly. It means our right triangle would have to be less 0.1mm to be considered degenerate. I tried to find a way to scale only the physics mesh, but it didn't seem possible.

This PR is a draft because it's not super well tested and needs some cleanup, but it seems to work very well with the basis tests I've done so far.

### DIFF
--- a/Source/CesiumRuntime/Private/CesiumGltfComponent.cpp
+++ b/Source/CesiumRuntime/Private/CesiumGltfComponent.cpp
@@ -3820,17 +3820,6 @@ void UCesiumGltfComponent::UpdateFade(float fadePercentage, bool fadingIn) {
   }
 }
 
-static bool isTriangleDegenerate(
-    const Chaos::FTriangleMeshImplicitObject::ParticleVecType& A,
-    const Chaos::FTriangleMeshImplicitObject::ParticleVecType& B,
-    const Chaos::FTriangleMeshImplicitObject::ParticleVecType& C) {
-  Chaos::FTriangleMeshImplicitObject::ParticleVecType AB = B - A;
-  Chaos::FTriangleMeshImplicitObject::ParticleVecType AC = C - A;
-  Chaos::FTriangleMeshImplicitObject::ParticleVecType Normal =
-      Chaos::FTriangleMeshImplicitObject::ParticleVecType::CrossProduct(AB, AC);
-  return (Normal.SafeNormalize() < 1.e-8f);
-}
-
 template <typename TIndex>
 #if ENGINE_VERSION_5_4_OR_HIGHER
 static Chaos::FTriangleMeshImplicitObjectPtr
--- a/Source/CesiumRuntime/Private/CesiumGltfComponent.cpp
+++ b/Source/CesiumRuntime/Private/CesiumGltfComponent.cpp
@@ -83,6 +83,15 @@ namespace {
 using TMeshVector2 = FVector2f;
 using TMeshVector3 = FVector3f;
 using TMeshVector4 = FVector4f;
+
+// We scale up the meshes because Chaos has a degenerate triangle
+// epsilon test in `TriangleMeshImplicitObject.cpp` that is almost laughably
+// too eager. Perhaps it would be fine if our meshes actually used units of
+// centimeters like UE, but they usually use meters instead. UE considering
+// a triangle that is ~10cm on each side to be degenerate is... infuriating.
+//
+// We scale up the collision meshes by a power-of-two factor so that only
+// the exponent portion of the floating point number is affected.
 constexpr double scaleFactor = 1024.0;
 } // namespace
 
@@ -3845,14 +3854,6 @@ BuildChaosTriangleMeshes(
   Chaos::TParticles<Chaos::FRealSingle, 3> vertices;
   vertices.AddParticles(vertexCount);
   for (int32 i = 0; i < vertexCount; ++i) {
-    // Scale up the collision meshes because Unreal has a degenerate triangle
-    // epsilon test in `TriangleMeshImplicitObject.cpp` that is almost laughably
-    // too eager. Perhaps it would be fine if our meshes actually used units of
-    // centimeters like UE, but they usually use meters instead. UE considering
-    // a triangle that is ~10cm on each side to be degenerate is... infuriating.
-    //
-    // We scale up the collision meshes by a power-of-two factor so that only
-    // the exponent portion of the floating point number is affected.
     vertices.X(i) = vertexData[i].Position;
   }
 
@@ -3868,15 +3869,8 @@ BuildChaosTriangleMeshes(
     int32 vIndex1 = indices[index0];
     int32 vIndex2 = indices[index0 + 2];
 
-    // if (!isTriangleDegenerate(
-    //         vertices.X(vIndex0),
-    //         vertices.X(vIndex1),
-    //         vertices.X(vIndex2))) {
     triangles.Add(Chaos::TVector<int32, 3>(vIndex0, vIndex1, vIndex2));
     faceRemap.Add(i);
-    //} else {
-    //  UE_LOG(LogCesium, Warning, TEXT("Degenerate"));
-    //}
   }
 
   TUniquePtr<TArray<int32>> pFaceRemap = MakeUnique<TArray<int32>>(faceRemap);
--- a/Source/CesiumRuntime/Private/CesiumGltfComponent.cpp
+++ b/Source/CesiumRuntime/Private/CesiumGltfComponent.cpp
@@ -83,6 +83,7 @@ namespace {
 using TMeshVector2 = FVector2f;
 using TMeshVector3 = FVector3f;
 using TMeshVector4 = FVector4f;
+constexpr double scaleFactor = 1024.0;
 } // namespace
 
 static uint32_t nextMaterialId = 0;
@@ -1309,6 +1310,9 @@ static void loadPrimitive(
       maxPosition = glm::dvec3(max[0], max[1], max[2]);
     }
 
+    minPosition *= scaleFactor;
+    maxPosition *= scaleFactor;
+
     primitiveResult.dimensions =
         glm::vec3(transform * glm::dvec4(maxPosition - minPosition, 0));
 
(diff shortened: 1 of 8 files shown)


arbertrary:
Very interesting find!

Seems to work well with the tilesets that were causing us collision problems in the past. 
Also seems to not have any performance repercussions as far as I as a user can see.

kring:
> Seems to work well with the tilesets that were causing us collision problems in the past.
Also seems to not have any performance repercussions as far as I as a user can see.

Good to hear. Thanks for trying it!

j9liu:
Something similar may be happening in Cesium for Unity as well: https://community.cesium.com/t/errors-in-playmode-after-updateing-to-unity-6/31892/7

kring:
This seems to be working well and is ready for review.

kring:
@csciguy8 can I bug you to review this one today, since Janine and Ashley are both out?

csciguy8:
You got it!

Source/CesiumRuntime/Private/CesiumPrimitive.h
```diff
@@ -101,6 +101,19 @@ class CesiumPrimitiveData {
...
+   * the Unreal mesh is populated.
+   *
+   * We scale up the meshes because Chaos has a degenerate triangle epsilon test
+   * in `TriangleMeshImplicitObject.cpp` that is almost laughably too eager.
+   * Perhaps it would be fine if our meshes actually used units of centimeters
+   * like UE, but they usually use meters instead. With a factor of 1.0, UE will
+   * consider a right triangle that is slightly less than ~10cm on each side to
+   * be degenerate.
+   */
+  static constexpr double positionScaleFactor = 1024.0;
```
csciguy8, line 115: Would be nice to see a comment as to why 1024 was chosen. From the PR you wrote "chosen so that only the exponent of the floating point number is affected, not the mantissa"

Source/CesiumRuntime/Private/Tests/CesiumFeatureIdSet.spec.cpp
```diff
@@ -542,7 +542,8 @@ void FCesiumFeatureIdSetSpec::Define() {
       FHitResult Hit;
       Hit.Component = pPrimitiveComponent;
       Hit.FaceIndex = 0;
-      Hit.Location = FVector_NetQuantize(0, -1, 0);
+      Hit.Location = FVector_NetQuantize(0, -1, 0) *
```
csciguy8, line 545: Nitpick here, the 0 components are unnecessarily getting multiplied by the scale factor

Source/CesiumRuntime/Private/Tests/CesiumFeatureIdTexture.spec.cpp
```diff
@@ -552,7 +552,8 @@ void FCesiumFeatureIdTextureSpec::Define() {
           ECesiumFeatureIdTextureStatus::Valid);
 
       FHitResult Hit;
-      Hit.Location = FVector_NetQuantize::Zero();
+      Hit.Location = FVector_NetQuantize::Zero() *
```
csciguy8, line 555: Similar nitpick here, any instance of `FVector_NetQuantize::Zero() ` or `FVector_NetQuantize{0, 0, 0}` doesn't need to actually need to use the scale factor

Source/CesiumRuntime/Private/CesiumGltfComponent.cpp
```diff
@@ -3849,13 +3859,8 @@ BuildChaosTriangleMeshes(
...
 
-    if (!isTriangleDegenerate(
-            vertices.X(vIndex0),
-            vertices.X(vIndex1),
-            vertices.X(vIndex2))) {
-      triangles.Add(Chaos::TVector<int32, 3>(vIndex0, vIndex1, vIndex2));
-      faceRemap.Add(i);
-    }
+    triangles.Add(Chaos::TVector<int32, 3>(vIndex0, vIndex1, vIndex2));
+    faceRemap.Add(i);
```
csciguy8, line 3863: Noticed you removed the call to `isTriangleDegenerate`, and rightly so. It looks unused now, and might want to consider removing that function completely so people don't go down this rabbit hole again.

csciguy8 (approved):
Nice work @kring !

Left some optional comments for you, but all looks good to go. Tested all unreal sample levels, including the ferry building windows, which was mentioned in the original issue.

kring:
Thanks for the review @csciguy8! I've removed the unused function and added a comment about the scale factor being a power of two so that it only affects the exponent.

I decided to leave in the unnecessary multiplications of zero by the scaling factor, though. I think the performance impact is irrelevant (especially because it's only in test code), and there's a bit of clarity to be gained by leaving it in. Particularly if someone in the future copies a test and modifies the vector value, there's less possibility of error if they don't also have to remember to add back in the multiplication by the scaling factor. Or, looking at it another way, the multiplication by the scaling factor in the tests is useful documentation that the scaling is needed in general, even if it happens to not be needed for some of the particular values used. But let me know if you feel strongly about it.

I'll merge this one CI passes.

csciguy8:
> But let me know if you feel strongly about it.

I have no strong feelings, all good.

WxpGiser:
I find a bug,in CesiumGltfComponent.cpp:
instanceTransforms[i] = glm::translate(
            instanceTransforms[i],
            translation);
should be:
instanceTransforms[i] = glm::translate(
            instanceTransforms[i],
            translation * CesiumPrimitiveData::positionScaleFactor);

j9liu:
Thank you for bringing this to our attention @WxpGiser ! Would you be able to open a PR with the changes described? That would allow us to merge into main ASAP. :pray:

WxpGiser:
I can't create new branch ,you just do it.

j9liu:
@WxpGiser We definitely want to do it, but we are cautious about copying code due to legal reasons. If you're unable to contribute a PR, could you please sign [a CLA](https://github.com/CesiumGS/cesium/blob/main/CONTRIBUTING.md#contributor-license-agreement-cla) so we can get this in ASAP?

WxpGiser:
Cesium Individual Contributor License Agreement ("Agreement") v1.0
Thank you so much! If you have a contribution pending review on GitHub, please let the reviewer know that you've signed the CLA.

j9liu:
Thanks @WxpGiser ! I just opened #1499 with this fix. :smile:

PR #1468, opened 2024-07-02
Fix bad_any_cast errors by disabling RTTI when building cesium-native

### LINKED ISSUE #1380: Quest 2/Android build: Unable to parse EXT_structural_metadata extension (or  perhaps any extension?)
I have a 3D Tiles dataset that utilizes EXT_structural_metadata in the gLTF models, however when processing the models I see 

[error] [TilesetContentManager.cpp:997] An unexpected error occurs when loading tile: bad any cast

I built cesium-native in debug to see what was going on and noticed this line was failing https://github.com/CesiumGS/cesium-native/blob/5efd7ba94f61f27a7663bb8ef558120e45bbe71f/CesiumGltfReader/generated/src/GeneratedJsonHandlers.cpp#L808. Puzzling to me as nothing looks wrong with that code to me, seems like something is getting corrupted?

This only happens when I build for the Quest 2, but if I run it on Windows (and or use Quest Link to display on the headset) there is no issue processing the models. 

Unfortunately I'm unable to share the model, is there any small sample out there that uses EXT_structural_metadata that I could test with?

kring:
Fixes #1380 

We're seeing `bad_any_cast` exceptions when loading tilesets with metadata, in release builds, on every platform except Windows. The cause appears to be that cesium-native is compiled with runtime type information (RTTI) enabled, but Cesium for Unreal - as required by Unreal Engine itself - is compiled with RTTI disabled. Windows isn't fussed about this, but apparently it causes subtle problems on Linux/Android/Apple.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,11 @@
 # Change Log
 
+### ? - ?
+
+##### Fixes :wrench:
+
+- Fixed a bug that could cause a `bad_any_cast` exception when trying to access glTF extensions. This commonly popped up when loading tilesets with metadata.
+
 ### v2.7.0 - 2024-07-01
 
 ##### Additions :tada:
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,7 +4,8 @@
 
 ##### Fixes :wrench:
 
-- Fixed a bug that could cause a `bad_any_cast` exception when trying to access glTF extensions. This commonly popped up when loading tilesets with metadata.
+- Fixed a bug that could cause a `bad_any_cast` exception when trying to access glTF extensions on non-Windows platforms. This commonly popped up when loading tilesets with metadata.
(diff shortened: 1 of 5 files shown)


kring:
This PR also includes a bug fix to the `GetInteger64` metadata functions on non-Windows platforms. None of them were working because on 64-bit clang platforms (definitely Linux, pretty sure about the others):

* `std::int64_t` is a typedef to `signed long int`.
* Unreal's `int64` is a typedef to `signed long long`.

Both are 64-bit and seemingly identical. But they're different types, so `IsMetadataInteger<int64>` evaluates to false. :upside_down_face: 

So this PR makes sure we use `int64_t` instead of `int64` when instantiating cesium-native's type traits templates.

j9liu:
Thanks @kring !

PR #1491, opened 2024-08-08
Add network profiling tests (Google P3D tiles)

csciguy8:
Adds some documentation and python scripts for investigating network performance when download Google P3D Tiles.


j9liu:
Closing for housekeeping purposes (and we don't remember the context for why this is draft), but we'll keep the branch up in case we decide to return to this.

PR #1497, opened 2024-08-14
Add `forEachRenderableTile`

j9liu:
I noticed that `Cesium3DTileset.cpp` has a lot of duplication when iterating over tiles with `UCesiumGltfComponent`. So, for convenience / reusability, I added a `forEachRenderableTile` template function that allows for a callback on each `UCesiumGltfComponent` in a collection of tiles.

I've updated a handful of functions to use this call. It can also be reused if we ever need to collectively transform / iterate over tiles in future functionality. I haven't noticed any clear performance differences, but let me know if something here is in bad form.


j9liu:
Forgot that only Unreal 5.4 is on C++ 20 🤦 I'll add overloads for the time being.

kring:
> Forgot that only Unreal 5.4 is on C++ 20

#1447 moves all UE versions to C++20.

j9liu:
Merged with main. This is ready for a look!

Issue #1496, opened 2024-08-14
Instanced Tileset not rendered right


Doenerlord:
Hi!

I'm currently using Unreal 5.4 with Cesium for Unreal 2.7.1 and also have tested Unreal 5.3 and Plugin versions since 2.5.

When I'm trying to load a instanced tileset containing some trees they are not displayed right. Some of them are squeezed into one location and none have any leaves, only the trunk. 
A sample screenshot of what it looks like in the editor:
![image](https://github.com/user-attachments/assets/e294c087-23d7-44e8-98e7-39fbc5ef0c8d)

In Cesium Ion the trees are displayed correctly as you can see in this screenshot:
![image](https://github.com/user-attachments/assets/14cd3e7f-cbe0-4e79-94c9-69cb2dfeb74f)

It does not matter wether I load the tileset locally or stream it in via cesium ion. I've tried multiple tilesets and different .glb files too.

Here is the tileset I've used in the screenhots. The trees are located near Karlsplatz, Vienna, Austria. (48.19951356208172, 16.372679666511534)
[SampleTrees.zip](https://github.com/user-attachments/files/16614441/SampleTrees.zip)

j9liu:
Hi @Doenerlord,

This is a known issue -- someone from the community identified the fix here: https://github.com/CesiumGS/cesium-unreal/pull/1465#issuecomment-2285559205

We're waiting on clearance to implement their changes so we can get a fix in ASAP. 🙏

j9liu:
@Doenerlord I confirmed that #1499 solves the placement issues, but the leaves are still a mystery. I'll make a note for our team to triage this when we get the chance.

kring:
I did a bit of digging into this. When we look at the root tile of this tileset in Unreal, we actually see the leaves only, no branches (it's only when we zoom in close that we get the opposite):

![Image](https://github.com/user-attachments/assets/854b0916-3a3c-4178-99c2-ecbced479cff)

The root tile has a glTF that looks like this (after it's converted from i3dm):

```
Node 0
  Node 1
    EXT_mesh_gpu_instancing
    Mesh 0
      Leaves primitive
  Node 2
    EXT_mesh_gpu_instancing
    Mesh 1
       Branches primitive
```

So, a single i3dm is being converted into two instances of the instancing extension. And all of the translations, rotations, and scales in the extension in Node 2 are zero, which is why we don't see any branches.

[...]

PR #1499, opened 2024-08-15
Add missing scale factor to instanced translations

j9liu:
Instanced translations were missing the `positionScaleFactor`, which made instances stack on top of each other incorrectly. Thanks @WxpGiser for the fix!

This fixes the translation errors in #1496, but there are still lingering issues, so I'll leave that open for now.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,11 @@
 # Change Log
 
+### ? - ?
+
+##### Fixes :wrench:
+
+- Fixed a bug introduced in the previous release that caused instanced tilesets to render incorrectly.
+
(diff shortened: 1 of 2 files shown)


kring:
Thanks @WxpGiser and @j9liu! And sorry to everyone that I missed this in #1465. This looks good to me, I'll merge it as soon as CI is happy.

timoore:
Sorry that I missed this... I don't think I want to scale the glTF instance translation directly this way. It should be build into the matrix that is currently called the yInvertMatrix. This is for consistency with the approach I described in my talk, at the very least, and may be more correct as well. I'll look at it in #1496.

PR #1510, opened 2024-08-31
Fix macOS build by freeing up disk space on the runner

kring:
The [latest update](https://github.com/actions/runner-images/compare/macos-14-arm64/20240818.4...macos-14-arm64/20240827.4#diff-5c04a529d3c8adf7a5f23afe544071dad1853e281c9c7b44cd8d626b6c57444d) to the GitHub Actions macos-14 runner has drastically less disk than the previous image version, mostly because more stuff is installed. So this PR uninstalls some of that stuff so we have enough space for our builds again.

See here:
https://github.com/actions/runner-images/issues/10511

### DIFF
--- a/.github/workflows/buildApple.yml
+++ b/.github/workflows/buildApple.yml
@@ -28,10 +28,13 @@ jobs:
       - name: Make some more disk space
         run: |
           df -h
+          ls /Applications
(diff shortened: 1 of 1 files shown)


kring:
Merging this to fix main.

PR #1515, opened 2024-09-04
Manually copy vertices to work around UE 5.3/5.4 bug.

### LINKED ISSUE #1513: UE 5.3 and 5.4 always create 8 sets of texture coordinates for our meshes, even though we usually only need 1 or 2
A user on the community forum noticed that when using Cesium for Unreal in UE 5.3 (and 5.4) to render their model, the frame rate is much lower than in UE 5.2:
https://community.cesium.com/t/fps-only-half-in-ue5-3-compare-to-ue5-2-3d-tiles-data-and-camera-position-is-same-draws-and-primitives-is-same-do-i-miss-some-configuration/34689

I initially couldn't reproduce it, but they very helpfully provided a project to demonstrate the problem. After some investigation, I learned that the immediate cause of the slowdown in UE 5.3 in that project is that in UE 5.3, the level is using _drastically_ more GPU memory than in 5.2. Once GPU memory is maxed out, the frame rate plummets because data needs to be streamed onto the GPU while rendering each frame.

After hours of spelunking through the UE source code, I found the reason: UE 5.3 introduced what I would consider a bug in this overload of `FStaticMeshVertexBuffer::Init`, which we use to create our meshes:

```cpp
void Init(const TArray<FStaticMeshBuildVertex>& InVertices, uint32 InNumTexCoords, bool bNeedsCPUAccess = true)
```

Notice it takes `InNumTexCoords` as the second parameter. In [UE 5.2](https://github.com/EpicGames/UnrealEngine/blob/5.2.1-release/Engine/Source/Runtime/Engine/Public/Rendering/StaticMeshVertexBuffer.h#L180), that parameter is used. In [UE 5.3](https://github.com/EpicGames/UnrealEngine/blob/5.3.2-release/Engine/Source/Runtime/Engine/Public/Rendering/StaticMeshVertexBuffer.h#L188), it is not. Instead, UE 5.3 creates `MAX_STATIC_TEXCOORDS` sets of texture coordinates, which is 8.

As a result, a typical Cesium for Unreal mesh uses 8x more GPU memory for texture coordinates in UE 5.3 than it should and than it does in UE 5.2.

This perhaps went unnoticed by the Unreal developers because UE 5.2 (but not 5.3) has a bug in the code that measures how much memory is used by a static mesh, undercounting the memory used by position data by twelve times (it counts each position as one byte instead of the twelve bytes it actually is in most cases). That bug was fixed in UE 5.3 (by [this commit](https://github.com/EpicGames/UnrealEngine/commit/93b51b22e346c2fddb5c78ae35488282023291ba)). So I guess if the developers saw the greatly increased memory usage in 5.3, they might have thought it to be red herring caused by the undercounting in 5.2, rather than an actual increase in 5.3.

The fix for this is relatively easy: use the `Init` method that takes a just vertex and texture coordinate count, and then manually set each attribute of each vertex. It may even be possible to avoid copying the glTF vertex data into the array of `FStaticMeshBuildVertex` instances at all, which would be nice.

kring:
Fixes #1513

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,7 +4,7 @@
 
 ##### Fixes :wrench:
 
-- Drastically reduced tile mesh memory usage in UE 5.3 and 5.4 by working around a bug that causes those engine versions to add more sets of texture coordinates than are necessary.
+- Drastically reduced tile mesh memory usage in UE 5.3 and 5.4 by working around a bug that causes those engine versions to add more texture coordinate sets than necessary.
 
 ### v2.8.0 - 2024-09-02
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,11 @@
 # Change Log
 
+### Next version (not released yet)
+
+##### Fixes :wrench:
+
+- Drastically reduced tile mesh memory usage in UE 5.3 and 5.4 by working around a bug that causes those engine versions to add more sets of texture coordinates than are necessary.
+
 ### v2.8.0 - 2024-09-02
 
 ##### Additions :tada:
--- a/Source/CesiumRuntime/Private/CesiumGltfComponent.cpp
+++ b/Source/CesiumRuntime/Private/CesiumGltfComponent.cpp
@@ -1628,11 +1628,40 @@ static void loadPrimitive(
       ColorVertexBuffer.Init(StaticMeshBuildVertices, false);
     }
 
-    LODResources.VertexBuffers.StaticMeshVertexBuffer.Init(
-        StaticMeshBuildVertices,
-        gltfToUnrealTexCoordMap.size() == 0 ? 1
-                                            : gltfToUnrealTexCoordMap.size(),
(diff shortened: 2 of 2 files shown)


CHANGES.md
```diff
@@ -1,5 +1,11 @@
 # Change Log
 
+### Next version (not released yet)
+
+##### Fixes :wrench:
+
+- Drastically reduced tile mesh memory usage in UE 5.3 and 5.4 by working around a bug that causes those engine versions to add more sets of texture coordinates than are necessary.
```
j9liu, line 7: Minor changelog nitpick for brevity :smile:

```suggestion
- Drastically reduced tile mesh memory usage in UE 5.3 and 5.4 by working around a bug that causes those engine versions to add more texture coordinate sets than necessary.
```

j9liu:
Thanks @kring ! Merging now.

PR #1517, opened 2024-09-05
Don't crash .build.cs files if native libs haven't been built.

kring:
The two .build.cs files have a line like this:

```csharp
string[] allLibs = Directory.GetFiles(libPath, libSearchPattern);
```

That line will throw an exception if `libPath` doesn't exist, which will be the case before cesium-native is built at all. This is mostly only a problem if you're trying to get Unreal to "Generate Visual Studio project files" before building Native. It's not too catastrophic, but slightly annoying, and easy to fix.


j9liu:
Looks good @kring ! For the record I haven't bothered to test this, but the change seems straightforward. Merging now.

PR #1520, opened 2024-09-09
Fixed build for Unreal Engine 5.4

### LINKED ISSUE #1519: Build fails on Unreal 5.4 using Cesium For Unreal 2.8.0
Hi there,

I am currently getting the following build error:
`Plugins\CesiumForUnreal\Source\CesiumRuntime\Private\UnrealAssetAccessors.cpp(154): error C4668: 'ENGINE_VERSION_5_4_OR_HIGHER' is not defined as a preprocessor macro, replacing with '0' for '#if/#elif'`

System information:
- Unreal Engine 5.4.4
- Cesium for Unreal 2.8.0
- Visual Studio 2022 14.38.33139 toolchain
- Windows 10.0.19041.0 SDK

I believe its because UnrealAssetAccessor doesn't include `CesiumCommon.h` which is where the preprocessor is defined.

I will make a PR now to resolve this issue!

KieranCoppins:
- Added include for CesiumCommon which contains the definition of the engine version preprocessor
- Fixes #1519

### DIFF
--- a/Source/CesiumRuntime/Private/UnrealAssetAccessor.cpp
+++ b/Source/CesiumRuntime/Private/UnrealAssetAccessor.cpp
@@ -1,6 +1,7 @@
 // Copyright 2020-2024 CesiumGS, Inc. and Contributors
 
 #include "UnrealAssetAccessor.h"
+#include "CesiumCommon.h"
 #include "Async/Async.h"
 #include "Async/AsyncWork.h"
 


j9liu:
Thanks so much @KieranCoppins ! I just ran into this problem myself actually, so it's great to get an early fix :smile:

PR #1521, opened 2024-09-11
Cache images across glTFs to avoid duplication

azrogers:
This implements the changes from CesiumGS/cesium-native#926 in Unreal. Now, when the same image is used by multiple glTFs, the texture data is only uploaded to the GPU once. Subsequent glTFs using the same image will create a new reference to the existing texture resource, reducing memory consumption.


kring (commented):
@azrogers I have some specific comments below, but I haven't looked at the code exhaustively.

More generally, I think the chains of futures here have gotten really complicated, and I'm not sure it's justified. From what I can tell, the only reason that any of the CesiumGltfComponent code has been futurified at all is because of the need to generate mipmaps, and the desire to avoid blocking worker threads (via waiting on a mutex) while some other thread generates mipmaps for a shared image.

Do I have that right? If so, I think a little restructuring will go a long way.

Currently, the mipmap generation starts right at the beginning of the process in `loadModelAnyThreadPart`. We get a SharedFuture for its completion. Unfortunately that `runInWorkerThread` continuation in `preprocessImage` is going to be run synchronously, because we're already in a worker thread (and AsyncSystem knows it), and so the mutex is going to be held for the entire duration of the mipmap generation. This will not only block all threads that need this same image, it will also block all glTF loading period (even across tilesets) due to the use of the global mutex. Ok, so that has to be fixed, or else all this complexity is pointless.

But let's say we fix that, without getting into the details of how we would do it....

Next we start the main part of the glTF loading process. Way down in the depths of creating of Unreal components from glTF node -> mesh -> primitive objects, we need to create some textures. And creating those textures requires first waiting on that previously-started mipmap generation. Which requires that process to be async (i.e., return a Future), and that requirement bubbles its way through everything.

I think we can drastically simplify this, with very little tradeoff, by simply moving the mipmap continuation earlier. `loadModelAnyThreadPart` would look something like this:

```cpp
static CesiumAsync::Future<TUniquePtr<UCesiumGltfComponent::HalfConstructed>>
loadModelAnyThreadPart(
    const CesiumAsync::AsyncSystem& asyncSystem,
    const glm::dmat4x4& transform,
    const CreateModelOptions& options,
    const CesiumGeospatial::Ellipsoid& ellipsoid) {
  return createMipMapsForAllTextures(...).thenInWorkerThread([...]() {
    doTheRestOfTheNormalProcessSynchronously(...);
  });
}
```

In other words, we wait (via continuation, not actually waiting) for mipmaps to be available for all of our images before we do the rest of the process, and we don't need Futures in any of the rest of the process at all.

Now, you might legitimately point out that this is a little less efficient. If we have multiple tiles sharing an image, we conceptually could have one thread generating mipmaps for that image while the other threads do the rest of the model loading process in parallel, and the design I proposed would preclude this. But in this scenario, we're no less efficient than we would be if each tile had its own image. And also, if we're worried about this inefficiency, shouldn't we be more worried about the same inefficiency while actually downloading the image? We could conceptually start creating Unreal meshes while downloading the shared images, too, not just while generating mipmaps for them.

Considering how much complexity it eliminates to ignore this small inefficiency, I think it's a really good tradeoff. Especially remembering that all those extra Futures flying around have a runtime cost, too.

Now, circling back, the trick to implement `createMipMapsForAllTextures` is to use a `Promise`. Roughly:
1. Lock the mutex
2. Get/add the extension. If it already exists, return its Future and we're done. Otherwise,
3. Create a Promise.
4. Get the Promise's Future and call `share` to make a SharedFuture from it
5. Assign the SharedFuture to the field on the extension
6. Unlock the mutex
7. Generate the mipmaps in the current thread.
8. Call `resolve` on the Promise.

This is the basic process needed to fix the problems in the current design as well.

Source/CesiumRuntime/Private/Cesium3DTileset.cpp
```diff
@@ -807,17 +815,18 @@ class UnrealResourcePreparer
       }
     }
 
+    CesiumGltf::SharedAsset<CesiumGltf::ImageCesium> imageAsset(image);
```
kring, line 818: Won't this copy the Image? That seems expensive. As I mentioned when I reviewed the cesium-native side of this, I don't really see much need for this whole SharedAsset / SharedContainer design. Having `Image` hold `IntrusivePointer<ImageCesium>` seems like it should be sufficient.

Source/CesiumRuntime/Private/Cesium3DTileset.cpp
```diff
@@ -1863,6 +1872,46 @@ void ACesium3DTileset::updateLastViewUpdateResultState(
...
   TRACE_CPUPROFILER_EVENT_SCOPE(Cesium::updateLastViewUpdateResultState)
 
+  if (this->DrawTileInfo) {
+    const UWorld* World = GetWorld();
+    check(World);
+
+    const TSoftObjectPtr<ACesiumGeoreference> Georeference = GetGeoreference();
+    check(Georeference);
+
+    for (auto& tile : result.tilesToRenderThisFrame) {
```
kring, line 1882: Don't use auto here. https://github.com/CesiumGS/cesium-native/blob/main/doc/style-guide.md#-auto

Source/CesiumRuntime/Private/Cesium3DTileset.cpp
```diff
@@ -1863,6 +1872,46 @@ void ACesium3DTileset::updateLastViewUpdateResultState(
...
+
+      FVector unrealCenter =
+          Georeference->TransformEarthCenteredEarthFixedPositionToUnreal(
+              VecMath::createVector(obb.getCenter()));
+
+      FString text = FString::Printf(
+          TEXT("ID %s (%p)"),
+          Cesium3DTilesSelection::TileIdUtilities::createTileIdString(
+              tile->getTileID())
+              .c_str(),
```
kring, line 1897: This isn't working, at least not with Cesium World Terrain:

alt text

Source/CesiumRuntime/Private/CesiumTextureUtility.cpp
```diff
@@ -352,54 +560,55 @@ TUniquePtr<LoadedTextureResult> loadTextureFromImageAndSamplerAnyThreadPart(
...
+bool getUseMipmapsIfAvailableFromSampler(const CesiumGltf::Sampler& sampler) {
   switch (sampler.minFilter.value_or(
       CesiumGltf::Sampler::MinFilter::LINEAR_MIPMAP_LINEAR)) {
   case CesiumGltf::Sampler::MinFilter::LINEAR_MIPMAP_LINEAR:
   case CesiumGltf::Sampler::MinFilter::LINEAR_MIPMAP_NEAREST:
   case CesiumGltf::Sampler::MinFilter::NEAREST_MIPMAP_LINEAR:
   case CesiumGltf::Sampler::MinFilter::NEAREST_MIPMAP_NEAREST:
-    useMipMapsIfAvailable = true;
+    return true;
     break;
```
kring, line 592: Unnecessary break now that there's a return.

Source/CesiumRuntime/Private/CesiumTextureUtility.cpp
```diff
@@ -670,4 +762,93 @@ TextureAddress convertGltfWrapTToUnreal(int32_t wrapT) {
...
+
+  // Generate mipmaps if needed.
+  // An image needs mipmaps generated for it if:
+  // 1. It is used by a Texture that has a Sampler with a mipmap filtering
+  // mode, and
+  // 2. It does not already have mipmaps.
+  // It's ok if an image has mipmaps even if not all textures will use them.
+  // There's no reason to have two RHI textures, one with and one without
+  // mips.
+  bool needsMipmaps;
```
kring, line 788: Could replace with a call to `getUseMipmapsIfAvailableFromSampler` now that you've factored that out.

Source/CesiumRuntime/Private/CesiumTextureUtility.cpp
```diff
@@ -670,4 +762,93 @@ TextureAddress convertGltfWrapTToUnreal(int32_t wrapT) {
...
+            sampler,
+            *model.images[texture.source].cesium);
+    if (optionalFuture.has_value()) {
+      futures.push_back(optionalFuture.value());
+    }
+  }
+
+  return asyncSystem.all(std::move(futures))
+      .thenImmediately(
+          []([[maybe_unused]] std::vector<CesiumGltf::ImageCesium*>&& results)
```
kring, line 849:
```suggestion
          [](std::vector<CesiumGltf::ImageCesium*>&& /*results*/)
```

From our style guide: 
https://github.com/CesiumGS/cesium-native/blob/main/doc/style-guide.md#%EF%B8%8F-other

Source/CesiumRuntime/Private/CesiumTextureResource.h
```diff
@@ -110,7 +111,7 @@ class FCesiumUseExistingTextureResource : public FCesiumTextureResourceBase {
 class FCesiumCreateNewTextureResource : public FCesiumTextureResourceBase {
 public:
   FCesiumCreateNewTextureResource(
-      CesiumGltf::ImageCesium&& image,
+      CesiumGltf::ImageCesium& image,
```
kring, line 114: The passed image is still being moved away, so the `&&` is more accurate. Why the change?

Source/CesiumRuntime/Private/CesiumGltfComponent.cpp
```diff
@@ -2267,119 +2230,101 @@ loadModelMetadata(LoadModelResult& result, const CreateModelOptions& options) {
...
+                    dummyNodeResult.meshResult,
+                    rootTransform,
+                    meshOptions,
+                    ellipsoid);
+              }
+            }
 
-  glm::dmat4x4 rootTransform = transform;
+            auto pResult =
+                MakeUnique<UCesiumGltfComponent::CreateOffGameThreadResult>();
```
kring, line 2322: The `MakeUnique` is unnecessary. Just use a simple value.

Source/CesiumRuntime/Private/CesiumGltfComponent.cpp
```diff
@@ -1848,26 +1828,20 @@ static void loadPrimitive(
...
-  const Model& model = *options.pNodeOptions->pModelOptions->pModel;
-  Mesh& mesh = *options.pMesh;
+  Model& model = *options.pNodeOptions->pModelOptions->pModel;
+  Mesh& mesh = model.meshes[options.meshIndex];
 
   result = LoadMeshResult();
   result->primitiveResults.reserve(mesh.primitives.size());
-  for (CesiumGltf::MeshPrimitive& primitive : mesh.primitives) {
-    CreatePrimitiveOptions primitiveOptions = {&options, &*result, &primitive};
+  for (int32_t i = 0; i < mesh.primitives.size(); i++) {
```
kring, line 1841:
```suggestion
  for (size_t i = 0; i < mesh.primitives.size(); i++) {
```

kring (changes requested):
It's definitely getting there @azrogers! I have lots of small comments, but the only big problem is the global lock during texture creation. Let me know if you want to discuss ways to solve that.

kring:
A few things before we can merge this, in order from rather important to mostly trivial. 😁 

* [ ] I have a level with CWT + Bing + a WMS, and there seems to be quite a large memory leak. Process memory increases quickly even just spinning in place. I spent some time trying to track this down but haven't had much success yet. Presumably it's the Unreal textures leaking, because we made changes here and because the entire thing is crazy complicated.
* [ ] What's going on with `Cesium3DTileset.spec.cpp`? It has both a unit test and a performance test (we usually separate these into different files), and both look more like a quick experiment than something that should live in the code base long term. Should we polish these up? Remove them?
* [ ] We need to update CHANGES.md.

kring:
The memory leak and macOS build problems are fixed. @azrogers can you please give this a self review today, and tick off the other two issues (plus test failures if there still are any), and we should be in good shape to merge it for the release tomorrow.

PR #1524, opened 2024-09-20
Fix incorrect metadata property values being sent to material

j9liu:
In our metadata handling code, I found some unfortunate typos that happened as a result of overzealous copy-pasting. For some reason the `encodedProperty.offset` value was being passed for the `scale`, `noData`, and `default` value properties, which is obviously wrong. 🙃  Plus, I had typoed a `+` sign as an `=` sign, so one parameter was never being set.

This PR does two things:
1. Fixes the aforementioned bugs
2. Moves the `Set____ParameterValues` functions from `CesiumGltfComponent.cpp` to `CesiumEncodedFeaturesMetadata.h`. This was not strictly necessary, but I think this organization is helpful for decreasing the complexity of `CesiumGltfComponent`, and for putting the code closer to the rest of the `CesiumEncodedFeaturesMetadata` functionality.

If you'd like, you can verify that these values are passed correctly using the CesiumJS sample dataset in `Specs\Data\Cesium3DTiles\Metadata\StructuralMetadata`. Notably the `employeeCount`, `year`, and `temperatureCelsius` properties will be affected.

For `temperatureCelsius` you can reproduce this result with these material nodes: 
| Result | Nodes | 
| ------ | -------- |
| image| image |
> You can also get the same result by dividing the **Raw Value** by 255, and lerp-ing that.

For `year` you can reproduce this result with these nodes, though it simply tests that the default value is passed in.

| Result | Nodes |
| ------- | ------- |
| image | image |

For `employeeCount` you can reproduce this result with these nodes:

| Result | Nodes |
| -------| -------- |
| image | image |

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,7 @@
 ##### Fixes :wrench:
 
 - Drastically reduced tile mesh memory usage in UE 5.3 and 5.4 by working around a bug that causes those engine versions to add more texture coordinate sets than necessary.
+- Fixed a bug where the `scale`, `noData`, and `default` values of a property in `EXT_strutural_metadata` were not correctly passed to the tileset's material.
 
 ### v2.8.0 - 2024-09-02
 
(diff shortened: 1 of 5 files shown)


kring:
Looks good, thanks @j9liu!

PR #1526, opened 2024-09-27
Handle SampleHeightMostDetailed on tilesets that completely fail to load, and add a test.

kring:
This depends on CesiumGS/cesium-native#952 so merge that first.

### DIFF
--- a/Source/CesiumRuntime/Private/Cesium3DTileset.cpp
+++ b/Source/CesiumRuntime/Private/Cesium3DTileset.cpp
@@ -142,8 +142,6 @@ void ACesium3DTileset::SampleHeightMostDetailed(
         position.Z));
   }
 
-  size_t count = positions.size();
-
   auto sampleHeights = [this, &positions]() mutable {
     if (this->_pTileset) {
       return this->_pTileset->sampleHeightMostDetailed(positions)
--- a/Source/CesiumRuntime/Private/Cesium3DTileset.cpp
+++ b/Source/CesiumRuntime/Private/Cesium3DTileset.cpp
@@ -142,19 +142,30 @@ void ACesium3DTileset::SampleHeightMostDetailed(
         position.Z));
   }
 
+  size_t count = positions.size();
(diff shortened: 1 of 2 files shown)


Source/CesiumRuntime/Private/Cesium3DTileset.cpp
```diff
@@ -142,19 +142,30 @@ void ACesium3DTileset::SampleHeightMostDetailed(
...
-                    std::vector<bool>(positions.size(), false),
+                    std::vector<bool>(count, false),
                     {"Could not sample heights from tileset because it has not "
                      "been created."}});
 
-  std::move(future).thenImmediately(
-      [this, OnHeightsSampled = std::move(OnHeightsSampled)](
-          Cesium3DTilesSelection::SampleHeightResult&& result) {
+  std::move(future)
+      .catchImmediately([positions = std::move(positions)](
```
j9liu, line 158: Checking my understanding: will this only trigger if `pTileset` is valid? (and therefore `std::move(positions)` isn't done twice if `pTileset == nullptr` ?)

j9liu (commented):
Thanks @kring ! I confirmed that this makes the future complete, and that I'm able to print out the warning on the screen. I just had one question but if it's not an actual issue, this can be merged.

Source/CesiumRuntime/Private/Cesium3DTileset.cpp
kring: Hmm yeah it will be fine, though it is a little sloppy. As you said, the first std::move only happens if the _pTileset is nullptr. The second, capturing for the `catchImmediately`, happens in all cases, so technically we're moving `positions` twice. But the it's harmless because a `createResolvedFuture` will never result in a rejected future, and so that catchImmediately's continuation will never be called in the case of a null tileset.

I'll see if I can reorganize this a little to be less sloppy and to require less analysis to be sure it's safe.

PR #1535, opened 2024-10-29
Initialize fields in FCesiumSampleHeightResult.

kring:
Unreal is reporting that the two fields in this class are not initialized. This is probably harmless, but good (and easy) to fix.

As reported here:
https://community.cesium.com/t/fcesiumsampleheightresult-errors/35906

### DIFF
--- a/Source/CesiumRuntime/Public/CesiumSampleHeightResult.h
+++ b/Source/CesiumRuntime/Public/CesiumSampleHeightResult.h
@@ -19,7 +19,7 @@ struct CESIUMRUNTIME_API FCesiumSampleHeightResult {
    * is false.
    */
   UPROPERTY(BlueprintReadWrite, Category = "Cesium")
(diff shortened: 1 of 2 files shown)


Source/CesiumRuntime/Public/CesiumSampleHeightResult.h
```diff
@@ -28,5 +28,5 @@ struct CESIUMRUNTIME_API FCesiumSampleHeightResult {
    * will have more information about the problem.
    */
   UPROPERTY(BlueprintReadWrite, Category = "Cesium")
-  bool SampleSuccess;
+  bool SampleSuccess{false};
```
j9liu, line 31: This is the biggest nitpick of all time, but in other classes or structs we tend to use `=` to initialize the default values, so I would like to change this to `bool SampleSuccess = false` 😅 Hopefully it's not too bad if I do that before merging

PR #1537, opened 2024-10-30
Instance feature metadata

timoore:
Support for accessing metadata in GPU instances in picking hit results.


kring (commented):
@timoore I tried this out with the instanced version of everybody's favorite model, and it works well. In fact, the existing Metadata level in the Samples can now query instanced models without any changes, which is nice!

However, I also found that that commonality with non-instanced metadata pretty much ends there. As soon as you want to use any of the lower-level, more powerful metadata access Blueprint APIs, we're stick with completely different paths for instanced and non-instanced models.

Now there are certainly details here that I don't understand, but I'm reasonably sure that this doesn't need to be so. For example, `Get Feature ID From Hit` looks to me like a concept that works (and is identical) across all types of models. So why do we now have two functions to do this? (confusingly, they have the same name, too)

Well, the immediate reason is that one needs "Primitive Features" and the other needs "Instance Features". But why are _those_ separate? Both are just containers for sets of feature IDs that are resolved in different ways:
1. By face -> vertex -> feature ID attribute lookup
2. By texture coordinate -> feature ID texture lookup
3. And now! By instance ID -> feature ID lookup

In fact, it seems reasonable that a single "hit" could return feature IDs from all three. There could be metadata attached to the instance, the vertex, and the texture.

What do you think, is that achievable?

So that's my main feedback on this PR. I also have some additional small comments below, based on only a partial review of the code so far.

j9liu:
I haven't looked deeply at this PR, but with the way that `EXT_instance_features` exists on a `node` versus `EXT_mesh_features` existing on a `primitive`, I feel like the separation between the two makes sense.

I also don't see something that disallows `EXT_instance_features` from instancing a mesh with `EXT_mesh_features`. I feel like you would need to differentiate between the two. Maybe you have two separate building instances, and you'll want their different instance IDs. But what if you want to get the feature ID of the window within an instance? Then it becomes confusing whether the returned feature ID is from `EXT_instance_features` or `EXT_mesh_features`

I feel like it makes more sense if we use "Instance ID" for `EXT_instance_features`, if that's acceptable terminology. Or if not "Instance ID", then very carefully differentiating the feature IDs of the _instances_ versus the feature IDs in the _mesh_. Using "Instance ID" as shorthand makes most sense for me, and you could rename the `Get Feature ID From Hit` with `CesiumInstanceFeatures` to `Get Instance ID From Hit`. Does that seem reasonable?

j9liu:
Hm now that I think about it, "Instance ID" is not an accurate descriptor for cases where you're assigning multiple instances the same feature ID. So maybe we have to use terminology like "Instance Feature ID" vs. "Mesh Feature ID" ?

kring:
@j9liu I'm not as well versed in the metadata system as you are, so I very well could be missing something. But I don't understand why an "Instance Feature ID" is fundamentally different from our two other feature IDs we already have: the vertex attribute feature ID, and the texture feature ID. I guess they come from different extensions, but I don't think users know or care about that. They just want to be able to say: what are the metadata properties associated with the thing I just clicked on? And right now, if you want typed metadata (not just strings), that requires coding up (in Blueprint, say) two completely different paths to access the two types of feature IDs.
 
Obviously, if that's unavoidable, then we'll just have to live with it. But it doesn't look unavoidable. As I understand it, a feature ID - no matter which of the three sources provided it - is fundamentally just an index into a property table.

lilleyse:
> I agree that functionality wise, they're pretty much the same, and distinctions between them are going to be confusing. And I'm not sure if `EXT_instance_features` + `EXT_mesh_features` is going to come up frequently enough to be of concern. (@lilleyse do you have any opinion?) I agree with trying to make it as easy as possible to get metadata from a tileset.

`EXT_instance_features` + `EXT_mesh_features` is unlikely to show up much in practice. I think it would be fine to assume that `EXT_instance_features` takes precedence over `EXT_mesh_features` if they both exist.

kring:
> But if we were to combine Blueprints for EXT_instance_features and EXT_mesh_features, that index might be ambiguous.

Ok, I see. That's a good point. We actually already have this problem, though, as it's currently implemented. `UCesiumMetadataPickingBlueprintLibrary::GetPropertyTableValuesFromHit` takes a `FeatureIDSetIndex`. If there's instance metadata, that index is an index into `EXT_instance_features`. If there's not, then it's an index into `EXT_mesh_features`.

I think we either need to (1) accept this, and apply it everywhere, meaning that there's just one `Get Feature ID From Hit` and it resolves the index in the same way.

OR (2) we need to declare this unacceptable and have a separate "namespace" for these two sets of feature ID indices. And if we go this way, I think that `Get Feature ID From Hit` should take an enumeration specifying which set of feature IDs to prefer when both exist, rather than have two entirely separate (but confusingly similar-looking in Blueprint) versions of this function.

It's also possible to start with (1) today, and add (2) in a later version of Cesium for Unreal. But currently this branch has the two separate functions, and I don't think we should ship it that way.

Another approach (which doesn't have to replace any of the options already mentioned) is to make it easy to enumerate the feature ID sets, wherever they come from, and then query by reference to one. In other words, avoid the need for feature ID set indices completely.

Source/CesiumRuntime/Private/LoadGltfResult.h
```diff
@@ -152,18 +165,45 @@ struct LoadPrimitiveResult {
...
+    meshResult.swap(other.meshResult);
+  }
+
   std::optional<LoadMeshResult> meshResult = std::nullopt;
   /**
    * Array of instance transforms, if any.
    */
   std::vector<FTransform> InstanceTransforms;
+  /**
+   * Instance features
```
j9liu, line 204: This would benefit from being more descriptive

```suggestion
   * Parses EXT_instance_features from a node. A pointer is used for shared ownership, since there can be multiple meshes and mesh primitives on a node.
```

Source/CesiumRuntime/Public/CesiumFeatureIdAttribute.h
```diff
@@ -61,6 +62,21 @@ struct CESIUMRUNTIME_API FCesiumFeatureIdAttribute {
       const int64 FeatureIDAttribute,
       const FString& PropertyTableName);
 
+  /**
+   * @brief Constructs a feature ID attribute instance from instance metadata.
+   *
+   * @param Model The model.
+   * @param Node The "mesh node" containing the feature ID attribute.
```
j9liu, line 69:
```suggestion
   * @param Node The node containing the feature ID attribute.
```

Source/CesiumRuntime/Public/CesiumInstanceFeatures.h
```diff
@@ -0,0 +1,118 @@
...
+UCLASS()
+class CESIUMRUNTIME_API UCesiumInstanceFeaturesBlueprintLibrary
+    : public UBlueprintFunctionLibrary {
+  GENERATED_BODY()
+
+public:
+  /**
+   * Gets the instance features of a glTF primitive component. If component is
+   * not a Cesium glTF primitive component, the returned features are empty.
+   */
```
j9liu, line 57:
```suggestion
   * Gets the instance features of an instanced glTF primitive component. If component is
   * not a Cesium glTF primitive component, or not instanced, then the returned features are empty.
   */
```

Source/CesiumRuntime/Public/CesiumInstanceFeatures.h
```diff
@@ -0,0 +1,118 @@
...
+      BlueprintCallable,
+      BlueprintPure,
+      Category = "Cesium|Primitive|Features")
+  static const TArray<FCesiumFeatureIdSet> GetFeatureIDSetsOfType(
+      UPARAM(ref) const FCesiumInstanceFeatures& InstanceFeatures,
+      ECesiumFeatureIdSetType Type);
+
+  /**
+   * Gets the feature ID associated with an instance number
+   *
```
j9liu, line 90:
```suggestion
   * Gets the feature ID associated with the instance at the specified index.
```

Source/CesiumRuntime/Public/CesiumInstanceFeatures.h
```diff
@@ -0,0 +1,118 @@
...
+  UFUNCTION(
+      BlueprintCallable,
+      BlueprintPure,
+      Category = "Cesium|Primitive|Features")
+  static const FCesiumInstanceFeatures&
+  GetInstanceFeatures(const UPrimitiveComponent* component);
+
+  /**
+   * Gets all the feature ID sets that are associated with the
+   * primitive.
```
j9liu, line 67:
```suggestion
   * Gets all the instance feature ID sets that are associated with the
   * primitive.
```

Source/CesiumRuntime/Public/CesiumInstanceFeatures.h
```diff
@@ -0,0 +1,118 @@
...
+
+  /**
+   * Gets the feature ID from the given line trace hit, assuming it
+   * has hit a glTF primitive component containing this CesiumPrimitiveFeatures.
+   *
+   * A primitive may have multiple feature ID sets, so this allows a feature ID
+   * set to be specified by index. This value should index into the array of
+   * CesiumFeatureIdSets in the CesiumPrimitiveFeatures. If the specified
+   * feature ID set index is invalid, this returns -1.
+   */
```
j9liu, line 109:
```suggestion
   * Gets the feature ID from the given line trace hit, assuming it
   * has hit a glTF primitive component containing this CesiumInstanceFeatures.
   *
   * A primitive may have multiple feature ID sets, so this allows a feature ID
   * set to be specified by index. This value should index into the array of
   * CesiumFeatureIdSets in the CesiumInstanceFeatures. If the specified
   * feature ID set index is invalid, this returns -1.
   */
```

Source/CesiumRuntime/Private/CesiumMetadataPickingBlueprintLibrary.cpp
```diff
@@ -148,10 +150,69 @@ bool UCesiumMetadataPickingBlueprintLibrary::FindUVFromHit(
...
+  const auto* pInstancedComponent =
+      Cast<UCesiumGltfInstancedComponent>(Hit.Component);
+  if (!IsValid(pInstancedComponent)) {
+    return std::nullopt;
+  }
+  const TSharedPtr<FCesiumInstanceFeatures> pInstanceFeatures =
+      pInstancedComponent->pInstanceFeatures;
+  if (!pInstanceFeatures) {
+    return TMap<FString, FCesiumMetadataValue>();
+  }
```
j9liu, line 165: See @kring 's comment of course, but if we're keeping these lines then these two `if` statements need to return the same thing. The point is that getting the instance metadata values fails because ultimately the component doesn't contain (proper) instance features. So the behavior should be consistent either way.

j9liu (changes requested):
@timoore I did a code-only pass for code design, clarity, and quality. We're on Thanksgiving break this week, so if this is supposed to be in by next week's release I ask that you address feedback ASAP so this can make it in!

Also please address @kring 's comments too, I'm seeing that some of them have not been resolved. If you have any questions or want to discuss something please don't hesitate to ask.

Source/CesiumRuntime/Public/CesiumInstanceFeatures.h
```diff
@@ -0,0 +1,118 @@
...
+  GENERATED_USTRUCT_BODY()
+
+public:
+  /**
+   * Constructs an empty primitive features instance.
+   */
+  FCesiumInstanceFeatures() : _instanceCount(0) {}
+
+  /**
+   * Constructs an instance feature.
```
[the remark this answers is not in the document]
timoore: I'm using "object" instead of "instance"; there's a few too many instances in play here!

Source/CesiumRuntime/Private/LoadGltfResult.h
```diff
@@ -34,10 +35,61 @@ namespace LoadGltfResult {
...
+ * should happen.
  */
 struct LoadPrimitiveResult {
 #pragma region Temporary render data
 
+  LoadPrimitiveResult(const LoadPrimitiveResult&) = delete;
+
+  LoadPrimitiveResult() {}
+
+  LoadPrimitiveResult(LoadPrimitiveResult&& other)
```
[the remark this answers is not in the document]
kring: The move constructor will be created by default in some situations. The rules are a little complicated, but I think roughly as long as you don't manually define a copy constructor or a destructor, you'll get a move constructor by default.

Source/CesiumRuntime/Private/CesiumFeatureIdSet.cpp
```diff
@@ -78,6 +119,16 @@ UCesiumFeatureIdSetBlueprintLibrary::GetAsFeatureIDAttribute(
   return EmptyFeatureIDAttribute;
 }
 
+const FCesiumFeatureIdAttribute&
+UCesiumFeatureIdSetBlueprintLibrary::GetAsFeatureIDInstanceAttribute(
```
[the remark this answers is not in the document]
timoore: I was aware that this seemed identical to `GetAsFeatureIDAttribute`, but that function returns an empty attribute if the type is not `Attribute`.  `GetAsFeatureIDInstanceAttribute` has the same behavior for instance attributes. Is it important to maintain that distinction when getting a `FCesiumFeatureIdAttribute`?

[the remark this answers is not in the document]
j9liu: I'm not aware of the reason why a `FCesiumFeatureIdAttribute` should behave differently between the instanced and non-instanced extensions. At the end of the day, it is just a buffer that someone could use to extract values from. Hence my suggestion to do option 2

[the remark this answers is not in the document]
j9liu: Plus you can always expand `GetAsFeatureIDAttribute` to return the attribute if the type is `Attribute || InstanceAttribute`. Just note it in the changelog

Source/CesiumRuntime/Public/CesiumInstanceFeatures.h
timoore: Function is deleted.

Source/CesiumRuntime/Private/CesiumFeatureIdSet.cpp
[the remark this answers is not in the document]
timoore: I don't think I understand all the implications of conflating feature IDs from attributes and feature IDs that come from the gpu instancing extension. I'd be happy to take a look in a separate cleanup issue, but I think it's out of scope for this PR.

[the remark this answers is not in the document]
j9liu: I disagree that this is out of scope for this PR. To me, it doesn't make much sense to add a redundant function that would be immediately removed in the next release.

I personally don't think it's a major lift to implement a rename / deprecation, though I acknowledge that it's probably due to having done it before. I'm happy to give pointers to the way we've handled deprecations if it's confusing.

Source/CesiumRuntime/Private/CesiumMetadataPickingBlueprintLibrary.cpp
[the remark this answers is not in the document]
timoore: My thinking was that this would be a cheap test-and-out if the hit result was not on an instance.

timoore: This was not clear and at least deserved a comment. The intent was to return std::nullopt if the primitive was not an instanced static mesh and therefore cause GetPropertyTableValuesFromHit() to use feature IDs from a primitive attribute. I was thinking that if the primitive was instanced and didn't have instance feature IDs, then it couldn't have feature IDs at all. But given our discussion above, that's not clearly the case.

timoore:
I've removed the `FCesiumInstanceFeatures` collection type, which I think resolves the complaints about redundancy and makes this PR a lot more compact. I've left `UCesiumFeatureIdSetBlueprintLibrary::GetAsFeatureIDInstanceAttribute()`; I guess I don't understand how it's cousin `GetAsFeatureIDAttribute()` is used today in blueprint functions and whether or not it's important for that function  to only return attributes from primitives.

Source/CesiumRuntime/Private/CesiumMetadataPickingBlueprintLibrary.cpp
timoore: I added a comment about the different return values and I hope it's clear now. The return of an empty `TMap` indicates a problem with the instance metadata; in that case I think it's appropriate to not continue searching for other metadata.

Source/CesiumRuntime/Private/CesiumMetadataPickingBlueprintLibrary.cpp
```diff
@@ -148,23 +150,86 @@ bool UCesiumMetadataPickingBlueprintLibrary::FindUVFromHit(
...
 TMap<FString, FCesiumMetadataValue>
 UCesiumMetadataPickingBlueprintLibrary::GetPropertyTableValuesFromHit(
     const FHitResult& Hit,
     int64 FeatureIDSetIndex) {
-  const UCesiumGltfPrimitiveComponent* pGltfComponent =
-      Cast<UCesiumGltfPrimitiveComponent>(Hit.Component);
-  if (!IsValid(pGltfComponent)) {
+  const UCesiumGltfComponent* pModel = nullptr;
+  if (const auto* pPrimComponent = Cast<UPrimitiveComponent>(Hit.Component);
+      !IsValid(pPrimComponent)) {
```
kring, line 214: Wow, somehow I missed until now that C++17 added optional initializers to if statements. 🤯 Good to know!

Source/CesiumRuntime/Private/CesiumFeatureIdSet.cpp
```diff
@@ -62,6 +63,42 @@ FCesiumFeatureIdSet::FCesiumFeatureIdSet(
...
+      _featureCount(InstanceFeatureID.featureCount),
+      _nullFeatureID(InstanceFeatureID.nullFeatureId.value_or(-1)),
+      _propertyTableIndex(InstanceFeatureID.propertyTable.value_or(-1)),
+      _label(FString(InstanceFeatureID.label.value_or("").c_str())) {
+  FString propertyTableName;
+
+  // For backwards compatibility with GetFeatureTableName.
+  const CesiumGltf::ExtensionModelExtStructuralMetadata* pMetadata =
+      InModel.getExtension<CesiumGltf::ExtensionModelExtStructuralMetadata>();
+  if (pMetadata && _propertyTableIndex >= 0) {
```
kring, line 81:
```suggestion
  if (pMetadata && this->_propertyTableIndex >= 0) {
```

Source/CesiumRuntime/Private/CesiumFeatureIdSet.cpp
```diff
@@ -136,9 +174,48 @@ int64 UCesiumFeatureIdSetBlueprintLibrary::GetFeatureIDForVertex(
   return -1;
 }
 
+namespace {
+struct FeatureIdFromAccessorView : public CesiumGltf::FeatureIdFromAccessor {
```
kring, line 178: This looks to be unused?

kring (commented):
Thanks for making all the changes @timoore! I have a few comments below, but just minor stuff because this looks really good to me overall. It would probably be good to get @j9liu's eyes on it too, though!

Source/CesiumRuntime/Private/LoadGltfResult.h
```diff
@@ -152,18 +159,47 @@ struct LoadPrimitiveResult {
...
+  LoadMeshResult(const LoadMeshResult&) = delete;
+
+  LoadMeshResult(LoadMeshResult&& other) {
+    primitiveResults.swap(other.primitiveResults);
+  }
+
+  LoadMeshResult& operator=(LoadMeshResult&& other) {
+    primitiveResults.swap(other.primitiveResults);
+    return *this;
+  }
```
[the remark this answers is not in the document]
timoore: I think you're correct. I got kind of fixated on the idea that the std::vector move constructor doesn't necessarily swap, but we can probably trust it to do the right thing.

Source/CesiumRuntime/Private/CesiumFeatureIdSet.cpp
timoore: Must be from some earlier hack...

kring:
Thanks @timoore, this looks good to me! @j9liu I'll let you hit merge since you reviewed it previously and also have a deeper understanding of the metadata system than I do.

kring:
"Hit merge" and also add more comments if necessary, of course!

j9liu:
@kring For sure, I'll take a look later today / tomorrow!

Source/CesiumRuntime/Public/CesiumPrimitiveFeatures.h
```diff
@@ -132,6 +150,19 @@ class CESIUMRUNTIME_API UCesiumPrimitiveFeaturesBlueprintLibrary
       int64 FaceIndex,
       int64 FeatureIDSetIndex = 0);
 
+  /**
+   * Gets the feature ID associated with an instance at an index.
+   *
```
j9liu, line 155: I'll push a commit to remove this extra line here.

Source/CesiumRuntime/Private/CesiumMetadataPickingBlueprintLibrary.cpp
```diff
@@ -148,23 +150,86 @@ bool UCesiumMetadataPickingBlueprintLibrary::FindUVFromHit(
   return true;
 }
 
+namespace {
+/*
+ * Returns std:nullopt if the component isnt an instanced static mesh or doesn't
```
j9liu, line 155:
```suggestion
 * Returns std:nullopt if the component isn't an instanced static mesh or doesn't
```

Source/CesiumRuntime/Private/CesiumMetadataPickingBlueprintLibrary.cpp
```diff
@@ -148,23 +150,86 @@ bool UCesiumMetadataPickingBlueprintLibrary::FindUVFromHit(
...
+getInstancePropertyTableValues(
+    const FHitResult& Hit,
+    const UCesiumGltfComponent* pModel,
+    int64 FeatureIDSetIndex) {
+  const auto* pInstancedComponent =
+      Cast<UCesiumGltfInstancedComponent>(Hit.Component);
+  if (!IsValid(pInstancedComponent)) {
+    return std::nullopt;
+  }
+  const TSharedPtr<FCesiumPrimitiveFeatures> pInstanceFeatures =
```
j9liu, line 170: Doesn't matter too much, but I think we're doing this in other parts of the code, so do it here for consistency.

```suggestion
  const TSharedPtr<FCesiumPrimitiveFeatures>& pInstanceFeatures =
```

Source/CesiumRuntime/Private/CesiumMetadataPickingBlueprintLibrary.cpp
```diff
@@ -148,23 +150,86 @@ bool UCesiumMetadataPickingBlueprintLibrary::FindUVFromHit(
...
-  if (!IsValid(pGltfComponent)) {
+  const UCesiumGltfComponent* pModel = nullptr;
+  if (const auto* pPrimComponent = Cast<UPrimitiveComponent>(Hit.Component);
+      !IsValid(pPrimComponent)) {
     return TMap<FString, FCesiumMetadataValue>();
+  } else {
+    pModel = Cast<UCesiumGltfComponent>(pPrimComponent->GetOuter());
+    if (!IsValid(pModel)) {
+      return TMap<FString, FCesiumMetadataValue>();
+    }
```
j9liu, line 220: Was this moved inside the `else` statement for any particular reason? Personally I avoid nested `if`s and `else`s, and this seems functionally the same if the `if` stays outside.

j9liu (commented):
I just have a few tweaks, but I'm going to push a commit myself to speed things up.

I'm not 100% happy about `CesiumPrimitiveFeatures` representing both primitive and instance IDs. Not because we need redundancy with `CesiumInstanceFeatures`, but I think the name `CesiumPrimitiveFeatures` is now inaccurate.

On the one hand, the node with `EXT_mesh_gpu_instancing` gets converted into an instanced mesh primitive under the hood, so it is technically still a primitive. On the other hand, I named `CesiumPrimitiveFeatures` after the fact that the `EXT_mesh_features` extension is on the glTF primitive. This scheme carries over to `EXT_structural_metadata`, where we have `CesiumModelMetadata` and `CesiumPrimitiveMetadata` to denote the difference.

But not sure what a happy replacement would be. Maybe `CesiumGltfFeatures`? Not saying this needs to be resolved -- it's out of scope for this PR. Maybe I'm just abnormally nitpicky about names :shrug:

I still notice something that should be addressed, but I will make my own follow-up PR to fix it.

Source/CesiumRuntime/Private/CesiumMetadataPickingBlueprintLibrary.cpp
j9liu: I moved this outside. But let me know if there's any reason why not

j9liu:
Merging after the Apple builds. I haven't played around with this in action but I assume @kring tested this enough when he reviewed 😛

PR #1543, opened 2024-11-11
Update build to work with changes in CesiumGS/cesium-native#962

kring:


kring:
I'm merging this so that cesium-unreal main continues to build with cesium-native main.

Issue #1544, opened 2024-11-11
Allow polygons in level instances to be used on persistent level's polygon raster overlay component


KieranCoppins:
# Request
I want to be able to select a cartographic polygon that exists inside a level instance on my `Cesium3DTileset` actor's `CesiumPolygonRasterOverlay` component.

# Use Case
I have a large building that is a part of a Cesium sub-level. As such, I need to use a cartographic polygon to cut out the building location in the 3D tile set. When I zoom outside of the sub-levels load radius, I am left with a large hole if the cartographic polygon is part of the persistent level. If the polygon is part of the sub-level then it can be loaded and unloaded dynamically with the polygon being added and removed.

Another use case is that I am using chunking in Unreal Engine to separate my buildings into separate chunks. If I do not include a chunk in my file, the same occurs as above; the building is gone, but the hole remains. By making the polygon part of the sub-level (and a part of the chunk) then the hole will not be there if the level is not loaded.

*I have a PR that will allow for this that I will submit shortly*

kring:
Fixed in #1545.

PR #1545, opened 2024-11-11
Use a soft object reference for cartographic polygons

KieranCoppins:
Use a soft object pointer to store references to cartographic polygons in other level instances (#1544 ).


kring:
Thanks for the PR @KieranCoppins! We'll review it as soon as we can.

kring:
Thanks @KieranCoppins!

PR #1552, opened 2024-11-19
Update for schema changes in cesium-native

kring:
Fixes the build after CesiumGS/cesium-native#969.

~Requires CesiumGS/cesium-native#999.~
Requires CesiumGS/cesium-native#1002

PR #1553, opened 2024-11-19
Fix `SampleHeightMostDetailed` crash on `Cesium3DTileset` with raster overlay

### LINKED ISSUE #1549: SampleHeightMostDetailed crashes when called on a tileset with a raster overlay before its first Tick
Reported here:
https://community.cesium.com/t/sampleheightmostdetailed-causes-crash-when-called-in-beginplay/36439

When a user calls `SampleHeightMostDetailed` on a `Cesium3DTileset` that does not yet have any underlying cesium-native `Tileset`, the system automatically calls `LoadTileset` in order to create it. But it doesn't also do the three "Resolve" operations that a normal Tick would do first:

https://github.com/CesiumGS/cesium-unreal/blob/v2.10.0/Source/CesiumRuntime/Private/Cesium3DTileset.cpp#L2124

```
this->ResolveGeoreference();
this->ResolveCameraManager();
this->ResolveCreditSystem();
```

This is fine for a tileset without any raster overlays. Sampling heights doesn't require any of these three systems. However, the raster overlay system assumes the presence of a CreditSystem. So if the tileset happens to have a raster overlay, then we'll crash.

We should:
1. Change `SampleHeightMostDetailed` to call these three Resolves before calling `LoadTileset`.
2. Fix the raster overlay crash when the credit system is missing.

Either of these changes would fix the immediate problem, but it's worth doing both.

For users experiencing this crash, an easy workaround is to manually call those three Resolve methods on the `Cesium3DTileset` before calling `SampleHeightMostDetailed`.

j9liu:
Fixes #1549. The cesium-native side isn't included yet since I didn't want to complicate things between PRs like #1552 and #1539.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,11 @@
 # Change Log
 
+### ? - 2024-12-02
+
+##### Fixes :wrench:
+
+- Fixed a crash that could occur when using `SampleHeightMostDetailed` on a `Cesium3DTileset` with a raster overlay.
+
 ### v2.10.0 - 2024-11-01
 
 ##### Additions :tada:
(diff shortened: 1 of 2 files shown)


kring:
Thanks @j9liu!

PR #1554, opened 2024-11-20
Fix raster overlay undefined behavior

### LINKED ISSUE #1551: CesiumPolygonRasterOverlay causes issues with multiple raster overlays
Reported on the forum here:
- https://community.cesium.com/t/cartographicpolygon-clipping-cause-terraintile-rendering-abnormal/36628
- https://community.cesium.com/t/cesiumworldterrain-with-multi-rasteroverlays-rendering-error/36636

I was able to reproduce this myself in UE 5.3, v2.10.0 by doing the following:

- In CesiumForUnrealSamples, go to 02_CesiumMelbourne.
- Add the WMS overlay from this [Sandcastle](https://sandcastle.cesium.com/?src=Web%20Map%20Service%20(WMS).html) to Cesium World Terrain, with key Overlay1.
- Zoom out and go to an area where the WMS is showing (over some of the blue dots). Rebase the georeference origin for convenience.
- Add a Cartographic Polygon over the area rendering the WMS. Then add it to a Polygon Raster Overlay on Cesium World Terrain.

You'll see some weird glitchy artifacts in the area affected by the polygon. When you remove the WMS, the artifacts won't disappear right away, but you can refresh and the clipping will appear as normal.

image

For curiosity's sake, I tried it the other way around, by adding the WMS after the polygon overlay. The polygon clipping works, but the WMS shows up as this blocky blue gradient over a lot of the tiles. 🤔 

image

I have not tested this, but this reportedly happens as far back as v2.8.0. The behavior works as expected in v2.6.0. It might be prudent to check if this happens with v2.7.0 -- that release introduces some CesiumPolygonRasterOverlay bugfixes, as well as custom ellipsoids.

kring:
This is a PR into #1552 so merge that first.
This depends on CesiumGS/cesium-native#1003 so merge that first, too.

Fixes #1551

### DIFF
--- a/Source/CesiumRuntime/Private/CesiumGltfComponent.cpp
+++ b/Source/CesiumRuntime/Private/CesiumGltfComponent.cpp
@@ -3387,6 +3387,10 @@ void UCesiumGltfComponent::AttachRasterTile(
                     EMaterialParameterAssociation::LayerParameter,
                     i),
                 translationAndScale);
+            check(
+                textureCoordinateID >= 0 &&
+                textureCoordinateID <
+                    primData.overlayTextureCoordinateIDToUVIndex.size());
             pMaterial->SetScalarParameterValueByInfo(
                 FMaterialParameterInfo(
                     "TextureCoordinateIndex",


kring:
Yeah, after the C++20 upgrade PRs were merged into cesium-native main, and main was merged into CesiumGS/cesium-native#1003, this became dependent on #1539. I merged that branch into here and changed the base branch accordingly.

PR #1561, opened 2024-11-29
Don't crash if an Image is missing its ImageAsset

kring:
Usually this happens because the `ImageAsset` failed to load, perhaps because the URL is invalid.

Also updates cesium-native to include CesiumGS/cesium-native#1016, so merge that first.


timoore (approved):

kring:
Thanks for the review @timoore!

PR #1570, opened 2024-12-13
Publish documentation to the web site

### LINKED ISSUE #1377: Publish Doxygen documentation online on release
While Cesium for Unreal currently has an API documentation that can be generated from the source using Doxygen, we don't seem to be publishing it anywhere currently. As noted in #689, this makes it hard to find out that documentation exists at all. If we published the docs, it would also help them show up on search engines, where lots of developers are going to look first. It seems to me like publishing the docs online is a pretty good win for developer experience compared to the amount of developer time required to upload the built docs after every release.

kring:
Fixes #1377 

This doesn't work yet due to S3 permissions.

### DIFF
--- a/.github/workflows/prod.yml
+++ b/.github/workflows/prod.yml
@@ -35,8 +35,8 @@ jobs:
       - name: Deploy to cesium.com
         if: ${{ github.ref_name == 'cesium.com' }}
         run: |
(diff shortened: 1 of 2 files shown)


kring:
This is working at the staging URL now, so I'm merging it!

PR #1573, opened 2024-12-20
Require UE 5.3+

kring:
This PR bumps all the versions, and removes some ifdefs that were needed to support UE 5.2.


azrogers:
If I can get my new laptop set up with Unreal by the end of the day, but no promises on that front 😅

j9liu:
That's fine @azrogers, I'll just merge so we can get this in earlier.

Thanks @kring !

PR #1574, opened 2024-12-20
Rename vertex-specific functions in `FCesiumFeatureIdAttribute`

j9liu:
With #1537 the `FCesiumFeatureIdAttribute` class gets used for both per-vertex and per-instance feature IDs. This makes the "vertex" verbiage in the API inaccurate.

This PR renames those functions so they are more generic. I added `Redirects` so that Blueprints can seamlessly pick up the new functions. If you'd like, you can test this by:

- Put the old nodes in the Blueprint of a level. Save, and then exit Unreal.
- Switch to this branch and open Unreal. Open the level Blueprint -- the nodes should be properly renamed.


Source/CesiumRuntime/Public/CesiumFeatureIdAttribute.h
```diff
@@ -131,27 +131,27 @@ class CESIUMRUNTIME_API UCesiumFeatureIdAttributeBlueprintLibrary
...
+   * attribute is invalid, this returns 0.
    */
   UFUNCTION(
       BlueprintCallable,
       BlueprintPure,
       Category = "Cesium|Features|FeatureIDAttribute")
   static int64
-  GetVertexCount(UPARAM(ref)
-                     const FCesiumFeatureIdAttribute& FeatureIDAttribute);
+  GetFeatureIDCount(UPARAM(ref)
```
j9liu, line 142: This might be a bit confusing on second glance -- I don't want this to convey the # of actual unique feature IDs versus. the length of the attribute. Let me tweak this

j9liu: Settled on `GetCount` with some additional documentation. Open to other alternatives too.

j9liu:
@kring do you mind giving this a review?

kring:
Looks good, thanks @j9liu!

PR #1576, opened 2024-12-23
Treat failed to load as unloaded sublevels

### LINKED ISSUE #1575: Cesium Sublevel doesn't load when unloaded when other sublevels are not in game files
# Description
In my project I have a globe level which contains sub levels of different venues around the world. Each sub level is packaged into its own [chunk](https://dev.epicgames.com/documentation/en-us/unreal-engine/preparing-assets-for-chunking-in-unreal-engine). When the application is deployed a chunks maybe removed that are not needed to help reduce file size. When a chunk is removed the other sublevels no longer load. If all chunks are present, the application works as expected.

# Reproduction Steps
- Have a project with a Globe level
- Create two sublevels (Sublevel A, Sublevel B)
- Associate Sublevel A with Chunk Id 1
- Associate Sublevel B with Chunk Id 2
- Package the project
- Remove Chunk 2 from the Content/Paks directory
- Launch the game
- Unload Sublevel A (if already loaded)
- Load Sublevel A
- Observe Sublevel A does not load

# Expected Behaviour
The sublevel should still load even if other sublevels are not available.

_I have a fix for this that I will raise a PR for soon._

KieranCoppins:
Fixes #1575 

If a sublevel fails to load, we should not block other sublevels from loading. We can treat it as being unloaded.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,10 @@
 
 - Removed support for Unreal Engine 5.2. Unreal Engine 5.3 or later is now required.
 
+##### Fixes :wrench:
+
+- Fixed a bug in `CesiumSubLevelSwitcherComponent` that could prevent all sub-levels from loading if a single sub-level failed to load.
+
 ### v2.11.0 - 2024-12-02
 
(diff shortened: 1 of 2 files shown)


kring:
Thanks @KieranCoppins!

PR #1579, opened 2025-01-06
Treat failed to load as unloaded sublevels Pt. 2

KieranCoppins:
I forgot an enum check in my previous PR #1576 

The bug still occurs unless if this is also done!


kring:
Thanks @KieranCoppins!

Issue #737, opened 2022-01-26
Allow setting the Authorization header for tilesets loaded from a URL


bmolynea-google:
Access to tilesets needs to be securely controlled for some applications. This would typically be done with OAuth.

Cesium Unreal should allow specifying the Authorization header for tileset requests made using the specified URL option. It must also be possible to update the header to allow refreshing the access token.

A public function may be added to `ACesium3DTileset` to set the Authorization header:
```
void SetAuthorizationHeader(FString InAuthorizationHeader);
```

For OAuth, the `InAuthorizationHeader` would look like:
```
Bearer <access token>
```

gitRuiZai:
I also get the same problem. How can I solve it?

lilleyse:
@azrogers can this be closed now that https://github.com/CesiumGS/cesium-unreal/pull/1581 is merged?

PR #1581, opened 2025-01-09
Add `RequestHeaders` property to `Cesium3DTileset`

azrogers:
To support use cases like #737, this change adds a `RequestHeaders` property to the `Cesium3DTileset`, which passes these headers through to the tileset loader for them to be attached to each request made. Depends on CesiumGS/cesium-native#1057.


Source/CesiumRuntime/Private/Cesium3DTileset.cpp
```diff
@@ -385,6 +385,31 @@ void ACesium3DTileset::SetTilesetSource(ETilesetSource InSource) {
...
 }
 
+bool CompareMaps(
+    const TMap<FString, FString>& Lhs,
+    const TMap<FString, FString>& Rhs) {
+  if (Lhs.Num() != Rhs.Num()) {
+    return false;
+  }
+
+  for (auto& [Key, Value] : Lhs) {
```
kring, line 395: I think this needs to be `const auto&` or else it will copy the key and value. I think.

Source/CesiumRuntime/Private/Cesium3DTileset.cpp
```diff
@@ -385,6 +385,31 @@ void ACesium3DTileset::SetTilesetSource(ETilesetSource InSource) {
...
+    }
+  }
+
+  return true;
+}
+
+void ACesium3DTileset::SetRequestHeaders(
+    const TMap<FString, FString>& InRequestHeaders) {
+  if (!CompareMaps(InRequestHeaders, this->RequestHeaders)) {
+    this->DestroyTileset();
```
kring, line 408: This handles the C++ and Blueprint cases, but when headers are changed on the Details panel, the tileset won't refresh. To fix that, add this new property to `PostEditChangeProperty`.

Source/CesiumRuntime/Private/Cesium3DTileset.cpp
```diff
@@ -1287,6 +1312,14 @@ void ACesium3DTileset::LoadTileset() {
 
   options.contentOptions.applyTextureTransform = false;
 
+  options.requestHeaders.reserve(this->RequestHeaders.Num());
+
+  for (auto& [Key, Value] : this->RequestHeaders) {
```
kring, line 1317: `const auto&` is better here too.

Source/CesiumRuntime/Private/Cesium3DTileset.cpp
```diff
@@ -1287,6 +1312,14 @@ void ACesium3DTileset::LoadTileset() {
...
+    options.requestHeaders.emplace_back(CesiumAsync::IAssetAccessor::THeader{
+        std::string(TCHAR_TO_UTF8(*Key)),
+        std::string(TCHAR_TO_UTF8(*Value))});
```
kring, line 1320: I don't think these explicit `std::string`s are needed.

kring (commented):
Works great, just a few nitpicky comments.

Source/CesiumRuntime/Private/Cesium3DTileset.cpp
```diff
@@ -385,6 +385,31 @@ void ACesium3DTileset::SetTilesetSource(ETilesetSource InSource) {
   }
 }
 
+bool CompareMaps(
```
kring, line 388: This should go in the anonymous namespace.

kring: Also something like `MapsAreEqual` would be a better name than `CompareMaps`, because the latter doesn't indicate as clearly what it means when the function returns true or false.

kring (commented):
Just a tiny comment. I'm going to make this change myself to save a cycle and get this merged.

kring:
Thanks @azrogers!

PR #1583, opened 2025-01-10
Update `testIonToken` to latest Unreal Samples token

azrogers:
Updating `SceneGenerationContext::testIonToken` to the Cesium for Unreal Samples v2.12.0 token to fix broken tests.


kring:
Thanks @azrogers!

PR #1589, opened 2025-01-21
Replace `uriparser` usage with `CesiumUtility::Uri`

azrogers:
With `uriparser` getting replaced with `ada` in CesiumGS/cesium-native#1072, we need to remove the usage of `uriparser` in Cesium for Unreal so we can remove it from the dependencies entirely. This change utilizes the `CesiumUtility::Uri::uriPathToNativePath` and `CesiumUtility::Uri::getPath` methods to resolve file URIs. As part of it, I needed to wrap includes of `CesiumUtility/Uri.h` in `THIRD_PARTY_INCLUDES_START` and `_END` macros as `ada` references multiple macros in `#if` directives, such as `ADA_USING_LIBRARY`, that aren't necessarily defined. This causes errors on MSVC. The `THIRD_PARTY_INCLUDES_` macros ignore errors in the included header files.


kring:
Thanks @azrogers!

PR #1590, opened 2025-01-22
Add `UpdateTimeZoneFromLongitude` to `ACesiumSunSky`

### LINKED ISSUE #1258: Compute CesiumSunSky time zone from longitude
The CesiumSunSky automatically gets its longitude and latitude from the CesiumGeoreference origin, ensuring that the sun position is accurate in the new Unreal coordinate system. However, we currently don't set the time zone. So a common user experience, for example in the [Adding Datasets](https://cesium.com/learn/unreal/unreal-datasets/) tutorial, looks like this:

1. CesiumGeoreference is somewhere, such as New York. The Time Zone property on the CesiumSunSky is set to -5, which is appropriate for New York. The Solar Time is set to 13, so it's about 1pm in New York, and the sun position reflects that.
2. The user wants to build their level in Melbourne. So maybe they add the Melbourne Photogrammetry tileset and double-click it. The tileset is visible, but upside-down because the CesiumGeoreference origin is still back in New York.
3. The user clicks the "Place Georeference Origin Here" button to move the origin near Melbourne.
4. Suddenly the scene goes completely dark. Time Zone -5 plus Solar Time 13 makes about 8AM UTC, which is about 7pm in Melbourne. It's dark there, and the CesiumSunSky accurately reflects that.

The weird thing here is that the user is required to manual enter the time zone. We inherited that from Unreal's SunSky, but there's no good reason for it when we know the longitude. Time zone isn't purely a function of longitude, of course (it's crazy complicated), but when paired with a Solar Time, longitude is really all we need. Solar Time and local time aren't the same thing anyway.

So, I think the user experience could be greatly improved by making CesiumSunSky automatically compute the time zone as a function of the longitude of the georeference origin. That way, after setting the new origin to Melbourne in the scenario above, it would be 1PM in Melbourne (roughly) rather than 1PM in New York.

azrogers:
Closes #1258. This PR adds a new method to `ACesiumSunSky` that updates the `TimeZone` property based on the given longitude. This is calculation is based on the simple formula of `LongitudeInDegrees / 15.0`, but this more or less approximates real time zones and when paired with solar time accomplishes the goal of "the sun should be at this position in the sky relative to the current location." `UpdateTimeZoneFromLongitude` can be called at runtime, but this change also makes it so that clicking "Place Georeference Origin Here" will automatically update the time zone based on the new longitude, reducing friction for new users.


CHANGES.md
```diff
@@ -5,6 +5,8 @@
 ##### Additions :tada:
 
 - Added `CesiumUrlTemplateRasterOverlay`, allowing a raster overlay to be added using tiles requested based on a specified URL template.
+- Added `UpdateTimeZoneFromLongitude` method to `ACesiumSunSky` to naively update the `ACesiumSunSky`'s `TimeZone` property based on a given longitude.
```
j9liu, line 8:
```suggestion
- Added `AdjustTimeZoneForLongitude` to `ACesiumSunSky` to set a reasonable `TimeZone` value at the given longitude.
```

CHANGES.md
```diff
@@ -5,6 +5,8 @@
...
+- The "Place Georeference Origin Here" button on `ACesiumGeoreference` will now update the time zone of any `ACesiumSunSky` instances using that georeference based on the new origin's longitude.
```
j9liu, line 9:
```suggestion
- The "Place Georeference Origin Here" button on `ACesiumGeoreference` will now adjust the time zone of the `ACesiumSunSky`(s) that reference it, based on the new origin's longitude. This improves user experience when moving the origin to locations where it is currently nighttime.
```

j9liu (changes requested):
Thanks @azrogers ! I'll test this locally in a bit, just wanted to leave these comments earlier than later.

CHANGES.md
azrogers: It's "when moving the origin to locations where it would be nighttime in the current time zone," right?

Source/CesiumRuntime/Public/CesiumSunSky.h
```diff
@@ -444,6 +444,21 @@ class CESIUMRUNTIME_API ACesiumSunSky : public AActor {
...
+   * every 15 degrees longitude equals 1 hour. This won't necessarily line up
+   * with the officially designated time zone at the location given.
+   *
+   * This method will call @ref UpdateSun automatically.
+   *
+   * @param InLongitude The longitude that the calculated time zone will be
+   * based on in degrees in the range [-180, 180].
+   */
+  UFUNCTION(CallInEditor, BlueprintCallable, Category = "Cesium")
+  void UpdateTimeZoneFromLongitude(double InLongitude);
```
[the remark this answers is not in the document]
azrogers: "Adjust" and "Update" seem like synonyms to me. What about "EstimateTimeZoneForLongitude" to be absolutely clear?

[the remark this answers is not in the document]
j9liu: I like `EstimateTimeZoneForLongitude`!

CHANGES.md
j9liu: Yeah that's better wording for what I was trying to say. :')

azrogers:
@j9liu Incorporated those edits!

kring:
Thanks @azrogers!

PR #1591, opened 2025-01-23
Remove decoy checkbox on "Enable Water Mask"

### LINKED ISSUE #1538: Cesium3DTileset's "Enable Water Mask" option has a decoy checkbox
image

I couldn't figure out why disabling the water mask wasn't working. Until I realized I was unchecking the first box, which is... well, I have no idea what it is. But I know now that it doesn't actually disable the feature. To disable that water mask, you have to keep that one checked, and uncheck the rightmost one.

We should figure out where this dodgy decoy checkbox is coming from and get rid of it.

azrogers:
Fixes #1538. There's currently a decoy checkbox on "Enable Water Mask" resulting from the `EditCondition` flag on the UPROPERTY. I believe this checkbox is actually controlling the `bIsMac` boolean that the `EditCondition` flag points to, but not only do we not want this to be a choice the user can make - the system doesn't suddenly become a Mac when the user unchecks the box! - but it also causes confusion between it and the actual "Enable Water Mask" checkbox that truly enables or disables the feature. This change uses `CanEditChange` to handle the `EditCondition` instead, removing the extra checkbox and also as a side effect removing the need for a `bIsMac` editor field.


j9liu:
@kring do you mind checking this one since it's Mac related? 😛

kring:
The checkbox is still correctly disabled on the Mac:
image

Also, if I remove all the Mac checks, then the water effect actually works now! This is UE 5.3 on an M4 Max MacBook Pro. I'll write a separate issue for us to consider re-enabling it.
image

kring:
Thanks @azrogers!

PR #1594, opened 2025-01-24
Fix build with GLM changes in Cesium Native

azrogers:
At the moment, `GLM_FORCE_SIZE_T_LENGTH` is still defined in the Unreal Build.cs, which - after we removed this define from Cesium Native - means linker errors result when trying to build Cesium for Unreal with the latest changes from Cesium Native. We can fix this by simply removing the define.

Also, I've been receiving errors related to using an experimental header without defining `GLM_ENABLE_EXPERIMENTAL`, despite the only place an experimental header is used in this codebase coming after a `#define GLM_ENABLE_EXPERIMENTAL` declaration. I don't know what's causing this - something to do with Unreal's unity build? - but it can be fixed by defining `GLM_ENABLE_EXPERIMENTAL` in the Build.cs.


kring:
Thanks @azrogers!

PR #1595, opened 2025-01-27
Fix crash in packaged build when tilesets are only created at runtime

### LINKED ISSUE #1507: Runtime crash related to CesiumIonServer in a packaged game when tilesets are _only_ created at runtime
Reported here:
https://community.cesium.com/t/fatal-error-while-spawning-cesium3dtileset/34641

When a Cesium3DTileset is added via the Editor, it will have a reference to a `CesiumIonServer` instance (usually the default SaaS one) and so Unreal will know to automatically include this instance in a packaged game.

However, if your level does _not_ have a Cesium3DTileset, but then you add one at runtime, Unreal won't know that the `CesiumIonServer` instance is required, and so it won't include it in the packaged game. As a result, the reference will be nullptr and the game will crash with an access violation.

This can be worked around by explicitly telling Unreal to include the CesiumIonServer. But it manifests as a mysterious crash to users, which is not great. We should address this by doing one or more of the following:

1. Tell the user at runtime that this is a problem and fail gracefully instead of crashing with an access violation.
2. Handle the possibility that there is no CesiumIonServer at runtime, and create one on the fly (like we already do in the Editor).
3. Always force the CesiumIonServer to be included in packaged games even if it doesn't appear to be used.

We should also consider if there are other assets that will be missing in this scenario.

azrogers:
Fixes #1507. At runtime, when not in the editor, `UCesiumIonServer::GetDefaultServer` relies on a `UCesiumIonServer` asset being included in the packaged game. This can cause a crash if tilesets are only being created at runtime, as Unreal doesn't understand that the `UCesiumIonServer` is required and won't include it in the build. This fix creates a `UCesiumIonServer` at runtime, like we do in the editor, to make sure the game doesn't crash. The one downside of this approach is that the user may have used a `UCesiumIonServer` with settings other than the default, but the one created at runtime will have default settings. In this case, however, they can use the workaround documented in #1507 where they can manually force Unreal to include their asset in the build.


j9liu:
Thanks @azrogers ! I'm weirdly struggling to test this because of the existing server in our samples project, but the code seems fair. Just waiting on CI to merge.

PR #1596, opened 2025-01-27
Check for null tileset in SampleHeightMostDetailed

### LINKED ISSUE #1562: SampleHeightMostDetailed Blueprint function crashes if called without a valid tileset
If nothing is connected to the `Tileset` pin on this node, and we run the Blueprint, it will crash the application. We just need a null check.

azrogers:
Fixes #1562. The `SampleHeightMostDetailed` blueprint function will currently crash if called without a valid `ACesium3DTileset` parameter. This PR adds a null check that returns an empty set of results and a warning if called with an invalid tileset.


kring:
Thanks @azrogers!

PR #1599, opened 2025-01-29
Fix matrix -> FTransform conversion for matrices with small scales

### LINKED ISSUE #1598: Models that are scaled down by a factor of 0.001 or smaller will not render in Unreal
Reported here:
https://community.cesium.com/t/i-cant-load-tileset-from-url-into-ue5/37461/3

A glTF model with a scale factor of about 0.001 or less (for example, where the vertex coordinates are expressed in millimeters), will not render in Cesium for Unreal. You will see errors like this:

```
LogPhysics: Warning: Scale for /Temp/Untitled_1.Untitled:PersistentLevel.Cesium3DTileset_1.CesiumGltfComponent_5.CesiumGltfPrimitiveComponent_0 has a component set to zero, which will result in a bad body instance. Scale:X=0.000 Y=0.000 Z=0.000
[[2025.01.28-23.37.43:506][ 33]]LogPhysics: Warning: Initialising Body : Scale3D is (nearly) zero: Actor: 'Cesium3DTileset' Component: '/Temp/Untitled_1.Untitled:PersistentLevel.Cesium3DTileset_1.CesiumGltfComponent_5.CesiumGltfPrimitiveComponent_0' 
```

The problem is that such a scale factor triggers a bug (IMO) in Unreal's FMatrix -> FTransform conversion where an overly-eager epsilon check considers the scale to be exactly 0.0 rather than its real value of about 1e-4.

The best solution I can think of is to implement our own FMatrix -> FTransform conversion (or more likely glm::mat4 -> FTransform) that doesn't have a dodgy epsilon.

kring:
Fixes #1598

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,6 +9,7 @@
 ##### Fixes :wrench:
 
 - Fixed another bug in `CesiumSubLevelSwitcherComponent` that could prevent all sub-levels from loading if a single sub-level failed to load.
+- Worked around a limitation in Unreal's `FMatrix` -> `FTransform` conversion that caused models using a small scale factor (such as where vertex positions are expressed in millimeters) to fail to render because their scale was treated as 0.0.
 
 ### v2.12.0 - 2025-01-02
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -12,7 +12,7 @@
 ##### Fixes :wrench:
 
 - Fixed another bug in `CesiumSubLevelSwitcherComponent` that could prevent all sub-levels from loading if a single sub-level failed to load.
-- Worked around a limitation in Unreal's `FMatrix` -> `FTransform` conversion that caused models using a small scale factor (such as where vertex positions are expressed in millimeters) to fail to render because their scale was treated as 0.0.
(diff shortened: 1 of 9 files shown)


CHANGES.md
```diff
@@ -12,6 +12,7 @@
 ##### Fixes :wrench:
 
 - Fixed another bug in `CesiumSubLevelSwitcherComponent` that could prevent all sub-levels from loading if a single sub-level failed to load.
+- Worked around a limitation in Unreal's `FMatrix` -> `FTransform` conversion that caused models using a small scale factor (such as where vertex positions are expressed in millimeters) to fail to render because their scale was treated as 0.0.
```
j9liu, line 15: Just some slight rewording:

```suggestion
- Worked around a limitation in Unreal's `FMatrix` -> `FTransform` conversion that prevented models with a small scale factor (e.g., where vertex positions are expressed in millimeters) from rendering because their scale was treated as 0.0.
```

j9liu:
Thanks @kring ! This looks good to me, I'll just do that one changelog edit and merge.

PR #1601, opened 2025-01-31
Add documentation for 3D Tiles integration into Cesium for Unreal

### LINKED ISSUE #1542: Create architecture diagram illustrating how 3D Tiles are streamed from Cesium Native
An extension of https://github.com/CesiumGS/cesium-native/issues/982, but with more focus on the Cesium for Unreal side.

It'd be great to show the integration points that we have with Cesium Native. How have we implemented `IAssetAccessor` and `IPrepareRendererResources`, how does `ACesium3DTileset` interface with cesium-native's `Tileset`, etc. Basically showing the data flow between Unreal and Native, and an overview of any additional work we do on the Unreal side to process that data (e.g. in `CesiumGltfComponent`).

kring:
Fixes #1542 

It basically explains how the Cesium Native integration points for 3D Tiles rendering are implemented in Unreal Engine.

As always, there's probably a lot more that can be said here, but I think this hits the most important points.

This PR also has some light refactoring, because sometimes it's easier to fix weird things than to describe them:

* Moved `UnrealResourcePreparer` out of `Cesium3DTileset.cpp` and into its own source files. Also renamed it to `UnrealPrepareRendererResources` to match the interface name.
* Renamed `LoadModelResult` and other similar types to `LoadedModelResult`. This is to match `LoadedTextureResult`, which is the only one among this family that is public.
* Separated `FCesiumUseExistingTextureResource` into two classes, the other called `FCesiumPreCreatedRHITextureResource`. Previously this class had two constructors that were used in two entirely different ways, for no really good reason.


Documentation/unreal-engine-3d-tiles.md
```diff
@@ -0,0 +1,77 @@
...
+
+In `prepareInLoadThread`, Cesium for Unreal receives a [Model](\ref CesiumGltf::Model) from Cesium Native and creates from it a `LoadedModelResult`. This struct holds a representation of the model that is as close to "fully renderable in Unreal Engine" as we can manage without creating any `UObjects`:
+
+* Normals and tangents are generated, if required.
+* Texture MipMaps are generated, if required.
+* Physics meshes are generated, if required.
+* Mesh vertex and index data for each `MeshPrimitive` are copied into an instance of Unreal's `FStaticMeshRenderData`.
+* An `FCesiumTextureResource` is created for each texture. This class is derived from Unreal's `FTextureResource` and is Unreal's low-level, render-thread representation of a texture.
+* Feature IDs and metadata that are made available to a material via the [UCesiumEncodedMetadataComponent](\ref UCesiumEncodedMetadataComponent) are turned into additional textures.
```
azrogers, line 49: This link fails to resolve because it's not named `UCesiumEncodedMetadataComponent` anymore - it's `UDEPRECATED_CesiumEncodedMetadataComponent`. I believe `UCesiumFeaturesMetadataComponent` is what we're using now instead, but @j9liu would know better than me.

azrogers:
Besides the one comment, it's looking good. I made a couple of copy edits but nothing major. Feel free to merge this once that comment is resolved!

kring:
Thanks @azrogers! It turned out `UCesiumFeaturesMetadataComponent` was missing from the docs, too, but for a different reason. Apparently constructs like this confuse the Doxygen parser and so any class decorated this way won't show up in the docs at all:

```
UCLASS(ClassGroup = (Cesium), meta = (BlueprintSpawnableComponent))
```

However, this works fine:

```
UCLASS(ClassGroup = Cesium, meta = (BlueprintSpawnableComponent))
```

So I changed a bunch of occurrences of this, and now a handful of types that were missing are showing up. Not sure if there's any way to make Doxygen report problems like this, rather than silently excluding the type. I couldn't find a way with a brief look.

PR #1602, opened 2025-01-31
Update cesium-native

kring:
Most importantly to incorporate CesiumGS/cesium-native#1089.


kring:
Merging this to unbreak main.

PR #1605, opened 2025-02-05
Enum metadata support

azrogers:
Requires CesiumGS/cesium-native#1085 to be merged before this. This change implements enum support for structural metadata in Unreal. The implementation is fairly simple, the scalar values of enums can be obtained using `UCesiumMetadataValueBlueprintLibrary::GetInteger` or `UCesiumMetadataValueBlueprintLibrary::GetInteger64`. One wrinkle is in obtaining metadata values as strings - the names corresponding to enum values can only be obtained with the corresponding `CesiumGltf::Enum` definition. This means that we can't just call `UCesiumMetadataValueBlueprintLibrary::GetString` on an enum `FCesiumMetadataValue`, since the metadata value itself doesn't contain enough information to resolve the name. Because of this, I've elected to only implement enum support for `UCesiumPropertyTablePropertyBlueprintLibrary::GetString`, as with the property table reference we have enough information to resolve the enum names. It would be nice to have full support for `GetString` on an enum value, but considering our design in Cesium Native I think this might be the best we can do.

### DIFF
--- a/Source/CesiumRuntime/Private/CesiumPropertyTableProperty.cpp
+++ b/Source/CesiumRuntime/Private/CesiumPropertyTableProperty.cpp
@@ -2,6 +2,7 @@
 
 #include "CesiumPropertyTableProperty.h"
 #include "CesiumGltf/MetadataConversions.h"
+#include "CesiumGltf/PropertyEnumValue.h"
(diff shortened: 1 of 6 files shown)


azrogers:
Closed in favor of #1613.

PR #1612, opened 2025-02-11
Add includes for compilation on Linux

timoore:
This was breaking, probably due to some quirk of the unity build


kring:
Thanks @timoore!

PR #1613, opened 2025-02-13
ENUM support treating enum values as scalars rather than their own distinct type

azrogers:
This is an alternative to #1605 and a companion to CesiumGS/cesium-native#1112. Here, instead of treating enums as a distinct property type with a class to wrap around the values, we just treat the values as scalars with a bit of information attached describing the enum they're related to. These enum definitions are stored on the schema in a private extension, and are passed to each value using a `TSharedPtr`.


Source/CesiumRuntime/Public/CesiumModelMetadata.h
```diff
@@ -35,6 +36,7 @@ struct CESIUMRUNTIME_API FCesiumModelMetadata {
 private:
   TArray<FCesiumPropertyTable> _propertyTables;
   TArray<FCesiumPropertyTexture> _propertyTextures;
+  TSharedPtr<FCesiumMetadataEnumCollection> _enumDefinitions;
```
kring, line 39:
```suggestion
  TSharedPtr<FCesiumMetadataEnumCollection> _pEnumDefinitions;
```

Source/CesiumRuntime/Private/CesiumPropertyTable.cpp
```diff
@@ -23,11 +24,23 @@ FCesiumPropertyTable::FCesiumPropertyTable(
...
                                         const std::string& propertyName,
                                         auto propertyValue) mutable {
     FString key(UTF8_TO_TCHAR(propertyName.data()));
-    properties.Add(key, FCesiumPropertyTableProperty(propertyValue));
+    TSharedPtr<FCesiumMetadataEnum> EnumDefinition;
+    if (EnumCollection.IsValid() && propertyValue.enumDefinition() != nullptr) {
+      const CesiumGltf::Enum* pEnumDefinition = propertyValue.enumDefinition();
+      if (pEnumDefinition->name) {
+        EnumDefinition = EnumCollection.Pin()->Get(
+            FString(UTF8_TO_TCHAR(pEnumDefinition->name->c_str())));
```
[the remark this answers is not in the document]
kring: I would suggest instead looking up the `ClassProperty` via the `propertyName`. I don't think I see anything preventing passing the `Schema` into this constructor if necessary.

Source/CesiumRuntime/Public/CesiumPropertyTableProperty.h
```diff
@@ -106,6 +118,7 @@ struct CESIUMRUNTIME_API FCesiumPropertyTableProperty {
 
   FCesiumMetadataValueType _valueType;
   bool _normalized;
+  TSharedPtr<FCesiumMetadataEnum> _enumDefinition;
```
kring, line 121:
```suggestion
  TSharedPtr<FCesiumMetadataEnum> _pEnumDefinition;
```

Source/CesiumRuntime/Private/CesiumPropertyTableProperty.cpp
```diff
@@ -1261,6 +1263,13 @@ FString UCesiumPropertyTablePropertyBlueprintLibrary::GetString(
           auto maybeString = CesiumGltf::
               MetadataConversions<std::string, decltype(value)>::convert(value);
 
+          if constexpr (CesiumGltf::IsMetadataInteger<ValueType>::value) {
+            if (EnumDefinition.IsValid()) {
+              return EnumDefinition->GetName(value).Get(
+                  UnrealMetadataConversions::toString(*maybeString));
```
kring, line 1269: Bit of a nitpick, but the way this is structured, it's always going to format the number as a string (and allocate memory on the heap for it), even though in pretty much all cases, that's going to be immediately thrown away because the value corresponds to a valid enum. It's only a tiny bit more work to avoid this unnecessary heap allocation (by manually checking for nullopt), and worth it IMO.

Source/CesiumRuntime/Public/CesiumMetadataValueType.h
```diff
@@ -106,10 +106,72 @@ enum class ECesiumMetadataType : uint8 {
   Mat3 = int(CesiumGltf::PropertyType::Mat3),
   Mat4 = int(CesiumGltf::PropertyType::Mat4),
   Boolean = int(CesiumGltf::PropertyType::Boolean),
-  Enum = int(CesiumGltf::PropertyType::Enum),
   String = int(CesiumGltf::PropertyType::String),
```
kring, line 109: I don't understand why `Enum` was removed. Since this is supposed to represent the "true" type from the extension, it's important that the Enum type be represented. I can see that `TypeToMetadataValueType` can't distinguish Scalar from Enum as written, but that can be solved by passing in an extra parameter. It looks like every place that `TypeToMetadataValueType` is called, there's also an `EnumDefinition` available.

Source/CesiumRuntime/Public/CesiumPropertyTextureProperty.h
```diff
@@ -109,6 +117,7 @@ struct CESIUMRUNTIME_API FCesiumPropertyTextureProperty {
 
   FCesiumMetadataValueType _valueType;
   bool _normalized;
+  TSharedPtr<FCesiumMetadataEnum> _enumDefinition;
```
kring, line 120:
```suggestion
  TSharedPtr<FCesiumMetadataEnum> _pEnumDefinition;
```

kring (commented):
Hmm it seems Visual Studio Code submitted some of my previous review comments separately. Reviewing in VS Code rather than on GitHub is really convenient, but it seems a bit...buggy.

Anyway! This all looks good to me, actually a lot cleaner than I expected. There's just the one place I noticed where the Enum-ness appears to be lost, and I _think_ it shouldn't be too hard to get it back (but let me know if I'm wrong).

I'm very interested in Janine's take, too, though.

Source/CesiumRuntime/Private/CesiumEncodedMetadataConversions.cpp
```diff
@@ -22,8 +22,9 @@ GetBestFittingEncodedType(FCesiumMetadataPropertyDetails PropertyDetails) {
     }
 
     if (type != ECesiumMetadataType::Boolean &&
-        type != ECesiumMetadataType::Scalar) {
-      // Only boolean and scalar array properties are supported.
+        type != ECesiumMetadataType::Scalar &&
+        type != ECesiumMetadataType::Enum) {
+      // Only boolean, scalar, and enum array properties are supported.
```
j9liu, line 27: Just want to make it clear: this code supports enums as a type that can be encoded into metadata styling. But we haven't really decided on how one would communicate the Value -> Name mapping in Unreal, right? So I hesitate to say that we support enums in an intuitive way, but I guess pushing the scalar values to the material is better than nothing

j9liu (changes requested):
@azrogers I haven't had time to test this with an actual model, but this is looking much better than before. I'll do a second pass with testing after you address the existing comments

Source/CesiumRuntime/Private/CesiumPropertyTableProperty.cpp
azrogers: I don't see a `GetString` method in `CesiumPropertyTextureProperty` to add this to, unless you mean adding that method?

Source/CesiumRuntime/Private/CesiumEncodedMetadataConversions.cpp
azrogers: Doesn't this PR make a decision on that front? If you give it an enum definition, the scalar value is interpreted as an enum. Otherwise, it's interpreted as a scalar.

Source/CesiumRuntime/Private/CesiumPropertyTableProperty.cpp
j9liu: You're right, sorry. I remember omitting it because `CesiumPropertyTextureProperty` doesn't support strings. But then I'm not sure how someone is supposed to know to resolve the enum value from this property? I suppose the workflow to resolve an enum would be to get it as an `FCesiumMetadataValue`, but I don't see that being accounted for in `GetValue`

azrogers: I've updated the `GetValue` methods throughout the codebase to pass through the enum definition. Let me know if I missed any.

Source/CesiumRuntime/Private/CesiumEncodedMetadataConversions.cpp
j9liu: There are two parts to metadata styling:

1. Writing the values to a texture
2. Auto-generating nodes to make meaning of those values in the material.

These changes are for Part 1, allowing enum values to be encoded on the GPU as scalars. But we haven't done part 2 for enums, and I think that should be out-of-scope for this PR. In which case, we can exclude these changes that are meant for part 1, so that things can be clear and consistent. Does that make sense?

Source/CesiumRuntime/Public/CesiumPropertyArray.h
```diff
@@ -102,39 +103,52 @@ struct CESIUMRUNTIME_API FCesiumPropertyArray {
...
-        ECesiumMetadataType(CesiumGltf::TypeToPropertyType<T>::value);
-    ECesiumMetadataComponentType componentType = ECesiumMetadataComponentType(
-        CesiumGltf::TypeToPropertyType<T>::component);
-    bool isArray = false;
-
-    _elementType = {type, componentType, isArray};
-  }
+      : FCesiumPropertyArray(
+            MoveTemp(value),
+            TSharedPtr<FCesiumMetadataEnum>(nullptr)) {}
```
j9liu, line 126: Same note about the `= nullptr` default value in these constructors.

Source/CesiumRuntime/Private/Tests/CesiumPropertyArray.spec.cpp
```diff
@@ -142,4 +142,51 @@ void FCesiumPropertyArraySpec::Define() {
       }
     });
   });
+
+  Describe("UCesiumPropertyArrayBlueprintLibrary::GetString", [this]() {
```
j9liu, line 146: `GetString` is actually deprecated 😅 but it doesn't hurt to leave a test anyways

j9liu (changes requested):
Thanks @azrogers for the work so far. I was able to test this out with some sample models from [3d-tiles-samples](https://github.com/CesiumGS/3d-tiles-samples/tree/a8d4e1f47fe8c7ffc44e78c5f9b970532f8c8bc2/glTF/EXT_structural_metadata/), and it works pretty well.

I have several comments throughout that I think should be addressed before merge. Also add this to CHANGES.md when you get the chance, thanks

Source/CesiumRuntime/Public/CesiumMetadataValue.h
```diff
@@ -175,61 +176,79 @@ struct CESIUMRUNTIME_API FCesiumMetadataValue {
...
+        TypeToMetadataValueType<CesiumGltf::PropertyArrayView<ArrayType>>(
+            pEnumDefinition);
   }
 
+  template <typename ArrayType>
+  explicit FCesiumMetadataValue(CesiumGltf::PropertyArrayCopy<ArrayType>&& Copy)
+      : FCesiumMetadataValue(
+            std::move(Copy),
+            TSharedPtr<FCesiumMetadataEnum>(nullptr)) {}
+
```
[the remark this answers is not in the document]
azrogers: It's not a new constructor, it was there before. If you mean the one above that has `pEnumDefinition`, it's just a copy of the lower pre-existing constructor with a `pEnumDefinition` parameter.

Source/CesiumRuntime/Private/CesiumPropertyTableProperty.cpp
```diff
@@ -790,6 +792,11 @@ TResult propertyTablePropertyCallback(
                      property,
                      valueType,
                      std::forward<Callback>(callback));
+  case ECesiumMetadataType::Enum:
```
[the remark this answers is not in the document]
azrogers: They can't be consistent because an enum can't be normalized.

Source/CesiumRuntime/Private/CesiumPropertyTableProperty.cpp
```diff
@@ -1258,6 +1266,15 @@ FString UCesiumPropertyTablePropertyBlueprintLibrary::GetString(
             CesiumGltf::IsMetadataString<ValueType>::value) {
           return UnrealMetadataConversions::toString(value);
         } else {
+          if constexpr (CesiumGltf::IsMetadataInteger<ValueType>::value) {
+            if (EnumDefinition.IsValid()) {
+              TOptional<FString> MaybeName = EnumDefinition->GetName(value);
+              if (MaybeName.IsSet()) {
+                return MaybeName.GetValue();
+              }
```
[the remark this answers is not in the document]
azrogers: As far as the Unreal side is concerned, an enum value is just a scalar with an enum definition. If the scalar has no enum definition, we should treat it as a regular scalar.

Source/CesiumRuntime/Public/CesiumMetadataEnum.h
```diff
@@ -0,0 +1,128 @@
...
+struct Enum;
+struct Schema;
+struct Model;
+} // namespace CesiumGltf
+
+/**
+ * A Blueprint-accessible wrapper for a glTF structural metadata enum
+ * definition.
+ */
+USTRUCT(BlueprintType)
```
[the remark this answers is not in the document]
azrogers: You're right, it's meant to apply to the `FCesiumMetadataEnum` below. Do we have any tests for the existing blueprint functionality beyond what's covered in the unit tests?

Source/CesiumRuntime/Public/CesiumMetadataValueType.h
```diff
@@ -186,30 +188,42 @@ struct CESIUMRUNTIME_API FCesiumMetadataValueType {
...
-    type =
-        ECesiumMetadataType(CesiumGltf::TypeToPropertyType<ArrayType>::value);
+    if (CesiumGltf::IsMetadataInteger<ArrayType>::value &&
+        pEnumDefinition != nullptr) {
+      type = ECesiumMetadataType::Enum;
+    } else {
+      type =
+          ECesiumMetadataType(CesiumGltf::TypeToPropertyType<ArrayType>::value);
+    }
     componentType = ECesiumMetadataComponentType(
```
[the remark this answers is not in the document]
azrogers: This isn't the component type of an enum property, this is a component type of an enum array property. We need to know the component type so that we can read the array properly. I can make the change for single enum values though if we want.

Source/CesiumRuntime/Private/CesiumPropertyTableProperty.cpp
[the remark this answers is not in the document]
azrogers: I've changed it to return the default value for when the name is missing, at least.

Source/CesiumRuntime/Private/CesiumPropertyTable.cpp
```diff
@@ -23,11 +24,31 @@ FCesiumPropertyTable::FCesiumPropertyTable(
...
+  propertyTableView.forEachProperty([&properties = _properties,
+                                     &Schema = *pExtension->schema,
+                                     &propertyTableView,
+                                     &EnumCollection](
                                         const std::string& propertyName,
                                         auto propertyValue) mutable {
     FString key(UTF8_TO_TCHAR(propertyName.data()));
-    properties.Add(key, FCesiumPropertyTableProperty(propertyValue));
+    const CesiumGltf::ClassProperty& classProperty =
+        *propertyTableView.getClassProperty(propertyName);
```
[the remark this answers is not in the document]
azrogers: Made the change for `CesiumPropertyTextureView` though.

Source/CesiumRuntime/Public/CesiumMetadataValue.h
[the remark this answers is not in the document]
j9liu: Ah sorry, I think the diff made it seem like it was added, when in reality it was just moved.

Source/CesiumRuntime/Private/CesiumPropertyTable.cpp
```diff
@@ -7,7 +7,8 @@ static FCesiumPropertyTableProperty EmptyPropertyTableProperty;
 
 FCesiumPropertyTable::FCesiumPropertyTable(
     const CesiumGltf::Model& Model,
-    const CesiumGltf::PropertyTable& PropertyTable)
+    const CesiumGltf::PropertyTable& PropertyTable,
+    const TSharedPtr<FCesiumMetadataEnumCollection>& EnumCollection)
```
j9liu, line 11:
```suggestion
    const TSharedPtr<FCesiumMetadataEnumCollection>& pEnumCollection)
```

Source/CesiumRuntime/Public/CesiumMetadataValueType.h
```diff
@@ -186,22 +189,34 @@ struct CESIUMRUNTIME_API FCesiumMetadataValueType {
...
-        CesiumGltf::TypeToPropertyType<T>::component);
+    if (CesiumGltf::IsMetadataInteger<T>::value && pEnumDefinition.IsValid()) {
+      type = ECesiumMetadataType::Enum;
+      componentType = ECesiumMetadataComponentType(
+          CesiumGltf::TypeToPropertyType<T>::component);
+    } else {
+      type = ECesiumMetadataType(CesiumGltf::TypeToPropertyType<T>::value);
+      componentType = ECesiumMetadataComponentType(
+          CesiumGltf::TypeToPropertyType<T>::component);
+    }
```
j9liu, line 219: Nitpick:

```suggestion
    if (CesiumGltf::IsMetadataInteger<T>::value && pEnumDefinition.IsValid()) {
      type = ECesiumMetadataType::Enum;
    } else {
      type = ECesiumMetadataType(CesiumGltf::TypeToPropertyType<T>::value);
    }
    componentType = ECesiumMetadataComponentType(
          CesiumGltf::TypeToPropertyType<T>::component);
```

j9liu (commented):
Looking good @azrogers ! The last few comments are just small things, otherwise this is good to go. 💪

azrogers:
@j9liu Should be all set!

j9liu:
Thanks @azrogers ! I'll merge once CI passes. 🙌

PR #1615, opened 2025-02-14
Add soak test for memory usage with external tilesets

azrogers:
Because merging a nearly year-old branch with the current `main` amounted to way more work than just copying over the changes, I decided to simply to the latter to bring #1415 up to date. Unfortunately my attempt to push those changes back into the original branch without overwriting the old history of the branch seemed to rewrite all the git history and make the whole thing into a mess, so I'm just creating a new PR for it instead of attempting to integrate it into the old one.


kring:
I'm moving this out of the release because it's failing CI and it's not pressing.

kring:
Thanks @azrogers!

PR #1616, opened 2025-02-19
Minor metadata reorganization

j9liu:
This PR:

- Moves the `FCesiumFeaturesMetadataDescription` into its own file, `CesiumFeaturesMetadataDescription.h`
- Makes `UCesiumFeaturesMetadataComponent` store a `FCesiumFeaturesMetadataDescription` directly, instead of copying its inner properties, with backwards compatibility.
- Moves a function in `CesiumEncodedMetadataConversions` out of the anonymous namespace, for future usability.
- Adds `MetadataTypeToString` and `MetadataComponentTypeToString` for string printing utility.

You can test for seamless compatibility using the `06_CesiumMetadata` sample level. The level should continue to work on this branch despite the changes.


Source/CesiumRuntime/Public/CesiumMetadataValueType.h
```diff
@@ -251,3 +251,22 @@ static size_t GetMetadataTypeByteSize(
 
   return byteSize;
 }
+
+static FString MetadataTypeToString(ECesiumMetadataType type) {
+  const UEnum* EnumPtr =
```
azrogers, line 256: Should be able to use `StaticEnum<ECesiumMetadataType>()` instead of looking it up by name here and down below.

azrogers:
Looks good! Just one minor suggestion to avoid looking up the enums via string.

j9liu:
Thanks @azrogers ! Updated with some tweaks so this is ready for another look.

azrogers:
Looks good! Thanks @j9liu!

PR #1620, opened 2025-02-26
Use the vckpkg-pkg branch of cesium-native without vcpkg

timoore:
Test the suitability of https://github.com/CesiumGS/cesium-native/pull/1026


kring:
The change in this PR is already effectively in main. Closing.

Issue #1609, opened 2025-02-07
Game packaged as an ipa crashes at startup on iOS


kring:
Mentioned on the community forum here:
https://community.cesium.com/t/solved-ios-apps-distributed-to-the-app-store-crash/29401
https://community.cesium.com/t/issues-with-cesium-in-ue5-4-on-ios/37756/3

Packaging a game and running it on an iPhone from Xcode works fine. But if we create an "archive" (ipa) and then run that on the device, it crashes very quickly after showing the splash screen. The immediate cause of the crash is that a std::string's destructor is called, and the memory it holds was not allocated by the standard `operator new`. That's because it was allocated from Cesium3DTileset.cpp, which, being compiled in the usual Unreal way, overrides `operator new` with its own custom allocator.

The fundamental problem is that the `basic_string` constructor that gets called is found in our Unreal game, so it uses Unreal's custom allocator, while the `basic_string` destructor is found in `libc++.dylib`, so it uses the standard allocator.

> Rant: this is why overriding operator new and delete is a terrible idea. Think you can get better performance with a custom allocator? Well, ok, I'm skeptical that that's true if it's a general purpose allocator, but go ahead if you want to try. But call your custom allocator explicitly from your code. Overriding the global new and delete makes it pretty much impossible for dynamically-allocated objects to cross .dll/.so/.dylib boundaries. Which maybe doesn't sound _too_ bad, until you consider that "objects" includes std::string, and ".dll/.so/.dylib" includes libc++.so/.dylib. It's pretty much just dumb luck that this works on other platforms where the C++ standard library is dynamically linked.

I spent a lot of time trying to understand why the constructor/destructor end up coming from two different places. Based on stepping through the disassembly with the xcode debugger (with the app running on the iPhone), I can see:

* Neither the constructor not the destructor are inlined. Both are actual function calls.
* The constructor implementation is found in our app binary.
* The destructor implementation is found in libc++.dylib.

Why does this happen? Well, in Apple's libc++ (taken from iOS SDK v18.2), the relevant std::string constructor is declared/defined like this:

```
  template <__enable_if_t<__is_allocator<_Allocator>::value, int> = 0>
  _LIBCPP_HIDE_FROM_ABI _LIBCPP_CONSTEXPR_SINCE_CXX20 basic_string(const _CharT* __s) {
     // ...
  }
```

The destructor looks like this:

```
[...]

PR #1621, opened 2025-02-27
Add instructions for symbolicating iOS stack traces.

kring:
Getting a useful stack trace from an Unreal app on an iOS device is surprisingly hard, but I worked out a way to do it while working on #1609 and wrote down some instructions.


azrogers:
Thanks @kring!

PR #1622, opened 2025-02-28
Parent actors created through the editor panel to the first selected georeference, if any

### LINKED ISSUE #1364: "Blank 3D Tiles Tileset" add button should account for selected georeference
If multiple `CesiumGeoreference` actors are in the scene, every "Blank 3D Tiles Tileset" added from the Quick Add panel will go under the first georeference in the scene. This happens even when a different georeference is selected. It would be a nice quality-of-life improvement if this accounted for the `CesiumGeoreference` that was selected.

azrogers:
Closes #1364. Currently, spawning an actor through the Cesium editor panel (such as the "Blank 3D Tileset" button) will always parent the new actor to the first georeference in the scene. This change makes it so that the first *selected* georeference, if any, will be chosen as the parent.


kring:
Thanks @azrogers!

PR #1627, opened 2025-03-05
Fix broken "Auto Fill" on `UCesiumFeaturesMetadataComponent`

j9liu:
Sorry that I missed this in #1616. I didn't push changes that rewired the new field of `CesiumFeaturesMetadataComponent` in `AutoFill()`, so the button doesn't work on `main`.

Since this is breaking it probably rewarrants a re-release 😬

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,11 @@
 # Change Log
 
+### ? - ?
+
+##### Fixes :wrench:
+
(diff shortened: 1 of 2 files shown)


j9liu:
I updated CHANGES.md and the plugin version in preparation for a patch release.

kring:
Thanks @j9liu!

PR #1631, opened 2025-03-15
User-defined credit text for tileset, with priority

Julot94:
- Add access to user-defined credit text in Unreal
- Add option to prioritize this user-defined credit text, by leveraging  [CesiumGS/cesium-native/1128](https://github.com/CesiumGS/cesium-native/pull/1128)


Julot94:
Abandoned - we will do things differently now that a new iTwin API is being developed to manage access to the Google tileset.

PR #1632, opened 2025-03-17
Improve organization of generated docs.

### LINKED ISSUE #1614: Organize tree of topics in the documentation
I suggest (though I'm not sure this is just right yet...):

* Cesium for Unreal
  * Change Log
  * User Guide - Add a link here to the tutorials on the main web site.
    * Cesium for Unreal v2.0 Upgrade Guide
    * Using Pre-Release Packages
  * Contribution Guide - rename to "Contributing"
    * Code of Conduct
  * Developer Resources - rename to "Contributors' Guide" - perhaps add an intro paragraph about these being resources for people that are modifying Cesium for Unreal itself, rather than people that are using it.
    * Developer Setup
      * Developer Setup for Windows
      * Developer Setup for macOS
      * Developer Setup for Linux
    * Cesium for Unreal Style Guide - rename to "Style Guide"
    * API Design Guide - rename to "API Design"
    * Performance Profiling
    * Releasing a new version of Cesium for Unreal
    * 3D Tiles in Unreal Engine
    * ~Packaging~ - delete this, it is too outdated to be useful. The GH Actions scripts are the authority for packaging.
    * ~Reference Frames~ - This is pretty outdated, too. It would be good to have something like this, but I think it's just misleading in its current form.

Some other small things to improve in the Cesium for Unreal docs:

* [ ] The [Developer Setup](https://cesium.com/learn/cesium-unreal/ref-doc/developer-setup-unreal.html) page has a broken `[TOC]` and broken `\subpage` links to the platform-specific instructions.
* [ ] Is it possible to make the tree on the left expand the selected topic? If so, I think we could eliminate the `[TOC]` on every page, because it would be mostly redundant.

kring:
Fixes #1614

Before:

image

After:

image


Documentation/packaging-guide.md
```diff
@@ -1,40 +0,0 @@
-# Packaging {#packaging-guide}
```
j9liu, line 1: Did this get moved somewhere?

Documentation/reference-frames.md
```diff
@@ -1,72 +0,0 @@
-# Reference Frames {#reference-frames}
```
j9liu, line 1: Same here -- was this moved to a new place?

j9liu (commented):
Thanks @kring ! I haven't tested the locally-built pages yet, but I have a few questions from a quick scan.

Documentation/packaging-guide.md
j9liu: Ah I just looked at the issue this is based on and see it was removed intentionally -- ignore this!

Documentation/reference-frames.md
j9liu: Ignore this too 😅

PR #1633, opened 2025-03-17
If gltf material is BLEND, then always choose transparent material.

AlexandreWalensa:
Previous version of the code would also check pbr.baseColorFactor of the material has an alpha channel, and force opaque material if not. In our case (tileset coming from Bentley mesh export service) this behavior is "wrong", because colors (and thus alpha) are defined per-vertex, so pbr.baseColorFactor may have no alpha. So this would result in using the opaque material and render transparency with dithering, which is very low quality transparency.

Actually, currently in our case pbr.baseColorFactor has its default value of (1,1,1,1), which has alpha, so this fix is not strictly needed.


kring:
Thanks @AlexandreWalensa! I merged in main and added a mention of the fix to CHANGES.md. Merging!

PR #1635, opened 2025-03-18
Metadata: retrieve a property by name in c++

Julot94:
This can be useful when we need to retrieve a given property for a large amount of faces, as it is much faster than calling GetMetadataValuesForFace for each face (which allocates a TMap and makes the same validity test each time)

### DIFF
--- a/Source/CesiumRuntime/Private/CesiumMetadataPropertyAccess.cpp
+++ b/Source/CesiumRuntime/Private/CesiumMetadataPropertyAccess.cpp
@@ -1,0 +1,45 @@
+// Copyright 2020-2023 CesiumGS, Inc. and Contributors
+
+#include "CesiumMetadataPropertyAccess.h"
+
+#include "CesiumFeatureIdSet.h"
+#include "CesiumModelMetadata.h"
+#include "CesiumPrimitiveFeatures.h"
+
+/*static*/
+const FCesiumPropertyTableProperty*
+FCesiumMetadataPropertyAccess::FindValidProperty(
+    const FCesiumPrimitiveFeatures& Features,
+    const FCesiumModelMetadata& Metadata,
+    const FString& PropertyName,
+    int64 FeatureIDSetIndex) {
+  const TArray<FCesiumFeatureIdSet>& featureIDSets =
+      UCesiumPrimitiveFeaturesBlueprintLibrary::GetFeatureIDSets(Features);
+
+  if (FeatureIDSetIndex < 0 || FeatureIDSetIndex >= featureIDSets.Num()) {
+    return nullptr;
+  }
+
+  const FCesiumFeatureIdSet& featureIDSet =
+      featureIDSets[FeatureIDSetIndex];
+  const int64 propertyTableIndex =
+      UCesiumFeatureIdSetBlueprintLibrary::GetPropertyTableIndex(
+          featureIDSet);
+
+  const TArray<FCesiumPropertyTable>& propertyTables =
+      UCesiumModelMetadataBlueprintLibrary::GetPropertyTables(Metadata);
+  if (propertyTableIndex < 0 || propertyTableIndex >= propertyTables.Num()) {
+    return nullptr;
+  }
+  const FCesiumPropertyTableProperty& propWithName =
+      UCesiumPropertyTableBlueprintLibrary::FindProperty(
+          propertyTables[propertyTableIndex],
+          PropertyName);
+  const ECesiumPropertyTablePropertyStatus status =
+      UCesiumPropertyTablePropertyBlueprintLibrary::
+          GetPropertyTablePropertyStatus(propWithName);
+  if (status != ECesiumPropertyTablePropertyStatus::Valid) {
+    return nullptr;
+  }
+  return &propWithName;
+}
--- a/Source/CesiumRuntime/Public/CesiumMetadataPropertyAccess.h
+++ b/Source/CesiumRuntime/Public/CesiumMetadataPropertyAccess.h
@@ -1,0 +1,24 @@
+// Copyright 2020-2023 CesiumGS, Inc. and Contributors
+
+#pragma once
+
+#include "Containers/UnrealString.h"
+
+struct FCesiumPrimitiveFeatures;
+struct FCesiumModelMetadata;
+struct FCesiumPropertyTableProperty;
+
+class CESIUMRUNTIME_API FCesiumMetadataPropertyAccess {
+
+public:
+  /**
+   * Retrieves a property by name.
+   * If the specified feature ID set does not exist or if the property table
+   * does not contain a property with that name, this function returns nullptr.
+   */
+  static const FCesiumPropertyTableProperty* FindValidProperty(
+      const FCesiumPrimitiveFeatures& Features,
+      const FCesiumModelMetadata& Metadata,
+      const FString& PropertyName,
+      int64 FeatureIDSetIndex = 0);
+};


j9liu (changes requested):
Thanks @Julot94 for the PR! I like the functionality that this adds, but I propose redefining it slightly. Could we rewrite the function definition like so?

```
 /**
   * Retrieves a property from the component by name.
   * If the specified feature ID set does not exist or if the property table
   * does not contain a property with that name, this function returns nullptr.
   */
  static const FCesiumPropertyTableProperty* FindValidPropertyTableProperty(
      const FCesiumPrimitiveFeatures& Features,
      const FCesiumModelMetadata& Metadata,
      const FString& PropertyName,
      int64 FeatureIDSetIndex = 0);
```

Since there are multiple types of "Property" (PropertyTableProperty, PropertyTextureProperty, etc.), it's better to be more explicit about the type in the function name.

Do you also have an example of where you would call this in code? We usually pass in a target `UPrimitiveComponent` instead of separate features / metadata structs, but I'd understand if that doesn't work for your use case. 

Depending on the above answer, I might propose reorganizing this to under a different class, or even making this accessible from Blueprints.

Julot94:
Thanks @j9liu for the review!
- I totally agree with the renaming of the function, which avoids ambiguity.
- For the 2nd point unfortunately, this would not fit our case: we plug our callback in `loadPrimitiveGameThreadPart`, and in this context, the relevant _Features_ we have is **pMesh**->Features (where pMesh is a _UCesiumGltfPrimitiveComponent_) but the relevant model meta-data is **pGltf**->Metadata (pMesh->Metadata exists too, but is of a different type, _FCesiumPrimitiveMetadata_, and in our case, it is empty).
I have to admit that I did not dig very much to understand the difference between `FCesiumPrimitiveMetadata `and `FCesiumModelMetadata`, but it's how it works in the tilesets generated by Bentley's mesh-export service...

j9liu:
@Julot94 Ah yes, sorry for the confusion 😅 I can explain if it helps!

`FCesiumMetadataPrimitive` was based on the old [`EXT_features_metadata` extension](https://github.com/CesiumGS/glTF/tree/3d-tiles-next/extensions/2.0/Vendor/EXT_feature_metadata), where the feature IDs and metadata properties were bundled together. This deprecated extension has been split into the [`EXT_mesh_features`](https://github.com/CesiumGS/glTF/tree/3d-tiles-next/extensions/2.0/Vendor/EXT_mesh_features) and [`EXT_structural_metadata`](https://github.com/CesiumGS/glTF/tree/3d-tiles-next/extensions/2.0/Vendor/EXT_structural_metadata) extensions respectively.

The `EXT_structural_metadata` extension can exist on both a mesh primitive and the base glTF model, so that's why `FCesiumPrimitiveMetadata` and `FCesiumModelMetadata` are separate. The model part contains the schema for the metadata classes, as well as the property tables and textures of the model. The primitive part contains references to the property textures and property attributes that a mesh primitive uses. For your use case, where you're mostly using property tables, the `FCesiumModelMetadata` is definitely more important.

> this would not fit our case: we plug our callback in loadPrimitiveGameThreadPart, and in this context, the relevant Features we have is pMesh->Features (where pMesh is a UCesiumGltfPrimitiveComponent) but the relevant model meta-data is pGltf->Metadata (pMesh->Metadata exists too, but is of a different type, FCesiumPrimitiveMetadata, and in our case, it is empty).

Ah, I see your reasoning. It's not obvious, but it _is_ possible to get the parent `UCesiumGltfComponent` of a `UCesiumGltfPrimitiveComponent` if you do:

https://github.com/CesiumGS/cesium-unreal/blob/79c9224d610c77b6bd74c6f391cd1aa815d63d62/Source/CesiumRuntime/Private/CesiumMetadataPickingBlueprintLibrary.cpp#L209-L219

That's definitely not obvious, though.

j9liu:
Sorry @Julot94, I completely dropped the ball on this. 😵‍💫 

Now that it's been a while, I think we should relocate this function to `UCesiumMetadataPickingBlueprintLibrary`. Consistency would be nice, but there's no real benefit to restricting the functions such that they must accept `Hit`s. It's a convenience library, and this function is fairly relevant.

So to summarize the feedback:

- Move the function to `UCesiumMetadataPickingBlueprintLibrary`
- Rename to `FindValidPropertyTableProperty`
- (Nitpick) Replace `Features` and `Metadata` with a `UPrimitiveComponent*` parameter

I also think it would also be easy to add a modifier to make this a Blueprint callable function, like so:
```
 /**
   * Retrieves a property table property from the component by name.
   * If the specified feature ID set does not exist or if the property table
   * does not contain a property with that name, this function returns nullptr.
   */
 UFUNCTION(
      BlueprintCallable,
      BlueprintPure,
      Category = "Cesium|Metadata|Picking")
  static const FCesiumPropertyTableProperty* FindValidPropertyTableProperty(
      const UPrimitiveComponent* Component,
      const FString& PropertyName,
      int64 FeatureIDSetIndex = 0);
```

What do you think?

Julot94:
Hi @j9liu,
It sounds great to me - I'll do the modifications tomorrow. Thank you for your feedback!

Julot94:
Hi @j9liu ,
The blueprint system does not accept a pointer here (I get an error `Inappropriate '*' on variable of type 'FCesiumPropertyTableProperty', cannot have an exposed pointer to this type.`).
I could return a const reference instead, a bit like in _UCesiumPropertyTableBlueprintLibrary::FindProperty_, and rename the function as _FindPropertyTableProperty_, but then, I would need a method such as `bool IsValid() const` in FCesiumPropertyTableProperty (currently, all is private in struct FITwinCesiumPropertyTableProperty), to test the validity of the returned table property. What do you think?

Julot94:
Also, it could be useful to keep the version working on separate Features and Metadata, at least for C++, for corner cases (during the creation of the glTF mesh component, typically, when the features have not yet been "moved" from _loadResult_ though this call:
`pMesh->Features = std::move(loadResult.Features);` in **loadPrimitiveGameThreadPart**

With only the version working on a UPrimitiveComponent, I would need to move those features sooner (or just very temporarily, before restoring loadResult...)
The version working on Features and Metadata would have no reason to be exposed to blueprints, but would avoid complications in our usage...
Would it be OK for you @j9liu ?

j9liu:
Hi @Julot94! Thanks for trying out my suggestions :pray:

> The blueprint system does not accept a pointer here (I get an error Inappropriate '*' on variable of type 'FCesiumPropertyTableProperty', cannot have an exposed pointer to this type.).

Ah my mistake... I thought I saw this work for `UClass` types, but the rules might not apply to `UStruct`s!

> I could return a const reference instead, a bit like in UCesiumPropertyTableBlueprintLibrary::FindProperty, and rename the function as FindPropertyTableProperty, but then, I would prefer adding a method such as bool IsValid() const in FCesiumPropertyTableProperty (currently, all is private in struct FCesiumPropertyTableProperty), to test the validity of the returned table property

That sounds like a great idea! Yes, let's do that instead :smile:

> Also, it could be useful to keep the version working on separate Features and Metadata, at least for C++, for corner cases (during the creation of the glTF mesh component, typically, when the features have not yet been "moved" from loadResult through this call:

We definitely don't want to make things harder for your use case; I think keeping the parameters separate is fine. We have methods to retrieve the features / model metadata from a `UPrimitiveComponent` anyways, and it's less cumbersome to do that manually than other steps in the process. :)

Julot94:
Hi @j9liu ,
I have implemented the suggested modifications. As you can see, I kept the version working on separate parameters for our case, to avoid the complications discussed yesterday, but I only exposed the one working on a  `UPrimitiveComponent` in blueprint, since it's more homogeneous with the other functions and easier to use for BP developers.

j9liu:
Thanks for the changes @Julot94! I found a header include bug when testing this locally, so I just pushed a commit for that. I also added a changelog entry and unit test while I was at it -- I felt bad about bothering you even more on this PR 😆

PR #1637, opened 2025-03-18
Refresh tileset: fixed crash when refreshing a tileset frantically

Julot94:
_tilesToHideNextFrame contains raw pointers on some tiles: empty it before destroying tiles, or we can get a crash in next Tick processing (happened when refreshing the Google Tileset from a custom UI several times, "too quickly")


kring:
Thanks @Julot94! I've just merged in main and updated CHANGES.md. Merging!

PR #1641, opened 2025-03-24
Fix broken polygon clipping on macOS and probably other Apple platforms

### LINKED ISSUE #1640: Broken clipping on macOS
Observed on macOS 15.3.2 on an M4 Max MacBook Pro in UE 5.5.3.

Clipping with CesiumCartographicPolygons doesn't work at all on this system. The behaviors are kind of dodgy and inconsistent. Here's a simple default clipping polygon being used to clip CWT + Bing. Instead of removing the pixels inside the square, it... changed their shading somehow?

Image

Zoom in closer and entire tiles are missing:

Image

I'm pretty sure I have seen clipping work recently on this system, but I think both macOS and Unreal Engine have been updated since then. I'm not sure if that's what caused it to break, or if something changed in our code, or if I just straight-up imagined it was working.

In any case, after spending entirely too long trying to debug this, I narrowed it down to the `CesiumSelectTexCoords` material function. It has a custom HLSL node that looks like this:

```
float2 texCoords[8] = {TexCoord0, TexCoord1, TexCoord2, TexCoord3, TexCoord4, TexCoord5, TexCoord6, TexCoord7};
int index = TexCoordIndex;
return texCoords[index];
```

It's incredibly unfortunate we need to do this kind of thing, but anyway... I noticed that if I cut the array down to only 4 elements, like this, the problem went away:

```
float2 texCoords[4] = {TexCoord0, TexCoord1, TexCoord2, TexCoord3};
int index = TexCoordIndex;
return texCoords[index];
```

I thought it might have to do with accessing some of the higher-numbered texture coordinates, but no. If we repeat the first four texture coordinate sets in the last four slots, we still have the bug:

```
float2 texCoords[8] = {TexCoord0, TexCoord1, TexCoord2, TexCoord3, TexCoord0, TexCoord1, TexCoord2, TexCoord3};
int index = TexCoordIndex;
return texCoords[index];
```

And it turns out the magic number is 4. If the `texCoords` array has 4 or fewer elements, no problem! If it has more, clipping doesn't work.

Here's the best part! This works perfectly:

```
float2 texCoords1[4] = {TexCoord0, TexCoord1, TexCoord2, TexCoord3};
float2 texCoords2[4] = {TexCoord4, TexCoord5, TexCoord6, TexCoord7};
int index = TexCoordIndex;
if (index < 4) return texCoords1[index];
else return texCoords2[index - 4];
```

So Metal (or something?) can't handle arrays of `float2` with more than four elements? I just can't even imagine...

To try to understand what's going on, I asked Unreal to show me the Metal code for the complete material instance. I did this twice, once with 4 elements in that array (and thus clipping working), and a second time with 5 elements (and thus clipping not working). Then I diffed the two sets of code. And.. yep! The only difference in all 7000 lines of the shader is the extra element in the array! So I can only imagine this is a Metal compiler or driver bug.

Probably-related clipping dodginess reported on the forum here:
https://community.cesium.com/t/issue-problems-with-cesium-and-ue-5-5-3-on-macos-ios/38984

kring:
Fixes #1640 
Lots of details in that issue.

I ended up fixing this by changing the `CesiumSelectTexCoords` custom HLSL node to this:

```hlsl
int index = round(TexCoordIndex);
if (index < 4) {
    if (index < 2) {
        return index < 1 ? TexCoord0 : TexCoord1;
    } else {
        return index < 3 ? TexCoord2 : TexCoord3;
    }
} else {
    if (index < 6) {
        return index < 5 ? TexCoord4 : TexCoord5;
    } else {
        return index < 7 ? TexCoord6 : TexCoord7;
    }
}
```

This avoids the bug and also seems to be a bit faster in my tests. While replacing an index into an 8-element array with three comparisons doesn't sound like it should be improvement, it's not too shocking that it is. This code is branching on a uniform (in CesiumJS this would be a compile-time constant), so this structure probably allows the compiler to notice that and optimize it better.

As a bonus, this PR also includes some new instructions for using UE 5.3's required Xcode version on a newer version of macOS that doesn't support it.


azrogers:
While I have no way of testing that this fixes the issue you reported, your thought process and code seems to make sense and I can at least confirm it doesn't break anything for me. Considering nobody else on our team has a mac anyways, I'm going to merge this!

PR #1648, opened 2025-04-08
Update for the "multiple views" changes in Cesium Native

kring:
This PR updates cesium-unreal to work with the changes in CesiumGS/cesium-native#1125.

### DIFF
--- a/Source/CesiumRuntime/Private/Cesium3DTileset.cpp
+++ b/Source/CesiumRuntime/Private/Cesium3DTileset.cpp
@@ -2062,11 +2062,22 @@ void ACesium3DTileset::Tick(float DeltaTime) {
   const Cesium3DTilesSelection::ViewUpdateResult* pResult;
   if (this->_captureMovieMode) {
     TRACE_CPUPROFILER_EVENT_SCOPE(Cesium::updateViewOffline)
-    pResult = &this->_pTileset->updateViewOffline(frustums);
+    pResult = &this->_pTileset->updateViewGroupOffline(
+        this->_pTileset->getDefaultViewGroup(),
+        frustums);
   } else {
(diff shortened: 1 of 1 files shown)


j9liu (commented):
Thanks @kring ! I went through some of the samples and didn't see anything of note. I "stress-tested" by repeatedly refreshing tilesets and switching levels to try and catch any state mismanagement, but all seems well. Credits are working too.

There's just that one unit test that's failing on CI. Otherwise, let me know if there's anything else to verify, e.g. anything related to performance.

kring:
Here's what's happening with that one test. It took me entirely too long to figure this out today.

Apparently, when running headless in UE 5.5 (but not earlier versions!), the engine seemingly ignores attempts to set the viewport width/height. i.e., this code in the test has no effect:

```
  Params.EditorPlaySettings->NewWindowWidth = viewportWidth;
  Params.EditorPlaySettings->NewWindowHeight = viewportHeight;
```

So the viewport ends up being 0x0. Ok, it's headless, so... whatever, right? Well, when looking for cameras, Cesium3DTileset ignores any that have a null viewport like this. So during the failing testing, we end up calling `updateView` with an empty list of ViewStates.

In previous versions, that would quickly do nothing and report it was 100% finished doing nothing. Because this test doesn't actually assert anything, other than that the tileset gets to 100% loaded eventually, the test would pass.

But now, calling `updateView` without any ViewStates effectively doesn't do an update. The view group maintains whatever state it had before. After calling `updateView` it remains at 0% complete, and the test eventually times out waiting for it to hit 100%.

I'm not sure what the right solution is here. Revert `updateView` to the old behavior? Try to fix the viewport sizing on UE 5.5 somehow? Remove the test since it's not testing anything anyway, really? I think I'll need to defer a decision until tomorrow.

j9liu:
Thanks for fixing CI @kring ! Merging now

PR #1651, opened 2025-04-09
Add support for loading and interacting with GeoJSON data in Blueprint

azrogers:
The companion to CesiumGS/cesium-native#1154. This PR adds blueprint functions for loading and interacting with GeoJSON data. In particular, it provides the ability to load GeoJSON from a string or from Cesium ion, the ability to query nodes, and the ability to obtain primitives as values usable in Unreal. 

One note is that access to the properties of a vector node requires blueprint functionality for interacting with JSON objects. I elected to push this responsibility onto the JSON Blueprint Utilities plugin, which is an official Unreal plugin but not enabled by default. Returning the `FJsonObjectWrapper` struct from a method is supported without this plugin enabled, but users will have to enable that plugin if they want to be able to do anything with it.

### DIFF
--- a/Source/CesiumRuntime/Private/Tests/CesiumVectorDocument.spec.cpp
+++ b/Source/CesiumRuntime/Private/Tests/CesiumVectorDocument.spec.cpp
@@ -1,0 +1,47 @@
+// Copyright 2020-2025 CesiumGS, Inc. and Contributors
+
+#include "CesiumVectorDocument.h"
+#include "CesiumVectorNode.h"
+#include "Misc/AutomationTest.h"
+
+BEGIN_DEFINE_SPEC(
+    FCesiumVectorDocumentSpec,
+    "Cesium.Unit.CesiumVectorDocument",
+    EAutomationTestFlags::EditorContext | EAutomationTestFlags::ClientContext |
+        EAutomationTestFlags::ProductFilter | EAutomationTestFlags::NonNullRHI)
+END_DEFINE_SPEC(FCesiumVectorDocumentSpec)
+
+void FCesiumVectorDocumentSpec::Define() {
+  Describe("UCesiumVectorDocumentBlueprintLibrary::LoadGeoJsonFromString", [this]() {
+    It("loads a valid GeoJSON document", [this]() {
+      FCesiumVectorDocument Document;
+      const bool bSuccess =
+          UCesiumVectorDocumentBlueprintLibrary::LoadGeoJsonFromString(
+              R"==({ "type": "Point", "coordinates": [1, 2, 3] })==",
+              Document);
+      TestTrue("LoadGeoJsonFromString Success", bSuccess);
+    });
+    It("fails to load an invalid GeoJSON document", [this]() {
+      AddExpectedError(
+          "Errors while loading GeoJSON from string:\n- Failed to parse GeoJSON: Missing a name for object member.");
+      FCesiumVectorDocument Document;
+      bool bSuccess =
+          UCesiumVectorDocumentBlueprintLibrary::LoadGeoJsonFromString(
+              R"==({ type: "Point", coordinates: [1, 2, 3] })==",
+              Document);
+      TestFalse("LoadGeoJsonFromString Success", bSuccess);
+      AddExpectedError(
+          "Errors while loading GeoJSON from string:\n- Unknown GeoJSON object type: 'Invalid'");
+      bSuccess = UCesiumVectorDocumentBlueprintLibrary::LoadGeoJsonFromString(
+          R"==({ "type": "Invalid", "coordinates": [] })==",
+          Document);
+      TestFalse("LoadGeoJsonFromString Success", bSuccess);
+      AddExpectedError(
+          "Errors while loading GeoJSON from string:\n- Failed to parse GeoJSON: Invalid value.");
+      bSuccess = UCesiumVectorDocumentBlueprintLibrary::LoadGeoJsonFromString(
+          R"==(<some_xml_idk />)==",
+          Document);
+      TestFalse("LoadGeoJsonFromString Success", bSuccess);
+    });
+  });
+}
--- a/Source/CesiumRuntime/Private/Tests/CesiumVectorDocument.spec.cpp
+++ b/Source/CesiumRuntime/Private/Tests/CesiumVectorDocument.spec.cpp
@@ -44,4 +44,279 @@ void FCesiumVectorDocumentSpec::Define() {
       TestFalse("LoadGeoJsonFromString Success", bSuccess);
     });
   });
+
+  Describe("UCesiumVectorNodeBlueprintLibrary::GetIdAsInteger", [this]() {
+    It("correctly interprets an integer ID", [this]() {
+      FCesiumVectorDocument Document;
+      TestTrue(
+          "LoadGeoJsonFromString Success",
+          UCesiumVectorDocumentBlueprintLibrary::LoadGeoJsonFromString(
+              R"==({ "type": "Feature", "id": 10, "geometry": null, "properties": null })==",
+              Document));
+      TestEqual(
+          "GetIdAsInteger",
+          UCesiumVectorNodeBlueprintLibrary::GetIdAsInteger(
+              UCesiumVectorDocumentBlueprintLibrary::GetRootNode(Document)),
+          10);
+    });
+    It("returns -1 when the ID is missing", [this]() {
+      FCesiumVectorDocument Document;
+      TestTrue(
+          "LoadGeoJsonFromString Success",
+          UCesiumVectorDocumentBlueprintLibrary::LoadGeoJsonFromString(
+              R"==({ "type": "Feature", "geometry": null, "properties": null })==",
+              Document));
+      TestEqual(
+          "GetIdAsInteger",
+          UCesiumVectorNodeBlueprintLibrary::GetIdAsInteger(
+              UCesiumVectorDocumentBlueprintLibrary::GetRootNode(Document)),
+          -1);
+    });
+  });
+
+  Describe("UCesiumVectorNodeBlueprintLibrary::GetIdAsString", [this]() {
+    It("correctly interprets a string ID", [this]() {
+      FCesiumVectorDocument Document;
+      TestTrue(
+          "LoadGeoJsonFromString Success",
+          UCesiumVectorDocumentBlueprintLibrary::LoadGeoJsonFromString(
+              R"==({ "type": "Feature", "id": "test", "geometry": null, "properties": null })==",
+              Document));
+      TestEqual(
+          "GetIdAsString",
+          UCesiumVectorNodeBlueprintLibrary::GetIdAsString(
+              UCesiumVectorDocumentBlueprintLibrary::GetRootNode(Document)),
+          "test");
+    });
+    It("stringifies an integer ID", [this]() {
+      FCesiumVectorDocument Document;
+      TestTrue(
+          "LoadGeoJsonFromString Success",
+          UCesiumVectorDocumentBlueprintLibrary::LoadGeoJsonFromString(
+              R"==({ "type": "Feature", "id": 10, "geometry": null, "properties": null })==",
+              Document));
+      TestEqual(
+          "GetIdAsString",
+          UCesiumVectorNodeBlueprintLibrary::GetIdAsString(
+              UCesiumVectorDocumentBlueprintLibrary::GetRootNode(Document)),
+          "10");
+    });
+    It("returns an empty string when the ID is missing", [this]() {
+      FCesiumVectorDocument Document;
+      TestTrue(
+          "LoadGeoJsonFromString Success",
+          UCesiumVectorDocumentBlueprintLibrary::LoadGeoJsonFromString(
+              R"==({ "type": "Feature", "geometry": null, "properties": null })==",
+              Document));
+      TestEqual(
+          "GetIdAsString",
+          UCesiumVectorNodeBlueprintLibrary::GetIdAsString(
+              UCesiumVectorDocumentBlueprintLibrary::GetRootNode(Document)),
+          "");
+    });
+  });
+
+  Describe("UCesiumVectorNodeBlueprintLibrary::GetChildren", [this]() {
+    It("returns an array of children", [this]() {
+      FCesiumVectorDocument Document;
+      TestTrue(
+          "LoadGeoJsonFromString Success",
+          UCesiumVectorDocumentBlueprintLibrary::LoadGeoJsonFromString(
+              R"==({
+                  "type": "FeatureCollection",
+                  "features": [
+                    { "type": "Feature", "id": "test", "geometry": null, "properties": null },
+                    { "type": "Feature", "id": "test2", "geometry": null, "properties": null }
+                  ]
+                 })==",
+              Document));
+      TArray<FCesiumVectorNode> Children =
+          UCesiumVectorNodeBlueprintLibrary::GetChildren(
+              UCesiumVectorDocumentBlueprintLibrary::GetRootNode(Document));
+      TestEqual("Children.Num()", Children.Num(), 2);
+      TestEqual(
+          "Children[0] Id",
+          UCesiumVectorNodeBlueprintLibrary::GetIdAsString(Children[0]),
+          "test");
+      TestEqual(
+          "Children[1] Id",
+          UCesiumVectorNodeBlueprintLibrary::GetIdAsString(Children[1]),
+          "test2");
+    });
+  });
+
+  Describe("UCesiumVectorNodeBlueprintLibrary::GetPrimitives", [this]() {
+    It("returns an array of primitives", [this]() {
+      FCesiumVectorDocument Document;
+      TestTrue(
+          "LoadGeoJsonFromString Success",
+          UCesiumVectorDocumentBlueprintLibrary::LoadGeoJsonFromString(
+              R"==({
+                  "type": "MultiPoint",
+                  "coordinates": [
+                    [ 1, 2, 3 ],
+                    [ 4, 5, 6 ]
+                  ]
+                 })==",
+              Document));
+      TArray<FCesiumVectorPrimitive> Primitives =
+          UCesiumVectorNodeBlueprintLibrary::GetPrimitives(
+              UCesiumVectorDocumentBlueprintLibrary::GetRootNode(Document));
+      TestEqual("Primitives.Num()", Primitives.Num(), 2);
+      TestEqual(
+          "Primitives[0] as point",
+          UCesiumVectorPrimitiveBlueprintLibrary::GetPrimitiveAsPoint(
+              Primitives[0]),
+          FVector(1, 2, 3));
+      TestEqual(
+          "Primitives[1] as point",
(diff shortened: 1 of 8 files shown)


Source/CesiumRuntime/Public/CesiumGeoJsonDocument.h
```diff
@@ -0,0 +1,117 @@
...
+
+private:
+  CesiumUtility::IntrusivePointer<CesiumVectorData::GeoJsonDocument> _document;
+
+  friend class UCesiumGeoJsonDocumentBlueprintLibrary;
+};
+
+/**
+ * @brief A Blueprint Function Library providing functions for interacting with
+ * a `FCesiumVectorDocument`.
```
j9liu:
```suggestion
 * a `FCesiumGeoJsonDocument`.
```

Source/CesiumRuntime/Public/CesiumGeoJsonDocument.h
```diff
@@ -0,0 +1,118 @@
...
+  UFUNCTION(
+      BlueprintCallable,
+      Category = "Cesium|Vector|Document",
+      meta =
+          (BlueprintInternalUseOnly = true,
+           DisplayName = "Load Vector Document from Cesium ion"))
+  static UCesiumLoadVectorDocumentFromIonAsyncAction* LoadFromIon(
+      int64 AssetId,
+      const FString& IonAccessToken,
+      const FString& IonAssetEndpointUrl = "https://api.cesium.com/");
```
kring: This should take a CesiumIonServer instead of an endpoint URL. The IonAccessToken should be optional, like it is on Cesium3DTileset. If not provided, the CesiumIonServer's token is used.

Source/CesiumRuntime/Public/CesiumGeoJsonDocument.h
```diff
@@ -0,0 +1,118 @@
...
+
+DECLARE_DYNAMIC_MULTICAST_DELEGATE_TwoParams(
+    FCesiumVectorDocumentIonLoadDelegate,
+    bool,
+    Success,
+    FCesiumGeoJsonDocument,
+    Document);
+
+UCLASS()
+class CESIUMRUNTIME_API UCesiumLoadVectorDocumentFromIonAsyncAction
```
kring: I'd like to also see a version of this that can load from an arbitrary URL.

Source/CesiumRuntime/Public/CesiumGeoJsonObject.h
```diff
@@ -0,0 +1,288 @@
...
+
+  /**
+   * @brief Returns the ID of the provided feature, or an empty string if no
+   * ID was present. If the ID is an integer, it will be converted to a string.
+   */
+  UFUNCTION(
+      BlueprintCallable,
+      BlueprintPure,
+      Category = "Cesium|Vector|Feature")
+  static FString GetIdAsString(const FCesiumGeoJsonFeature& InFeature);
```
kring, line 148: Let's add a function to determine which ID type it is. `IsIdInteger` or somesuch.

Source/CesiumRuntime/Public/CesiumGeoJsonObject.h
```diff
@@ -0,0 +1,288 @@
...
+ * @brief A Blueprint Funciton Library for interacting with `FCesiumVectorNode`
+ * values.
+ */
+UCLASS()
+class UCesiumGeoJsonObjectBlueprintLibrary : public UBlueprintFunctionLibrary {
+  GENERATED_BODY()
+
+public:
+  UFUNCTION(BlueprintCallable, BlueprintPure, Category = "Cesium|Vector|Object")
+  static bool IsValid(const FCesiumGeoJsonObject& InObject);
```
kring, line 275: Add doc for this function and the others in this class.

Source/CesiumRuntime/Public/CesiumGeoJsonObject.h
```diff
@@ -0,0 +1,288 @@
...
+      Meta = (ExpandEnumAsExecs = "Branches"))
+  static FBox
+  GetBoundingBox(const FCesiumGeoJsonObject& InObject, EHasValue& Branches);
+
+  UFUNCTION(BlueprintCallable, BlueprintPure, Category = "Cesium|Vector|Object")
+  static FJsonObjectWrapper
+  GetForeignMembers(const FCesiumGeoJsonObject& InObject);
+
+  UFUNCTION(BlueprintCallable, BlueprintPure, Category = "Cesium|Vector|Object")
+  static FVector GetObjectAsPoint(const FCesiumGeoJsonObject& InObject);
```
kring, line 311: Can we provide a way to enumerate over all the points in the document? I'm not sure if that would go here (and return a TArray), or (probably better) look more like the CesiumGeoJsonObjectIterator.

Source/CesiumRuntime/Public/CesiumGeoJsonObjectIterator.h
```diff
@@ -0,0 +1,58 @@
...
+
+UCLASS()
+class UCesiumGeoJsonObjectIteratorFunctionLibrary
+    : public UBlueprintFunctionLibrary {
+  GENERATED_BODY()
+
+public:
+  UFUNCTION(BlueprintCallable, Category = "Cesium|Vector|Object")
+  static FCesiumGeoJsonObjectIterator
+  Iterate(const FCesiumGeoJsonObject& Object);
```
kring, line 43: Add doc.

Source/CesiumRuntime/Public/CesiumGeoJsonDocument.h
```diff
@@ -0,0 +1,118 @@
...
+  /**
+   * @brief Creates a `FCesiumGeoJsonDocument` wrapping the provided
+   * `CesiumVectorData::GeoJsonDocument`.
+   */
+  FCesiumGeoJsonDocument(
+      std::shared_ptr<CesiumVectorData::GeoJsonDocument>&& document)
+      : _document(std::move(document)) {}
+
+private:
+  std::shared_ptr<CesiumVectorData::GeoJsonDocument> _document;
```
kring:
```suggestion
  std::shared_ptr<CesiumVectorData::GeoJsonDocument> _pDocument;
```

Source/CesiumRuntime/Private/CesiumGeoJsonObject.cpp
```diff
@@ -0,0 +1,412 @@
...
+    }
+    TSharedPtr<FJsonValue> operator()(const double& value) {
+      return MakeShared<FJsonValueNumber>(value);
+    }
+    TSharedPtr<FJsonValue> operator()(const std::uint64_t& value) {
+      return MakeShared<FJsonValueNumberString>(FString::FromInt(value));
+    }
+    TSharedPtr<FJsonValue> operator()(const std::int64_t& value) {
+      return MakeShared<FJsonValueNumberString>(FString::FromInt(value));
+    }
```
kring: The vast majority of integers would be better represented losslessly as a double. The `losslessNarrow` function can be used to separate the ones that can become doubles from the ones that must be strings.

Source/CesiumRuntime/Private/CesiumGeoJsonObject.cpp
```diff
@@ -0,0 +1,412 @@
...
+      }
+      return MakeShared<FJsonValueArray>(MoveTemp(values));
+    }
+    TSharedPtr<FJsonValue>
+    operator()(const CesiumUtility::JsonValue::Object& value) {
+      TSharedPtr<FJsonObject> obj = MakeShared<FJsonObject>();
+      for (const auto& [k, v] : value) {
+        obj->SetField(UTF8_TO_TCHAR(k.c_str()), jsonValueToUnrealJsonValue(v));
+      }
+      return MakeShared<FJsonValueObject>(MoveTemp(obj));
```
kring, line 116: This looks redundant.

```suggestion
      return obj;
```

Source/CesiumRuntime/Private/CesiumGeoJsonObject.cpp
```diff
@@ -0,0 +1,412 @@
...
+  }
+  TSharedPtr<FJsonObject> object = MakeShared<FJsonObject>();
+  if (InFeature._feature->properties) {
+    for (const auto& [k, v] : *InFeature._feature->properties) {
+      object->SetField(UTF8_TO_TCHAR(k.c_str()), jsonValueToUnrealJsonValue(v));
+    }
+  }
+
+  FJsonObjectWrapper wrapper;
+  wrapper.JsonObject = object;
```
kring: Minor optimization.
```suggestion
  wrapper.JsonObject = MoveTemp(object);
```

Source/CesiumRuntime/Public/CesiumGeoJsonObject.h
```diff
@@ -0,0 +1,288 @@
...
+   * @brief Creates a new `FCesiumGeoJsonFeature` wrapping the provided
+   * `CesiumVectorData::GeoJsonFeature`.
+   */
+  FCesiumGeoJsonFeature(
+      const std::shared_ptr<CesiumVectorData::GeoJsonDocument>& document,
+      const CesiumVectorData::GeoJsonFeature* feature)
+      : _document(document), _feature(feature) {}
+
+private:
+  std::shared_ptr<CesiumVectorData::GeoJsonDocument> _document;
```
kring:
```suggestion
  std::shared_ptr<CesiumVectorData::GeoJsonDocument> _pDocument;
```

Source/CesiumRuntime/Public/CesiumGeoJsonObject.h
```diff
@@ -0,0 +1,288 @@
...
+  const CesiumVectorData::GeoJsonFeature* _feature;
```
kring:
```suggestion
  const CesiumVectorData::GeoJsonFeature* _pFeature;
```

Source/CesiumRuntime/Private/CesiumGeoJsonObject.cpp
```diff
@@ -0,0 +1,412 @@
...
+  return InObject._pDocument != nullptr && InObject._pObject != nullptr;
+}
+
+ECesiumGeoJsonObjectType UCesiumGeoJsonObjectBlueprintLibrary::GetObjectType(
+    const FCesiumGeoJsonObject& InObject) {
+  if (!InObject._pDocument || !InObject._pObject) {
+    return {};
+  }
+
+  return (ECesiumGeoJsonObjectType)InObject._pObject->getType();
```
kring, line 198: Given this conversion via cast, it would add a bit of safety to add static assertions that the enum values are equivalent. That way a change on either side (other than adding a new enum value at the end) would be flagged as a compiler error.

Source/CesiumRuntime/Private/CesiumGeoJsonObject.cpp
```diff
@@ -0,0 +1,412 @@
...
+    return {};
+  }
+
+  TSharedPtr<FJsonObject> object = MakeShared<FJsonObject>();
+  for (const auto& [k, v] : InObject._pObject->getForeignMembers()) {
+    object->SetField(UTF8_TO_TCHAR(k.c_str()), jsonValueToUnrealJsonValue(v));
+  }
+
+  FJsonObjectWrapper wrapper;
+  wrapper.JsonObject = object;
```
kring:
```suggestion
  wrapper.JsonObject = MoveTemp(object);
```

Source/CesiumRuntime/Private/CesiumGeoJsonObject.cpp
```diff
@@ -0,0 +1,412 @@
...
+  return wrapper;
+}
+
+namespace {
+FVector degreesToVector(const glm::dvec3& coordinates) {
+  return FVector(coordinates.x, coordinates.y, coordinates.z);
+}
+} // namespace
```
kring: Maybe this function used to do more, but at this point I'd suggest just replacing it with `VecMath::createVector`.

Source/CesiumRuntime/Public/CesiumGeoJsonObjectIterator.h
```diff
@@ -0,0 +1,58 @@
...
+#include "CesiumGeoJsonObjectIterator.generated.h"
+
+USTRUCT(BlueprintType)
+struct FCesiumGeoJsonObjectIterator {
+  GENERATED_BODY()
+
+  FCesiumGeoJsonObjectIterator() : _object(), _iterator() {}
+
+  FCesiumGeoJsonObjectIterator(const FCesiumGeoJsonObject& object)
+      : _object(object), _iterator(*_object.getObject()) {}
```
kring: Move implementation to .cpp file.

Content/VectorIteratorMacros.uasset
kring, line 1: Do you think it could make sense to make "Loop Over Iterator" take a GeoJsonObject instead of an iterator? It would save a seemingly-unnecessary step in the Blueprint:
image

The macro would probably benefit from a rename at that point, too. "Iterate Over GeoJSON Objects" maybe? 

And this filename is a bit generic, too. `CesiumVectorIteratorMacros` maybe? Assuming we may want to put other iteration macros in there eventually; it's not specific to GeoJSON.

Source/CesiumRuntime/Public/CesiumGeoJsonDocument.h
kring: I just realized we don't have a function like this in cesium-native, either. If there's not a good reason for that, we should add it there, too.

kring (commented):
Thanks @azrogers! Lots of small comments below, but this looks really good!

Source/CesiumRuntime/Private/Tests/CesiumVectorDocument.spec.cpp
```diff
@@ -0,0 +1,157 @@
...
+#include "Misc/AutomationTest.h"
+
+BEGIN_DEFINE_SPEC(
+    FCesiumVectorDocumentSpec,
+    "Cesium.Unit.CesiumVectorDocument",
+    EAutomationTestFlags::EditorContext | EAutomationTestFlags::ClientContext |
+        EAutomationTestFlags::ProductFilter | EAutomationTestFlags::NonNullRHI)
+END_DEFINE_SPEC(FCesiumVectorDocumentSpec)
+
+void FCesiumVectorDocumentSpec::Define() {
```
j9liu: Change instances of `VectorDocument` -> `GeoJsonDocument` throughout this file

Source/CesiumRuntime/Public/CesiumGeoJsonDocument.h
```diff
@@ -0,0 +1,118 @@
...
+ * a `FCesiumVectorDocument`.
+ */
+UCLASS()
+class UCesiumGeoJsonDocumentBlueprintLibrary
+    : public UBlueprintFunctionLibrary {
+  GENERATED_BODY()
+
+public:
+  /**
+   * @brief Attempts to load a `FCesiumVectorDocument` from a string containing
```
j9liu: There's more instances of "Vector" instead of GeoJSON; I'd do a find & replace throughout the docs

Source/CesiumRuntime/Public/CesiumGeoJsonObject.h
```diff
@@ -0,0 +1,288 @@
+// Copyright 2020-2025 CesiumGS, Inc. and Contributors
+
+#pragma once
+
+#include "CesiumUtility/IntrusivePointer.h"
+#include "CesiumVectorData/GeoJsonDocument.h"
+#include "CesiumVectorData/GeoJsonObject.h"
```
j9liu, line 7: angle brackets?

Source/CesiumRuntime/Private/CesiumGeoJsonDocument.cpp
```diff
@@ -0,0 +1,94 @@
+#include "CesiumGeoJsonDocument.h"
+#include "CesiumRuntime.h"
+
+#include "CesiumUtility/Result.h"
```
j9liu:
```suggestion
#include <CesiumUtility/Result.h>
```

Source/CesiumRuntime/Private/CesiumGeoJsonObject.cpp
```diff
@@ -0,0 +1,412 @@
+#include "CesiumGeoJsonObject.h"
+
+#include "CesiumGeospatial/Cartographic.h"
```
j9liu, line 3:
```suggestion
#include <CesiumGeospatial/Cartographic.h>
```

Source/CesiumRuntime/Public/CesiumGeoJsonDocument.h
```diff
@@ -0,0 +1,118 @@
+// Copyright 2020-2025 CesiumGS, Inc. and Contributors
+
+#pragma once
+
+#include "CesiumGeoJsonObject.h"
+#include "CesiumUtility/IntrusivePointer.h"
+#include "CesiumVectorData/GeoJsonDocument.h"
```
j9liu, line 7:
```suggestion
#include <CesiumUtility/IntrusivePointer.h>
#include <CesiumVectorData/GeoJsonDocument.h>
```

Source/CesiumRuntime/Private/CesiumGeoJsonObject.cpp
azrogers: For whatever reason, `FJsonValueObject` and `FJsonObject` are distinct types and we seem to need to manually convert one to the other like this.

Source/CesiumRuntime/Public/CesiumGeoJsonObject.h
azrogers: I don't think we've decided on what we're doing about this in Unreal. The Cesium Native style is de facto to use angle brackets (though we've yet to write it down anywhere), but our Unreal code uses quotes a lot more than angle brackets, and all the code from Epic uses quotes instead of angle brackets.

Source/CesiumRuntime/Private/CesiumGeoJsonObject.cpp
azrogers: I'm not sure what you mean. How would we assert that the enums are the same?

kring: You can find an example of it here:
https://github.com/CesiumGS/cesium-unreal/blob/v2.16.1/Source/CesiumRuntime/Private/Cesium3DTileset.cpp#L1015

azrogers: Ah, I see. I've added it here.

azrogers:
@kring @j9liu I think that's all the review comments taken care of, minus the `#include` question (are we committing to angle bracket includes in Unreal)?

kring:
Looks like we still have some VectorDocuments:
image

kring:
There's a lot of "vector document" in a bunch of other places as well, including filenames and parameter names (which matter because they're shown as output in names in Blueprint). Can I bug you to do a pass over the code to replace all those occurrences, and then I'll look for any that might have been missed while I'm reviewing?

azrogers:
@kring I believe I've caught all of them, let me know if you find any others.

kring:
Thanks @azrogers!

PR #1652, opened 2025-04-09
Proof of concept for multiple views

kring:
This is some extremely hacky code that demonstrates how the new multiple view support in CesiumGS/cesium-native#1125 can be used in Unreal. Don't even think about merging it, but it might be fun to play with.

In this branch, you can create multiple editor viewports, and each will select and render its own set of tiles:
image

Notice that some of the viewports are near the city center, so they have high detail, and yet the zoomed out view in the lower-left doesn't show excessive detail. Each view is selecting its own level-of-detail.

To set this up, press the button in the top-right corner of the viewport to show four viewports:
image

Then switch one or more viewports to perspective:
image

And turn on real-time mode in that viewport:
image

If you click on the CesiumCameraManager Actor, you'll see a View Groups property. It's automatically populated from all the Editor viewports, and you can set the "Weight" of each, which controls how much tile loading that view gets to do compared to the others.

Play-in-Editor mode won't work at all in this branch, so just stick to Editor viewports.


kring:
I've updated this branch to support orthographic Editor viewports, too, using the cesium-native change in CesiumGS/cesium-native#1135.

Unreal orthographic views seem really inflexible, though. It doesn't support arbitrary orthographic views, only top, bottom, left, right, front, and back. The "top" view is useful, but all the other directions are pretty unnatural for viewing geospatial scenes, especially because there doesn't seem to be control over the near and far distances.

image

kring:
If you try the orthographic views, note that they default to "Wireframe" and that the "Lit" mode doesn't work, seemingly due to a lack of HDR support in the orthographic view. Setting the viewport to "Unlit" is a good choice.

j9liu:

PR #1657, opened 2025-04-14
Fix KTX-Software submodule not finding bash when git-for-windows installed in non-default path

### LINKED ISSUE #1656: Change in KTX-Software overlay prevents finding bash when git-for-windows is installed in a non-default path
Symptom is a build error when ezvcpkg tries to install KTX-Software using the overlay port:
```
*** Content of log file: C:\(...)\.ezvcpkg\2024.11.16\buildtrees\ktx\config-x64-windows-unreal-dbg-err.log:
CMake Error at C:/Dev/CMake/share/cmake-3.31/Modules/FindPackageHandleStandardArgs.cmake:233 (message):
  Could NOT find Bash (missing: BASH_EXECUTABLE)

  Hint: The project() command has not yet been called.  It sets up
  system-specific search paths.
Call Stack (most recent call first):
  C:/Dev/CMake/share/cmake-3.31/Modules/FindPackageHandleStandardArgs.cmake:603 (_FPHSA_FAILURE_MESSAGE)
  cmake/modules/FindBash.cmake:96 (find_package_handle_standard_args)
  CMakeLists.txt:11 (find_package)
```

I do have git/bash installed (from gitforwindows.org) but in `C:\Dev\Git` instead of the default `C:\Program Files\Git` path. And indeed KTX's `FindBash.cmake` looks for the bash executable like this:
```
# First, look if GIT bash is installed
find_program (
    BASH_EXECUTABLE
    bash
PATHS
    # Additional paths for Windows
    "C:\\Program Files\\Git\\bin"
NO_SYSTEM_ENVIRONMENT_PATH
)

if(NOT BASH_EXECUTABLE)
  # Fallback search in default locations
  find_program (
      BASH_EXECUTABLE
      bash
  )
endif()
```
The first `find_program` fails because of Git's non-default installation path, the second because the default and recommended option when installing git-for-windows is _not_ to put bash and other executables in the `PATH` as that would override Windows executables in a Command prompt (only a minimal set of git wrappers from `your_install_dir\Git\cmd` is added).

I guess this is why the vcpkg port's maintainer chose to fetch msys2's bash in the portfile, which @kring removed in https://github.com/CesiumGS/cesium-unreal/commit/4c5c5488d256ab37ba759450f34775b709ce35c0 because the links were broken at the time.

Even though it is ultimately a bug in FindBash.cmake IMO, for more robustness and developer-friendliness I would like to revert the commit (and updating the links), even though I agree it is annoying to have to maintain hardcoded tool versions and SHA's in portfile.

GhisBntly:
Fixes #1656, see detailed problem and fix descriptions there.


kring:
Thanks @GhisBntly!

PR #1668, opened 2025-05-05
Use custom physics state creation to avoid Chaos scale complaints.

### LINKED ISSUE #1659: Picking issues with tilesets representing "small" models (and massive logging)
### What happened?

Hello,
Since a recent version (we experienced it with 2.14.1, but are unsure when the change was integrated exactly), we encounter some picking issues and massive warnings in logs with tilesets representing a "small" geometry (a building, typically).

The warning that we get in UE looks like this:

`LogPhysics: Warning: Initialising Body : Scale3D is (nearly) zero: Actor: 'xxx' Component: 'xxxx' `

And when this happens, it is no longer possible to pick the corresponding primitives (using UE generic trace methods such as UWorld::**LineTraceMultiByObjectType**).

After comparing with a previous working version, we found out that it was caused by the scaling by CesiumPrimitiveData::**positionScaleFactor** (currently defined as 1024.0), which scales up the internal coordinates but (logically ;-)) scales down the primitive's transformation, which then becomes "too small" for UE, which then discards the collision mesh (or at least makes the tracing miss some impacts).

Here is a link to a commit dealing with this scaling factor:
https://github.com/CesiumGS/cesium-unreal/commit/22769c2

One possible workaround could consist in making this scaling factor depend on the tileset's dimensions, but I guess @kring would have more context about the issue that made you introduce this factor (it would be bad to get a regression on the cases this factor was fixing...)

### Environment

[Cesium](url) for Unreal version: 2.14.1
Unreal Engine Version: 5.3
Operating System: Windows 11 Enterprise

### Reproduction steps

1. As explained, it depends on the dimensions of the model (I have a few examples coming from iTwin)
2.
3.
...

### Supporting evidence

_No response_

kring:
Fixes #1659 

Chaos gets persnickety about models with scales outside of its rather narrow expected range. When a model has a scale smaller than 1e-4, Chaos complains loudly (`LogPhysics: Warning: Initialising Body : Scale3D is (nearly) zero`) and resets the scale to an incorrect value. I believe this to be an Unreal bug, and have reported it to Epic, but even if they fix it quickly that won't help with older versions of UE.

This comes up a lot because, to work around a _different_ Chaos scale issue, we currently multiply all vertex positions by 1024.0 and then add in a 1/1024.0 scaling to return the model to its normal size (see #1465). So that gets us pretty close to the 1e-4 scaling size limitation. Fortunately, it's also mitigated by the fact that glTF uses meters and UE uses centimeters, so we also scale _up_ by a factor of 100.0. This means everything works pretty well for models where the internal units are meters, as they often are. But a model that internally uses millimeters, for example, is likely to have a spammy log and broken physics.

This PR works around the UE bug by overriding `UPrimitiveComponent::OnCreatePhysicsState`. When it detects a problematic scale, it nullifies that scale (replaces it with 1.0), calls the normal `FBodyInstance::InitBody`, and then fixes the scale afterward. Fortunately, UE is fine with setting whatever scale you like after the body instance is initialized; it only complains when it _initially_ has a small scale.

Draft because I'd like to write a test or two for this.

This PR does _not_ handle instanced meshes. I started trying to implement it in this branch:
https://github.com/CesiumGS/cesium-unreal/tree/physics-scale-instancing

I don't have an instanced tileset that has this problem, so I artificially created it by setting `positionScaleFactor` to a large value (2^32). That makes the instance positions much larger and cancels that out with a small scale, triggering the bug. But unfortunately I wasn't able to get this to work with instances. Something seems to be going wrong with updating the body instances with the correct scale, because collision with the instances is very broken in that branch. The camera collides with invisible geometry, and we see lots of log messages like this:

> LogPhysics: Warning: (Invalid Normal from ConvertQueryImpactHit) Non-normalized OutResult.Normal from hit conversion: X=0.000 Y=0.000 Z=0.000 (Component- CesiumGltfPrimitiveComponent /Game/CesiumSamples/Maps/06/UEDPIE_0_06_CesiumMetadata.06_CesiumMetadata:PersistentLevel.Cesium3DTileset_1.CesiumGltfComponent_235.CesiumGltfPrimitiveComponent_0)
LogPhysics: Warning: Start Loc(X=-2038.967 Y=4593.831 Z=-1804.676), End Loc(X=-2053.065 Y=4598.326 Z=-1804.676), Hit Loc(X=-2039.846 Y=4594.112 Z=-1804.676), ImpactNormal(X=0.000 Y=0.000 Z=0.000) NormalPreSafeNormalize(X=0.000 Y=0.000 Z=0.000)

I used `Design Tiler - Snowdon Towers Sample Architectural Instanced`, Cesium ion asset ID 2707515 from the CesiumGS account, to test it.

CC @Julot94


kring:
I wrote a test and this should be ready to go now.

Julot94:
btw, I can confirm that it has resolved our issues (both the warnings and the picking issues with such models).
Thank you @kring

Source/CesiumRuntime/Private/Tests/Cesium3DTileset.spec.cpp
```diff
@@ -18,6 +18,10 @@
 #include <Cesium3DTilesSelection/TilesetSharedAssetSystem.h>
 #include <CesiumAsync/ICacheDatabase.h>
 
+THIRD_PARTY_INCLUDES_START
```
j9liu, line 21: Just curious what this macro is used for? (Why does it apply to `Uri.h` but not the other includes form cesium-native here?

j9liu (approved):
Thanks @kring! Changes look fine to me, I haven't noticed any different behavior in our sample levels either. Just had one question but it doesn't have to hold up this PR.

Were you going to address the instanced physics issues in a different PR?

Source/CesiumRuntime/Private/Tests/Cesium3DTileset.spec.cpp
kring: Arguably we should wrap all the cesium-native headers in that. But it's necessary here because one of the ada-url headers triggers warnings (which become errors) in Unreal builds.

kring:
> Were you going to address the instanced physics issues in a different PR?

I'm currently not planning to address it at all. It's difficult to fix, I'm not aware of any real-world case where it comes up, and with any luck it'll be fixed in Unreal Engine itself soon anyway.

I just wrote an issue for it though: https://github.com/CesiumGS/cesium-unreal/issues/1671

j9liu:
Just waiting on CI to pass and then I'll merge. Thanks!

Issue #1658, opened 2025-04-14
Extension points for mesh- and material-related customizations


GhisBntly:
### Feature

This is a proposal for API additions needed to implement mesh- and material-related customizations beyond the current cesium-unreal capabilities.

## Our use case

We need to process the rendered cesium tiles with advanced materials that cannot be done with the current capabilities, and also because of a limitation of Unreal Engine shaders:
  - Our material shader needs to test an ID on each vertex to determine the shading: the ID is an unsigned 64bits integer, which Unreal shaders do not support, and we cannot simply cast to 32 bits because high-order bits contain relevant information.
  - Other external data not present in the tileset's metadata also need to be correlated with the ID to determine what shading characteristics to apply at any given time, which adds specific requirements:
    - The custom shading may need to change the material Blend Mode to Translucent
    - The custom shading may need the "total" bounding box in the scene (as far as the currently loaded tiles allow to determine) for each ID (which identifies a consistent sub-mesh). The rough estimate given by the first loaded and thus coarser tiles is usually sufficient.
      - (@me/Bentley) Is it really important? Including those as extra metadata in the generated tileset is probably preferrable, ultimately.

Given these requirements, this is the solution we developed, inspired from metadata-based material customization in cesium-unreal:
  - When tiles are loaded and Unreal meshes are built, we inspect each vertex to map its 64bits ID (found in the glTF tiles' metadata) to its 32bits FeatureID
  - We may need to force the use of `BaseTranslucent` instead of `BaseOpaque` as parent of the UDynamicMaterialInstance created by cesium-unreal
  - For each tile, we create textures which will be indexed by FeatureIDs (baked into the mesh UVs) in our custom material to apply the desired shading according to the external (non-tileset) data mentioned.
  - Managing the per-tile information for meshes, materials, textures and ID vs. FeatureIDs mappings means we need some basic notifications on tiles lifecycle: when a tile has finished loading, when its visibility changes, and when it is about to be unloaded.

## Implementation in our cesium-unreal's fork

Translating to the cesium-unreal APIs, those are extension points we added (temp names from our forked implementation):
1. In `loadPrimitive`, `BakeFeatureIDsInVertexUVs` = hook allowing API user to enforce the baking of FeatureIDs into a UV layer, without requiring a CesiumFeaturesMetadataComponent
   - The rationale is unclear, I guess we wanted to avoid creating a component because we didn't need any actual encoding of metadata, we only need to access the FeatureIDs in the UVs.
[...]

PR #1669, opened 2025-05-09
[enhancement#1658] mesh, material and tile loading callbacks

GhisBntly:
Proposed implementation for https://github.com/CesiumGS/cesium-unreal/issues/1658

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,8 @@
 
 - Added a Cesium -> Geocoder -> Geocode Blueprint function, making it easy to query the Cesium ion geocoder.
 - Added `UCesiumMetadataPickingBlueprintLibrary::FindPropertyTableProperty` to search for a `FCesiumPropertyTableProperty` by name on a given `UPrimitiveComponent`.
+- Added a property `bAllowMeshBuffersCPUAccess` to `ACesium3DTileset` actors to keep the buffers of Unreal meshes created by the tileset in CPU memory, in order to have access to them in-game. Defaults to false.
+- Added the class `CesiumMeshBuildCallbacks`: when an implementation is registered on a tileset (with `ACesium3DTileset::SetMeshBuildCallbacks`), its functions will be called at various points in a tile's lifecycle, like when a mesh component is created, when a material is instanced, when the tile changes visibility, when it is unloaded, etc.
 
 ##### Fixes :wrench:
 
--- a/Source/CesiumRuntime/Private/CesiumGltfComponent.cpp
+++ b/Source/CesiumRuntime/Private/CesiumGltfComponent.cpp
@@ -9,6 +9,7 @@
 #include "CesiumGltfPrimitiveComponent.h"
 #include "CesiumGltfTextures.h"
 #include "CesiumMaterialUserData.h"
+#include "CesiumMeshBuildCallbacks.h"
 #include "CesiumRasterOverlays.h"
 #include "CesiumRuntime.h"
 #include "CesiumTextureUtility.h"
@@ -1253,6 +1254,9 @@ static void loadPrimitive(
     }
   }
 
+  primitiveResult.MeshBuildCallbacks =
+      options.pMeshOptions->pNodeOptions->pModelOptions->MeshBuildCallbacks;
+
   auto normalAccessorIt = primitive.attributes.find("NORMAL");
   CesiumGltf::AccessorView<TMeshVector3> normalAccessor;
   bool hasNormals = false;
@@ -1668,6 +1672,12 @@ static void loadPrimitive(
     computeTangentSpace(StaticMeshBuildVertices);
   }
 
+  // For iTwin scene mapping mechanism (used both for Synchro 4D schedules and
+  // selection highlight), we need to access vertex data from the CPU (in
+  // packaged game, if we don't set this flag, the data can become inaccessible
+  // at any time...)
+  const bool& bNeedsCPUAccess = options.pMeshOptions->pNodeOptions->pModelOptions->allowMeshBuffersCPUAccess;
+
   {
     TRACE_CPUPROFILER_EVENT_SCOPE(Cesium::InitBuffers)
 
@@ -1679,12 +1689,12 @@ static void loadPrimitive(
 
     LODResources.VertexBuffers.PositionVertexBuffer.Init(
         StaticMeshBuildVertices,
-        false);
+        bNeedsCPUAccess);
 
     FColorVertexBuffer& ColorVertexBuffer =
         LODResources.VertexBuffers.ColorVertexBuffer;
     if (hasVertexColors) {
-      ColorVertexBuffer.Init(StaticMeshBuildVertices, false);
+      ColorVertexBuffer.Init(StaticMeshBuildVertices, bNeedsCPUAccess);
     }
 
     uint32 numberOfTextureCoordinates =
@@ -1697,7 +1707,7 @@ static void loadPrimitive(
     vertexBuffer.Init(
         StaticMeshBuildVertices.Num(),
         numberOfTextureCoordinates,
-        false);
+        bNeedsCPUAccess);
 
     // Manually copy the vertices into the buffer. We do this because UE 5.3
     // and 5.4 have a bug where the overload of `FStaticMeshVertexBuffer::Init`
@@ -1744,6 +1754,9 @@ static void loadPrimitive(
 
   {
     TRACE_CPUPROFILER_EVENT_SCOPE(Cesium::SetIndices)
+    if (bNeedsCPUAccess) {
+      LODResources.IndexBuffer.TrySetAllowCPUAccess(true);
+    }
     LODResources.IndexBuffer.SetIndices(
         indices,
         StaticMeshBuildVertices.Num() >= std::numeric_limits<uint16>::max()
@@ -2452,7 +2465,8 @@ static void SetGltfParameterValues(
     const CesiumGltf::MaterialPBRMetallicRoughness& pbr,
     UMaterialInstanceDynamic* pMaterial,
     EMaterialParameterAssociation association,
-    int32 index) {
+    int32 index,
+    CesiumMeshBuildCallbacks const* meshBuildCallbacks) {
   for (auto& textureCoordinateSet : loadResult.textureCoordinateParameters) {
     pMaterial->SetScalarParameterValueByInfo(
         FMaterialParameterInfo(
@@ -2670,6 +2684,11 @@ static void SetGltfParameterValues(
     pMaterial->SetVectorParameterValueByInfo(
         FMaterialParameterInfo("emissiveFactor", association, index),
         FVector(1.0f, 1.0f, 1.0f));
+  }
+
+  // Extra material customizations
+  if (meshBuildCallbacks) {
+    meshBuildCallbacks->CustomizeGltfMaterial(material, pbr, pMaterial, association, index);
   }
 }
 
@@ -3130,14 +3149,34 @@ static void loadPrimitiveGameThreadPart(
   }
 #endif
 
-  UMaterialInstanceDynamic* pMaterial;
+  // Move this right now: CreateMaterial may need them!
+  // "Safe" even though loadResult is still used later, because the methods used
+  // during material setup (SetGltfParameterValues, etc.) below do not use these
+  // members.
+  primData.Features = std::move(loadResult.Features);
+  primData.Metadata = std::move(loadResult.Metadata);
+
+  UMaterialInstanceDynamic* pMaterial = nullptr;
+  TSharedPtr<CesiumMeshBuildCallbacks> MeshBuildCallbacks =
+      loadResult.MeshBuildCallbacks.Pin();
   {
     TRACE_CPUPROFILER_EVENT_SCOPE(Cesium::SetupMaterial)
-
-    pMaterial = UMaterialInstanceDynamic::Create(
-        pBaseMaterial,
-        nullptr,
-        ImportedSlotName);
+    ensure(pBaseMaterial);
+    if (MeshBuildCallbacks) {
+      // Possibility to override the material for this primitive
+      pMaterial = MeshBuildCallbacks->CreateMaterial(
+          *pCesiumPrimitive,
+          pBaseMaterial,
+          nullptr,
+          ImportedSlotName);
+    }
+    ensure(pBaseMaterial);
+    if (!pMaterial) {
+      pMaterial = UMaterialInstanceDynamic::Create(
+          pBaseMaterial,
+          nullptr,
+          ImportedSlotName);
+    }
 
     pMaterial->SetFlags(
         RF_Transient | RF_DuplicateTransient | RF_TextExportTransient);
@@ -3148,7 +3187,8 @@ static void loadPrimitiveGameThreadPart(
         pbr,
         pMaterial,
         EMaterialParameterAssociation::GlobalParameter,
-        INDEX_NONE);
+        INDEX_NONE,
+        MeshBuildCallbacks.Get());
     SetWaterParameterValues(
         model,
         loadResult,
@@ -3195,7 +3235,8 @@ static void loadPrimitiveGameThreadPart(
           pbr,
           pMaterial,
           EMaterialParameterAssociation::LayerParameter,
-          0);
+          0,
+          MeshBuildCallbacks.Get());
 
       // Initialize fade uniform to fully visible, in case LOD transitions
       // are off.
@@ -3249,9 +3290,6 @@ static void loadPrimitiveGameThreadPart(
       }
     }
   }
-
-  primData.Features = std::move(loadResult.Features);
-  primData.Metadata = std::move(loadResult.Metadata);
 
   primData.EncodedFeatures = std::move(loadResult.EncodedFeatures);
   primData.EncodedMetadata = std::move(loadResult.EncodedMetadata);
@@ -3328,6 +3366,11 @@ static void loadPrimitiveGameThreadPart(
     TRACE_CPUPROFILER_EVENT_SCOPE(Cesium::RegisterComponent)
     pMesh->RegisterComponent();
   }
+
+  // Call the observer callback (if any) once all is done
+  if (MeshBuildCallbacks) {
+    MeshBuildCallbacks->OnMeshConstructed(*pGltf, *pCesiumPrimitive);
+  }
 }
 
 /*static*/ CesiumAsync::Future<UCesiumGltfComponent::CreateOffGameThreadResult>
@@ -3352,7 +3395,7 @@ UCesiumGltfComponent::CreateOffGameThread(
     UMaterialInterface* pBaseTranslucentMaterial,
     UMaterialInterface* pBaseWaterMaterial,
     FCustomDepthParameters CustomDepthParameters,
-    const Cesium3DTilesSelection::Tile& tile,
+    Cesium3DTilesSelection::Tile& tile,
     bool createNavCollision) {
   TRACE_CPUPROFILER_EVENT_SCOPE(Cesium::LoadModel)
 
@@ -3366,6 +3409,7 @@ UCesiumGltfComponent::CreateOffGameThread(
   // }
 
   UCesiumGltfComponent* Gltf = NewObject<UCesiumGltfComponent>(pTilesetActor);
+  Gltf->pTile = &tile;
   Gltf->SetMobility(pTilesetActor->GetRootComponent()->Mobility);
   Gltf->SetFlags(RF_Transient | RF_DuplicateTransient | RF_TextExportTransient);
 
@@ -3394,10 +3438,12 @@ UCesiumGltfComponent::CreateOffGameThread(
     encodeMetadataGameThreadPart(*Gltf->EncodedMetadata_DEPRECATED);
   }
 
+  LoadGltfResult::LoadedPrimitiveResult* pAnyPrimResult = nullptr;
   for (LoadedNodeResult& node : pReal->loadModelResult.nodeResults) {
     if (node.meshResult) {
       for (LoadedPrimitiveResult& primitive :
            node.meshResult->primitiveResults) {
+        pAnyPrimResult = &primitive;
         loadPrimitiveGameThreadPart(
             model,
             Gltf,
@@ -3412,9 +3458,26 @@ UCesiumGltfComponent::CreateOffGameThread(
     }
   }
 
+  if (pAnyPrimResult && pAnyPrimResult->MeshBuildCallbacks.IsValid()) {
+    pAnyPrimResult->MeshBuildCallbacks.Pin()->OnTileConstructed(
+        tile.getTileID());
+    Gltf->VisibilityChangedObserver =
+        [MeshBuildCallbacks = pAnyPrimResult->MeshBuildCallbacks,
+         TileId = tile.getTileID()](bool visible) {
+          if (MeshBuildCallbacks.IsValid())
+            MeshBuildCallbacks.Pin()->OnVisibilityChanged(TileId, visible);
+        };
+  }
+
   Gltf->SetVisibility(false, true);
   Gltf->SetCollisionEnabled(ECollisionEnabled::NoCollision);
   return Gltf;
+}
+
+void UCesiumGltfComponent::OnVisibilityChanged() {
+  USceneComponent::OnVisibilityChanged();
+  if (VisibilityChangedObserver)
+    VisibilityChangedObserver(GetVisibleFlag());
 }
 
 UCesiumGltfComponent::UCesiumGltfComponent() : USceneComponent() {
@@ -3445,6 +3508,20 @@ UCesiumGltfComponent::UCesiumGltfComponent() : USceneComponent() {
   this->Transparent1x1 = ConstructorStatics.Transparent1x1.Object;
 
   PrimaryComponentTick.bCanEverTick = false;
+}
+
+const FCesiumModelMetadata& UCesiumGltfComponent::GetModelMetadata() const {
+  return Metadata;
+}
+
+const Cesium3DTilesSelection::TileID& UCesiumGltfComponent::GetTileID() const {
+  return pTile->getTileID();
+}
+
+void UCesiumGltfComponent::SetRenderReady(bool bToggle) {
+  if (pTile) {
+    pTile->setRenderEngineReadiness(bToggle);
+  }
 }
 
 void UCesiumGltfComponent::UpdateTransformFromCesium(
--- a/Source/CesiumRuntime/Private/CesiumGltfComponent.h
+++ b/Source/CesiumRuntime/Private/CesiumGltfComponent.h
@@ -5,6 +5,7 @@
 #include "Cesium3DTilesSelection/Tile.h"
 #include "Cesium3DTileset.h"
 #include "CesiumEncodedMetadataUtility.h"
+#include "CesiumLoadedTile.h"
 #include "CesiumModelMetadata.h"
 #include "Components/PrimitiveComponent.h"
 #include "Components/SceneComponent.h"
@@ -12,6 +13,7 @@
 #include "CustomDepthParameters.h"
 #include "EncodedFeaturesMetadata.h"
 #include "Interfaces/IHttpRequest.h"
+#include "Templates/Function.h"
 #include <CesiumAsync/SharedFuture.h>
 #include <glm/mat4x4.hpp>
 #include <memory>
@@ -56,7 +58,7 @@ struct FRasterOverlayTile {
 };
 
 UCLASS()
-class UCesiumGltfComponent : public USceneComponent {
+class UCesiumGltfComponent : public USceneComponent, public ICesiumLoadedTile {
   GENERATED_BODY()
 
 public:
@@ -87,7 +89,7 @@ public:
       UMaterialInterface* BaseTranslucentMaterial,
       UMaterialInterface* BaseWaterMaterial,
       FCustomDepthParameters CustomDepthParameters,
-      const Cesium3DTilesSelection::Tile& tile,
+      Cesium3DTilesSelection::Tile& tile,
       bool createNavCollision);
 
   UCesiumGltfComponent();
@@ -103,6 +105,8 @@ public:
 
   UPROPERTY(EditAnywhere, Category = "Rendering")
   FCustomDepthParameters CustomDepthParameters{};
+
+  Cesium3DTilesSelection::Tile* pTile = nullptr;
 
   FCesiumModelMetadata Metadata{};
   EncodedFeaturesMetadata::EncodedModelMetadata EncodedMetadata{};
@@ -131,10 +135,18 @@ public:
   virtual void SetCollisionEnabled(ECollisionEnabled::Type NewType);
 
   virtual void BeginDestroy() override;
+  virtual void OnVisibilityChanged() override;
+
+  // from ICesiumLoadedTile
+  const FCesiumModelMetadata& GetModelMetadata() const override;
+  const Cesium3DTilesSelection::TileID& GetTileID() const override;
+  void SetRenderReady(bool bToggle) override;
 
   void UpdateFade(float fadePercentage, bool fadingIn);
 
 private:
   UPROPERTY()
   UTexture2D* Transparent1x1 = nullptr;
+
+  TFunction<void(bool /*visible*/)> VisibilityChangedObserver;
 };
--- a/Source/CesiumRuntime/Public/Cesium3DTileset.h
+++ b/Source/CesiumRuntime/Public/Cesium3DTileset.h
@@ -37,6 +37,7 @@ class ACesiumCameraManager;
 class UCesiumBoundingVolumePoolComponent;
 class CesiumViewExtension;
 struct FCesiumCamera;
+class CesiumMeshBuildCallbacks;
 
 namespace Cesium3DTilesSelection {
 class Tileset;
@@ -339,6 +340,19 @@ public:
    */
   UFUNCTION(BlueprintCallable, Category = "Cesium")
   void InvalidateResolvedCameraManager();
+
+  /**
+   * Whether to configure the Unreal mesh buffers to allow access from CPU. If
+   * this is false, the buffers can be freed from CPU memory at any time after
+   * they have been moved to GPU memory: set to true if you need to access these
+   * buffers for your game logic.
+   */
+  UPROPERTY(
+      EditAnywhere,
+      BlueprintGetter = GetAllowMeshBuffersCPUAccess,
+      BlueprintSetter = SetAllowMeshBuffersCPUAccess,
+      Category = "Cesium")
+  bool bAllowMeshBuffersCPUAccess = false;
 
   /**
    * The maximum number of pixels of error when rendering this tileset.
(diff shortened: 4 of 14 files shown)


kring (commented):
Thanks for opening this PR @GhisBntly! I have a bunch of small comments, but overall it looks great.

Source/CesiumRuntime/Public/Cesium3DTileset.h
```diff
@@ -340,6 +341,19 @@ class CESIUMRUNTIME_API ACesium3DTileset : public AActor {
...
+   * this is false, the buffers can be freed from CPU memory at any time after
+   * they have been moved to GPU memory: set to true if you need to access these
+   * buffers for your game logic.
+   */
+  UPROPERTY(
+      EditAnywhere,
+      BlueprintGetter = GetAllowMeshBuffersCPUAccess,
+      BlueprintSetter = SetAllowMeshBuffersCPUAccess,
+      Category = "Cesium")
+  bool bAllowMeshBuffersCPUAccess = false;
```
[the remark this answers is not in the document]
GhisBntly: Agreed, it would indeed be much better to use the `CesiumGltf::Model`!

We access the mesh buffers (read-only), to:
1. determine boundings boxes of mesh subsets depending on vertex metadata
2. extract some of these subsets as individual static mesh components, for example when they need to be transformed (translated/rotated) independently from the rest of the original mesh.

When we wrote this code, we didn't realize it meant keeping a copy of data CPU-side (until we got crashes in Shipping builds^^) + we probably were not aware that `CesiumGltf::Model` was still available after the tile was loaded into Unreal.

I will see if I can handle bounding boxes without accessing Unreal buffers. Mesh extraction should not be a concern as I have good hope to switch entirely to our new solution involving glTF "tuning" soon, so that in the end there will be no reason to merge anything extraction-related.

Source/CesiumRuntime/Public/Cesium3DTileset.h
```diff
@@ -1371,6 +1408,9 @@ class CESIUMRUNTIME_API ACesium3DTileset : public AActor {
 
   int32 _tilesetsBeingDestroyed;
 
+  // Optional implementation of the extension callbacks
+  TWeakPtr<CesiumMeshBuildCallbacks> _meshBuildCallbacks;
```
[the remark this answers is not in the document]
GhisBntly: On closer scrutiny, I think the weak ptr was there because the callbacks could be used in a worker thread (in loadPrimitive). With BakeFeatureIDsInVertexUVs gone, it is no longer the case and it would seem I can indeed make the class an UCLASS, use a raw ptr in the tileset, and remove the CreateModelOptions ptr copy and related boilerplate.

Source/CesiumRuntime/Public/CesiumMeshBuildCallbacks.h
```diff
@@ -0,0 +1,104 @@
...
+struct Material;
+struct MaterialPBRMetallicRoughness;
+} // namespace CesiumGltf
+namespace Cesium3DTilesSelection {
+class Tile;
+}
+
+/** Extension points for customizations requiring mesh and tile properties and
+ * lifecycle information. All methods are called from the game thread.*/
+class CESIUMRUNTIME_API CesiumMeshBuildCallbacks {
```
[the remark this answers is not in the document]
GhisBntly: I was also looking for another name, I'm grateful for the native speaker's suggestion ;^^
I also made it match an UINTERFACE and our implem an UOBJECT, I think it will work with regard to lifetime and garbage collection.
I didn't make the methods Blueprint-accessible for the moment, maybe I should, because I think some signatures are forbidden in BPs, so I may have problems with backward compatibility later :-/

Source/CesiumRuntime/Public/CesiumMeshBuildCallbacks.h
```diff
@@ -0,0 +1,104 @@
...
+  virtual void OnMeshConstructed(
+      ICesiumLoadedTile& LoadedTile,
+      ICesiumLoadedTilePrimitive& TilePrim) = 0;
+
+  /**
+   * Called at the end of all static mesh components' construction for a given
+   * tile.
+   */
+  virtual void
+  OnTileConstructed(const Cesium3DTilesSelection::TileID& TileID) = 0;
```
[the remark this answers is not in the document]
GhisBntly: Sure, all methods are now passed either an `ICesiumLoadedTilePrimitive` (which now has a `GetLoadedTile()` method), or an `ICesiumLoadedTile`

Source/CesiumRuntime/Private/CesiumGltfComponent.cpp
```diff
@@ -2452,7 +2465,8 @@ static void SetGltfParameterValues(
     const CesiumGltf::MaterialPBRMetallicRoughness& pbr,
     UMaterialInstanceDynamic* pMaterial,
     EMaterialParameterAssociation association,
-    int32 index) {
+    int32 index,
+    CesiumMeshBuildCallbacks const* meshBuildCallbacks) {
```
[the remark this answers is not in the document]
GhisBntly: I have searched for all such occurrences that I had added and fixed them.

Source/CesiumRuntime/Private/CesiumGltfComponent.h
```diff
@@ -104,6 +106,8 @@ class UCesiumGltfComponent : public USceneComponent {
   UPROPERTY(EditAnywhere, Category = "Rendering")
   FCustomDepthParameters CustomDepthParameters{};
 
+  Cesium3DTilesSelection::Tile* pTile = nullptr;
```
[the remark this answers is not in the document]
kring: Thanks for the explanation @GhisBntly! I'm still a little confused as to why it's necessary to delay, though.

`UpdateTextureRegions` enqueues a command to the renderer thread, that makes sense. But as long as any draw commands for that tile are executed by the render thread _after_ that command enqueued by `UpdateTextureRegions`, I don't think any particular synchronization or delay should be necessary. Certainly this is true when we create textures, at least.

Is it possible something else was happening here? Like the `UpdateTextureRegions` wasn't invoked until after the tile was drawn with the non-updated texture somehow?

Or is there more going on in `UpdateTextureRegions` than I expect?

Source/CesiumRuntime/Public/CesiumMeshBuildCallbacks.h
```diff
@@ -0,0 +1,104 @@
...
+      ICesiumLoadedTilePrimitive& TilePrim,
+      UMaterialInterface*& InOut_pChosenBaseMaterial,
+      UObject* InOuter,
+      FName const& Name) = 0;
+
+  /**
+   * Customize the Unreal material instance, depending on the glTF material
+   * definition.
+   */
+  virtual void CustomizeGltfMaterial(
```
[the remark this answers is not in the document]
kring: How about providing a method to get the `UCesiumMaterialUserData` from a given material from within `CreateMaterial`? That would be more powerful than the current setup, because you could customize other material layers, not just the base / glTF layer.

By the way, for new code, we can more or less ignore the GlobalParameter setting. That's only used for background compatibility with materials created in very old version of Cesium for Unreal. All new materials should be using layers.

Source/CesiumRuntime/Public/CesiumMeshBuildCallbacks.h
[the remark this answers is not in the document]
kring: > to implement the interface in a BP instead of C++ code, is that even possible?

That's usually done with a BlueprintNativeEvent. This video looks decent at a glance, but I didn't watch the whole thing:
https://dev.epicgames.com/community/learning/tutorials/2mB/unreal-engine-blueprintnativeevent

It'll be a fair bit of work to expose all the necessary C++ types to Blueprints. So rather than do that, I think it's ok to just provide a C++ interface for now, and expose separate Blueprint events that can be subscribed to later. Even if we were to do both now, we probably wouldn't want the C++ and Blueprint APIs to be identical.

Source/CesiumRuntime/Public/CesiumMeshBuildCallbacks.h
```diff
@@ -0,0 +1,104 @@
...
+  virtual void
+  OnTileConstructed(const Cesium3DTilesSelection::TileID& TileID) = 0;
+
+  /**
+   * Called when changing the visibility of any UCesiumGltfComponent, ie usually
+   * several times per tile (when the tileset selection leads to showing or
+   * hiding a whole tile).
+   */
+  virtual void OnVisibilityChanged(
+      const Cesium3DTilesSelection::TileID& TileID,
```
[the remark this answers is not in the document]
GhisBntly: Wow, I really didn't expect two tiles to have the same GLB content URI :-/ Wouldn't both tiles load all the same geometries then, or can they differ by their bounding volume, filtering their content spatially?
I find it disturbing that there is no way to identify a tile that is not yet or no longer present in memory.

I removed the TileID from the signatures in this PR anyway, only keeping `GetTileID` in `ICesiumLoadedTile` but for information purposes only (officially - in our internal code I'll still have to use it until I have refactored the set of optimizations that used it...).

Source/CesiumRuntime/Public/CesiumMeshBuildCallbacks.h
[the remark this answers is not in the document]
GhisBntly: I had a look at that, and I still see it as two extension points (see commit):

1. `CreateMaterial`, like now, which mostly allows to change the `pBaseMaterial`.
2. `CustomizeMaterial`, which could indeed be more useful than `CustomizeGltfMaterial` by passing it the `UCesiumMaterialUserData` (and removing the parameter association and index parameters). I still see this one as a separate method because I think it makes more sense to do all the generic component's material setup (from `pMaterial->SetFlags(...)` until the end of the `if (pCesiumData) { ... }` part) _before_ calling the customization endpoint, in case one would want to overwrite something. Although we don't: our customizations are independent from the rest of the material setup.

I could come up with a way to have everything in `CreateMaterial` nonetheless, using a lambda for example (to hide some Private stuff that's needed by material setup), but it would be a deviation from current practices I guess.

Source/CesiumRuntime/Public/CesiumMeshBuildCallbacks.h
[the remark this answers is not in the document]
kring: > Wouldn't both tiles load all the same geometries then, or can they differ by their bounding volume, filtering their content spatially?

Different bounding volumes wouldn't filter the content, but different tiles could have a different transform to put the same GLB at different locations. It's definitely not a common thing to do.

> I find it disturbing that there is no way to identify a tile that is not yet or no longer present in memory.

I guess we've never had a need? And also I'm not sure how we could even provide one.

For in-memory use, another possibility I can think of is to add an extension to the `Tile`. That could help you avoid coming up with a key to use in a map or similar. We do something like that to associate an Unreal texture with a glTF texture:
https://github.com/CesiumGS/cesium-unreal/blob/v2.16.0/Source/CesiumRuntime/Private/CesiumTextureUtility.cpp#L33

If you need this identifier to persist across runs or be stable across machines... I guess you could use the "path" to the tile as a unique identifier, i.e. first child, third child, second child, first child. That will work as long as the tileset itself doesn't change (but that's definitely not guaranteed, either).

The tile's content URI (which becomes the tile ID) is of course pretty good. If you're caching derived data or customizing content, it's probably fine, you just need to be aware of the (unlikely) caveats: 1) it's not _guaranteed_ to be unique (though it almost always is), and 2) if the tileset structure or tile content changes, your previous data might become invalid or disassociated.

Source/CesiumRuntime/Public/Cesium3DTileset.h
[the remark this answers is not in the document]
GhisBntly: I have removed the Tileset's CPU access flag from this PR, so I'm closing this conversation.
I am mostly done getting rid of our two use cases so I'm confident it will not be a problem to use the Gltf model instead of Unreal vertices.

Source/CesiumRuntime/Public/CesiumMeshBuildCallbacks.h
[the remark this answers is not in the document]
GhisBntly: We don't need to persist anything tile-related, we just have this optimization to keep some data mapped to tiles independently of them being cycled (loaded, unloaded, reloaded, etc.) by cesium during navigation.

When you mentioned the pointers being stable, do you mean that the tiles vector is allocated only once for a tileset, and after that a given tile will always have the same pointer, even if it is unloaded then reloaded?

Anyway, for the kind of tiles we need this optimization for (iTwin), I don't think we will ever have the case of different tiles placing the same glb at different places. And even if we had, the glTF metadata that we rely on would be the same for both instances, so mapping to the tile URL would still work.

GhisBntly:
Note: I had to add `GetVertexPositionScaleFactor()` to the `UCesiumLoadedTile` interface in order to get rid of CPU access to Unreal memory buffers in our implementation of the lifecycle events receiver: the function replicates the logic hardcoded in `loadPrimitive` (multiplication by the internal `CesiumPrimitiveData::positionScaleFactor` as well as Y axis inversion).
I could have added a vertex visiting method instead in `UCesiumLoadedTilePrimitive`, but it didn't really seem worth the trouble(?)

GhisBntly:
I mark this PR as ready for review, although I'm still unsure about `SetRenderReady`. I will try to confirm (or debunk ;-)) our understanding of the problem with a simpler example tomorrow. In particular, I'm indeed keen on seeing whether the problem could also happen with cesium-created textures :-/

GhisBntly:
About `SetRenderReady`, note that I also considered using it to avoid blocking the game thread for too long, because each tile needs substantial post-processing to be made "render-ready" (parsing vertex metadata, map this data to internal domain data, create optimization structures for 4D schedule replay, and finally create the textures for the custom material shader...).

The method that does all that in our code already supports a time budget: even though I do not use it for this purpose at the moment, I may want to do so in the future to smooth out tile transitions, which may be a valid reason to include this feature?

GhisBntly:
Sadly, I was unable to reproduce the specific issue that led us to defer material texture assignments, but it doesn't mean the problem no longer happens, as it was witnessed only in rare circumstances, and the project data I had used to investigate it no longer exists :-(
Still, I would really prefer keeping `SetRenderReady`, if only to allow for better distribution of the tile post-process load on our end over several game ticks.

j9liu:
Sorry for the delay on these PRs @GhisBntly :( I'll take a look at this sometime this week!

CHANGES.md
```diff
@@ -46,6 +46,7 @@ In addition to the above, this release updates [cesium-native](https://github.co
 
 - Added a Cesium -> Geocoder -> Geocode Blueprint function, making it easy to query the Cesium ion geocoder.
 - Added `UCesiumMetadataPickingBlueprintLibrary::FindPropertyTableProperty` to search for a `FCesiumPropertyTableProperty` by name on a given `UPrimitiveComponent`.
+- Added the class `CesiumMeshBuildCallbacks`: when an implementation is registered on a tileset (with `ACesium3DTileset::SetMeshBuildCallbacks`), its functions will be called at various points in a tile's lifecycle, like when a mesh component is created, when a material is instanced, when the tile changes visibility, when it is unloaded, etc.
```
kring: Move this up to the correct version. Also, `CesiumMeshBuildCallbacks` has been renamed.

Source/CesiumRuntime/Public/Cesium3DTileset.h
```diff
@@ -1371,6 +1385,8 @@ class CESIUMRUNTIME_API ACesium3DTileset : public AActor {
 
   int32 _tilesetsBeingDestroyed;
 
+  ICesium3DTilesetLifecycleEventReceiver* _lifecycleEventReceiver;
```
kring:
```suggestion
  ICesium3DTilesetLifecycleEventReceiver* _pLifecycleEventReceiver;
```

kring: Also, in order for this pointer to be trackable by the garbage collector, this needs to be declared a `UPROPERTY`.

```suggestion
  // Make this visible to the garbage collector, but don't save/load/copy it.
  UPROPERTY(Transient, DuplicateTransient, TextExportTransient)
  TObjectPtr<ICesium3DTilesetLifecycleEventReceiver> _pLifecycleEventReceiver;
```

Source/CesiumRuntime/Public/Cesium3DTilesetLifecycleEventReceiver.h
```diff
@@ -0,0 +1,100 @@
...
+   * @param Name The name for the new material, as used by creation functions
+   *like NewObject
+   * @return Material instance created. If null, a material will be created by
+   *the caller based on the pBaseMaterial passed, ie. as if this method had not
+   *been called.
+   */
+  virtual UMaterialInstanceDynamic* CreateMaterial(
+      ICesiumLoadedTilePrimitive& TilePrim,
+      UMaterialInterface* pDefaultBaseMaterial,
+      const FName& Name) = 0;
```
kring: This method has an implementation, so I don't think there's any reason to declare it pure virtual.

Source/CesiumRuntime/Public/Cesium3DTilesetLifecycleEventReceiver.h
```diff
@@ -0,0 +1,100 @@
...
+   * @param Outer The outer for the new material, as used by creation functions
+   *like NewObject
+   * @param Name The name for the new material, as used by creation functions
+   *like NewObject
+   * @return Material instance created. If null, a material will be created by
+   *the caller based on the pBaseMaterial passed, ie. as if this method had not
+   *been called.
+   */
+  virtual UMaterialInstanceDynamic* CreateMaterial(
+      ICesiumLoadedTilePrimitive& TilePrim,
```
kring: Avoid abbreviations. Call it `TilePrimitive`.

Source/CesiumRuntime/Public/CesiumLoadedTile.h
```diff
@@ -0,0 +1,54 @@
...
+
+UINTERFACE()
+class UCesiumLoadedTile : public UInterface {
+  GENERATED_BODY()
+};
+class ICesiumLoadedTile {
+  GENERATED_BODY()
+public:
+  /** Get the tile identifier: this is informational only, as there is no
+   * guarantee of unicity */
```
kring: "unicity" isn't wrong, but it's unusual. "uniqueness" is more common.

```suggestion
   * guarantee of uniqueness. */
```

Source/CesiumRuntime/Public/CesiumLoadedTile.h
```diff
@@ -0,0 +1,54 @@
...
+#include "CesiumLoadedTile.generated.h"
+
+class ACesium3DTileset;
+class UStaticMeshComponent;
+
+UINTERFACE()
+class UCesiumLoadedTile : public UInterface {
+  GENERATED_BODY()
+};
+class ICesiumLoadedTile {
```
kring, line 28: Add a doc comment for the interface itself.

Source/CesiumRuntime/Public/Cesium3DTilesetLifecycleEventReceiver.h
```diff
@@ -0,0 +1,100 @@
...
+};
+/** Extension points for customizations requiring mesh and tile properties and
+ * lifecycle information. All methods are called from the game thread.*/
+class CESIUMRUNTIME_API ICesium3DTilesetLifecycleEventReceiver {
+  GENERATED_BODY()
+public:
+  /**
+   * Allows to override the base material from which the given tile primitive's
+   *material instance will be created. The default implementation of this method
+   *can be used to create the material when no customization is needed.
```
kring: I think this description doesn't quite capture the purpose of this method after some of the later changes. Suggestion:


```suggestion
   * Creates a material instance for a given tile primitive.
   *
   * The default implementation simply calls `UMaterialInstanceDynamic::Create`
   * with the default base material. Overriding this method is useful when a
   * different base material should be selected based on properties of the
   * primitive.
```

Also, a nitpick: Add a space after the `*` in doc comments.

Source/CesiumRuntime/Public/Cesium3DTilesetLifecycleEventReceiver.h
```diff
@@ -0,0 +1,100 @@
...
+   *like NewObject
+   * @param Name The name for the new material, as used by creation functions
+   *like NewObject
+   * @return Material instance created. If null, a material will be created by
+   *the caller based on the pBaseMaterial passed, ie. as if this method had not
+   *been called.
+   */
+  virtual UMaterialInstanceDynamic* CreateMaterial(
+      ICesiumLoadedTilePrimitive& TilePrim,
+      UMaterialInterface* pDefaultBaseMaterial,
```
kring: I'm going to sound like I'm contradicting myself here, relative to some other comments... but this should be named:

```suggestion
      UMaterialInterface* DefaultBaseMaterial,
```

The reason is that this method is part of the public API, and the public API should follow Unreal naming conventions.

Code that is not part of the public API can follow either the Unreal conventions or the cesium-native conventions, and we're not very consistent about which. But all our code should be consistent with one or the other.

Source/CesiumRuntime/Public/Cesium3DTilesetLifecycleEventReceiver.h
```diff
@@ -0,0 +1,100 @@
...
+  /**
+   * Allows to override the base material from which the given tile primitive's
+   *material instance will be created. The default implementation of this method
+   *can be used to create the material when no customization is needed.
+   * @param TilePrim Loaded tile primitive for which a material is needed
+   * @param pDefaultBaseMaterial Default chosen base material. May be ignored if
+   *the method chooses to create the mesh material based on a custom base
+   *material.
+   * @param Outer The outer for the new material, as used by creation functions
+   *like NewObject
```
kring: There's no such parameter.

Source/CesiumRuntime/Public/Cesium3DTilesetLifecycleEventReceiver.h
```diff
@@ -0,0 +1,100 @@
...
+   * @param Name The name for the new material, as used by creation functions
+   *like NewObject
+   * @return Material instance created. If null, a material will be created by
+   *the caller based on the pBaseMaterial passed, ie. as if this method had not
+   *been called.
```
kring: I think I'd prefer to simplify this by _requiring_ that this method return a valid material instance. That allows us to delete the code that checks for nullptr (replace with an assertion that it's not nullptr), and doesn't really make things any harder for users because they only have to call the base implementation.

Source/CesiumRuntime/Public/Cesium3DTilesetLifecycleEventReceiver.h
```diff
@@ -0,0 +1,100 @@
...
+   * @param glTFmaterialPBR Parameters for this primitive's material defining the
+   * metallic-roughness material model from Physically-Based Rendering (PBR)
+   * methodology
+   */
+  virtual void CustomizeMaterial(
+      ICesiumLoadedTilePrimitive& TilePrim,
+      UMaterialInstanceDynamic& Material,
+      const UCesiumMaterialUserData* pCesiumData,
+      const CesiumGltf::Material& glTFmaterial,
+      const CesiumGltf::MaterialPBRMetallicRoughness& glTFmaterialPBR);
```
kring:
```suggestion
      ICesiumLoadedTilePrimitive& TilePrimitive,
      UMaterialInstanceDynamic& Material,
      const UCesiumMaterialUserData* CesiumData,
      const CesiumGltf::Material& GltfMaterial,
      const CesiumGltf::MaterialPBRMetallicRoughness& GltfMaterialPBR);
```

Maybe we can remove `GltfMaterialPBR`? I don't think it adds much, since it's usually just a reference to the one in `GltfMaterial`. Or a reference to a default-constructed `MaterialPBRMetallicRoughness` if there isn't one in `GltfMaterial`.

Source/CesiumRuntime/Public/Cesium3DTilesetLifecycleEventReceiver.h
```diff
@@ -0,0 +1,100 @@
...
+
+  /**
+   * Called at the end of the static mesh component construction.
+   * @param TilePrim Loaded tile primitive being constructed
+   */
+  virtual void
+  OnTileMeshPrimitiveConstructed(ICesiumLoadedTilePrimitive& TilePrim) = 0;
```
kring: Apologies for the nitpicking here...

```suggestion
  /**
   * Called after a `MeshPrimitive` in a tile's glTF is loaded. This method is called at the
   * end of the load process, after construction of the static mesh component that will
   * render the primitive.
   *
   * @param TilePrimitive Tile primitive that has just been loaded.
   */
  virtual void
  OnTileMeshPrimitiveLoaded(ICesiumLoadedTilePrimitive& TilePrimitive) = 0;
```

Source/CesiumRuntime/Public/Cesium3DTilesetLifecycleEventReceiver.h
```diff
@@ -0,0 +1,100 @@
...
+
+  /**
+   * Called at the end of all static mesh components' construction for a given
+   * tile.
+   * @param LoadedTile The tile that has just been loaded
+   */
+  virtual void OnTileConstructed(ICesiumLoadedTile& LoadedTile) = 0;
```
kring:
```suggestion
  /**
   * Called after a new tile has been loaded. This method is called after
   * `OnTileMeshPrimitiveLoaded` has been called for all of the tile's primitives.
   *
   * @param Tile The tile that has just been loaded
   */
  virtual void OnTileLoaded(ICesiumLoadedTile& Tile) = 0;
```

Source/CesiumRuntime/Public/Cesium3DTilesetLifecycleEventReceiver.h
```diff
@@ -0,0 +1,100 @@
...
+   * @param visible New visibility flag being applied
+   */
+  virtual void
+  OnVisibilityChanged(ICesiumLoadedTile& LoadedTile, bool visible) = 0;
+
+  /**
+   * Called before a tile is destroyed (when it is unloaded, typically).
+   * @param LoadedTile The tile which is about to be unloaded
+   */
+  virtual void BeforeTileDestruction(ICesiumLoadedTile& LoadedTile) = 0;
```
kring:
```suggestion
  /**
   * Called before a tile is unloaded.
   * @param Tile The tile that is about to be unloaded
   */
  virtual void OnTileUnloading(ICesiumLoadedTile& Tile) = 0;
```

Source/CesiumRuntime/Public/Cesium3DTilesetLifecycleEventReceiver.h
```diff
@@ -0,0 +1,100 @@
...
+
+  /**
+   * Called when changing the visibility of any UCesiumGltfComponent, ie usually
+   * several times per tile (when the tileset selection leads to showing or
+   * hiding a whole tile).
+   * @param LoadedTile The tile which visibility is being toggled
+   * @param visible New visibility flag being applied
+   */
+  virtual void
+  OnVisibilityChanged(ICesiumLoadedTile& LoadedTile, bool visible) = 0;
```
kring: `UCesiumGltfComponent` is private, so we shouldn't mention it. And some other nitpicks.

```suggestion
  /**
   * Called when a tile is shown or hidden. This may be called zero or more times per tile.
   * @param Tile The tile for which visibility is being toggled
   * @param visible New visibility flag being applied
   */
  virtual void
  OnTileVisibilityChanged(ICesiumLoadedTile& Tile, bool visible) = 0;
```

Source/CesiumRuntime/Public/Cesium3DTilesetLifecycleEventReceiver.h
kring: Actually, should any of these methods be pure virtual? It'll be easier for users if we provide default, do-nothing implementations, and users can override only the ones they care about.

Source/CesiumRuntime/Public/CesiumLoadedTile.h
```diff
@@ -0,0 +1,54 @@
...
+   * guarantee of unicity */
+  virtual const Cesium3DTilesSelection::TileID& GetTileID() const = 0;
+  /** Scaling factor to be applied (component-wise multiplication) to glTF
+   * vertices of this tile's model to obtain the values represented in their
+   * matching mesh component (see @{link
+   * UCesiumLoadedTilePrimitive::GetMeshComponent). */
+  virtual FVector GetGltfToUnrealLocalVertexPositionScaleFactor() const = 0;
+  virtual ACesium3DTileset& GetTilesetActor() = 0;
+  virtual const FCesiumModelMetadata& GetModelMetadata() const = 0;
+  virtual void SetRenderReady(bool bToggle) = 0;
```
kring: These are all part of the public API, so they should have doc comments.

Source/CesiumRuntime/Public/CesiumLoadedTile.h
```diff
@@ -0,0 +1,54 @@
...
+};
+
+UINTERFACE()
+class UCesiumLoadedTilePrimitive : public UInterface {
+  GENERATED_BODY()
+};
+// Not merged with ICesiumPrimitive because this is Public whereas
+// ICesiumPrimitive is Private
+class ICesiumLoadedTilePrimitive {
```
kring, line 75: Add doc comments to this class and its methods.

Source/CesiumRuntime/Public/CesiumLoadedTile.h
```diff
@@ -0,0 +1,54 @@
...
+  GENERATED_BODY()
+public:
+  virtual ICesiumLoadedTile& GetLoadedTile() = 0;
+  virtual UStaticMeshComponent& GetMeshComponent() = 0;
+  virtual const FCesiumPrimitiveFeatures& GetPrimitiveFeatures() const = 0;
+  virtual const FCesiumPrimitiveMetadata& GetPrimitiveMetadata() const = 0;
+  virtual const CesiumGltf::MeshPrimitive* GetMeshPrimitive() const = 0;
+  virtual std::optional<uint32_t>
+  FindTexCoordIndexForGltfAttribute(int32_t accessorIndex) const = 0;
```
kring: Isn't it a GltfAccessor, not a GltfAttribute? Also, avoid abbreviations.

```suggestion
  FindTextureCoordinateIndexForGltfAccessor(int32_t accessorIndex) const = 0;
```

Source/CesiumRuntime/Public/CesiumLoadedTile.h
```diff
@@ -0,0 +1,54 @@
...
+  virtual const Cesium3DTilesSelection::TileID& GetTileID() const = 0;
+  /** Scaling factor to be applied (component-wise multiplication) to glTF
+   * vertices of this tile's model to obtain the values represented in their
+   * matching mesh component (see @{link
+   * UCesiumLoadedTilePrimitive::GetMeshComponent). */
+  virtual FVector GetGltfToUnrealLocalVertexPositionScaleFactor() const = 0;
+  virtual ACesium3DTileset& GetTilesetActor() = 0;
+  virtual const FCesiumModelMetadata& GetModelMetadata() const = 0;
+  virtual void SetRenderReady(bool bToggle) = 0;
+};
```
kring, line 60: I'm surprised there's no method to get the `CesiumGltf::Model`. Particularly because `ICesiumLoadedTilePrimitive` provides a way to get the `CesiumGltf::MeshPrimitive`.

Source/CesiumRuntime/Public/CesiumMetadataValue.h
```diff
@@ -867,7 +867,7 @@ class CESIUMRUNTIME_API UCesiumMetadataValueBlueprintLibrary
  * These should be moved to UCesiumMetadataValueBlueprintLibrary if those types
  * become compatible with Blueprints in the future.
  */
-class CESIUMRUNTIME_API CesiumMetadataValueAccess {
+class CESIUMRUNTIME_API FCesiumMetadataValueAccess {
```
kring: Why add the `F` prefix? This isn't a USTRUCT.

kring:
> The method that does all that in our code already supports a time budget: even though I do not use it for this purpose at the moment, I may want to do so in the future to smooth out tile transitions, which may be a valid reason to include this feature?

cesium-native already has (and cesium-unreal already uses) a main thread loading time limit:
https://github.com/CesiumGS/cesium-native/blob/v0.49.0/Cesium3DTilesSelection/include/Cesium3DTilesSelection/TilesetOptions.h#L302

Also, if you can do any of the work in a worker thread, then that process is allowed to take place over multiple frames, with a `Future` signaling when it's complete.

Does the "render ready" mechanism satisfy a use-case that those two mechanisms don't already satisfy?

> Sadly, I was unable to reproduce the specific issue that led us to defer material texture assignments, but it doesn't mean the problem no longer happens, as it was witnessed only in rare circumstances, and the project data I had used to investigate it no longer exists :-(

It makes me a little uneasy, because the use-case is unclear (at least to me) and its the sort of thing that has a moderately high potential for error by causing violation of assumptions in the tile selection algorithm. From what I can see, this PR doesn't actually use the "render ready" flag at all. So it should be pretty easy to extract that part into a separate PR, right?

kring (commented):
Lots of small comments here, but once again, it looks great overall!

Source/CesiumRuntime/Public/CesiumMetadataValue.h
GhisBntly: Good catch, I was indeed supposed to remove the `F` from _our_ version, not add it to _yours_ ! 😮

GhisBntly:
> Also, if you can do any of the work in a worker thread, then that process is allowed to take place over multiple frames, with a `Future` signaling when it's complete.
> 
> Does the "render ready" mechanism satisfy a use-case that those two mechanisms don't already satisfy?

No you're right, my time budget use case doesn't add anything indeed.

> It makes me a little uneasy, because the use-case is unclear (at least to me) and its the sort of thing that has a moderately high potential for error by causing violation of assumptions in the tile selection algorithm. From what I can see, this PR doesn't actually use the "render ready" flag at all. So it should be pretty easy to extract that part into a separate PR, right?

I understand, I will remove it from this PR. I have started testing again our app without this deferred texture setup system, I guess if it proves stable and no longer show any bug like the one we had, we may remove it entirely on our side too in the future.

Source/CesiumRuntime/Public/CesiumLoadedTile.h
GhisBntly: Indeed the method existed (we use it), it was apparently lost in a merge...

GhisBntly:
@kring , I believe I have accounted for all comments so far - ready for next round ;^^

Source/CesiumRuntime/Public/CesiumLoadedTile.h
```diff
@@ -0,0 +1,79 @@
...
+/** Access to the data loaded for a given glTF primitive of an {@link
+ * ICesiumLoadedTile}.
+ */
+class ICesiumLoadedTilePrimitive {
+  GENERATED_BODY()
+public:
+  /** The {@link ICesiumLoadedTile} the primitive belongs to. */
+  virtual ICesiumLoadedTile& GetLoadedTile() = 0;
+  /** The Unreal mesh component built to represent the glTF primitive. */
+  virtual UStaticMeshComponent& GetMeshComponent() = 0;
```
kring: Should this be a pointer, in order to allow for the posibility that a primitive is not represented as a static mesh component?

kring:
Thanks for all your work - and patience! - on this @GhisBntly!

PR #1677, opened 2025-05-20
Implement `CesiumRasterOverlays::VectorDocumentRasterOverlay` in Unreal

azrogers:
This is the Cesium for Unreal counterpart to CesiumGS/cesium-native#1163. That PR will need to be merged before this one is.


azrogers:
@kring Pinging you to let you know this is in a PR if you want to give it a try!

azrogers:
What I assume to be a shared_ptr issue has appeared in the most recent round of changes - somehow a double free is happening to the GeoJsonDocument, but I can't tell where that might be happening.

azrogers:
Looks like iOS is building successfully! But now it's running into curl issues, like Android.

kring:
@azrogers I'm using this branch and trying to recreate this screenshot:
image

I've created a `CesiumGeoJsonDocumentRasterOverlay` and I'm able to render the lines. Now I'm trying to style them.

I've added this to my level blueprint:
image

And that event function looks like this:
image

But what I can't figure out is how to set the style on particular features. Looking only briefly at the code, it looks like it might not be hooked up yet. Is that right?

azrogers:
@kring I've added blueprint functions for working with GeoJsonObject styling so you should be able to set it up now!

kring:
I tried to add a raster overlay with [this GeoJSON](https://digital.atlas.gov.au/datasets/digitalatlas::gas-pipelines-1/about) (click Download and then GeoJSON), loaded from a file URL. It doesn't report any errors in the log, but nothing appears. Debugging it _very briefly_, I think the immediate problem is that `GlobeRectangleFromObjectVisitor` isn't implemented for FeatureCollection, which is the top-level object type in this file.

kring:
Oh sorry, I'm wrong. `GlobeRectangleFromObjectVisitor` is fine, it doesn't need to be implemented for FeatureCollection. The problem was that `BoundingRegionBuilder::expandToIncludeGlobeRectangle` has a couple of bugs. It's not computing the longitude correctly, because the `<` and `>` operators are flipped. And it wasn't setting `_longitudeRangeIsEmpty` to false. With those two changes (which I pushed to cesium-native `vector-rasterizer` branch), this GeoJSON is working well!

kring:
Mip generation doesn't appear to be working correctly. Only the first mip level has lines drawn. I believe the only reason this isn't visible is because, with the default raster overlay SSE of 2, every pixel ends up sampling the base mip level. But if you set the Max SSE to 0.5, lines start fading out and disappearing.

I confirmed that lines aren't drawn except in the base mip level using RenderDoc. Here's the base mip:
image

And here's the first mip level:
image

kring:
The problem was that `primitivesRendered.assign({false});` needs to be `primitivesRendered.assign(primitivesRendered.size(), false);`. With that change, the lines stay a nice constant width as you zoom out, it looks great!

kring:
I'm seeing some shifting at different LODs. It's a little subtle in this gif due to the low frame rate, more obvious live, but hopefully you can see it. The lines appear to move relative to the globe surface when the LOD shifts.
line-shift
This only happens when the raster overlay's projection is set to Web Mercator. Setting it to Geographic avoids the problem, or at least drastically reduces it. Most likely this comes down to whether the lines are straight in the geographic projection versus the web mercator projection versus in reality (i.e., geodesics). But I need to think through it a bit more.

kring:
Here's a more obvious example where two different LODs side-by-side make a line disjoint:
image
(I've suspended update and zoomed in closer to make it easier to see in the screenshot)

kring:
Thanks @azrogers!

PR #1679, opened 2025-05-23
Update cesium-native

kring:
It looks like the vcpkg upgrade in CesiumGS/cesium-native#1178 breaks the Unreal build on every platform other than Windows. 🤦

Draft until all the platforms are working again...


kring:
I'm still trying to understand what's going on with Linux, but the Android build failure is actually a pretty big problem, and I don't know how to solve it.

The immediate problem is that the Android NDK r25b, which is the one we use across all Unreal versions, has shaky support for C++20. It's _compiler_ support is solid, but for it has an old version of libc++. As explained here:
https://github.com/android/ndk/issues/1530

We're largely aware of this ourselves, and so we avoid using C++20 features like ranges because we know they'll cause us problems.

Unfortunately, the rest of the world doesn't really care about our problems, and quite reasonably assumes that things like ranges and concepts are fair game to use when targeting C++20. And occurrences of that will only increase over time, as demonstrated by this new problem in ada-url.

NDK r26 came out awhile ago, and upgrades the libc++ version finally, which would fix the problem for us, so maybe we can upgrade? Nope, Unreal Engine 5.3 is quite inflexible about its Android NDK version (https://github.com/EpicGames/UnrealEngine/blob/5.3.2-release/Engine/Source/Programs/UnrealBuildTool/Platform/Android/AndroidPlatformSDK.Versions.cs):

```
protected override void GetValidVersionRange(out string MinVersion, out string MaxVersion)
{
	MinVersion = "r25b";
	MaxVersion = "r25b";
}
```

We could change this to allow a newer version, and it _might_ work, but we'd have to convince Epic to make this change, too, if we want to publish on Fab. So that seems like a non-starter.

Fortunately, UE 5.4 allows through r26c (but not the latest, r26d), and UE 5.5 allows through r27. So I _think_ we'll be ok once we can drop support for UE 5.3.

In the meantime, the only possible solution I can think of is to downgrade ada-url to an earlier version from before they started using the features our Android toolchain can't support.

kring:
I've fixed this by downgrading ada-url for now (in other words: kicking the can down the road), so this should be ready to merge now. Other changes in this PR:

1. Set `CMAKE_POLICY_VERSION_MINIMUM` while building tidy-html5 so that it builds with CMake 4.
2. Removed the `asyncplusplus` and `zlib-ng` overlay ports. They're no longer needed after upgrading to a newer vcpkg release.
3. Updated the `ktx` overlay port to match the one in the new vcpkg release (but we still need Unreal-specific modifications).

j9liu (approved):
Thanks @kring ! I don't have non-Windows systems to test this on, but I trust CI would let us know if anything was wrong.

PR #1681, opened 2025-05-27
Fix handling of Dynamic Material Instances when creating tile meshes

### LINKED ISSUE #1569: Tileset rendering does not work correctly with a Dynamic Material Instance
Reported here:
https://community.cesium.com/t/cesium-for-unreal-3dtiles-dynamicmaterial/37034

To reproduce, create a Level Blueprint like this:
https://blueprintue.com/blueprint/oyczy--l/
image

Click Play-in-Editor, and the tileset will render, but it will be missing its raster overlays (among other things).

The reason is that the new material doesn't have `UCesiumMaterialUserData` attached to it, so it doesn't know the mapping between material layer name (such as `Overlay0`) and the material layer index used to set material parameters. Adding `UCesiumMaterialUserData` is not an option, because the API necessary to find the layer names is only available in the Editor.

The best solution, I think, is to modify the code starting here:
https://github.com/CesiumGS/cesium-unreal/blob/v2.11.0/Source/CesiumRuntime/Private/CesiumGltfComponent.cpp#L3024
So that, rather than just checking the `pBaseAsMaterialInstance` for `UCesiumMaterialUserData`, it also walks the `Parent` chain looking for a parent material that has it.

j9liu:
## Description

This PR resolves an issue with Dynamic Material Instances that were assigned to tilesets at runtime. This was because of the `CesiumMaterialUserData` that our materials rely on for mapping material layers; they aren't instantiated for dynamic material instances. Now when tiles are created, we search the instance's parent hierarchy for the user data.

## Issue number or link

Fixes #1569 

## Author checklist

## Testing plan

In the first level of the Cesium for Unreal Samples, reproduce the Blueprint in #1569. Check that Cesium World Terrain renders correctly with the raster overlay.

If you'd like to test the dynamic material instance behavior, copy this Blueprint -- and remove the raster overlay on CWT so that you can see the color come through.

Image

### DIFF
--- a/Source/CesiumRuntime/Private/CesiumGltfComponent.cpp
+++ b/Source/CesiumRuntime/Private/CesiumGltfComponent.cpp
@@ -3164,7 +3164,7 @@ static void loadPrimitiveGameThreadPart(
   UMaterialInstanceDynamic* pMaterial;
   {
     TRACE_CPUPROFILER_EVENT_SCOPE(Cesium::SetupMaterial)
-       
+
     UMaterialInstanceDynamic* pBaseAsMaterialInstanceDynamic =
         Cast<UMaterialInstanceDynamic>(pBaseMaterial);
     UMaterialInstance* pParentMaterialInstance =


kring:
Thanks @j9liu!

PR #1683, opened 2025-05-28
Access to glTF tuner and tuned tiles versions in the Unreal plugin

GhisBntly:
Goes with https://github.com/CesiumGS/cesium-native/pull/1186


kring:
Thanks @GhisBntly!

PR #1685, opened 2025-05-29
Add basic support for voxel rendering and styling

j9liu:
## Description

Depends on https://github.com/CesiumGS/cesium-native/pull/1188/.

This PR adds support for loading and rendering tilesets with `3DTILES_content_voxels`, parsing glTFs with `EXT_primitive_voxels` payloads.

**TODO**: screenshot

This is a meaty set of changes, so I'll add a written walkthrough of the changes in a follow-up comment.

## Author checklist

~- [ ] I have added or updated unit tests to ensure consistent code coverage as necessary.~

## Remaining Tasks

I'm not happy with the way voxels are currently selected for rendering in `VoxelResources`. It's a weird compromise between CesiumJS's `VoxelTraversal` and the cesium-native traversal, and it falls apart for larger datasets.

## Testing plan

**TODO**


j9liu:
# Voxel Shader Walkthrough (OUTDATED 07/09)

This is a breakdown of `CesiumVoxelTemplate.hlsl`, which is used to generate custom shaders for styling voxel metadata. This file is _tall_ -- a lot of components could be separated into their own `.h` files if they could be included by Unreal's material compiler, but alas.

In lieu of that, you'll see a lot of `struct`s recursively encapsulate utility functions. I try to make the divisions more noticeable using comments. You'll also see a lot of `if`s branching, but the majority of them are there to check uniforms.

I'm doing my best to approach it from top-down, but it might also help to start at `MAIN FUNCTION BODY` and work your way up as you encounter new stuff.

<details>
<summary><h2>Shape Intersections</h2></summary>

Voxel grid shape affects the intersection math in shader. It's not as apparent in this PR because this only supports box voxels. But it establishes the framework for future ellipsoid and cylinder support.

A ray is passed into this shader -- it is taken into the shape's "Unit" space where everything is normalized from `[-1, 1]`. This initially checks if the ray even intersects the shape in the first place. If not, exit early.

Then, coordinates are transformed into "UV" space (from [0, 1]) to then traverse the Voxel Octree in a mathematically reasonable way (explained later).

### UV Space Complexity

There is a distinction between "vanilla UV" space vs. "shape UV" space, and it won't be apparent from this PR because boxes are so simple.
- Vanilla UV space can be thought of as from [0, 1], as expected.
- Shape UV space is space dependent. For instance, Shape UV space for cylinders would take the vanilla UV coordinates and redefine them with respect to the cylinder's (radius, angle, height). For ellipsoids, the output is (longitude, latitude, altitude).

</details>
<details><summary><h2>Voxel Octree</h2></summary>

Once we have ray intersection in Shape UV coordinates, we can use it to traverse the voxel octree. This octree is just the one indicated by implicit tiling in the tileset. We're trying to figure out which tile our intersection point is in so we can sample the correct tile's voxel data. An octree node corresponds to the implicit tile with the same octree tile ID / coordinates.

The `Octree` struct makes use of the texture generated by `VoxelOctree.h`. The tree structure is encoded into a texture, where each node is basically 9 numbers. This number is duo flag + index combo where the flag indicates the nature of the index. (See the comments for `struct Node`).

`OctreeTraversal` keeps track of which node is being traversed. `BeginTraversal` starts this traversal at the root node, calling `TraverseToLeaf`. This function will go down the tree to find the leaf node that contains the point. The math involves finding the extends of a tile in UV space and testing whether the intersection falls inside it.

A leaf node is assumed to have an associated data texture with it (explained below). Once we reach a leaf node with data, we start marching subsequent voxels using `ResumeTraversal`. The voxel intersections are also handled in Shape UV space. If the extent of the tile has been reached, we go back up the tree towards a common parent to find the next valid leaf tile.

</details>

<details><summary><h2>Voxel Data Textures</h2></summary>

The `DataTextures` struct makes use of the megatextures produced by `VoxelDataTextures.h`. If a tile has a slot in the megatexture (because it successfully uploaded its voxel data), then it will keep track of that slot's index. When a tile is encountered by `OctreeTraversal`, this index is used to get the correct bounds from the megatexture to query. `DataTextures` will automatically compute where to sample the megatexture based on the tile's coordinates and the voxel within it.

This part must be auto-generated by `CesiumVoxelMetadataComponent`, so you'll see a couple `%s` symbols where the boilerplate would be filled. It retrieves the values and puts them in `CustomShaderProperties` for use in the user's custom shader.

</details>

<details><summary><h2>Custom Shader</h2></summary>

Autogenerated by `CesiumVoxelMetadataComponent`, this just pastes the user code as-is so they can act on the metadata values passed through the `CustomShaderProperties` struct.

There is also an additional area for additional functions. For example, if you want to map values to a color ramp, it is easiest to create a separate function that can map a value to an index in a common array.

</details>

<details><summary><h2>Putting it all together</h2></summary>

The main body of the shader does the following:

- Initializes the octree + data textures, mapping the input variables to the member variables on the structs
- Tests whether the ray intersects the grid shape at all; if not, exit early
- Repeat as necessary:
  - Finds and converts the ray intersection in Vanilla UV space to Shape UV space
  - Traverses the octree to find the corresponding tile with voxel data
  - Samples the data textures to retrieve data at the intersected voxel
  - Passes data to custom shader for user styling
  - If the voxel is not opaque, step along the ray, and keep sampling to accumulate enough color to exit
- Return result color (including opacity)
</details>

j9liu:
I could keep picking apart at this, but I think this would benefit more from a fresh pair of eyes. I'm keeping this as a draft but I would appreciate any and all reviews :pray:

j9liu:
Turns out that UE 5.4 has drastically improved the UX of Custom HLSL nodes in materials!

It's now possible to define extra `#include` paths in for the shader, so I don't need to use one uber shader anymore. I've reorganized the code so that everything is separated into meaningful pieces. Hopefully this makes things easier to review, too ^^

kring:
I was able to get a basic thing working:
image

There were a few things I didn't realize, and they took me awhile to figure out:
* The voxel material layer must be the bottommost layer in the material. I initially thought, for some reason, that it had to be layered on top of a `ML_CesiumGltf` like our other materials. And when that didn't work, I assumed it was because I misnamed the material layer. Only after trying and failing to find the expected material layer name in the code did I realize it was hard-coded to layer 0.
* The shape of this dataset isn't apparent unless your material sets the alpha to 0.0 when Temp_C == Temp_C_NO_DATA. I initially tried to just return a solid color and wondered why all I got was a rectangle.
* But even after the above, I still got a rectangle. Because my material had a Blend Mode of "Masked" instead of "Translucent".

So a lot of this wouldn't have been a problem if we had a tutorial. But if the generated material layer isn't really meant to be used in a layered material, perhaps we should just generate a material directly? That would have eliminated two out of my three errors above.

Also, I'm wondering if the "Voxel Raymarch" custom expression could have an output node for each of the `CustomShaderProperties`? The way the styling of other types of tiles works. Then users could implement their logic using the shader graph. Also, it seems like pressing the "Auto Generate" button again clobbers any edits to that custom expression node, so the chance of losing the user's work is pretty high with the way things are set up now.

j9liu:
Thanks for the feedback @kring ! I agree with your points about setting up the material and generating it as a full material instead of as a layer.

> Also, I'm wondering if the "Voxel Raymarch" custom expression could have an output node for each of the CustomShaderProperties? [...] pressing the "Auto Generate" button again clobbers any edits to that custom expression node, so the chance of losing the user's work is pretty high...

Yeah, I acknowledge this system isn't great. The problem is that raymarching must be done in a `for` loop to actually find valid voxels in the first place, and I don't see how to enable that from material nodes. I wish there was a way separate the raymarch logic from the node-based styling, I don't know how you'd re-call the raymarching logic (and not only that, to stop it once it's reached an efficiently saturated output). I can experiment with this some more, but I'm not super hopeful I'll find a better solution...

j9liu:
Tried hacking a for-loop by using a couple custom nodes in a material, but it looks like Unreal doesn't like it:
```
LogMaterial: Warning: [AssetLog] /Engine/Transient.Material_0: Failed to compile Material for platform PCD3D_SM5, Default Material will be used in game.
    Expression is part of a cycle. Please make sure the material graph is acyclic.
```

Makes sense though... it'd probably end up nesting calls like `functionA(functionB(functionC(functionA...` etc.

kring:
Works well on macOS / Metal (UE 5.5):
image

Shaders/Private/CesiumVoxelTemplate.usf
```diff
@@ -0,0 +1,224 @@
...
+struct CustomShaderProperties
+{
+%s
+};
+
+struct CustomShader
+{
+%s
+  
+  float4 Shade(CustomShaderProperties Properties)
```
kring: Why not make this Shade function a member of `CustomShaderProperties` instead? Then the properties can be accessed without qualifying them with `Properties.`.

Source/CesiumRuntime/Public/CesiumVoxelMetadataComponent.h
```diff
@@ -0,0 +1,157 @@
...
+   * The custom shader code to apply to each voxel that is raymarched.
+   */
+  UPROPERTY(
+      EditAnywhere,
+      Category = "Cesium",
+      Meta =
+          (TitleProperty = "Custom Shader",
+           DisplayAfter = "CustomShaderPreview",
+           MultiLine = true))
+  FString CustomShader = TEXT("return 1;");
```
kring: I suggest changing the default value to something like `return float4(1.0, 1.0, 1.0, 0.02);`. That makes it much clearer that this is a color, and it makes the default visualization look at a bit more voxely.

Source/CesiumRuntime/Public/CesiumVoxelMetadataComponent.h
```diff
@@ -0,0 +1,157 @@
...
+
+  /**
+   * Any additional functions to include for use in the custom shader.
```
kring:
```suggestion
   * Any additional functions to include for use in the custom shader. The HLSL code provided here is included verbatim in the generated material.
```

Source/CesiumRuntime/Private/Cesium3DTileset.cpp
```diff
@@ -1157,6 +1174,24 @@ void ACesium3DTileset::LoadTileset() {
           TCHAR_TO_UTF8(*dbFile));
 #endif
 
+  this->_pTileset->getRootTileAvailableEvent().thenImmediately([thiz = this]() {
```
kring: This logic looks like it needs to run in the main thread, so `thenImmediately` is dangerous.

kring: Also, `thiz = this` is pointless because `thiz` has the same type as `this`. You might as well just capture `this`.

There's a bit of danger here though in that the `Cesium3DTileset` could get garbage collected before the root tile is available. And I don't think there's anything that would prevent the root tile available event from still being raised. I think a simple `IsValid(this)` - and early return - should be sufficient to guard against that. We do the same thing in `SampleHeightMostDetailed`.

Source/CesiumRuntime/Private/Cesium3DTileset.cpp
```diff
@@ -2077,31 +2120,37 @@ void ACesium3DTileset::Tick(float DeltaTime) {
...
-        (pRenderContent &&
-         pRenderContent->getLodTransitionFadePercentage() >= 1.0f)) {
-      _tilesToHideNextFrame.push_back(pTile);
-    }
-  }
+  if (this->_pVoxelRendererComponent) {
+    this->_pVoxelRendererComponent->UpdateTiles(
+        pResult->tilesToRenderThisFrame,
+        pResult->tileScreenSpaceErrorThisFrame);
+  } else {
```
kring: A single tileset could have both voxel and non-voxel content, right? In that cause this else would be wrong, I think.

Source/CesiumRuntime/Private/Cesium3DTileset.cpp
```diff
@@ -2320,6 +2370,53 @@ void ACesium3DTileset::RuntimeSettingsChanged(
...
+        LogCesium,
+        Error,
+        TEXT(
+            "Tileset %s contains voxels but is missing a metadata schema to describe its contents."),
+        *this->GetName())
+    return;
+  }
+
+  const FCesiumVoxelClassDescription* pVoxelClassDescription =
+      this->_voxelClassDescription ? &(*this->_voxelClassDescription) : nullptr;
```
kring: It's not clear to me why this `_voxelClassDescription` is squirreled away earlier, and then accessed here. Why not get the description directly from the voxel metadata component here?

kring:
Here are some initial review comments. There's a lot here and I haven't been through everything yet. Overall it looks great, though!

* I noticed that the tileset disappears when I get too close to it (inside its bounding volume, presumably). How realistic is it to avoid that? I was hoping to be able to fly through the voxel field. 😀

* I can get this weird lighting effect just by rotating the camera. I say it's lighting related because it goes away in the "Unlit" view. Any idea what this is?
lighting-flash

This is my (not particularly well considered) shade function: 

```
return Properties.Temp_C_NO_DATA == Properties.Temp_C ? float4(0.0, 0.0, 0.0, 0.0) : lerp(float4(0.0, 1.0, 0.0, 0.02), float4(0.0, 0.0, 1.0, 0.02), (Properties.Temp_C - 10.0) / 100.0);
```

Source/CesiumRuntime/Private/Cesium3DTileset.cpp
j9liu: I was copying what we're doing for `CesiumFeaturesMetadataComponent` (and the corresponding `_featuresMetadataDescription` variable). But I see your point in both cases 😁

Source/CesiumRuntime/Private/Cesium3DTileset.cpp
j9liu: Most of our early datasets only contain voxels, but I see your point and suppose there's no harm in doing both.

j9liu:
Finally getting around to @kring's comments...

> I noticed that the tileset disappears when I get too close to it (inside its bounding volume, presumably). How realistic is it to avoid that? I was hoping to be able to fly through the voxel field. 😀

This should be possible if you check **Two Sided** on the resulting material! Here's a closeup from being inside the cube but still seeing the voxels:
image

This also allows you to fly into the voxel volume, but you just end up seeing a solid color of whatever voxel you're in. Maybe that's expected for this particular dataset??

image

> I can get this weird lighting effect just by rotating the camera. I say it's lighting related because it goes away in the "Unlit" view. Any idea what this is?

I was still seeing this even after I switched to the Unlit lighting mode. It appears to be the result of the atmosphere's interaction with translucent materials. Here's the difference after I toggle `CesiumSunSky`:

| Before | After |
| - | - |
| image | image |

The sharp line across the back is the result of material being two-sided, but the light is still there when it's toggled off:

image

So I'm not sure how to get rid of this interaction. Maybe this is an artifact of using a cube mesh for rendering, rather than doing a separate post process / draw pass. I would be willing to explore the latter, but that will probably be another week or so of effort. Is it bad enough to hold up this PR?

Source/CesiumRuntime/Private/CesiumGltfComponent.cpp
```diff
@@ -1479,348 +1482,340 @@ static void loadPrimitive(
...
-
-  if (primitive.hasExtension<CesiumGltf::ExtensionKhrGaussianSplatting>()) {
-    // Data in a gaussian splat is handled differently from other mesh data.
-    primitiveResult.GaussianSplatData =
-        MakeUnique<FCesiumGltfGaussianSplatData>(model, primitive);
-  } else {
-    TUniquePtr<FStaticMeshRenderData> RenderData =
-        MakeUnique<FStaticMeshRenderData>();
-    RenderData->AllocateLODResources(1);
+  TUniquePtr<FStaticMeshRenderData> pRenderData =
```
j9liu, line 1484: This restores the `loadPrimitive` code so it only deals with typical 3D mesh data; the splats route has been moved to `loadGaussianSplats`

Source/CesiumRuntime/Private/CesiumGltfComponent.cpp
```diff
@@ -3095,37 +3284,15 @@ static void loadPrimitiveGameThreadPart(
...
-        NewObject<UCesiumGltfGaussianSplatComponent>(pGltf, componentName);
-    pGaussianSplat->Data = MoveTemp(*loadResult.GaussianSplatData.Release());
-    pGaussianSplat->SetupAttachment(pGltf);
-    pGaussianSplat->RegisterComponent();
-    pGaussianSplat->SetMobility(pGltf->Mobility);
-    pGaussianSplat->RegisterWithSubsystem();
-    pCesiumPrimitive = pGaussianSplat;
-    createMesh = false;
-  } else if (meshPrimitive.mode == CesiumGltf::MeshPrimitive::Mode::POINTS) {
+  if (meshPrimitive.mode == CesiumGltf::MeshPrimitive::Mode::POINTS) {
```
j9liu, line 3283: Same story here; splats / voxels have their own `load___GameThread` function

Source/CesiumRuntime/Private/CesiumVoxelRendererComponent.cpp
```diff
@@ -0,0 +1,696 @@
...
+
+      f(i, pVoxelComponent);
+    }
+  }
+}
+} // namespace
+
+void UCesiumVoxelRendererComponent::UpdateTiles(
+    const std::vector<Cesium3DTilesSelection::Tile::ConstPointer>& VisibleTiles,
+    const std::vector<double>& VisibleTileScreenSpaceErrors) {
```
j9liu, line 535: **Context:** This approach tries to match CesiumJS, which uses a custom traversal algorithm. Their algorithm uses a priority queue based on screen space error to retain parent detail more aggressively (avoiding LOD skips and holes).

I'm happy to take other suggestions or experiment with this, esp. because I feel like it's somewhat wasteful to do sort based on SSE every frame when Native is already handling tiles to us. At the same time, there needs to be a heuristic to determine when it's acceptable to kick a tile from its megatexture slot, since that visibility info is not sync'd 1-to-1 with Native's.

j9liu:
Happy to mark this as ready for another review! I left a few comments that explain some of the decisions made, but please let me know if there's anything else I can clear up. I'm also happy to do a "pair programming" session to walk through some of the larger chunks of code.

@azrogers, would love to have your eyes on this, especially since you've dabbled in non-traditional rendering techniques (i.e. splats) and I figure you'll bring a fresh perspective
@kring, if you get the time to do another pass, I would greatly appreciate that too!

Content/GaussianSplatting/M_CesiumSplatMaterial.uasset
azrogers, line 1: I would be wary about moving this - it's in the `GaussianSplatting` folder for a reason. This folder is specified in [`DirectoriesToAlwaysCook`](https://github.com/CesiumGS/cesium-unreal/blob/main/Config/Game.ini).  This is because the splat system is never saved to the level, so Unreal won't know it's supposed to be referenced and won't include it in a shipping build.

Shaders/Private/CesiumRayIntersection.usf
```diff
@@ -0,0 +1,101 @@
...
+
+struct Ray
+{
+  float3 Origin;
+  float3 Direction;
+};
+
+struct Intersection
+{
+  float t;
```
azrogers, line 27: What is `t`? Is it possible to give this a descriptive name like `Origin` or `Direction` or `Normal` have?

Shaders/Private/CesiumShaderConstants.usf
```diff
@@ -0,0 +1,10 @@
+// Copyright 2020-2026 CesiumGS, Inc. and Contributors
+
+/*===========================
+  CesiumShaderConstants.usf: Definitions of common constants for Cesium shaders.
+=============================*/
+
+#define CZM_INFINITY 5906376272000.0 // Distance from the Sun to Pluto in meters.
```
azrogers: Why is this set to the distance from the Sun to Pluto? Like, it's a very far distance, but it seems pretty arbitrary. Is there something in the math that limits how big this number can be? If not, we should just use the same value as `std::numeric_limits<float>::max()` in C++, which I believe you can get with `asfloat(0x7F7FFFFF)` (instead of writing out the full float representation).

Shaders/Private/CesiumShape.usf
```diff
@@ -0,0 +1,96 @@
...
+    
+  /**
+  * Tests whether the input ray (Unit Space) intersects the shape.
+  */
+  RayIntersections Intersect(in Ray R)
+  {
+    RayIntersections result;
+    
+    [branch]
+    switch (ShapeConstant)
```
azrogers, line 56: This seems like a fairly lightweight switch statement to be given the `[branch]` attribute. Seems like flattening it would probably be more performant. Though I suppose that there will be more shapes added in the other PRs so it won't be as lightweight then.

Shaders/Private/CesiumVoxelOctree.usf
```diff
@@ -0,0 +1,305 @@
...
+=============================================================================*/
+
+#include "CesiumShape.usf"
+
+// An octree node is encoded by 9 texels.
+// The first texel contains the index of the node's parent.
+// The other texels contain the indices of the node's children.
+#define TEXELS_PER_NODE 9
+
+#define OCTREE_MAX_LEVELS 32 // Hardcoded because HLSL doesn't like variable length loops
```
azrogers, line 14: It'll actually handle them just fine if you use the `[loop]` attribute on the `for` statement. Worth performance testing.

Shaders/Private/CesiumRayIntersection.usf
azrogers: I'm open to the argument that we just have to accept that this is what mathematics has decided to call a variable in this role, because I'm having trouble thinking of a descriptive name that isn't wordy (like `NormalizedPosition`).

Source/CesiumRuntime/Private/CesiumVoxelMetadataComponent.cpp
```diff
@@ -0,0 +1,1032 @@
...
+    }
+  }
+
+  NodeX = DataSectionX;
+  NodeY = DataSectionY;
+
+  // Inspired by HLSLMaterialTranslator.cpp. Similar to MaterialTemplate.ush,
+  // CesiumVoxelTemplate.usf contains "%s" formatters that will be replaced
+  // with generated code.
+  FLazyPrintf LazyPrintf(*ResourceLibrary.ShaderTemplate);
```
azrogers: I wonder if named arguments might work, either via `fmt` or via Unreal's `FString::Format`. I can see the curly brackets getting confused for formatting parameters. But if possible it would make the templates clearer and easier to edit in the future.

Shaders/Private/CesiumVoxelTemplate.usf
```diff
@@ -0,0 +1,227 @@
...
+
+// Pass textures to the VoxelMegatextures struct.
+%s
+
+float CurrentT = Intersections.Entry.t;
+float EndT = Intersections.Exit.t;
+float3 PositionUV = R.Origin + CurrentT * R.Direction;
+
+// The Jacobian is necessary to compute voxel intersections with respect to the grid's shape.
+float3x3 JacobianT;
```
azrogers, line 174: Image

Shaders/Private/CesiumVoxelTemplate.usf
```diff
@@ -0,0 +1,227 @@
...
+  }
+
+  PositionUV = R.Origin + CurrentT * R.Direction;
+  PositionShapeUVSpace = Octree.GridShape.ConvertUVToShapeUVSpace(PositionUV, JacobianT);
+  Octree.ResumeTraversal(PositionShapeUVSpace, Traversal, Sample);
+  NextIntersection = Octree.GetNextVoxelIntersection(Sample, RawDirection, Intersections, JacobianT, CurrentT);
+}
+
+// Convert the alpha from [0,ALPHA_ACCUMULATION_MAX] to [0,1]
+AccumulatedResult.a /= ALPHA_ACCUMULATION_MAX;
```
azrogers: Is there a reason that `ALPHA_ACCUMULATION_MAX` isn't just 1.0?

Source/CesiumRuntime/Private/CesiumGltfComponent.cpp
```diff
@@ -96,6 +99,8 @@ class HalfConstructedReal : public UCesiumGltfComponent::HalfConstructed {
 };
 } // namespace
 
+const int32_t VoxelPrimitiveMode = 2147483647;
```
azrogers: It doesn't feel like this should be defined here. It feels like it should be in Cesium Native somewhere. Seems a bit bizarre to me that voxels use a custom primitive mode that goes against the spec instead of using something like `POINTS`, but whatever.

Source/CesiumRuntime/Private/CesiumGltfComponent.cpp
```diff
@@ -1888,6 +1884,181 @@ static void loadIndexedPrimitive(
...
+        Warning,
+        TEXT(
+            "Voxel primitive padding does not match the padding specified by the tileset. Skipped."));
+    return;
+  }
+
+  const CesiumGltf::ExtensionMeshPrimitiveExtStructuralMetadata*
+      pPrimitiveMetadata = primitive.getExtension<
+          CesiumGltf::ExtensionMeshPrimitiveExtStructuralMetadata>();
+  if (!pPrimitiveMetadata || pPrimitiveMetadata->propertyAttributes.empty()) {
```
azrogers, line 1980: Is it really necessary to have metadata to render voxels? Like, as a requirement of the spec?

Source/CesiumRuntime/Private/CesiumGltfComponent.cpp
```diff
@@ -1888,6 +1884,181 @@ static void loadIndexedPrimitive(
...
+        primitive.mode);
+    return;
+  }
+
+  const CesiumGltf::Model& model =
+      *options.pMeshOptions->pNodeOptions->pModelOptions->pModel;
+  const CesiumGltf::Mesh& mesh = model.meshes[options.pMeshOptions->meshIndex];
+
+  double scale = 1.0 / CesiumPrimitiveData::positionScaleFactor;
+  glm::dmat4 scaleMatrix = glm::dmat4(
```
azrogers: Might make sense to make this scale matrix itself a constant, since it's used in both loading splats and loading primitives.

Source/CesiumRuntime/Private/CesiumMetadataComponent.cpp
```diff
@@ -0,0 +1,60 @@
+// Copyright 2020-2025 CesiumGS, Inc. and Contributors
+
+#include "CesiumMetadataComponent.h"
+#include "Cesium3DTileset.h"
+
+#include <Cesium3DTilesSelection/Tileset.h>
+#include <Cesium3DTilesSelection/TilesetMetadata.h>
+
+void UCesiumMetadataComponent::SyncStatistics() {
+  if (this->_syncInProgress)
```
azrogers, line 10: I don't think we ever put this in the style guide but I strongly feel like putting brackets around one-line statements is the way to go. It can produce some pretty nasty bugs if people aren't careful.

Source/CesiumRuntime/Private/CesiumVoxelMetadataComponent.cpp
```diff
@@ -0,0 +1,1032 @@
...
+      normalizedHlslType = GetHlslTypeForEncodedType(
+          Property.EncodingDetails.Type,
+          ECesiumEncodedMetadataComponentType::Float);
+      FString rawPropertyName = PropertyName + MaterialPropertyRawSuffix;
+
+      // If the property is normalized, the encoded type actually corresponds to
+      // the raw data values. A second line for the normalized value is added.
+      // e.g., "uint8 myProperty_RAW; float myProperty;"
+      // clang-format off
+      DeclareShaderProperties += "\t" + encodedHlslType + " " + rawPropertyName +
```
azrogers: You can do `DeclareShaderProperties.Appendf("\t%s %s;\n\t%s %s;", encodedHlslType, rawPropertyName, normalizedHlslType, PropertyName)` which is a bit cleaner and possibly more efficient (not that performance is really that crucial here).

Source/CesiumRuntime/Private/CesiumVoxelRendererComponent.cpp
```diff
@@ -0,0 +1,697 @@
...
+      nullptr,
+      FName("VoxelMaterial"));
+  pVoxelMaterial->SetFlags(
+      RF_Transient | RF_DuplicateTransient | RF_TextExportTransient);
+
+  EVoxelGridShape shape = pVoxelComponent->Options.gridShape;
+
+  pVoxelMaterial->SetTextureParameterValueByInfo(
+      FMaterialParameterInfo{
+          UTF8_TO_TCHAR("Octree"),
```
azrogers: You can just use `TEXT("Octree")` here. It will output a `wchar_t` literal, instead of `UTF8_TO_TCHAR` which has to allocate a whole new struct and convert from UTF-8 to UCS-2 or whatever the platform's `wchar_t` is. Same goes for all the other literals.

Source/CesiumRuntime/Private/CesiumVoxelRendererComponent.cpp
azrogers: For now I think it's probably fine. I don't think sort performance will start to be an issue until we're trying to display tens or hundreds of thousands of tiles at the same time, which seems pretty unlikely.

Source/CesiumRuntime/Private/CesiumVoxelRendererComponent.cpp
```diff
@@ -0,0 +1,697 @@
...
+    // moving too, so in a relative sense the object isn't actually moving.
+    // This isn't a perfect assumption, of course.
+    this->_pMeshComponent->SetRelativeTransform_Direct(transform);
+    this->_pMeshComponent->UpdateComponentToWorld();
+    this->_pMeshComponent->MarkRenderTransformDirty();
+  }
+}
+
+double UCesiumVoxelRendererComponent::computePriority(double sse) {
+  return 10.0 * sse / (sse + 1.0);
```
azrogers: What's the rationale for calculating priority this way instead of just using the sse?

Source/CesiumRuntime/Private/VoxelMegatextures.cpp
```diff
@@ -0,0 +1,382 @@
...
+  }
+
+  glm::uvec3 textureDimensions = this->_tileCountAlongAxes * slotDimensions;
+
+  this->_maximumTileCount = this->_tileCountAlongAxes.x *
+                            this->_tileCountAlongAxes.y *
+                            this->_tileCountAlongAxes.z;
+
+  // Initialize the data slots.
+  this->_slots.resize(this->_maximumTileCount, Slot());
```
azrogers, line 117: Seems like we could use `CesiumUtility::DoublyLinkedList` instead of reimplementing a linked list here?

Source/CesiumRuntime/Private/CesiumMetadataEncodingDetails.cpp
```diff
@@ -149,3 +149,21 @@ FCesiumMetadataEncodingDetails::GetBestFitForProperty(
...
+    const FCesiumMetadataPropertyDetails& PropertyDetails) {
+  TArray<ECesiumEncodedMetadataConversion> result;
+  if (PropertyDetails.Type == ECesiumMetadataType::Invalid) {
+    return result;
+  }
+
+  result.Reserve(2);
+  result.Add(ECesiumEncodedMetadataConversion::Coerce);
+
+  if (PropertyDetails.Type == ECesiumMetadataType::String) {
```
azrogers, line 163: It's not a big deal but it's probably worth removing the `Reserve` here - allocating twice the memory necessary for every metadata type is probably less efficient than doing an extra allocation just for the string types.

Source/CesiumRuntime/Private/CesiumVoxelMetadataComponent.cpp
```diff
@@ -0,0 +1,1032 @@
...
+UCesiumVoxelMetadataComponent::UCesiumVoxelMetadataComponent() {
+  // Structure to hold one-time initialization
+  struct FConstructorStatics {
+    ConstructorHelpers::FObjectFinder<UTexture> DefaultVolumeTexture;
+    FConstructorStatics()
+        : DefaultVolumeTexture(
+              TEXT("/Engine/EngineResources/DefaultVolumeTexture")) {}
+  };
+  static FConstructorStatics ConstructorStatics;
+  _pDefaultVolumeTexture = ConstructorStatics.DefaultVolumeTexture.Object;
```
azrogers: `this->`

Source/CesiumRuntime/Private/VoxelMegatextures.cpp
```diff
@@ -0,0 +1,382 @@
...
+  ([pTexture = data.pTexture,
+    format = data.encodedFormat.format,
+    &property,
+    updateRegion,
+    texelSizeBytes = data.texelSizeBytes,
+    pData](FRHICommandListImmediate& RHICmdList) {
+    FTextureResource* pResource =
+        IsValid(pTexture) ? pTexture->GetResource() : nullptr;
+    if (!pResource)
+      return;
```
azrogers, line 204: As per previous comment, add braces to this.

Source/CesiumRuntime/Private/VoxelMegatextures.cpp
```diff
@@ -0,0 +1,382 @@
...
+        pData);
+  });
+}
+
+/*static*/ void FVoxelMegatextures::incrementalWriteToTexture(
+    const FCesiumPropertyAttributeProperty& property,
+    const FVoxelMegatextures::TextureData& data,
+    const FUpdateTextureRegion3D& updateRegion) {
+  if (!data.pTexture)
+    return;
```
azrogers, line 226: And here, and below too.

Source/CesiumRuntime/Private/CesiumFeaturesMetadataComponent.cpp
```diff
@@ -2644,9 +2556,3 @@ void UCesiumFeaturesMetadataComponent::PostLoad() {
 
   Super::PostLoad();
 }
```
azrogers, line 2558: It doesn't look like an additional call to replace `SyncStatistics` was added anywhere else in the file to do what this `OnRegister` method was previously doing. Was it just unneeded in the first place?

Shaders/Private/CesiumShape.usf
j9liu: There are two other shape options to add. Personally I like how `switch` statements read more than chained `if`s, but I could go either way. What do you think?

Shaders/Private/CesiumShaderConstants.usf
j9liu: Hah, I stole this from CesiumJS so you'll have to ask the person who added that :shrug: It is a bit arbitrary, though, so I'll go ahead and change the value.

Source/CesiumRuntime/Private/CesiumGltfComponent.cpp
j9liu: Hm yeah, I thought I left a comment about that here but maybe I removed it by accident. I can move it.

> Seems a bit bizarre to me that voxels use a custom primitive mode that goes against the spec instead of using something like POINTS, but whatever.

It's weird because voxels are just metadata blobs without explicit positions. Maybe we could use `POINTS` and suggest a fallback for inferring the point positions relative to the voxel grid. Will have to put a note on the spec for that: https://github.com/KhronosGroup/glTF/pull/2496

Source/CesiumRuntime/Private/CesiumGltfComponent.cpp
j9liu: It isn't required by the spec, but not sure how Unreal could act upon the voxel attributes otherwise.

CesiumJS doesn't live by this requirement because it can freely pass arbitrary vertex attributes to its shaders as it detects them. In Unreal, we are bound to whatever attributes Unreal allows us to pass, and can't just add custom vertex attributes on the fly. There has to be an element of user intervention to even act on the metadata, so given what we do in `UCesiumFeaturesMetadataComponent`, this seemed like the best way to maintain consistency.

Source/CesiumRuntime/Private/CesiumMetadataComponent.cpp
j9liu: Think I picked up this habit from reviewing too many of @kring's PRs 😛

Source/CesiumRuntime/Private/CesiumFeaturesMetadataComponent.cpp
j9liu: `OnRegister` got moved to the `UCesiumMetadataComponent` base class, which itself calls `SyncStatistics`. So it is indeed needed, but with inheritance I think this should be covered. (Did I read your question right?)

Source/CesiumRuntime/Private/CesiumVoxelRendererComponent.cpp
j9liu: Honestly just copy paste from CesiumJS. I think this was weighting SSE along with other metrics (specifically for time-dynamic voxels), but you're right, there's no reason to use it here.

Source/CesiumRuntime/Private/VoxelMegatextures.cpp
j9liu: Unfortunately we need to maintain the slots' order so we can free the right ones based on their indices. If I used `CesiumUtility::DoublyLinkedList` that order would be lost, and it seems redundant to store both the `std::vector` of slots (to preserve index order) and a linked list for connectivity.

Shaders/Private/CesiumVoxelTemplate.usf
j9liu: This is also a relic of copy paste from CesiumJS, where it changes depending on whether it's the render pass vs. the picking pass. Not useful here, so I'll remove it.

Shaders/Private/CesiumRayIntersection.usf
j9liu: I put a comment that should hopefully explain it. LMK if you think it needs more help

Source/CesiumRuntime/Private/CesiumVoxelMetadataComponent.cpp
j9liu: I replaced it with `FString::Format`, though all the examples I saw used `{0}, {1}` etc. I'm not sure how much of an improvement that is, but LMK if I should change this further.

Source/CesiumRuntime/Private/VoxelMegatextures.cpp
j9liu: Thinking about this again just to be fair to the idea, we could maintain two `CesiumUtility::DoublyLinkedList`s for occupied/empty slots, and then have a `std::vector` of pointers to maintain that index -> slot mapping. Would that be more worth it? ~Guess it depends on the ratio of `sizeof(struct)` vs. a pointer...~

EDIT: I think using a `DoublyLinkedList` is essentially moving the `pHead/pNext` memory to that list, which is fine, but then we're adding `N` more memory for the pointers in `std::vector`. So it feels like it's not worth, unless I'm missing something

Source/CesiumRuntime/Private/CesiumGltfComponent.cpp
j9liu: Opened https://github.com/CesiumGS/cesium-native/pull/1329 for this.

Shaders/Private/CesiumVoxelOctree.usf
j9liu: I didn't notice a performance difference either way, but it could potentially matter on a larger dataset (especially with #1711).

For now I've used `[loop]` on `ResumeTraversal` where there's a clear limit on the number of steps up. For `BeginTraversal`, we don't pass the max number of octree levels to the shader. We could do so, of course, but not sure if we gain much from it atm. I'm thinking I'll keep the constant around, then more seriously test eliminating this in #1711. Does that sound good?

j9liu:
> there's some sort of issue where duplicate material nodes are generated by CesiumVoxelMetadata, so that's the major thing to fix here.

Finally found the reason why it was happening -- doing `pMaterial->GetExpressionCollection()->RemoveExpression` isn't enough to actually remove the node from the material. I think there's some cache/lifetime magic happening under the hood that I wasn't accounting for. Thankfully `UMaterialEditingLibrary` exists to do it correctly for me :)

I believe I've addressed all the comments (or at least responded to them). Ready for another review @azrogers !

Source/CesiumRuntime/Private/CesiumVoxelMetadataComponent.cpp
azrogers: [It is possible to use named arguments](https://dev.epicgames.com/documentation/en-us/unreal-engine/API/Runtime/Core/FString/Format), though the Unreal website has zero information about it. I believe I've found snippets of code using it online in the past. Not worth holding up the review over, though, ordered arguments is fine.

j9liu:
I'm resolving the merge conflict + debugging the unit tests, one sec...

azrogers:
Thanks @j9liu!

PR #1687, opened 2025-06-02
Support Materials, not just MaterialInstances, and fix normal flip when computing tangents

### LINKED ISSUE #1686: 15_CesiumArchitecturalMetadataStyling level shows floors in v2.16.1
Because the floors are visible, it's hard to see the pipes. I'm not sure what has changed to make these floors appear.

Image

### LINKED ISSUE #1688: Tiles with water have incorrect normals
It's clearly visible in the "World Normal" buffer visualization:
Image

I think this is regression since v2.16.0.

kring:
Fixes #1686 
Fixes #1688

The modifications to support dynamic material instances also inadvertently broke the case where the user-provided material is not a material instance at all.

Also, the restructuring of the mesh creation introduced a bug where the Y component of the normals was inadvertently flipped when computing tangents.

No changelog entry because this is a regression since the last release.

I'll be merging this myself because we need to fix these two regressions for today's release.


j9liu:
Whoops, sorry for the trouble and thanks for adding those fixes!😅

PR #1690, opened 2025-06-04
Metadata: Add a way to propagate the NullFeatureId to the GPU

Julot94:
when using UCesiumFeaturesMetadataComponent, typically. This can be useful if one wants to apply a specific treatment to vertices associated with no Feature, in the shader (by default, such vertices are associated to the feature of ID 0, which is usually a valid value!)

## Author checklist

- [ ] I have submitted a [Contributor License Agreement](https://github.com/CesiumGS/community/tree/main/CLAs) (only needed once).

### DIFF
--- a/Source/CesiumRuntime/Private/CesiumFeaturesMetadataComponent.cpp
+++ b/Source/CesiumRuntime/Private/CesiumFeaturesMetadataComponent.cpp
@@ -317,9 +317,10 @@ void AutoFillFeatureIdSetDescriptions(
       pDescription->PropertyTableName = getNameForPropertyTable(propertyTable);
     }
 
-    pDescription->bHasNullFeatureId =
-        UCesiumFeatureIdSetBlueprintLibrary::GetNullFeatureID(featureIDSet) >
-        -1;
+    const int64 nullFeatureId =
+        UCesiumFeatureIdSetBlueprintLibrary::GetNullFeatureID(featureIDSet);
+    pDescription->NullFeatureId = (nullFeatureId > -1) ? nullFeatureId : -1;
+    pDescription->bHasNullFeatureId = (nullFeatureId > -1);
 
     if (type == ECesiumFeatureIdSetType::Texture) {
       FCesiumFeatureIdTexture featureIdTexture =
--- a/Source/CesiumRuntime/Private/CesiumGltfComponent.cpp
+++ b/Source/CesiumRuntime/Private/CesiumGltfComponent.cpp
@@ -193,6 +193,11 @@ void accumulateFeaturesMetadataAccessors(
               primitive,
               attributeName,
               gltfToUnrealTexCoordMap));
+      if (encodedFeatureIDSet.nullFeatureId) {
+        primitiveResult.AccessorToNullFeatureIdMap.emplace(
+            accessorIndex,
+            encodedFeatureIDSet.nullFeatureId.value());
+      }
     } else if (encodedFeatureIDSet.texture) {
       int64 setIndex = encodedFeatureIDSet.texture->textureCoordinateSetIndex;
       std::string attributeName = "TEXCOORD_" + std::to_string(setIndex);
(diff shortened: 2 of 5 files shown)


Julot94:
This is more a safety measure we had written when we were baking the feature Ids in UVs in our plugin, independently from _UCesiumFeaturesMetadataComponent_. Now that we use _UCesiumFeaturesMetadataComponent_, we just reported it - as it seemed useful to be able to distinguish the case "no feature" from the feature #0, which is a valid one, inside the shader.

j9liu:
Thanks for the PR @Julot94 ! One question before I review:

When I wrote `UCesiumFeaturesMetadataComponent`, my assumption was that a set of vertices not associated with feature could just use `nullFeatureId` in `EXT_primitive_features`. But it sounds like you're trying to account for primitives that don't contain `EXT_primitive_features` as well -- is that correct? If so, I think this is a worthy addition for sure!

Julot94:
Hi @j9liu ,
Thank you for your quick feedback. Sorry, I don't find any reference to `EXT_primitive_features`, searching [here](https://github.com/KhronosGroup/glTF/tree/main/extensions), but maybe I'm not looking at the right place? In the cesium-native repository, I just found `ExtensionExtInstanceFeaturesFeatureId`, and we do not use it either - so I guess we are in the case which you described: `UCesiumFeaturesMetadataComponent ` without `EXT_primitive_features` (but I might be wrong, since I did not find that extension...)

j9liu:
Sorry for the confusion @Julot94, I meant `EXT_mesh_features`. My mind confused it with the `FCesiumPrimitiveFeatures` struct in Unreal 🤦

Julot94:
Hi @j9liu ,
Thank you for your answer. In fact, there are 2 very distinct aspects in this PR, and I should have detailed them more.

- on the one hand, you are right, the modification in `FCesiumFeatureIdSetDescription ` (addition of the property `NullFeatureId`) was done to enforce the value of the nullFeatureId value in case it is not specified from the source glTF file. It may not be necessary at all, even though it was a quick way for me to make it work in our case, without having to request a change in the iTwin Mesh Export Service...
- on the other hand, I modified the `copyFeatureIds` functions to take the nullFeatureId into account when we get a negative value, instead of falling back to 0 as you currently do with
`glm::max(featureId, 0.0f)`
So instead of using 0.0f when UCesiumFeatureIdAttributeBlueprintLibrary::GetFeatureID returns -1, I thought it safer to use the nullFeatureId (if any) defined for the corresponding set. But maybe the `glm::max(featureId, 0.0f)` was just added for robustness for a case which is not supposed to happen (ie. in reality, if the glTF model is conformant to the spec, there should *never* be a negative value here), and you added the test just for safety... Even in such case, it could be preferable not to associate those "degenerated" vertices to a valid feature, and prefer assigning them the nullFeatureId.
However, since the modification is not totally free (as I needed to introduce a map to recover the correct nullFeatureId, following the recent refactoring of this code), we could just abandon the idea.

j9liu:
Thank you for the detailed explanation, @Julot94! I think I understand the intentions of this PR now.

I agree that it doesn't make conceptual sense to use zero as a default feature ID. The reason for it due to our material styling implementation in `UCesiumFeaturesMetadataComponent`. Property tables are represented by textures in Unreal materials, and feature IDs are used to sample them at the corresponding indices. So I avoid passing in negative feature IDs because I don't want to access textures with a negative index.

However, even with the above rationale, this is definitely not the best solution. Perhaps this would work better if:

- `bHasNullFeatureId` is used to make the material generate an if statement to handle a potential null feature ID, like so:
image
- We can force these nodes to _always_ generate. Then, `bHasNullFeatureId` doesn't have to exist.
- The default value used in `copyFeatureIds` can be -1 instead of 0. Same for the generated `_NULL_ID` parameter in the Unreal materials.

How does that sound? I acknowledge that it is a bit more work than your current solution, but I think it would handle both use cases nicely. And I'm willing to open a PR for this so you don't have to learn a new system in our plugin :smile:

Julot94:
It sound good to me! (I guess the impact of the additional IF statement would be negligible...)

j9liu:
@Julot94 I opened #1697 for this. Let me know if it works for your use case!

> (I guess the impact of the additional IF statement would be negligible...)

Yeah, I'm not a fan of `if`s in shaders either 😅 But I think one `if` statement is reasonable, especially if most / all threads go down the same branch.

Julot94:
Thank you @j9liu ! Yes, it will work for our case. So I will close this PR now :-)

PR #1691, opened 2025-06-05
Add support for Unreal Engine 5.6

kring:
Draft because it's not completely working yet.

PR #1692, opened 2025-06-07
Add Unreal Engine 5.6 for all non-Apple platforms

kring:
This is the same as #1691, minus the Apple platforms. It seems Apple may take a little longer to get working, but I'd like to get a build for UE 5.6 on the other platforms out as soon as possible.


kring:
Closing this in favor of #1691.

Issue #1689, opened 2025-06-03
Add support for property attribute properties


j9liu:
### Feature

Currently we don't support access to property attributes, nor the styling of them. It shouldn't be too much of a lift to add support for `FCesiumPropertyAttribute` and `FCesiumPropertyAttributeProperty`. This would also simplify a lot of things for #1685, so I'm tempted to just do this off-the-bat.

Property attribute styling is less straightforward. 😕 There isn't an easy way to pass custom vertex attributes to Unreal materials without writing a custom vertex factory / shader (which is a lot of effort), and also isn't dynamic to work with the same way that Unreal materials are. Some options are...

- **Attempt to pass properties through unused UV channels.** We already do this for per-vertex feature IDs. However, metadata properties are more arbitrary, so we could have something like vec4 properties which would take up 2 full UV sets. There is a hard limit on the number of properties we could support because we could only pass up to 16 scalar values, and this space is further compromised if the model comes with its own UV coordinates.
- **Copy what we did for property tables.** Upload the entire attribute buffer to a texture, and then access the values by index. The vertex index must be stored in a UV channel because there is no "Vertex Index" node in Unreal material graphs, but it only takes one UV channel to enable an arbitrary amount of properties.

The latter seems best for the most immediate value and reliability. I think in that case, it also wouldn't be too much of a lift.

j9liu:
### API Design Nitpicks
In the glTF's top-level `EXT_structural_metadata`, a property attribute just maps each property name to the attribute ID it should look for in the glTF, e.g.:

```json
"propertyAttributes": [{
        "class": "movement",
        "properties": {
          "direction": {
            "attribute": "_DIRECTION",
          },
          "magnitude": {
            "attribute": "_MAGNITUDE",
          }
        }
      }]
```

Then, a glTF primitive's `EXT_structural_metadata` specifies the indices of the property attribute that it uses:

```json
  "primitives": [
    {
      "mode:": 0, 
      "attributes": {
        "POSITION": 0,
        "_DIRECTION": 1,
        "_MAGNITUDE": 2,
      },
      "extensions": {
        "EXT_structural_metadata": {
          "propertyAttributes": [0]
        }
      }
    }
  ]
```

[...]

PR #1694, opened 2025-06-11
Store property attributes in `FCesiumPrimitiveMetadata`

j9liu:
## Description

This PR is in support for #1685. This generalizes access of voxel property attributes regardless of accessor structure (because that's all handled by `AccessorView` under the hood). It also enable would enable non-`uint8`/`float` variables to be passed to the data textures.

Summary:

- Adds `FCesiumPropertyAttribute` and `FCesiumPropertyAttributeProperty`
- Adds `BlueprintLibrary` functions to retrieve values from those properties
- Adds arrays of `FCesiumPropertyAttribute` to `FCesiumPrimitiveMetadata`. See [this comment](https://github.com/CesiumGS/cesium-unreal/issues/1689#issuecomment-2937163258) for justification on this approach.
- Deprecates `FCesiumPrimitiveMetadataBlueprintLibrary::GetPropertyAttributeIndices`. This wasn't doing anything helpful to begin with :sweat:
- Introduces various style fixes in the other metadata files.
- Adds `GetInteger64` for `FCesiumPropertyTextureProperty`. Although int64's aren't directly supported, this enables `uint32` properties to be losslessly retrieved.

## Issue number or link

Partially addresses #1689, but not fully.

## Author checklist

- [X] I have added or updated unit tests to ensure consistent code coverage as necessary.

## Testing plan

This is difficult because there's no good way to implement property attribute picking (how would you pick a single vertex from a model?), and I didn't want to increase the scope of this to property attribute styling. I hope the unit tests can inspire enough confidence :')


Source/CesiumRuntime/Private/Tests/CesiumPropertyTableProperty.spec.cpp
```diff
@@ -26,7 +26,7 @@ void FCesiumPropertyTablePropertySpec::Define() {
           UCesiumPropertyTablePropertyBlueprintLibrary::
               GetPropertyTablePropertyStatus(property),
           ECesiumPropertyTablePropertyStatus::ErrorInvalidProperty);
-      TestEqual<int64>(
+      TestEqual<int64_t>(
```
j9liu, line 29: This diff is large, but I basically swept for `int64` -> `int64_t`, since I remember `int64` having a dubious definition on some platforms.

Additionally, I switched out `static_cast<int64_t>()` for `int64_t()` in accordance with the cesium-native [style guide](https://cesium.com/learn/cesium-native/ref-doc/style-guide.html):

> Use constructor notation for very safe integer conversions (where the range is checked) rather than `static_cast`.

kring: Cross-link to the related bug we had awhile back:
https://github.com/CesiumGS/cesium-unreal/pull/1468#issuecomment-2204840573

Source/CesiumRuntime/Private/CesiumPropertyAttribute.cpp
```diff
@@ -0,0 +1,127 @@
...
+}
+
+/*static*/ TMap<FString, FCesiumMetadataValue>
+UCesiumPropertyAttributeBlueprintLibrary::GetMetadataValuesAtIndex(
+    UPARAM(ref) const FCesiumPropertyAttribute& PropertyAttribute,
+    int64 Index) {
+  TMap<FString, FCesiumMetadataValue> values;
+  if (Index < 0) {
+    return values;
+  }
```
kring: I'm curious about the reason for this check. If I'm reading this right, an Index less than 0 will return an empty map. But an Index after the last vertex will return a map containing all of the properties and an empty/error FCesiumMetadataValue for each. I don't have a strong preference for either behavior, but I don't see any reason they should be different.

kring (commented):
Looks great! I just have one small comment.

Source/CesiumRuntime/Private/CesiumPropertyAttribute.cpp
j9liu: That's definitely not intentional, thanks for the catch!

j9liu:
Thanks @kring for the review! This is ready for another look.

kring:
Thanks @j9liu!

PR #1697, opened 2025-06-17
Change default null feature ID to `-1` instead of `0`

j9liu:
## Description

Alternative to #1690. Previously, `0` was used a default value for feature IDs that were encoded as UV coordinates. However, `0` can be a valid feature ID value, so this is misleading. This PR changes how null feature IDs are treated by `UCesiumFeatureMetadataComponent`.

- The `If` node for checking null feature IDs is *always* included now. This means that `bHasNullFeatureId` is no longer needed.
- The default null feature ID is `-1`, and if a feature ID set has an explicit `nullFeatureId` value, it will override this default.

Although branching in shaders is not typically ideal, I hope a single `If` node is reasonable enough. If a feature ID set always supplies non-null feature IDs, then I think it would stick to the same branch anyways? 🤔 

## Author checklist

- ~[] I have updated the documentation as necessary.~

## Testing plan

1. Make sure `06_CesiumMetadata` works as expected. **San Francisco Ferry Building** specifies an explicit null feature ID, so that should continue to work.
2. Test the `UCesiumFeaturesMetadataComponent` workflow with **New York City 3D Buildings**. To my knowledge, this dataset does not contain any null feature IDs, and we can confirm that by attempting to render a very salient color.

My material graph, if it is helpful:

image

image


kring:
Thanks @j9liu!

PR #1704, opened 2025-07-03
Basic support for glTF LINES primitive mode

GhisBntly:
This is the result of hacking quick support for these lines in order to test a model I really wanted to see but that only had lines in it... ,^^
But I'm opening a draft PR in order to get feedback already, because I found it almost suspicious that lines weren't supported if it were that easy! Anything I'm missing?
I find this feature useful even without attenuation support, what do you think? I won't have much more time to work on this as it was never really part of my assignments...

TODO:
  - [ ] Only LINES for the moment, _ie_ lists of segments: wouldn't be hard to support glTF line loops and strips I guess, but I'd need a dataset to test that.
  - [ ] "Basic support" = no support for rendering with "attenuation" like point clouds: haven't looked at the custom vertex factory and shader yet to see how hard it would be to support lines?
  - [ ] Probably need to change the name of CesiumGltfPointsComponent and proxy, unless it makes more sense to duplicate them.
  - [ ] Update `CHANGES.md`

Sample result: no screen capture because this is user data, sorry! Will try to find lines in one of our own models.

### DIFF
--- a/Source/CesiumRuntime/Private/CesiumGltfComponent.cpp
+++ b/Source/CesiumRuntime/Private/CesiumGltfComponent.cpp
@@ -1328,6 +1328,7 @@ getIndices(const TIndexAccessor& indicesView, int32 primitiveMode) {
     break;
   case CesiumGltf::MeshPrimitive::Mode::TRIANGLES:
   case CesiumGltf::MeshPrimitive::Mode::POINTS:
+  case CesiumGltf::MeshPrimitive::Mode::LINES:
   default:
     indices.SetNum(static_cast<TArray<uint32>::SizeType>(indicesView.size()));
     for (int32 i = 0; i < indicesView.size(); ++i) {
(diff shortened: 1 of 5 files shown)


j9liu (changes requested):
Thanks @GhisBntly for the PR! I don't think it would be much of a lift to add support for `LINES`, so feel free to keep working on this!

I would first suggest creating a separate `CesiumGltfLinesComponent` to detangle this implementation from the point one. You might find this #1037 a useful reference; it was the first PR where we added point support without worrying about attenuation.

The `SceneProxy` stuff is all extra to support point attenuation. Basically we turn the points into billboarded quads to render them at a larger size, giving the point cloud a more opaque appearance. I'm not sure what line attenuation would look like, but it might help if you have any examples ^^

GhisBntly:
Thanks for the prompt feedback, @j9liu.
I was hoping a similar "line attenuation" would be able to render lines with a width greater than 1 pixel, but clearly billboards won't do ;^^ So I'll just do what you say and it'll be good enough for now 👍

j9liu:
I'm about to open a PR for this, so I'm going to close this one. But thank you for opening this in the first place! 😄

j9liu:
Replaced by #1736

PR #1718, opened 2025-08-04
Add missing CesiumRuntime.h include for v2.18.0

byjtew:
## Description

Missing declaration of `LogCesium` in _CesiumPropertyAttribute.cpp_

## Issue number or link

Tell me if you want an issue linked, I'll open one.

## Author checklist

## Remaining Tasks

Setup a compilation CI in the sample project for example would avoid that kind of oversights.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,7 +2,9 @@
 
 ### ???
 
-- Add missing CesiumRuntime.h include in CesiumPropertyAttribute.cpp. Broke compilation of previous v2.18.0 release on Windows.
+##### Fixes :wrench:
+
+- Added a missing `CesiumRuntime.h` include in `CesiumPropertyAttribute.cpp` that broke compilation in v2.18.0 on Windows.
 
 ### v2.18.0 - 2025-08-01
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,9 @@
 # Change Log {#changes}
 
+### ???
+
+- Add missing CesiumRuntime.h include in CesiumPropertyAttribute.cpp. Broke compilation of previous v2.18.0 release on Windows.
+
 ### v2.18.0 - 2025-08-01
 
 ##### Breaking Changes :mega:
--- a/Source/CesiumRuntime/Private/CesiumPropertyAttribute.cpp
+++ b/Source/CesiumRuntime/Private/CesiumPropertyAttribute.cpp
@@ -1,8 +1,8 @@
 // Copyright 2020-2024 CesiumGS, Inc. and Contributors
 
 #include "CesiumPropertyAttribute.h"
-#include <CesiumGltf/PropertyAttributeView.h>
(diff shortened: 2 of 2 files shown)


Source/CesiumRuntime/Private/CesiumPropertyAttribute.cpp
```diff
@@ -2,6 +2,7 @@
 
 #include "CesiumPropertyAttribute.h"
 #include <CesiumGltf/PropertyAttributeView.h>
+#include "CesiumRuntime.h"
```
j9liu: Swap order so the third-party include is at the bottom:

```suggestion
#include "CesiumRuntime.h"
#include <CesiumGltf/PropertyAttributeView.h>
```

CHANGES.md
```diff
@@ -1,5 +1,9 @@
 # Change Log {#changes}
 
+### ???
+
+- Add missing CesiumRuntime.h include in CesiumPropertyAttribute.cpp. Broke compilation of previous v2.18.0 release on Windows.
```
```suggestion

##### Fixes :wrench:

- Added a missing `CesiumRuntime.h` include in `CesiumPropertyAttribute.cpp` that broke compilation in v2.18.0 on Windows.
```

j9liu (commented):
Sorry for breaking things again @byjtew, and thanks for the fix! I just have two small suggestions; once those are addressed this will be good to merge  ^^

byjtew:
Done @j9liu

j9liu:
Thanks @byjtew !

kring:
I did test packaging on Windows for v2.18.0, and this didn't show up for me. Probably because Unreal's unity build isn't entirely deterministic with which source files it groups together, and certain combinations can mask a problem like this. So I'm not entirely sure how we can avoid problems like this in the future.

In any case, thanks for the PR @byjtew!

byjtew:
> I did test packaging on Windows for v2.18.0, and this didn't show up for me. Probably because Unreal's unity build isn't entirely deterministic with which source files it groups together, and certain combinations can mask a problem like this. So I'm not entirely sure how we can avoid problems like this in the future.

I know that's an actual pain to enforce. I am happy to test your future releases though, idk if you have some group of testers.

kring:
Thanks for the offer @byjtew! We don't have a formal group of testers, and in fact the window between a release candidate build and the actual release is usually just a few hours.

One thing you could do, though, if you're up for it, is install and test a build from the `main` branch in the day or two before each release. Releases always happen on the first business day of the month, so a couple of days before that would be ideal. That won't catch every problem, because sometimes merges into `main` happen at the last minute, but it should definitely help!

Instructions for installing a build from a branch can be found here:
https://cesium.com/learn/cesium-unreal/ref-doc/using-prerelease.html#download-a-pre-release-version-from-a-branch-or-commit

byjtew:
I'll do that 👌

PR #1720, opened 2025-08-08
Use `const-get-root-tile` branch of cesium-native

kring:
This PR updates cesium-unreal to work with the breaking changes to cesium-native in this PR:
https://github.com/CesiumGS/cesium-native/pull/1221


j9liu (approved):
Looks good @kring ! I brought in the latest changes on cesium-native main, so I'll wait for CI to pass before merge.

PR #1726, opened 2025-08-20
Bump Windows SDK Version on UE 5.4 CI

j9liu:
## Description

Windows SDK version `10.0.18362` is no longer supported as of version `20250811.1.0` on the Github Windows runner, which is the version that the Unreal 5.4 builds use. See https://github.com/actions/runner-images/issues/12776.

Although that version is the one officially used by Epic's build farm, the [release notes](https://dev.epicgames.com/documentation/en-us/unreal-engine/unreal-engine-5.4-release-notes?application_version=5.4) suggest it can still build with newer versions. So this bumps the version to the minimum version supported by the new runner image. The Windows UE 5.4 build passes and produces an artifact, and I haven't seen any unexpected behavior while going through the Cesium for Unreal Samples. So... maybe it's fine? 😁

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -49,9 +49,9 @@ jobs:
       # and using them ensures we're compiling our plugin in the exact same way that Unreal Engine itself is compiled.
       cmake-generator: "Visual Studio 17 2022"
       cmake-toolchain: "version=14.34"
-      cmake-platform: "x64,version=10.0.18362.0"
+      cmake-platform: "x64,version=10.0.19041.0"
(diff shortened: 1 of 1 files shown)


kring:
There's also this line in the UE 5.4 release notes:

> Bump windows version from 18362 to 19041 to meet the minimum spec for GameInput API

So maybe Epic has already moved to 19041 for the build farm, and the release notes are wrong?

In any case, I think this should be fine. There's a chance it will mean the plugin no longer works on older versions of Windows 10 or something, but that's probably ok. Thanks for implementing it @j9liu!

PR #1732, opened 2025-08-27
Fix crash when loading GeoJSON from Cesium ion with null server

j9liu:
## Description

This PR prevents a crash in `UCesiumLoadGeoJsonDocumentFromUrlAsyncAction` from dereferencing a `nullptr` Cesium ion server. It falls back on the default Cesium ion server (just like `Cesium3DTileset`) if one is not explicitly specified. I argue that it does not make sense for GeoJSON loading to deviate from existing behavior and expectations.

Additionally this PR:

- Reorders the function parameters so that the server is last (and thus considered more advanced, less necessary). Although this is a minor break of the C++ API, it does not break parsing of Blueprint graphs.
- Removes redundant cohens from error messages.

## Author checklist

## Testing plan

In any level, open the **Level Blueprint** and run the **Load GeoJSON Document From Cesium Ion** node in **BeginPlay**.

image

Without this branch it crashes, but with this branch it proceeds with the loading process.

### DIFF
--- a/Source/CesiumRuntime/Public/CesiumGeoJsonDocument.h
+++ b/Source/CesiumRuntime/Public/CesiumGeoJsonDocument.h
@@ -62,7 +62,7 @@ class UCesiumGeoJsonDocumentBlueprintLibrary
 
 public:
   /**
-   * @brief Attempts to load a `FCesiumGeoJsonDocument` from a string containing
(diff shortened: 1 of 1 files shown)


j9liu:
Also noticing some stray `@brief`s that are bleeding into the Blueprint popups. I'm removing those as well.

image

azrogers:
Thanks @j9liu!

PR #1736, opened 2025-09-05
Add support for glTF line primitives

j9liu:
## Description

This PR adds support for the `LINES`, `LINE_LOOP`, and `LINE_STRIP` primitive modes in glTF.

Apparently the only line mode supported in Unreal is `PT_LineList`, so the other two have to be converted into line lists with duplicate indices.

image

## Author checklist

## Remaining Tasks

The normals are off somehow for this model and I haven't figured out why? I added a clause to disable flat normal generation for non-triangle models (since that doesn't make much sense), but there's still some lighting variation among the lines themselves. You can see it when the sun angle is really extreme. 🤔 

image

## Testing plan

Load this tileset under a georeference with **True Origin**: [MeshPrimitiveModes.zip](https://github.com/user-attachments/files/20353260/MeshPrimitiveModes.zip)

Reference screenshot for this sample model here: https://github.com/KhronosGroup/glTF-Sample-Assets/tree/main/Models/MeshPrimitiveModes

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,11 @@
 # Change Log {#changes}
 
+### ? - ?
+
+##### Additions :tada:
+
+- Added support for rendering glTFs with line primitives.
+
 ### v2.19.1 - 2025-09-02
 
 ##### Fixes :wrench:
--- a/Source/CesiumRuntime/Private/CesiumGltfComponent.cpp
+++ b/Source/CesiumRuntime/Private/CesiumGltfComponent.cpp
@@ -683,14 +683,21 @@ static void setUnlitNormals(
         glm::dvec4(
             VecMath::createVector3D(FVector(positionBuffer.VertexPosition(i))),
             1.0));
-    glm::dvec3 normal = ellipsoid.geodeticSurfaceNormal(positionFixed);
+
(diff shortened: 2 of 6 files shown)


j9liu:
After a closer look, I believe the weird shading results from the ellipsoid-based computation of unlit normals.

The dataset is small and at the center of the Earth, so the "up" direction for those points are at very far-apart angles even within the same primitive. Thus the shading varies greatly even between points on the same line. Similar normals can be seen on the point cloud primitive as well, which relies on the same unlit normal generation. The reason that the triangles don't suffer from this is because we can actually compute flat normals for those meshes.

I did find that some of the points resulted in `NaN`s, so I tried to add a clause to prevent that. But it doesn't affect the mesh appearance anyways in this case. Then I tried to add some georeferencing to the sample tileset to prove it would look better, but for some reason I couldn't get it to show up after fiddling with transforms. I'd have to hash out more precise values :/

So instead of fighting it, I tried visualizing a 3D Tileset generated from [NYC Open Data](https://opendata.cityofnewyork.us/) instead:

image 

It even works with metadata styling! Though, this particular dataset wasn't generated with real metadata, so I've arbitrarily colored the roads based on feature ID.

image

There's some z-fighting in the distance, admittedly, but I'm not sure how much that can be helped. At any rate, this is ready for review.

kring:
Thanks @j9liu!

PR #1740, opened 2025-09-29
Update cesium-native and doc

kring:
* Update cesium-native to head of main. The riskiest change here is CesiumGS/cesium-native#1251.
* ~Changed a use of `fmt::format` to `std::format` because the former is causing a compile warning-as-error in Unreal after upgrading.~ (`std::format` won't work on macOS, for annoying reasons)
* Added `anl` to the link line for CesiumEditor on Linux, because the updated version of cpp-httplib uses `getaddrinfo_a` on that platform, and that function is found in `anl`.
* Updates `Documentation/developer-setup-windows-gotchas.md` to mention the new `VCPKG_PLATFORM_TOOLSET_VERSION` environment variable, and to recommend setting it in the VS Code cmake kit file.

### DIFF
--- a/Documentation/developer-setup-windows-gotchas.md
+++ b/Documentation/developer-setup-windows-gotchas.md
@@ -48,7 +48,7 @@ If you've followed the instructions above, and Unreal Build Tool is now building
 
 vcpkg has hard-coded logic to choose the very latest version of the compiler that you have installed. It completely ignores all the usual ways that different compiler versions can be selected, such as setting the defaults file (`"C:\Program Files\Microsoft Visual Studio\2022\Professional\VC\Auxiliary\Build\Microsoft.VCToolsVersion.default.props"`), setting environment variables or running the `vcvarsall` script.
 
-The easiest way to select a compiler for vcpkg to use is to set the `VCPKG_PLATFORM_TOOLSET_VERSION` environment variable. Despite the name, this is not built-in vcpkg functionality, but is instead something that our `x64-windows-unreal.cmake` triplet file looks for explicitly. To set it, in a PowerShell window, run:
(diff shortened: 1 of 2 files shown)


Documentation/developer-setup-windows-gotchas.md
```diff
@@ -46,13 +46,30 @@ With that, all builds invoked by Unreal Build Tool should use the chosen compile
...
+vcpkg has hard-coded logic to choose the very latest version of the compiler that you have installed. It completely ignores all the usual ways that different compiler versions can be selected, such as setting the defaults file (`"C:\Program Files\Microsoft Visual Studio\2022\Professional\VC\Auxiliary\Build\Microsoft.VCToolsVersion.default.props"`), setting environment variables or running the `vcvarsall` script.
+
+The easiest way to select a compiler for vcpkg to use is to set the `VCPKG_PLATFORM_TOOLSET_VERSION` environment variable. Despite the name, this is not built-in vcpkg functionality, but is instead something that our `x64-windows-unreal.cmake` triplet file looks for explicitly. To set it, in a PowerShell window, run:
```
j9liu: **NITPICK**

```suggestion
The easiest way to select a compiler for vcpkg to use is to set the `VCPKG_PLATFORM_TOOLSET_VERSION` environment variable. Despite the name, this is not built-in vcpkg functionality, but is instead something that our `x64-windows-unreal.cmake` triplet file looks for explicitly. To set it, run the following in a PowerShell window:
```

Source/CesiumRuntime/CesiumRuntime.Build.cs
```diff
@@ -77,6 +77,11 @@ public CesiumRuntime(ReadOnlyTargetRules Target) : base(Target)
 
         PublicAdditionalLibraries.AddRange(allLibs);
 
+        if (Target.Platform == UnrealTargetPlatform.Linux)
```
j9liu: It's worth copying the comment from `CesiumEditor.Build.cs`:

```suggestion
        // On Linux, cpp-httplib uses getaddrinfo_a, which is in the anl library.
        if (Target.Platform == UnrealTargetPlatform.Linux)
```

j9liu (approved):
Thanks @kring ! Just two small comments that I can merge myself.

PR #1741, opened 2025-09-29
Add CesiumGoogleMapTilesRasterOverlay

kring:
This is the Unreal side of CesiumGS/cesium-native#1254.

In addition to a new raster overlay type, this also extends `CesiumIonRasterOverlay` to allow a free-form JSON string to be passed as an `options` query parameter.

### DIFF
--- a/Source/CesiumRuntime/Private/CesiumGoogleMapTilesRasterOverlay.cpp
+++ b/Source/CesiumRuntime/Private/CesiumGoogleMapTilesRasterOverlay.cpp
@@ -108,6 +108,11 @@ JsonValue::Array getStyles(const TArray<FString>& styles) {
 std::unique_ptr<CesiumRasterOverlays::RasterOverlay>
 UCesiumGoogleMapTilesRasterOverlay::CreateOverlay(
     const CesiumRasterOverlays::RasterOverlayOptions& options) {
+  if (this->Key.IsEmpty()) {
+    // We must have a key to create this overlay.
+    return nullptr;
+  }
+
   return std::make_unique<CesiumRasterOverlays::GoogleMapTilesRasterOverlay>(
       TCHAR_TO_UTF8(*this->MaterialLayerKey),
       CesiumRasterOverlays::GoogleMapTilesNewSessionParameters{
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,8 @@
 ##### Additions :tada:
 
 - Added support for rendering glTFs with line primitives.
+- Added `UCesiumGoogleMapTilesRasterOverlay`.
+- Added `AssetOptions` property to `UCesiumIonRasterOverlay`.
 
 ### v2.19.1 - 2025-09-02
 


Source/CesiumRuntime/Public/CesiumIonRasterOverlay.h
```diff
@@ -54,6 +54,17 @@ class CESIUMRUNTIME_API UCesiumIonRasterOverlay : public UCesiumRasterOverlay {
       AdvancedDisplay)
   UCesiumIonServer* CesiumIonServer;
 
+  /**
+   * Extra options to pass to the Cesium ion when accessing the asset. This
```
j9liu:
```suggestion
   * Extra options to pass to Cesium ion when accessing the asset. This
```

Source/CesiumRuntime/Private/CesiumGoogleMapTilesRasterOverlay.cpp
```diff
@@ -0,0 +1,124 @@
...
+
+} // namespace
+
+std::unique_ptr<CesiumRasterOverlays::RasterOverlay>
+UCesiumGoogleMapTilesRasterOverlay::CreateOverlay(
+    const CesiumRasterOverlays::RasterOverlayOptions& options) {
+  return std::make_unique<CesiumRasterOverlays::GoogleMapTilesRasterOverlay>(
+      TCHAR_TO_UTF8(*this->MaterialLayerKey),
+      CesiumRasterOverlays::GoogleMapTilesNewSessionParameters{
+          .key = TCHAR_TO_UTF8(*this->Key),
```
j9liu, line 119: Can we avoid creating the raster overlay if the key is empty? Similar to how an ion raster overlay is not created if the asset ID is 0.

Source/CesiumRuntime/Public/CesiumGoogleMapTilesRasterOverlay.h
```diff
@@ -0,0 +1,171 @@
...
+  Satellite,
+
+  /**
+   * The standard Google Maps painted map tiles.
+   */
+  Roadmap,
+
+  /**
+   * Terrain imagery. When selecting terrain as the map type, you must
+   * also set the `LayerTypes` property to "Roadmap".
```
j9liu:
```suggestion
   * Terrain imagery. When selecting terrain as the map type, you must
   * also add "Roadmap" to the `LayerTypes` property.
```

Source/CesiumRuntime/Private/CesiumGoogleMapTilesRasterOverlay.cpp
```diff
@@ -0,0 +1,124 @@
...
+      }
+    }
+  }
+
+  if (mapType == EGoogleMapTilesMapType::Terrain && !hasRoadmap) {
+    UE_LOG(
+        LogCesium,
+        Warning,
+        TEXT(
+            "When the MapType is set to Terrain, the LayerType must contain Roadmap."));
```
j9liu:
```suggestion
            "When the MapType is set to Terrain, LayerTypes must contain Roadmap."));
```

j9liu (commented):
Thanks @kring ! Can you update `CHANGES.md` for this when you get the chance?

I also found two things while testing:

- Sometimes `Error response 500` is printed in the console for the `Roadmap` map type. But the data seems to load fine otherwise. Is this expected? I see it happen on these parameters: `&zoom=23&west=-180.000000&south=-90.000000&east=180.000000&north=90.000000` 
- There appears to be an extra dot in the editor credits, but it goes away during runtime. Not the end of the world though.
image

kring:
> Sometimes Error response 500 is printed in the console for the Roadmap map type.

Yep, Google consistently returns 500 errors for level 23 of the viewport service for the Roadmap map type. Levels 22 and 24 are fine. AFAICT it's a Google problem.

kring:
> There appears to be an extra dot in the editor credits, but it goes away during runtime. Not the end of the world though.

Hmm I haven't seen that. I did see two Cesium ion logos, though. I think because I was pulling assets from both production and staging, and each provided a credit. Given my poor ability to debug it at the moment, I suggest we write an issue for it and don't let it hold up this PR or release.

kring:
Thanks for the review @j9liu, this should be ready!

kring:
Merging this, too.

PR #1743, opened 2025-10-01
adding small function to control collision response by channel on Cesium3DTileset

V2iDanAbend:
## Description

Adds a very small function that allow the Cesium3DTileset to have the Collision response change per collision channel which is present in standard Unreal StaticMeshActors. 

This is extremely useful for changing the collision response for just pawns but still retain hit responses for other channels such as vehicles. An example use case is to set the Pawn collision response to ignore if the player pawn is within a small CartographicPolygon that would otherwise not have the partially clipped tile collisions removed. All other actors in the world will still have normal collision response to the tileset for game play purposes. 

image

## Issue number or link

Related to my question here about finding interior of CartographicPolygon: https://community.cesium.com/t/best-way-to-check-if-object-is-inside-or-outside-of-a-cartographicpolygon/42926?u=v2i

## Author checklist

- [ x ] I have submitted a [Contributor License Agreement](https://github.com/CesiumGS/community/tree/main/CLAs) (only needed once).
- [ x ] I have done a full self-review of my code.
- [ x ] I have updated `CHANGES.md` with a short summary of my change (for user-facing changes).
- [ x ] I have added or updated unit tests to ensure consistent code coverage as necessary.
- [ x ] I have updated the documentation as necessary.

## Remaining Tasks

No remaining tasks that I know of.

## Testing plan

Can be called in C++ or Blueprint. Accepts array or single object reference. Tested with player Pawn and Vehicle channel dynamic response changes.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,7 @@
 # Change Log {#changes}
 
+- adding a function to change the collision response of a Cesium3DTileset by collision channel.
+
 ### v2.20.0 - 2025-10-01
 
 ##### Additions :tada:
--- a/Source/CesiumRuntime/Private/Cesium3DTileset.cpp
+++ b/Source/CesiumRuntime/Private/Cesium3DTileset.cpp
@@ -146,10 +146,11 @@ void ACesium3DTileset::SampleHeightMostDetailed(
   positions.reserve(LongitudeLatitudeHeightArray.Num());
 
   for (const FVector& position : LongitudeLatitudeHeightArray) {
-    positions.emplace_back(CesiumGeospatial::Cartographic::fromDegrees(
-        position.X,
-        position.Y,
-        position.Z));
+    positions.emplace_back(
+        CesiumGeospatial::Cartographic::fromDegrees(
+            position.X,
+            position.Y,
+            position.Z));
   }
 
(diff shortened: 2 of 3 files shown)


kring:
Thanks for the PR @V2iDanAbend! This looks great, just one problem that I can see:

When a collision response is changed, there's some code like this meant to apply the change:

```
          Prim->SetCollisionEnabled(ECollisionEnabled::NoCollision);
          Prim->SetCollisionEnabled(ECollisionEnabled::QueryAndPhysics);
```

The problem with this is that not all components should have `QueryAndPhysics` enabled. The total set of GltfComponents includes both currently-visible tiles and also tiles that are loaded but not currently visible. For example, a low-detail tile may exist (but be hidden and have physics turned off) while high-detail tiles are shown. But turning collision off and then on, you'll end up enabling it for tiles for which it was originally off. This will likely lead to phantom collisions with invisible tiles.

The solution, I think, is to only toggle `QueryAndPhysics` off and back on if it was on to begin with.

kring:
Hmm also CI is reporting that `UpdateCollisionProfile` isn't a valid function...?

j9liu:
Hi @V2iDanAbend, are you able to address Kevin's comments above? We'd love to get this PR into our next release if possible!

PR #1746, opened 2025-10-13
Don't remove postgresql-common-dev because it isn't installed anymore

kring:
This package appears to no longer be installed in the GitHub image, so trying to remove it causes the build to fail. As a result, all of our Linux builds are currently failing in main.


kring:
I'm going to merge this in order to fix CI in main.

PR #1754, opened 2025-11-04
SceneCaptures: take the show-only / hidden list of actors into consideration

Julot94:
A scene capture component can be configured to render only a given list of actors (or to ignore a specific list of actors). 
Take those options into consideration in order to avoid requesting a huge amount of data for a 3D tileset if it is not even seen by the capture component.

Following a crash report: in the user's scene, a scene capture component configured to render only a specific actor was causing the Cesium tileset to request a _huge_ amount of data (wheras the tileset was not seen by the capture component)

## Author checklist

### DIFF
--- a/Source/CesiumRuntime/Private/Cesium3DTileset.cpp
+++ b/Source/CesiumRuntime/Private/Cesium3DTileset.cpp
@@ -1503,6 +1503,22 @@ std::vector<FCesiumCamera> ACesium3DTileset::GetSceneCaptures() const {
       continue;
     }
 
+    // Take the show-only / hidden list of actors into consideration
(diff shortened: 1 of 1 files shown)


j9liu:
Thanks for the PR @Julot94, and sorry for the delayed response!

It's looking like our camera system needs deeper reworking (#1768), so we're going to see if we can develop a more generalized solution that also addresses this. But we'll keep this PR open in case we come around to it.

j9liu:
Closing this now that #1781 has been merged. There is now an explicit `SceneCaptures` array to dynamically add/remove captures for use. Let us know if that works for your use case!

PR #1755, opened 2025-11-04
Fixes for Unreal 5.7

azrogers:
The Unreal 5.7 preview is here, so I went through and updated everything to work with the new version. Most of the changes involve texture loading, as some of the deprecated paths we were using are now fully removed. 

Still need to get 5.7 working on CI. Will probably wait for full release for that one.

### DIFF
--- a/extern/unreal-linux-toolchain.cmake
+++ b/extern/unreal-linux-toolchain.cmake
@@ -9,16 +9,35 @@ SET(CMAKE_AR $ENV{UNREAL_ENGINE_COMPILER_DIR}/bin/llvm-ar)
 SET(CMAKE_BUILD_WITH_INSTALL_RPATH on)
 SET(CMAKE_POSITION_INDEPENDENT_CODE on)
 
-# Before Unreal Engine 5.6, libcxx was found in a weird place. In 5.6, it's found in that same place,
-# plus also under the regular sysroot. In 5.7, it's only available under the regular sysroot.
-# Rather than try to figure out what version of UE we're targeting, just always add the legacy location
-# resolution after the modern one.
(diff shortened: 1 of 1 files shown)


Source/CesiumRuntime/Public/CesiumCommon.h
```diff
@@ -5,4 +5,10 @@
 #include "Runtime/Launch/Resources/Version.h"
 
 #define ENGINE_VERSION_5_5_OR_HIGHER                                           \
-  (ENGINE_MAJOR_VERSION > 5 || ENGINE_MINOR_VERSION >= 5)
+  (ENGINE_MAJOR_VERSION > 5 && ENGINE_MINOR_VERSION >= 5)
```
kring: `||` was correct, `&&` is wrong. Think about how it will evaluate for 5.4, 5.5, and 6.1.

azrogers: I guess this should be `>= 5` or `> 4`, but wouldn't `ENGINE_MAJOR_VERSION > 5 || ENGINE_MINOR_VERSION >= 5` also trigger for, say, Unreal 4.27?

azrogers: Ah, I see. It should be `(ENGINE_MAJOR_VERSION == 5 && ENGINE_MINOR_VERSION >= 5) || (ENGINE_MAJOR_VERSION > 5)`

kring: Yeah, the original formulation of this won't work at all in 4.27. But neither will anything else in this code base. 😁

kring:
Thanks @azrogers!

PR #1762, opened 2025-11-14
Replace `AutoFill` with better property-picking UI

j9liu:
## Description

Depends on #1757, #1765, so merge those first.

In `UCesiumFeaturesMetadataComponent`, the "AutoFill" function is a quick way to grab all of the metadata and feature ID sets on visible tiles, but it does not provide an easy editing experience. The nested storage of properties makes navigating the component's UI very cumbersome, and it gets worse for datasets with huge amounts of varying properties (think Cesium OSM Buildings).

This PR replaces "Auto Fill" with a new window that allows for incremental addition of feature ID sets / properties to the tileset.

image

There is room for UI improvements that probably won't be in scope for this PR:

- Search bar
- Sorting
- Multi-selection

## Issue number or link

No direct issue, but this is to support some of the UI concerns in #1685.

## Author checklist

## Testing plan

TODO

- [x] San Francisco Ferry Building
- [x] Cesium OSM Buildings
- [x] [Muscatatuk (property textures)](https://sandcastle.cesium.com/?id=custom-shaders-property-textures)
- [ ] ~Feature ID texture with KHR_texture_transform~
- [ ] ~Property texture property with KHR_texture_transform~
- [ ] glTF with offset / scale / noData / default properties
- [ ] Instanced features / metadata


j9liu:
Still have to test a few cases, but I'll mark this ready for review anyways just to get the ball rolling.

kring:
This is a huge improvement!

Using metadata is still hard, though. I'm going to dump what I see as some of the usability problems. It's definitely not necessarily to fix all (or any) of these as part of this PR:

1. I can never remember what to name the Material Layer, and have to go to the tutorial every time. Our users probably don't even realize it matters. Can we help them somehow? Perhaps generate the Material Instance in addition to the Material Layer?
2. Metadata converted from B3DM JSON often uses different types to represent the same property, because it chooses the smallest type that losslessly represents the range of values within that particular tile. Here an especially bad example from Cesium OSM Buildings:
image
But as far as I can tell, it doesn't matter much which of these we choose. After adding one, we can even change it to type "String" and values make it through to the requested texture type as expected. So... why do we force this distinction on the users? There must be some scenario where it matters, I guess.

3. It would be really helpful if the Add Property panel allowed us to see a dump of the values associated with a given property. Just a big list by feature ID would be great.
4. This might just be a data problem, but I noticed with the Ferry building tileset that `_FEATURE_ID_TEXTURE_0` is associated with `labels1` in some tiles and `labels2` in others. 
image
That seems like it will be very hard for users to work with. If this is a tileset problem, maybe we can fix the tileset? Or, if this is something that is expected to come up fairly often, then perhaps we need to do some work in Unreal. For example, rather than making the user choose `_FEATURE_ID_TEXTURE_0` they can instead choose `labels1` and it we internally choose the appropriate feature ID texture for that table within each tile?

Here's what the Ferry building looks like if I switch it to use `_FEATURE_ID_TEXTURE_0` and `labels2` instead of `labels1`. Looks like the clock and some other bits end up unshaded? I guess I'd really need to dig into the data to understand what's happening here.
image

Source/CesiumEditor/Private/CesiumFeaturesMetadataViewer.cpp
```diff
@@ -0,0 +1,1185 @@
...
+      .AutoHeight()
+      .Padding(0.0f, 10.0f)
+      .VAlign(VAlign_Bottom)
+      .HAlign(HAlign_Center)
+          [SNew(SButton)
+               .ButtonStyle(FCesiumEditorModule::GetStyle(), "CesiumButton")
+               .TextStyle(FCesiumEditorModule::GetStyle(), "CesiumButtonText")
+               .ContentPadding(FMargin(1.0, 1.0))
+               .HAlign(EHorizontalAlignment::HAlign_Center)
+               .Text(FText::FromString(TEXT("Sync with View")))
```
kring: I think the word "Sync" doesn't convey what's happening here very well, mostly because it doesn't give a hint about the direction of the sync. How about "Refresh from Current View"?

Source/CesiumRuntime/Public/CesiumFeaturesMetadataComponent.h
```diff
@@ -27,15 +27,11 @@ class CESIUMRUNTIME_API UCesiumFeaturesMetadataComponent
...
-   * existing metadata.
-   *
-   * Warning: Using Auto Fill may populate the description with a large amount
-   * of metadata. Make sure to delete the properties that aren't relevant.
+   * Opens a window to add feature ID sets and metadata properties from the
+   * current view of the tileset.
    */
   UFUNCTION(CallInEditor, Category = "Cesium")
-  void AutoFill();
+  void AddProperties();
```
kring, line 37: Suggest decorating this with:
```
UFUNCTION(CallInEditor, Category = "Cesium", Meta = (DisplayName = "Add Properties"))
```

So that the Editor displays it as "Add Properties" instead of "AddProperties". It would be nice if it did this automatically like it does for regular properties, but alas.

Same for `GenerateMaterial` below.

j9liu:
Thanks @kring for all your feedback!

> I can never remember what to name the Material Layer, and have to go to the tutorial every time. Our users probably don't even realize it matters. Can we help them somehow? Perhaps generate the Material Instance in addition to the Material Layer?

Agree, and admittedly I have been peeved by the small (but annoying) burden of cloning a material instance + setting this new material layer. I'll look into what's possible here; hopefully it's not a big lift to implement!

> Metadata converted from B3DM JSON often uses different types to represent the same property, because it chooses the smallest type that losslessly represents the range of values within that particular tile. [...] But as far as I can tell, it doesn't matter much which of these we choose.

Yeah, I could see how having multiple "Scalar" property instances is confusing. Component type is not a big issue at the end of the day. The most important thing here is distinguishing the base type of the property: for example, "Scalar" vs. "String", or even "Scalar Array" vs. "Scalar". But I see your point that, if everything is going to attempt to be coerced for a certain type, then that information might not be too useful to the user.

I acknowledge that the whole "multiple property" business is a little silly. It's technically necessary, because as of today there isn't mandatory cohesion between glTF metadata in a tileset (I've ranted a fair share about it in https://github.com/CesiumGS/3d-tiles/issues/828). But it may not be practically necessary... I'll play around with the UI and see if I can simplify things further.

> It would be really helpful if the Add Property panel allowed us to see a dump of the values associated with a given property. Just a big list by feature ID would be great.

That's a cool idea! Though I have to think about how to reasonably implement that, since there can be numerous instances of a property (with the same type) across a view of a tileset. Maybe the first 3 discovered instances or something?

> This might just be a data problem, but I noticed with the Ferry building tileset that _FEATURE_ID_TEXTURE_0 is associated with labels1 in some tiles and labels2 in others.

I can take a deeper look at that to confirm. This is another gripe of mine -- `EXT_mesh_features` doesn't have any mandatory identifier for feature ID sets, and they just exist in an arbitrary order in an array, so using their index is an identifier is a flaky way of distinguishing them from each other. But if the feature ID set does not supply an optional `label`, then that's all we got. (See my other rant in https://github.com/CesiumGS/3d-tiles/issues/829.)

j9liu:
> Component type is not a big issue at the end of the day. The most important thing here is distinguishing the base type of the property: for example, "Scalar" vs. "String", or even "Scalar Array" vs. "Scalar".

So this isn't as clean as I thought, unfortunately. I remembered that component type is a very big deal for property textures, because it has to generate different code to assemble the values from the texture, depending on the component type:

```
  switch (PropertyDetails.ComponentType) {
  case ECesiumMetadataComponentType::Float32:
    code += OutputName + " = asfloat(" + PropertyResultName + ");\n";
    break;
  case ECesiumMetadataComponentType::Int8:
  case ECesiumMetadataComponentType::Int16:
  case ECesiumMetadataComponentType::Int32:
    code += OutputName + " = asint(" + PropertyResultName + ");\n";
  default:
    code += OutputName + " = " + PropertyResultName + ";\n";
    break;
  }
```

It's unideal that this mandates the user making some decision about the precise details of the property that probably don't make a practical difference. I wonder if it's worth adding `if` statements to the code that take in constants for the component type, which would eliminate the need to know the property's component type. I'm sure this sentiment applies to other aspects of `FCesiumMetadataPropertyDetails`, too.

This wouldn't be a hard change, I don't think, though I worry about bloating the size of this PR even further 😅

j9liu:
Added a remove button upon @kring's suggestion: 

image

j9liu:
Hm I can't seem to reproduce the Ferry Building bug. I've actually seen it before, but now I'm only seeing a one-to-one mapping between the feature ID sets and their property tables:

image

Source/CesiumEditor/Private/CesiumFeaturesMetadataViewer.h
```diff
@@ -0,0 +1,269 @@
...
+     * specified (e.g., by details queried from
+     * UCesiumFeatureMetadataComponent).
+     */
+    TSharedPtr<ECesiumEncodedMetadataComponentType>
+        pEncodedComponentTypeSelection;
+  };
+
+  /**
+   * An instance of a CesiumGltf::PropertyTableProperty for a particular glTF in
+   * the tileset.
```
j9liu:
```suggestion
   * An instance of a CesiumGltf::PropertyTableProperty or
   * CesiumGltf::PropertyTextureProperty for a particular glTF in the tileset.
```

j9liu:
This is ready for another look @kring ! Not sure why Mac CI failed to upload the plugin artifact, but the build itself seems to have succeeded.

kring:
> I've actually seen it before, but now I'm only seeing a one-to-one mapping between the feature ID sets and their property tables

I see it in the default view (press 1) in the 06_CesiumMetadata level:
image

It could be viewport size and DPI dependent, though. So if you don't see it, try zooming in or out for the next LOD and you probably will.

kring:
The UI gets a bit shifty when adding/removing properties:
shifty

kring:
Looks good to me other than that UI shift thing. If it's difficult to fix we can probably live with that, even.

j9liu:
So it turns out some of the feature ID sets in San Francisco ferry building have a `featureCount` of 0. The code was ignoring them since they're kind of useless, but since it's resulting in UI confusion, I'll let it count empty feature ID sets as valid.

kring:
Thanks @j9liu!

Issue #1764, opened 2025-11-18
Picking and styling broken for KHR_texture_transform samples


j9liu:
### What happened?

This was working in #1360 and #1367, but must have broken in sometime between then.

Feature ID styling seems completely broken with this KHR_texture_transform sample model: https://github.com/CesiumGS/cesium/tree/main/Specs/Data/Models/glTF-2.0/FeatureIdTextureWithTextureTransform/glTF

Additionally, attempting to pick the texture results in a crash. Looking at the debug stack, it's because the image data no longer exists and has `size` = 0. So I suspect that this image is getting moved 

### Environment

Cesium for Unreal version: v2.21.0
Unreal Engine Version: 5.4, 5.5
Operating System: Windows 11

### Reproduction steps

1. Create a new **CesiumGeoreference** actor, and set it to **True Origin**
2. Load the linked glTF through a simple tileset.json.
3. Add **CesiumFeaturesMetadataComponent** to it.
4. Autofill + generate the material. Recreate this shader:
<img width="1071" height="1247" alt="Image" src="https://github.com/user-attachments/assets/694c4a45-e802-4d3a-bb15-9ff0fb281072" />
5. Create a material instance that uses this material layer, then set it on the tileset. Observe that it's monochrome.
6. For picking, copy the linetrace code in `06_CesiumMetadata`'s Level Blueprint, then attach the following nodes:
<img width="1586" height="982" alt="Image" src="https://github.com/user-attachments/assets/6631c816-6b6e-4a67-90df-1be19ff42774" />
7. Hit Play and click on the texture. The Editor will crash.

### Supporting evidence

| Error | Expected | 
| ----- | --------- |
| <img width="1069" height="1140" alt="Image" src="https://github.com/user-attachments/assets/3eb83af1-d505-424e-ab20-dcee563d8c11" /> | <img width="898" height="953" alt="Image" src="https://github.com/user-attachments/assets/47e25ef5-69b2-450c-b895-3a79debc7f6e" /> |

PR #1765, opened 2025-11-19
Remove `KHR_texture_transform` distinction in `UCesiumFeaturesMetadataComponent`

### LINKED ISSUE #1760: Ignore `KHR_texture_transform` distinction for textures in `UCesiumFeaturesMetadataComponent`
Currently, `FCesiumFeatureIDTextureDescription` and `FCesiumPropertyTexturePropertyDescription` have a property called `bHasKhrTextureTransform`, which records whether the texture contains a `KHR_texture_transform` extension.

This was added so "Generate Material" could create additional nodes to parse the offset / scale from `KHR_texture_transform`. However, there are reasonable defaults for these properties (0 and 1), and having this distinction doesn't add much value to the user IMO. I argue we should do away with this property, and instead _always_ generate offset and scale nodes for the textures that fall back on default values.

j9liu:
## Description

Depends on #1757, so merge that first.

This change removes the non-meaningful `bHasKhrTextureTransform` properties from `FCesiumFeatureIdSetDescription` and `FCesiumPropertyTexturePropertyDescription`. Now, it will always generate nodes to intercept any scale / offset / rotation parsed from `KHR_texture_transform`.

This simplifies some of the implementation in #1762, so IMO it is a worthwhile and easy change. It also won't break any existing materials because there are already correct default values when sampling feature ID textures anyway.

## Issue number or link

Closes #1760.

## Author checklist

## Testing plan

Although #1764 exists, it's still possible to test the actual _generation_ of nodes in the material. Also open `06_CesiumMetadata` to check that it still works, since it uses feature ID textures without the `KHR_texture_transform` extension.


kring:
@j9liu I tried this test:

1. Open the 06_CesiumMetadata level of the Cesium for Unreal Samples.
2. Click the San Francisco Ferry Building actor, then the CesiumFeaturesMetadata component attached to it.
3. Click the "GenerateMaterial" button.
4. Click Save in the material editor that opens up.

This makes the model turn grey:
image

I can't see any errors or warnings or anything relevant in the log.

That's unexpected, right?

kring:
I guess the problem is that everything has been disconnected:
image

I thought user connections were supposed to be preserved when regenerating the material, but maybe I'm misremembering.

j9liu:
> I thought user connections were supposed to be preserved when regenerating the material, but maybe I'm misremembering.

They're supposed to, but for some reason the `If` node does not reliably re-build those connections. I believe you can see this behavior on main, too (and my bad for not writing up an issue about it earlier!)

kring:
Looks good other than those `If` node connections being lost. I wrote a separate issue for that:
#1772 

Thanks @j9liu!

PR #1770, opened 2025-12-04
Fix UE 5.5+ GetInputsView deprecation warnings

jayyabsley:
## Description

Fixed deprecation warnings when compiling with Unreal Engine 5.5+ by replacing 
deprecated `GetInputsView()` calls with `FExpressionInputIterator` while maintaining 
backward compatibility with Unreal Engine 5.4.

### Changes Made:
- Added `#include "CesiumCommon.h"` for version detection macros
- Created version-conditional helper lambda `IterateInputs` in `ClearAutoGeneratedNodes()`
- Updated all 7 occurrences of `GetInputsView()` to use the new helper
- For UE 5.5+: Uses `FExpressionInputIterator`
- For UE 5.4: Continues using `GetInputsView()`

This eliminates 7 compiler warnings without changing functionality.

## Issue number or link

Fixes deprecation warnings reported in UE 5.5-5.7 compilation.

## Author checklist

- [x] I have submitted a [Contributor License Agreement](https://github.com/CesiumGS/community/tree/main/CLAs) (only needed once).
- [x] I have done a full self-review of my code.
- [x] I have updated `CHANGES.md` with a short summary of my change (for user-facing changes).
- [ ] I have added or updated unit tests to ensure consistent code coverage as necessary.
- [ ] I have updated the documentation as necessary.

## Testing plan

1. **UE 5.4.4**: Verified code compiles without errors using legacy `GetInputsView()` path
2. **UE 5.7.1**: Verified code compiles without deprecation warnings using new `FExpressionInputIterator` path

PR #1771, opened 2025-12-05
Remove code and CI for UE 5.4

j9liu:
## Description

Now that our new minimum version is UE 5.5, this removes all `#if` directives that preserved compatibility with UE 5.4.

Builds off of #1770.

## Author checklist

- [ ] ~I have updated `CHANGES.md` with a short summary of my change (for user-facing changes).~

## Testing plan

Although a holistic samples spot-check would be great, make sure that in particular:

- Point cloud shading still works
- Material generation maintains the same level of functionality

### DIFF
--- a/.github/workflows/buildAndroid.yml
+++ b/.github/workflows/buildAndroid.yml
@@ -77,7 +77,7 @@ jobs:
           rm -r -fo extern
       - name: Overwrite plugin engine version
         run: |
-          ((Get-Content -path CesiumForUnreal.uplugin -Raw) -replace '"EngineVersion": "5.3.0"','"EngineVersion": "${{ inputs.unreal-engine-version }}"') | Set-Content -Path CesiumForUnreal.uplugin
+          ((Get-Content -path CesiumForUnreal.uplugin -Raw) -replace '"EngineVersion": "5.5.0"','"EngineVersion": "${{ inputs.unreal-engine-version }}"') | Set-Content -Path CesiumForUnreal.uplugin
       - name: Customize BuildConfiguration.xml
         run: |
           mkdir -p "$env:USERPROFILE\AppData\Roaming\Unreal Engine\UnrealBuildTool"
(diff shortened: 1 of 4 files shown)


j9liu:
The following docs need non-trivial edits, so I haven't touched them yet:

- developer-setup-linux.md
- developer-setup-windows-gotchas.md (are these all applicable still with 5.5?)

developer-setup-osx.md could also use a look from a dedicated Mac user :')

j9liu:
- [x] @kring - from a glance, do you know if the Windows gotchas still apply to 5.5? (is it just a matter of changing version numbers, or something else?)
- [ ] @timoore - can you check out what needs to change for 5.5 in `developer-setup-linux.md`?
- [x] @david-lively - can you verify that my changes to `developer-setup-osx.md` are correct for Unreal 5.5?

kring:
@j9liu I updated the gotchas document. UE 5.5 (tested), 5.6 (not tested, but probably works), and 5.7 (14.44 is the documented build server version) all work just fine with 14.44, so in some ways this is simpler than it used to be. But the gotchas document still tries to explain how everything fits together.

I also updated the default path the the UE installation in CMakeLists.txt.

Documentation/developer-setup-linux.md
```diff
@@ -13,11 +13,11 @@ After compiling Unreal Engine, set the following environment variables in your `
 
 ```bash
 export UNREAL_ENGINE_DIR=<path_to_unreal_engine>
-export UNREAL_ENGINE_COMPILER_DIR=$UNREAL_ENGINE_DIR/Engine/Extras/ThirdPartyNotUE/SDKs/HostLinux/Linux_x64/v22_clang-16.0.6-centos7/x86_64-unknown-linux-gnu
+export UNREAL_ENGINE_COMPILER_DIR=$UNREAL_ENGINE_DIR/Engine/Extras/ThirdPartyNotUE/SDKs/HostLinux/Linux_x64/v25_clang-18.1.0-rockylinux8/x86_64-unknown-linux-gnu
 export UNREAL_ENGINE_LIBCXX_DIR=$UNREAL_ENGINE_DIR/Engine/Source/ThirdParty/Unix/LibCxx
 ```
 > [!note]
-> `v22_clang-16.0.6-centos7` is correct for Unreal Engine v5.4. It may be different for other versions of Unreal Engine. See the [Unreal documentation](https://dev.epicgames.com/documentation/en-us/unreal-engine/linux-development-requirements-for-unreal-engine?application_version=5.4) for the equivalent for your version of Unreal Engine.
+> `v25_clang-18.1.0-rockylinux8` is correct for Unreal Engine v5.6. It may be different for other versions of Unreal Engine. See the [Unreal documentation](https://dev.epicgames.com/documentation/en-us/unreal-engine/linux-development-requirements-for-unreal-engine?application_version=5.6) for the equivalent for your version of Unreal Engine.
```
j9liu: @timoore We're still on Unreal 5.5 as our base version

timoore: Does that matter here? The comment is mostly for illustration. I have 5.6 installed locally, but I can go get 5.5 and check if you think we shouldn't refer to anything later than 5.5.

j9liu: Historically we've kept the base version consistent throughout our different platform documentation, and I think we should maintain that pattern to avoid possible confusion.

j9liu:
Not sure why CI is failing for Windows / Android... seems like while building cesium-native ezvcpkg is failing at the configure step, saying that it has 503s for certain packages' links?

kring:
I made some last changes:

1. Modified the CI scripts to correctly update the version in the uplugin prior to building. It seems we forgot to do this for the 5.3 deprecation, and it didn't really matter because UE does it itself. But still a good improvement.
2. Removed the attempted log printing on failure. It's not needed and wasn't working in most cases anyway, because it was looking for a log with 5.3 in the filename.

I'm going to merge this!

Issue #1700, opened 2025-06-30
`GetPrimitiveFeatures` doesn't work for instanced tilesets


j9liu:
### What happened?

`GetPrimitiveFeatures` returns an empty primitive features for instanced tilesets. This is because of these lines of code:

https://github.com/CesiumGS/cesium-unreal/blob/1e9f442cf34762fdae47753cced15ec2791c5665/Source/CesiumRuntime/Private/CesiumPrimitiveFeatures.cpp#L45-L51

It should also be checking for an `UCesiumGltfInstancedComponent`, and returning the attached features if found.

### Environment

Cesium for Unreal version: v2.16.1
Unreal Engine Version: 5.4
Operating System: Windows

### Reproduction steps

1. Open the sample level `06_CesiumMetadata`.
2. Load the [Instanced Orientation](https://github.com/CesiumGS/cesium/tree/main/Apps/SampleData/Cesium3DTiles/Instanced/InstancedOrientation) tileset from URL.
3. In the Level Blueprint, recreate this Blueprint:![Image](https://github.com/user-attachments/assets/f06de157-a292-43fd-8b6b-67074794b6ab)
4. Run the level. Click on any of the cubes. You should see a -1 printed no matter what cube you press. However, the metadata table will be populated with the Height property.

![Image](https://github.com/user-attachments/assets/5deb5047-d9fa-4887-9433-993cd9676850)

PR #1776, opened 2025-12-12
Fix various features/metadata getters for instanced meshes

j9liu:
## Description

After #1762 was merged, I tried using the new UI on an i3dm tileset, only to find that none of it showed up. It turned out that most of the functions being used did not correctly handle `UCesiumGltfInstancedComponent`. I had _thought_ I solved #1700, but it turns out I was referencing the wrong primitive features object and that slipped through review 🙃 

I really wish we could clean this up so you don't have to do two cast checks -- or at least, make some changes to standardize features/metadata storage -- but I suppose that's out of scope.

## Author checklist

- [X] I have added or updated unit tests to ensure consistent code coverage as necessary.

## Testing plan

Try one of the [i3dm sample models](https://github.com/CesiumGS/cesium/tree/main/Apps/SampleData/Cesium3DTiles/Instanced) in CesiumJS. The Add Properties UI should correctly report having a feature ID set and a metadata property:

Screenshot 2025-12-12 113317


j9liu:
I'm also pushing an unrelated UI fix for when properties have special qualifiers (default value, etc.) because I was using the wrong units:

image

kring:
Thanks @j9liu!

PR #1778, opened 2025-12-22
Support styling with tileset statistics

### LINKED ISSUE #1751: Add interface for tileset statistics
### Feature

`TilesetMetadata` in cesium-native parses the [statistics](https://github.com/CesiumGS/cesium-native/blob/6da0610f3c83c9f543e8384faf7b7d9025c68081/Cesium3DTilesSelection/include/Cesium3DTilesSelection/TilesetMetadata.h#L59) property if it is present in a tileset. However, we don't expose this to the user at our runtimes or provide any way to act on this at runtime. 

It would be nice to create ways to act on them, and even use them while styling with `CesiumFeaturesMetadataComponent` (like what CesiumJS does). For example, you could style property values in a gradient between provided min -> max values.

My one concern is user interface, and of course the way that **"Auto Fill"** works. Currently it just parses the glTF extensions in the visible tiles, which are not guaranteed to have the same properties across multiple tiles. Each property can have its own min/max range as well in [`EXT_structural_metadata`](https://github.com/CesiumGS/glTF/blob/3d-tiles-next/extensions/2.0/Vendor/EXT_structural_metadata/schema/class.property.schema.json) -- and then _on top of that_, each INSTANCE of a property can have its own min/max override. So.... this becomes messy very quickly. There are probably things we can do to manage this (e.g. checkbox that says to override the tileset min/max with the glTF min/max), but it's still pretty nebulous.

It might also be worth revisiting a UI / UX idea we had in #1190 to revamp the **Auto Fill** function. Instead of autofilling the properties, it could open a separate pop up with a list of the properties in scene. Users can scroll through all instances, search for the ones they're looking for, then add them manually to the component, instead of adding all 200+ properties from OSM buildings for example. The tileset metadata could be a separate view that you can then choose whether or not to incorporate into your material in some way. Just some thoughts 🤷

j9liu:
## Description

Depends on #1782, so merge that first.

This PR exposes tileset-level metadata statistics for styling with `UCesiumFeaturesMetadataComponent`. This will be incorporated into #1685 for improved voxel metadata UI.

image

## Issue number or link

Fixes #1751.

## Author checklist

## Remaining Tasks

- [x] Implement + test non-min/max statistics.
- [x] Do some additional property type testing.
- [x] Troubleshoot some Editor state management with the values on `UCesiumFeaturesMetadataComponent`

## Testing plan

1. Load [TilesetWithStatistics.zip](https://github.com/user-attachments/files/24301283/TilesetWithStatistics.zip) in an Unreal level.
2. Add `UCesiumFeaturesMetadataComponent` and open the **Add Properties** panel.
3. Add the feature ID attribute at the top of the panel. Then, add any of the properties from the property table. Also add the corresponding min/max statistics from the tileset. For the example below I used the **Longitude** and **Height** properties.
4. Hit **Generate Material** and observe how nodes are generated for the statistics in the material.
5. The quickest way to visualize the min/max range is to use the `RemapValuesNormalized` node. Here's the graph I used to achieve the below visual output:<br>image<br>
image
6. Check that hitting **Generate Material** again preserves the graph structure.

### DIFF
--- a/Source/CesiumRuntime/Private/CesiumFeaturesMetadataComponent.cpp
+++ b/Source/CesiumRuntime/Private/CesiumFeaturesMetadataComponent.cpp
@@ -2248,6 +2248,9 @@ void GenerateMaterialNodes(
   }
 
   NodeX += BeginSectionX + MaximumSectionX;
+  // If no nodes were generated, this prevents the next nodes from spawning on
+  // top of each other.
+  NodeX = FMath::Max(NodeX, 1.5 * Incr);
 
   UMaterialExpressionSetMaterialAttributes* SetMaterialAttributes = nullptr;
   for (const TObjectPtr<UMaterialExpression>& ExistingNode :


Source/CesiumRuntime/Private/EncodedFeaturesMetadata.h
```diff
@@ -96,6 +91,21 @@ static const FString MaterialPropertyNoDataSuffix = "_NO_DATA";
 static const FString MaterialPropertyDefaultValueSuffix = "_DEFAULT";
 static const FString MaterialPropertyHasValueSuffix = "_HAS_VALUE";
 
+/**
+ * Property Statistic Semantic:
+ * Prefix + PropertyEntityName + PropertyName + "_SEMANTIC"
```
j9liu: This should actually be:
```suggestion
 * ClassId + PropertyId + "_SEMANTIC"
```

Source/CesiumRuntime/Private/EncodedFeaturesMetadata.h
```diff
@@ -147,16 +157,6 @@ static const FString MaterialTextureRotationSuffix = "_TX_ROTATION";
...
-/**
- * @brief Generates a HLSL-safe name for a feature ID set. This is used by the
- * material for parameter nodes.
- *
- * @param FeatureIDSet The feature ID set
- * @param FeatureIDTextureCounter The counter representing how many feature ID
- * textures have been seen in the primitive thus far. Will be incremented by
- * this function if the given feature ID set is a texture.
- */
-FString getMaterialNameForFeatureIDSet();
```
j9liu, line 159: This function was never used, and also never had a definition! Not sure what happened, but removing it didn't introduce any problems.

j9liu:
I'm struggling to convert a glb -> gltf on my machine so I can add more buffers (having trouble with various CLI tools), so I don't have a test tileset with vector metadata. But this is what the view is for vector / scalar array statistics: 
image

kring:
Each time I hit "Refresh", another set of statistics appears. And the panel isn't scrollable:

refresh-statistics

I also noticed another problem that is probably related. If I click "Add Properties" on a tileset with statistics, and then _without closing the panel_ click "Add Properties" on a different tileset that does not have statistics, then the "glTF Features" and "glTF Metadata" sections are updated correctly for the new tileset, but the "Tileset Statistics" section still shows the old tileset's info.

Source/CesiumRuntime/Public/Cesium3DTileset.h
```diff
@@ -1210,11 +1210,6 @@ class CESIUMRUNTIME_API ACesium3DTileset : public AActor {
     return this->_pTileset.Get();
   }
```
kring, line 1212: Technically a breaking change from C++, so we should mention it in the changelog.

kring:
The latest Apple55 and Apple56 builds in this branch took a lot longer (166 min and 145 min, respectively) compared to the latest builds in main (102 min and 71 min respectively).

I'm running them again and hoping it was just a fluke, but Clang on these low-memory Apple systems doesn't handle big templates well, so I'm worried that's the cause.

kring:
Maybe not an important use-case at all, but if I add a couple of statistics without adding any feature IDs or properties, and hit Generate Material, it generates a couple of nodes right on top of each other:
image

Source/CesiumRuntime/Public/Cesium3DTileset.h
```diff
@@ -1210,11 +1210,6 @@ class CESIUMRUNTIME_API ACesium3DTileset : public AActor {
...
 
-  const std::optional<FCesiumFeaturesMetadataDescription>&
-  getFeaturesMetadataDescription() const {
-    return this->_featuresMetadataDescription;
-  }
-
```
kring, line 1217: The comment above about a breaking change should go here. I'm not sure what I did. 😆

CHANGES.md
```diff
@@ -14,6 +23,7 @@
 - Added `ACesiumCartographicPolygon::SetPolygonPoints` to shape a polygon spline from a list of points in cartographic or Earth-Centered, Earth-Fixed space.
 - Re-enabled `ACesium3DTileset::EnableWaterMask` for macOS devices.
 
+
```
```suggestion

```

j9liu:
> The latest Apple55 and Apple56 builds in this branch took a lot longer (166 min and 145 min, respectively) compared to the latest builds in main (102 min and 71 min respectively).
>
> I'm running them again and hoping it was just a fluke, but Clang on these low-memory Apple systems doesn't handle big templates well, so I'm worried that's the cause.

The templating was definitely a factor. I pushed a commit that completely removed it from the `JsonValue` constructor in `CesiumMetadataValue`, and the builds have gone down to <= 2hrs :')

Admittedly, I hacked this so that the underlying data is either `double`, `int64_t`, or `uint64_t`. The `FCesiumMetadataValueType` may report something smaller than what is stored. But ultimately, the underlying C++ type is only important to the conversion code, and for scalar conversions this shouldn't have an impact.

j9liu:
I believe I addressed your comments @kring! Not sure what your bandwidth looks like, but do let me know if I should tag someone else to review.

kring:
Looks good, thanks @j9liu!

PR #1779, opened 2025-12-23
Add support for Cesium ion login token refresh

kring:
clang-tidy makes this diff look bigger than necessary. The changes are fairly straightforward:

- Update cesium-native, to get CesiumGS/cesium-native#1292 (merge that first).
- Handle the breaking changes in the above where `CesiumIonClient::Connection::authorize` returns errors via `Result` rather than by rejecting the promise.
- Persist the refresh token provided by Cesium ion, in addition to the login token.
- If a request fails due to an invalid login token, flip back to the "disconnected" state so that the user is asked to log in again. This will happen when the login token and refresh token are both expired (or are somehow invalid).

The user isn't informed when a login token refresh happens. But if it fails (because the refresh token also expired, for example), they'll see this in the log:

```
LogCesiumEditor: Error: Failed to get valid access token:
- [Error] Token refresh failed. Please sign into Cesium ion again.
- [Error] The server returned an error code: 403 (Code NoValidToken)
```

And the Cesium panel UI will flip back to the disconnected state. I thought a popup would just be annoying.

### DIFF
--- a/Source/CesiumEditor/Private/CesiumIonSession.cpp
+++ b/Source/CesiumEditor/Private/CesiumIonSession.cpp
@@ -280,7 +280,7 @@ void CesiumIonSession::resume() {
 
         CesiumUtility::Result<LoginToken> tokenResult =
             userAccessToken.IsEmpty()
-                ? LoginToken("", -1)
+                ? LoginToken("", std::nullopt)
                 : LoginToken::parse(TCHAR_TO_UTF8(*userAccessToken));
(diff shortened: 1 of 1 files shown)


j9liu (approved):
Thanks @kring! I should have set up a dev connection since this isn't out in production, but I'm not sure I'll have time in my remaining break, and I trust that you and @azrogers covered enough ground 👍

Issue #1768, opened 2025-12-02
Extend CesiumCameraManager to control which Unreal cameras are used by tilesets


kring:
With the help of a member of the community, we added a `CesiumCameraManager` to Cesium for Unity awhile back. It gives uses control over which cameras are auto-discovered, and also allows them to manually tell the system to use particular cameras that it wouldn't otherwise discover itself.

This comment describes how it works:
https://github.com/CesiumGS/cesium-unity/pull/421#issuecomment-2337986794

We should add something similar to Cesium for Unreal. We already have a CesiumCameraManager in Unreal, but a) it doesn't allow any of the cameras that are auto-discovered to be excluded, and b) it only allows additional cameras to be very manually specified.

Related to #1754 and #1744.

j9liu:
Closed by #1781

PR #1781, opened 2026-01-05
Additional cameras

timoore:
Add an option to the camera manager to use (or not) the player's camera for tile loading, and also expose the table of additional cameras.

See #1768

### DIFF
--- a/Source/CesiumRuntime/Public/CesiumCamera.h
+++ b/Source/CesiumRuntime/Public/CesiumCamera.h
@@ -35,13 +35,21 @@ struct CESIUMRUNTIME_API FCesiumCamera {
 
 public:
   /**
-   * @brief Source of camera parameters
+   * @brief The desired source of camera parameters.
    */
   UPROPERTY(EditAnywhere, BlueprintReadWrite, Category = "Cesium")
   ECameraParameterSource ParameterSource = ECameraParameterSource::Manual;
 
   /**
-   * @brief Source camera component, if any.
+   * @brief The pixel dimensions of the viewport.
+   */
+  UPROPERTY(EditAnywhere, BlueprintReadWrite, Category = "Cesium")
+  FVector2D ViewportSize;
+
+  /**
+   * @brief The camera component from which to retrieve parameters. Due to
+   * UI limitations, this property must be set through Blueprints or C++. Used
+   * when ParameterSource is set to Camera Component.
    */
   UPROPERTY(
       EditAnywhere,
@@ -53,13 +61,8 @@ public:
   TSoftObjectPtr<UCameraComponent> CameraComponent;
 
   /**
-   * @brief The pixel dimensions of the viewport.
-   */
-  UPROPERTY(EditAnywhere, BlueprintReadWrite, Category = "Cesium")
-  FVector2D ViewportSize;
-
-  /**
-   * @brief The Unreal location of the camera.
+   * @brief The Unreal location of the camera. Used when Parameter Source is set
+   * to Manual.
    */
   UPROPERTY(
       EditAnywhere,
@@ -70,7 +73,8 @@ public:
   FVector Location;
 
   /**
-   * @brief The Unreal rotation of the camera.
+   * @brief The Unreal rotation of the camera. Used when Parameter Source is set
+   * to Manual.
    */
   UPROPERTY(
       EditAnywhere,
@@ -81,7 +85,8 @@ public:
   FRotator Rotation;
 
   /**
-   * @brief The horizontal field of view of the camera in degrees.
+   * @brief The horizontal field of view of the camera in degrees. Used when
+   * Parameter Source is set to Manual.
    */
   UPROPERTY(
       EditAnywhere,
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -15,6 +15,7 @@
 - Added `DetectTransformChanges` to `UCesiumGlobeAnchorComponent` to toggle whether the globe anchor should update itself after changes in its Actor's transform.
 - Added `UsePlayerCameras`, `UseEditorCameras`, and `UseSceneCapturesInLevel` to `ACesiumCameraManager` to toggle the automatic detection of cameras in the level for tile selection.
 - Added `ACesiumCameraManager::SceneCaptures` for more explicit control over which `ASceneCapture2D`s are used for tile selection.
+- Added the option for `FCesiumCamera` to use parameters from a given `UCameraComponent` assigned through C++ or Blueprints.
 
 ##### Fixes :wrench:
 


j9liu:
Thanks @timoore for the PR! This is a good opportunity to discuss the interface design, and others can chime in as well with their own input.

As written in #1768:

> We already have a CesiumCameraManager in Unreal, but a) it doesn't allow any of the cameras that are auto-discovered to be excluded, and b) it only allows additional cameras to be very manually specified.

I'm seeing that most of the changes in this PR intend to address (a), but I'm not sure if they're the right call. The new `useMainCamera` setting simply toggles whether `GetPlayerCameras` is queried at all, but all the returned cameras will still be added. There needs to be more granular control than this. (I also see that `useSceneViewCameraInEditor` was added but not actually used in the case of `GetEditorCameras`, but we would run into similar concerns here too.)

Second, we might want to revisit if the current setup is good to begin with to address (b), before committing to exposing the `cameras` map (which maps arbitrary camera IDs to `FCesiumCamera`s).

Deprecating `FCesiumCamera` would be nice, but it seems to exist out of necessity. Unfortunately we don't seem to have a simple `Camera` class like Unity, and we have to gather separate views from `ASceneCapture2D`, `APlayerCameraManager`, and `FEditorViewportClient`. There's also the possibility of custom camera classes in plugins that we don't want to introduce as dependencies, but that people otherwise want to hook up to the camera manager.

Still, is it worth creating explicit lists for `ASceneCapture2D`s, `APlayerCameraManager`, etc. on `CesiumCameraManager` so they can be easily added in the Editor? IMO, I think it is worth it to at least save people the hassle of having to do this via Blueprints script. And because there are existing Blueprints to add / remove elements from a list, we wouldn't have to maintain our own Blueprint functions for that.

We might ultimately still keep our current Blueprints + `FCesiumCamera` map, but we need to consider design options would improve user experience.

timoore:
> Thanks @timoore for the PR! This is a good opportunity to discuss the interface design, and others can chime in as well with their own input.
> 
> As written in #1768:
> 
> > We already have a CesiumCameraManager in Unreal, but a) it doesn't allow any of the cameras that are auto-discovered to be excluded, and b) it only allows additional cameras to be very manually specified.
> 
> I'm seeing that most of the changes in this PR intend to address (a), but I'm not sure if they're the right call. The new `useMainCamera` setting simply toggles whether `GetPlayerCameras` is queried at all, but all the returned cameras will still be added. There needs to be more granular control than this. (I also see that `useSceneViewCameraInEditor` was added but not actually used in the case of `GetEditorCameras`, but we would run into similar concerns here too.)

I'm not actually sure what is meant by the "main camera" in UE, so kind of seized on the player cameras as being equivalent.

> 
> Second, we might want to revisit if the current setup is good to begin with to address (b), before committing to exposing the `cameras` map (which maps arbitrary camera IDs to `FCesiumCamera`s).

Right; this seemed to be the heart of #1768, as blueprint access to the camera list is already kind of supported.
> 
> Deprecating `FCesiumCamera` would be nice, but it seems to exist out of necessity. Unfortunately we don't seem to have a simple `Camera` class like Unity, and we have to gather separate views from `ASceneCapture2D`, `APlayerCameraManager`, and `FEditorViewportClient`. There's also the possibility of custom camera classes in plugins that we don't want to introduce as dependencies, but that people otherwise want to hook up to the camera manager.
> 
> Still, is it worth creating explicit lists for `ASceneCapture2D`s, `APlayerCameraManager`, etc. on `CesiumCameraManager` so they can be easily added in the Editor? IMO, I think it is worth it to at least save people the hassle of having to do this via Blueprints script. And because there are existing Blueprints to add / remove elements from a list, we wouldn't have to maintain our own Blueprint functions for that.
> 
> We might ultimately still keep our current Blueprints + `FCesiumCamera` map, but we need to consider design options would improve user experience.

Source/CesiumRuntime/Public/CesiumCameraManager.h
```diff
@@ -17,6 +39,38 @@ class CESIUMRUNTIME_API ACesiumCameraManager : public AActor {
   GENERATED_BODY()
 
 public:
+  /**
+   * @brief Determines whether the cameras attached to PlayerControllers should
+   * be used for Cesium3DTileset culling and level-of-detail.
+   */
+  UPROPERTY(EditAnywhere, BlueprintReadWrite, Category = "Cesium")
+  bool usePlayerCameras = true;
```
j9liu: [Per our style guide](https://cesium.com/learn/cesium-unreal/ref-doc/style-guide.html#naming), public variables should use `PascalCase`.

```suggestion
  bool UsePlayerCameras = true;
```

Source/CesiumRuntime/Public/CesiumCamera.h
```diff
@@ -25,7 +25,6 @@ struct CESIUMRUNTIME_API FCesiumCamera {
    */
   UPROPERTY(BlueprintReadWrite, Category = "Cesium")
   FVector2D ViewportSize;
-
```
j9liu, line 28: Undo this?

Source/CesiumRuntime/Public/CesiumCameraManager.h
```diff
@@ -17,6 +39,38 @@ class CESIUMRUNTIME_API ACesiumCameraManager : public AActor {
...
+  /**
+   * @brief Array of auxilliary cameras.
+   */
+  UPROPERTY(EditAnywhere, BlueprintReadWrite, Category = "Cesium")
+  TArray<FAuxiliaryCamera> otherCameras;
+
+  /**
+   * @brief Array of explicit scene capture actors.
+   */
+  UPROPERTY(EditAnywhere, BlueprintReadWrite, Category = "Cesium")
```
j9liu, line 56: This should be editable only when `UseSceneCapturesInScene` is disabled, and vice versa. Look at the [`EditCondition`](https://unreal-garden.com/docs/uproperty/#editcondition) macro for how to do this.

Source/CesiumRuntime/Public/CesiumCameraManager.h
```diff
@@ -2,12 +2,34 @@
...
+ */
+USTRUCT(BlueprintType)
+struct FAuxiliaryCamera {
+  GENERATED_BODY()
+public:
+  /**
+   * @brief the viewport
+   */
+  UPROPERTY(EditAnywhere, BlueprintReadWrite, Category = "Cesium")
+  FVector2D ViewportSize;
```
j9liu: I understand why this was added, but does it make sense to introduce yet _another_ struct that people have to use when `FCesiumCamera` already exists?

This could be improved by removing this struct and just using an array of `UCameraComponent`s on `CesiumCameraManager`. You can use a reasonable default for the viewport size by doing `GEngine->GameViewport->Viewport->GetSizeXY()`. Here's a forum post from a user who did that: https://community.cesium.com/t/problems-with-tileset-occlusion/39652

Source/CesiumRuntime/Public/CesiumCameraManager.h
```diff
@@ -17,6 +39,38 @@ class CESIUMRUNTIME_API ACesiumCameraManager : public AActor {
...
+   * @brief Search for the scene capture actors.
+   */
+  UPROPERTY(EditAnywhere, BlueprintReadWrite, Category = "Cesium")
+  bool UseSceneCapturesInScene = true;
+
+  /**
+   * @brief Array of auxilliary cameras.
+   */
+  UPROPERTY(EditAnywhere, BlueprintReadWrite, Category = "Cesium")
+  TArray<FAuxiliaryCamera> otherCameras;
```
j9liu:
```suggestion
  TArray<TObjectPtr<UCameraComponent>> AdditionalCameras;
```

Source/CesiumRuntime/Public/CesiumCameraManager.h
```diff
@@ -17,6 +39,38 @@ class CESIUMRUNTIME_API ACesiumCameraManager : public AActor {
...
+
+  /**
+   * @brief Array of explicit scene capture actors.
+   */
+  UPROPERTY(EditAnywhere, BlueprintReadWrite, Category = "Cesium")
+  TArray<TObjectPtr<AActor>> SceneCaptureActors;
```
j9liu: Why not specify the scene capture class?

```suggestion
  TArray<TObjectPtr<ASceneCapture2D>> SceneCaptures;
```

Source/CesiumRuntime/Public/CesiumCameraManager.h
```diff
@@ -72,6 +126,11 @@ class CESIUMRUNTIME_API ACesiumCameraManager : public AActor {
 
   virtual void Tick(float DeltaTime) override;
 
+  /**
+   * @brief Return a list of all cameras handled by the manager.
+   */
+  std::vector<FCesiumCamera> GetAllCameras() const;
```
j9liu, line 130: This function is at odds with the existing "GetCameras" function on this class -- needs clearer naming

Source/CesiumRuntime/Public/CesiumCameraManager.h
```diff
@@ -17,6 +39,38 @@ class CESIUMRUNTIME_API ACesiumCameraManager : public AActor {
...
+   * In a game, this property has no effect.
+   */
+  UPROPERTY(EditAnywhere, BlueprintReadWrite, Category = "Cesium")
+  bool UseEditorCameras = true;
+
+  /**
+   * @brief Search for the scene capture actors.
+   */
+  UPROPERTY(EditAnywhere, BlueprintReadWrite, Category = "Cesium")
+  bool UseSceneCapturesInScene = true;
```
j9liu:
```suggestion
  /**
   @brief Whether to find and use all scene captures within the level for Cesium3DTileset culling and level-of-detail.
   */
  UPROPERTY(EditAnywhere, BlueprintReadWrite, Category = "Cesium")
  bool UseSceneCapturesInLevel = true;
```

Source/CesiumRuntime/Public/CesiumCameraManager.h
timoore: I don't know what a reasonable default for the viewport would be. One of the main drivers for having additional cameras is to influence tile selection and LOD, and IMHO the viewport dimension is a critical part of that.

Source/CesiumRuntime/Public/CesiumCameraManager.h
timoore: Good question! "A polymorphic  haze."

Source/CesiumRuntime/Public/CesiumCameraManager.h
timoore: Happy to hear suggestions. GetCameras() is a legacy blueprint function, GetAllCameras() seems to be descriptive.

Source/CesiumRuntime/Public/CesiumCameraManager.h
j9liu: Just to write down what I said on call, I don't see a point in adding if we're not formally deprecating `FCesiumCamera`. Otherwise, we're copying Unreal's problem of having duplicate camera definitions that aren't consumed in the same way.

Our options are:

1. Formally deprecate `FCesiumCamera`, including the Blueprint associated with it. Replace it with `FAuxiliaryCamera`.
2. Remove `FAuxiliaryCamera`, but make it easier for people to construct `FCesiumCamera`s by exposing them as an array in the Details panel.

I would like to hear more concrete iteration so I'm not just coming up with everything myself :) Perhaps you could post on the community forum and ask for feedback here. My one reservation is that there may be a use case that can't be covered if we go with Option 1, and we should avoid breaking users' applications if we can help it.

Source/CesiumRuntime/Public/CesiumCameraManager.h
j9liu: I don't understand how `GetCameras` is considered legacy here when the Blueprint isn't formally deprecated?

Source/CesiumRuntime/Private/Cesium3DTileset.cpp
```diff
@@ -1493,9 +1497,15 @@ std::vector<FCesiumCamera> ACesium3DTileset::GetSceneCaptures() const {
...
-  UGameplayStatics::GetAllActorsOfClass(this, SceneCapture2D, sceneCaptures);
+  ACesiumCameraManager* pCameraManager = this->ResolvedCameraManager;
+  if (pCameraManager && pCameraManager->UseSceneCapturesInLevel) {
+    static TSubclassOf<ASceneCapture2D> SceneCapture2D =
+        ASceneCapture2D::StaticClass();
+    UGameplayStatics::GetAllActorsOfClass(this, SceneCapture2D, sceneCaptures);
+  }
+  for (auto sceneCaptureActor : pCameraManager->SceneCaptures) {
+    sceneCaptures.Push(sceneCaptureActor.Get());
+  }
```
j9liu: This `for` loop should be an `else`, right? Because we only use `pCameraManager->SceneCaptures` when `UseSceneCapturesInLevel` is false.

Source/CesiumRuntime/Public/CesiumCameraManager.h
j9liu: Still need to resolve this comment. Having both the internal `_cameras` map and the `AdditionalCameras` array is duplicative.

But, the `AdditionalCameras` approach that you added seems more convenient and scaleable going forward. Let's talk about a deprecation plan for `GetCameras()`, `AddCamera`, `UpdateCamera`, and `RemoveCamera` sometime tomorrow. It'll just be a matter of adding some macros.

Source/CesiumRuntime/Public/CesiumCamera.h
```diff
@@ -11,6 +12,19 @@
 
 #include "CesiumCamera.generated.h"
 
+UENUM(BlueprintType)
+enum class ECameraParameterSource : uint8 {
+  /**
+   * Camera parameters are set directly
+   */
+  Local UMETA(DisplayName = "Local Pose"),
```
j9liu: I think "Manual" might be clearer?

```suggestion
  /**
   * Camera parameters are set explicitly by the user.
   */
  Manual UMETA(DisplayName = "Manual"),
```

Source/CesiumRuntime/Public/CesiumCamera.h
j9liu: All the manually-specified properties need an `EditAnywhere`. They are sadly missing in the Editor without it:

Image

The `EditCondition` macro should be used here too.

Source/CesiumRuntime/Public/CesiumCameraManager.h
```diff
@@ -17,6 +19,43 @@ class CESIUMRUNTIME_API ACesiumCameraManager : public AActor {
...
+  TArray<FCesiumCamera> AdditionalCameras;
+
+  /**
+   * @brief Array of explicit scene capture actors.
+   */
+  UPROPERTY(
+      EditAnywhere,
+      BlueprintReadWrite,
+      Category = "Cesium",
+      Meta = (EditCondition = "!UseSceneCapturesInLevel", EditConditionHides))
```
j9liu: Sorry I didn't catch this earlier -- we prefer to let variables get greyed out (uneditable) rather than hiding them completely. See ["Conditional Properties"](https://cesium.com/learn/cesium-unreal/ref-doc/api-design.html#conditional-properties) in the API design guide.

```suggestion
      Meta = (EditCondition = "!UseSceneCapturesInLevel"))
```

j9liu: Also, I would paste `SceneCaptures` _above_ `AdditionalCameras` so the dependency on `UseSceneCapturesInLevel` is more clear. See ["Organization"](https://cesium.com/learn/cesium-unreal/ref-doc/api-design.html#organization) in the API design guide.

j9liu (commented):
This is looking better @timoore! I left some comments.

With your changes, I think it's appropriate to do a deprecation plan. The ["Deprecation and Backwards Compatibility"](https://cesium.com/learn/cesium-unreal/ref-doc/api-design.html#deprecation-and-backwards-compatibility) section in the API design guide should tell you what to do (which won't be much). We don't need to do any `CoreRedirects` or custom versioning, just the "Deprecation Specifiers" that mark things as deprecated. People see the warning and change their implementation themselves.

Source/CesiumRuntime/Private/Cesium3DTileset.cpp
timoore: That wasn't clear to me, but does make sense. I will change it to follow those semantics.

Source/CesiumRuntime/Public/CesiumCamera.h
```diff
@@ -20,28 +34,55 @@ struct CESIUMRUNTIME_API FCesiumCamera {
...
+  /**
+   * @brief Source of camera parameters
+   */
+  UPROPERTY(EditAnywhere, Category = "Cesium")
+  ECameraParameterSource ParameterSource = ECameraParameterSource::Manual;
+
+  /**
+   * @brief Source camera component, if any.
+   */
+  UPROPERTY(EditAnywhere, Category = "Cesium")
```
j9liu: This needs an `EditCondition` so it only is active when `ParameterSource` is `CameraComponent`.

Source/CesiumRuntime/Public/CesiumCameraManager.h
```diff
@@ -30,48 +69,58 @@ class CESIUMRUNTIME_API ACesiumCameraManager : public AActor {
...
-   * @param Camera The current state for the new camera.
-   * @return The generated ID for this camera. Use this ID to refer to the
-   * camera in the future when calling UpdateCamera.
+   * @brief DEPRECATED
+   * @deprecated The AdditionalCameras array should be used for directly storing
+   * and accessing cameras.
    */
+  UE_DEPRECATED(
+      "Cesium For Unreal v2.0",
+      "Use the AdditionalCameras array instead.")
```
j9liu: This does not deprecate a `UFUNCTION`. If you use the CesiumCameraManager via Blueprints, you can still add these nodes to the graph:

Image

Please see [Deprecation Specifiers](https://cesium.com/learn/cesium-unreal/ref-doc/api-design.html#deprecation-specifiers) for the correct method.

Source/CesiumRuntime/Public/CesiumCameraManager.h
```diff
@@ -17,6 +19,43 @@ class CESIUMRUNTIME_API ACesiumCameraManager : public AActor {
...
+
+  /**
+   * @brief Array of explicit scene capture actors.
+   */
+  UPROPERTY(
+      EditAnywhere,
+      BlueprintReadWrite,
+      Category = "Cesium",
+      Meta = (EditCondition = "!UseSceneCapturesInLevel"))
+  TArray<TObjectPtr<ASceneCapture2D>> SceneCaptures;
```
j9liu, line 51: Can you cut + paste this so it is _above_ `AdditionalCameras`? This changes the order in which the variables appear in the Details panel, and makes it clearer that `UseSceneCapturesInLevel` is linked to `SceneCaptures`.

Source/CesiumRuntime/Public/CesiumCameraManager.h
```diff
@@ -30,48 +69,58 @@ class CESIUMRUNTIME_API ACesiumCameraManager : public AActor {
...
 
   virtual void Tick(float DeltaTime) override;
 
+  /**
+   * @brief Return a list of all cameras handled by the manager.
+   */
+  std::vector<FCesiumCamera> GetAllCameras() const;
+
 private:
   int32 _currentCameraId = 0;
```
j9liu, line 136: Might be worth leaving a comment here that these variables are for the legacy camera implementation.

Source/CesiumRuntime/Private/Cesium3DTileset.cpp
```diff
@@ -1491,10 +1495,16 @@ std::vector<FCesiumCamera> ACesium3DTileset::GetSceneCaptures() const {
   // where users can volunteer cameras to be used with the tile selection as
   // needed?
```
j9liu: We can delete this comment now that the `SceneCaptures` array exists.

Source/CesiumRuntime/Private/Cesium3DTileset.cpp
```diff
@@ -1491,10 +1495,16 @@ std::vector<FCesiumCamera> ACesium3DTileset::GetSceneCaptures() const {
...
-  static TSubclassOf<ASceneCapture2D> SceneCapture2D =
-      ASceneCapture2D::StaticClass();
-  UGameplayStatics::GetAllActorsOfClass(this, SceneCapture2D, sceneCaptures);
-
+  ACesiumCameraManager* pCameraManager = this->ResolvedCameraManager;
+  if (pCameraManager && pCameraManager->UseSceneCapturesInLevel) {
+    static TSubclassOf<ASceneCapture2D> SceneCapture2D =
+        ASceneCapture2D::StaticClass();
+    UGameplayStatics::GetAllActorsOfClass(this, SceneCapture2D, sceneCaptures);
+  } else {
```
j9liu: Guard against possible nullptr dereference:

```suggestion
  } else if (pCameraManager) {
```

Source/CesiumRuntime/Private/CesiumCameraManager.cpp
```diff
@@ -114,3 +114,17 @@ bool ACesiumCameraManager::UpdateCamera(
...
 }
+
+std::vector<FCesiumCamera> ACesiumCameraManager::GetAllCameras() const {
+  std::vector<FCesiumCamera> result;
+  for (auto camera : this->_cameras) {
+    result.push_back(camera.Value);
+  }
+  for (const auto& otherCamera : this->AdditionalCameras) {
+    if (!(otherCamera.FieldOfViewDegrees == 0.0 ||
+          !otherCamera.CameraComponent)) {
```
j9liu: I think inverting this logic makes it more readable?

```suggestion
    if (otherCamera.CameraComponent || otherCamera.FieldOfViewDegrees > 0.0) {
```

Source/CesiumRuntime/Private/Tests/CesiumCameraManager.spec.cpp
```diff
@@ -32,59 +32,87 @@ void FCesiumCameraManagerSpec::Define() {
        });
   });
 
-  Describe("AddCamera", [this]() {
+  Describe("AdditionalCameras", [this]() {
     It("should add and remove a single camera", [this]() {
```
j9liu, line 36: This isn't testing much now that we're using a `TArray` instead of a loaded Blueprint interface. It would be better to test if each boolean setting works, and also if the `AdditionalCameras` array works when all other settings are disabled.

A small challenge is how to check that the viewport is empty. One option is to test whether any `CesiumGltfComponents` get created in the scene after one tick of the camera. `Cesium3DTileset.spec.cpp` has examples for how it set up such test cases.

j9liu (changes requested):
Thanks @timoore for the changes! There's still a few things that need to be addressed before merge. LMK if you have any questions.

Source/CesiumRuntime/Public/CesiumCameraManager.h
```diff
@@ -30,49 +69,68 @@ class CESIUMRUNTIME_API ACesiumCameraManager : public AActor {
...
-   * camera in the future when calling UpdateCamera.
+   * @brief DEPRECATED
+   * @deprecated The AdditionalCameras array should be used for directly storing
+   * and accessing cameras.
    */
+  UE_DEPRECATED(
+      "Cesium For Unreal v2.0",
+      "Use the AdditionalCameras array instead.")
   UFUNCTION(BlueprintCallable, Category = "Cesium")
   int32 AddCamera(UPARAM(ref) const FCesiumCamera& Camera);
```
j9liu: This still hasn't been deprecated properly.

j9liu:
While doing some final testing, I found that I couldn't actually assign a component to the **Camera Component** field in the Details panel. Despite having a Camera Actor in the scene, it didn't seem to be picking up the reference.

However, based on some activity online, this appears to be a known limitation in Unreal...

- https://www.potion8.com/post/issues-when-adding-actor-components-to-character-in-c
- https://forums.unrealengine.com/t/editable-sub-object-component-pointers/314612/13

The closest thing is[ `FComponentReference`](https://dev.epicgames.com/documentation/en-us/unreal-engine/API/Runtime/Engine/FComponentReference), but the API returns a simple `UActorComponent` pointer and doesn't seem to restrict the type.

I'm reaching out with questions about whether we're doing something wrong here, but if this is a known limitation, I think at this point we could just proceed with the merge and acknowledge the bug when it comes up.

j9liu:
So it seems that `FComponentReference` is the best way to search for camera components in the editor. It has UI built for it already as long as you use the right `Meta` specifiers.

image

However, it's not exactly the same as a `UCameraComponent*`, and using it in Blueprints/code is a little clunky. It doesn't actually point to the component itself, but rather stores the name of the component + the Actor it wants to find the component on. Then, the component is retrieved by name. This is reflected by the Blueprint:

image

The UI is an obvious improvement, but it's unfortunate that it downgrades the Blueprints/C++ experience at the same time. One possible way forward is to provide convenience functions for creating the `FCesiumCamera`, so that instead of using the built-in "Make" node, it can create one from a given `UCameraComponent`.

j9liu:
Merging now!

PR #1782, opened 2026-01-07
Simplify storage of array values in `FCesiumMetadataValue`

j9liu:
## Description

This PR is broken out of #1778 to improve the diff for review.

Previously, `FCesiumMetadataValue` defined a huge `swl::variant` that included array and non-array property types. This PR simplifies that, allowing a `FCesiumPropertyArray` to be stored directly by `FCesiumMetadataValue` to avoid duplication.

Since `FCesiumPropertyArray` isn't trivially copyable or destroyable (at least when building for Android), it's stored in an optional separate from that large variant definition. However, since the variant will be in `swl::monostate`, all the conversion functions should fall back to the default value as expected. The tests are all passing, so I don't seem to have introduced any unexpected behavior :D

## Author checklist

- ~[ ] I have updated `CHANGES.md` with a short summary of my change (for user-facing changes).~
- ~[] I have added or updated unit tests to ensure consistent code coverage as necessary.~

### DIFF
--- a/Source/CesiumRuntime/Private/CesiumMetadataValue.cpp
+++ b/Source/CesiumRuntime/Private/CesiumMetadataValue.cpp
@@ -19,12 +19,10 @@ FCesiumMetadataValue&
 FCesiumMetadataValue::operator=(FCesiumMetadataValue&& rhs) = default;
 
 FCesiumMetadataValue::FCesiumMetadataValue(const FCesiumMetadataValue& rhs)
-    : _value(),
+    : _value(rhs._value),
       _arrayValue(rhs._arrayValue),
       _valueType(rhs._valueType),
-      _pEnumDefinition(rhs._pEnumDefinition) {
-  swl::visit([this](const auto& value) { this->_value = value; }, rhs._value);
-}
+      _pEnumDefinition(rhs._pEnumDefinition) {}
 
 FCesiumMetadataValue&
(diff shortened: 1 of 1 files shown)


Source/CesiumRuntime/Private/CesiumMetadataValue.cpp
```diff
@@ -14,22 +20,10 @@ FCesiumMetadataValue::operator=(FCesiumMetadataValue&& rhs) = default;
...
-            this->_value = decltype(value)(this->_storage);
-          } else {
-            this->_value = value;
-          }
-        } else {
-          this->_value = value;
-        }
-      },
-      rhs._value);
+  swl::visit([this](const auto& value) { this->_value = value; }, rhs._value);
```
kring: I don't think this visit is needed anymore, because it only existed previously in order to make sure the array storage was copied. Just copy `_value` in the member initializer list like the other properties.

j9liu: Right! For some reason I thought it was needed because Intellisense gives this fun error:

image

But it seems to build anyways, so I guess that was a false alarm :shrug:

j9liu:
@kring This is ready for another look!

kring:
Thanks @j9liu!

PR #1785, opened 2026-01-09
Defer `DestroyTileset` to the next `Tick` when `RefreshTileset` is called

### LINKED ISSUE #1784: `Cesium3DTileset` can accidentally delete its `Tileset` mid-tick
### What happened?

@david-lively ran into a fun bug with this Blueprint:

Image

`ACesium3DTileset::DestroyTileset` immediately resets the pointer holding the underlying native `Tileset`. At the beginning of its `Tick` function, it checks whether this pointer is null and attempts to create the `Tileset` before proceeding, returning early if it can't.

However, it's possible to invoke `DestroyTileset` by calling `RefreshTileset` and thus accidentally null the pointer mid-`Tick`. How might this happen? Well, just chain a `RefreshTileset` through any asynchronous task that gets handled by `asyncSystem.dispatchMainThreadTasks()`, which itself is called during `Tileset::updateViewGroup()`. This happened after a "Load GeoJSON from Cesium ion" call, but it could likely happen after a "Sample Height Most Detailed" call as well.

~The solution may be to delay calling `Reset` on the pointer in `DestroyTileset`, and instead defer that to `IsReadyForFinishDestroy`, after all the async main thread tasks are complete.~

### Environment

Cesium for Unreal version: v2.22.1
Unreal Engine Version: 5.5
Operating System: Windows 11

### Reproduction steps

1. Start with `01_CesiumWorld` in the Cesium for Unreal Samples.
2. Recreate the Blueprint shown above.
3. Hit Play.

### Supporting evidence

_No response_

j9liu:
## Description

This implementation is inspired by what we do in Cesium for Unity. Instead of calling `DestroyTileset` immediately in `RefreshTileset`, this queues the function to happen on the next `Tick()`. This does not affect object destruction when the Actor itself is destroyed.

## Issue number or link

Fixes #1784.

## Author checklist

## Testing plan

Go through the **Reproduction Steps** in #1784 and confirm they don't result in a crash.

### DIFF
--- a/Source/CesiumRuntime/Private/Cesium3DTileset.cpp
+++ b/Source/CesiumRuntime/Private/Cesium3DTileset.cpp
@@ -306,7 +306,7 @@ ACesiumCreditSystem* ACesium3DTileset::ResolveCreditSystem() {
 
 void ACesium3DTileset::InvalidateResolvedCreditSystem() {
   this->ResolvedCreditSystem = nullptr;
-  this->RefreshTileset();
+  this->DestroyTileset();
 }
 
 TSoftObjectPtr<ACesiumCameraManager>
@@ -338,7 +338,7 @@ ACesiumCameraManager* ACesium3DTileset::ResolveCameraManager() {
 
 void ACesium3DTileset::InvalidateResolvedCameraManager() {
   this->ResolvedCameraManager = nullptr;
-  this->RefreshTileset();
+  this->DestroyTileset();
 }
 
 void ACesium3DTileset::RefreshTileset() { this->_destroyOnNextTick = true; }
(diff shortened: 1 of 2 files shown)


j9liu:
I saw that the height sampling tests were failing, and I don't know if I understand the logistics of how the `Tick` prevented the test from succeeding. But I saw that `RefreshTileset` was being invoked during `ResolveCreditSystem`, so I changed it to a direct `DestroyTileset` call that mimicked the old behavior.

Before that, I moved the `_destroyOnNextTick = false` line to the top of `DestroyTileset`, and that seemed to work too. But it seems better to unset that in the actual `Tick` function.

j9liu:
> I don't know if I understand the logistics of how the Tick prevented the test from succeeding.

Silly me... it's because `DestroyTileset` calls the `Tileset` destructor, which immediately fails all height requests! And although `ResolveCreditSystem` is called right away in `SampleHeightMostDetailed`, the `RefreshTileset` won't be processed until the next `Tick` as expected.

Issue #1351, opened 2024-02-03
Set initial points of CesiumCartographicPolygon based on world-coords of screen corners


corybarr:
When initially creating a `CesiumCartographicPolygon` in Unreal, I wanted to clip a building. The control corners and verts were outside of the camera bounds:

![image](https://github.com/CesiumGS/cesium-unreal/assets/1040194/e79d87c9-3f82-4a88-89d4-587c6e213182)

If screen coords with some padding (maybe (0, 0), (0.75, 0), (0.75, 0.75), (0, 0.75)) were projected to world coords, it would allow users of the feature to quickly set up the initial corners based on the camera composition.

david-lively:
Closed as part of https://github.com/CesiumGS/cesium-unreal/pull/1787

PR #1787, opened 2026-01-26
Create poly in viewport

david-lively:
## Description

This PR modifies the way that `ACesiumCartographicPolygon` actors are created from the Cesium editor panel. Previously, the actor was placed at the origin with a square side length of 20000 units. With this change, the actor is translated to the intersection of the camera view ray and the Z=0 plane, and the spline vertices are scaled to fit within the viewport. 

image

## Issue number or link

https://github.com/CesiumGS/cesium-unreal/issues/1351

## Author checklist

~~- [ ] I have submitted a [Contributor License Agreement](https://github.com/CesiumGS/community/tree/main/CLAs) (only needed once).~~

## Testing plan

Open the 01_CesiumWorld scene. 
Point the camera at the ground or anywhere below the horizon. 
Add a Cartographic Polygon via the Cesium panel. 
Observe that the new `Cesium Cartographic Polygon` is visible within the viewport. 
Point the camera above the horizon.
Add a Cartographic Polygon via the Cesium panel. 
Observe that the actor is created at the origin with default extents.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,7 @@
 
 - `ACesiumCartographicPolygon`s created via the Cesium editor panel are sized and translated to be visible in the active editor viewport.
 - Added `ACesiumCartographicPolygon::SetPolygonPoints` to shape a polygon spline from a list of points in cartographic or Earth-Centered, Earth-Fixed space.
+- Re-enabled `ACesium3DTileset::EnableWaterMask` for macOS devices.
 
 ##### Fixes :wrench:
 
--- a/Documentation/release-process.md
+++ b/Documentation/release-process.md
@@ -4,7 +4,7 @@ This is the process we follow when releasing a new version of Cesium for Unreal 
 
 ## Prepare Cesium Native for Release
 
-[cesium-native](https://github.com/CesiumGS/cesium-native) is used as a git submodule of cesium-unreal. So the first step in releasing Cesium for Unreal is to make sure that Cesium Native is ready for release. See the [Cesium Native release guide](#native-release-process) for instructions. You may want to wait to create the new Cesium Native tag until after you have tested the Cesium for Unreal release.
+[cesium-native](https://github.com/CesiumGS/cesium-native) is used as a git submodule of cesium-unreal. So the first step in releasing Cesium for Unreal is to make sure that Cesium Native is ready for release. See the [Cesium Native release guide](https://github.com/CesiumGS/cesium-native/blob/main/doc/topics/release-process.md) for instructions. You may want to wait to create the new Cesium Native tag until after you have tested the Cesium for Unreal release.
 
 ## Prepare Cesium for Unreal for Release
 
--- a/Source/CesiumRuntime/Private/Cesium3DTileset.cpp
+++ b/Source/CesiumRuntime/Private/Cesium3DTileset.cpp
@@ -1077,11 +1077,7 @@ void ACesium3DTileset::LoadTileset() {
   options.contentOptions.generateMissingNormalsSmooth =
       this->GenerateSmoothNormals;
 
-  // TODO: figure out why water material crashes mac
-#if PLATFORM_MAC
-#else
   options.contentOptions.enableWaterMask = this->EnableWaterMask;
-#endif
 
   CesiumGltf::SupportedGpuCompressedPixelFormats supportedFormats;
   supportedFormats.ETC1_RGB = GPixelFormats[EPixelFormat::PF_ETC1].Supported;
@@ -2298,14 +2294,6 @@ void ACesium3DTileset::PostEditImport() {
   this->DestroyTileset();
 }
 
-bool ACesium3DTileset::CanEditChange(const FProperty* InProperty) const {
-  if (InProperty->GetFName() ==
-      GET_MEMBER_NAME_CHECKED(ACesium3DTileset, EnableWaterMask)) {
-    // Disable this option on Mac
-    return PlatformName != TEXT("Mac");
-  }
-  return true;
-}
 #endif
 
 void ACesium3DTileset::BeginDestroy() {
--- a/Source/CesiumRuntime/Private/CesiumGltfComponent.cpp
+++ b/Source/CesiumRuntime/Private/CesiumGltfComponent.cpp
@@ -3152,12 +3152,6 @@ static void loadPrimitiveGameThreadPart(
                CesiumGltf::Material::AlphaMode::BLEND;
   };
 
-#if PLATFORM_MAC
-  // TODO: figure out why water material crashes mac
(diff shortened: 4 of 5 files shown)


Source/CesiumEditor/Private/CesiumEditor.cpp
```diff
@@ -748,7 +748,12 @@ AActor* FCesiumEditorModule::SpawnBlankTileset() {
 }
 
 AActor* FCesiumEditorModule::SpawnCartographicPolygon() {
-  return SpawnActorWithClass(ACesiumCartographicPolygon::StaticClass());
+  ACesiumCartographicPolygon* actor = static_cast<ACesiumCartographicPolygon*>(
```
j9liu: Remember the `p` prefix for pointers!

```suggestion
  ACesiumCartographicPolygon* pActor = static_cast<ACesiumCartographicPolygon*>(
```

Source/CesiumRuntime/Public/CesiumCartographicPolygon.h
```diff
@@ -81,6 +81,14 @@ class CESIUMRUNTIME_API ACesiumCartographicPolygon : public AActor {
   // AActor overrides
   virtual void PostLoad() override;
 
+  /**
+   * Set the spline points to a square centered where the editor camera view ray
+   * intersects the ground and sized to fit within thee viewport.
+   * @returns boolean indicating whether the spline points and transform were
+   * successfully recomputed.
```
j9liu:
```suggestion
   * @returns Whether the spline points and transform were successfully recomputed.
```

Source/CesiumRuntime/Private/CesiumCartographicPolygon.cpp
```diff
@@ -2,14 +2,66 @@
...
 #include <glm/glm.hpp>
 
 using namespace CesiumGeospatial;
 
+bool ACesiumCartographicPolygon::ResetSplineAndCenterInEditorViewport() {
+#if WITH_EDITOR
+  if (!GEditor)
+    return false;
+
+  UUnrealEditorSubsystem* editorSubsystem =
```
j9liu:
```suggestion
  UUnrealEditorSubsystem* pEditorSubsystem =
```

Source/CesiumRuntime/Private/CesiumCartographicPolygon.cpp
```diff
@@ -18,13 +70,19 @@ ACesiumCartographicPolygon::ACesiumCartographicPolygon() : AActor() {
...
+      {10000.0f, -10000.0f, 0},
+      {10000.0f, 10000.0f, 0},
+      {-10000.0f, 10000.0f, 0},
+  };
+
+  this->Polygon->SetSplinePoints(points, ESplineCoordinateSpace::Local);
+
+#if WITH_EDITOR
+  if (GEditor)
+    GEditor->RedrawAllViewports(true);
```
j9liu: Should this be done after `this->MakeLinear` below?

j9liu: Also, I have no idea if Copilot's comment is worth any salt, but it's worth commenting the reason you added this here.

Source/CesiumEditor/Private/CesiumEditor.cpp
```diff
@@ -748,7 +748,12 @@ AActor* FCesiumEditorModule::SpawnBlankTileset() {
...
+      SpawnActorWithClass(ACesiumCartographicPolygon::StaticClass()));
+  if (actor) {
+    actor->ResetSplineAndCenterInEditorViewport();
+  }
```
j9liu: This approach looks fine. We only want to do this when we add a polygon via the Editor panel.

Source/CesiumRuntime/Public/CesiumCartographicPolygon.h
```diff
@@ -81,6 +81,14 @@ class CESIUMRUNTIME_API ACesiumCartographicPolygon : public AActor {
...
+   */
+  bool ResetSplineAndCenterInEditorViewport();
```
j9liu, line 91: This name looks good to me. You could probably get the point across with a shorter name, but it's nitpicking at that point.

```suggestion
  bool ResetSplineForEditorViewport();
```

Source/CesiumRuntime/Private/CesiumCartographicPolygon.cpp
```diff
@@ -2,14 +2,66 @@
...
+  FVector viewDirection = viewRotation.Vector();
+
+  // Check if the ray and plane are parallel first to avoid division by zero
+  if (FMath::Abs(FVector::DotProduct(viewDirection, FVector::UpVector)) <
+      DBL_EPSILON) {
+    // Ray is parallel to the plane, no single intersection point.
+    return false;
+  }
+
+  const FPlane groundPlane(FVector::UpVector);
```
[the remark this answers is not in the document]
david-lively: Sticking with creating the polygon at Z=0 to be consistent with behavior before this PR.

kring:
A couple of user-facing things I noticed. In UE 5.7, in case it matters.

1. When you add a new Cesium Cartographic Polygon from the Quick Add, it is immediately selected in the Outliner. But the polygon doesn't appear in the viewport until you click some _other_ object in the Outliner, and then click back. This may not be a new problem, I'm not sure. But it had me thinking something was broken, for longer than I'd like to admit!
2. It's pretty easy to maneuver the camera below Z=0, and when you do, the polygon appears back at the origin instead of somewhere in the current viewport. Why cast the camera view diretion to Z=0 instead of casting it into the terrain surface?

Source/CesiumRuntime/Public/CesiumCartographicPolygon.h
kring: I'd suggest marking this entire function `WITH_EDITOR`. No reason to have a function available outside the editor that does nothing.

kring (commented):
Looks good to me! One minor comment below. I also tried a bunch of edge-casey things:
 
1. Creating a polygon while underground. No intersection, so we get default behavior.
2. Creating a polygon while zoomed out so far that the Earth is tiny. No intersection (the Earth is too far away) so we get default behavior.
3. Creating a polygon while physics meshes are turned off. No intersection (there's nothing to intersect with), so we get default behavior.
4. Create a cube in front of the camera, so the camera view ray hits the cube instead of the terrain. Polygon gets created near the cube, which seems fine I guess.

That all seems ok to me.

I was initially concerned that the polygon actor created far from the origin has a rotation applied to it. But this is just done by the CesiumGlobeAnchor component, and the raster overlay seems to account for it just fine.

j9liu:
Thanks @david-lively! I'm going to go ahead and merge ahead of today's release.

Issue #1611, opened 2025-02-11
Use Unreal's version of tinyxml2 in UE 5.5


kring:
Starting in UE 5.5, Epic added tinyxml2 headers and libs to `C:\Program Files\Epic Games\UE_5.5\Engine\Source\ThirdParty\TinyXML2`. By using these in our builds (using a vcpkg overlay port), rather than bringing our own, we should be able to avoid conflicts with the NDisplay plugin.

https://community.cesium.com/t/cesium-tinyxml2-with-ndisplay-error-on-launch/38323

We'll have to continue to bring our own tinxml2 in older UE versions, though.

LightPig1995:
In which version of the plugin has this modification been implemented?

kring:
It has not been implemented yet @LightPig1995, that's why the issue is still open.

j9liu:
Closed by #1790

PR #1790, opened 2026-02-03
Add vcpkg overlay to use Unreal's tinyxml2

timoore:
This uses tinyxml2 from Unreal's ThirdParty tree instead of building the library in vcpkg, which can cause issues. Doesn't work on Android yet; untested on Mac and IOS.

### DIFF
--- a/.github/workflows/buildLinux.yml
+++ b/.github/workflows/buildLinux.yml
@@ -60,14 +60,14 @@ jobs:
           unreal-program-name: ${{ inputs.unreal-program-name }}
           aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
           aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
-      - name: Configure vcpkg caching
-        uses: ./.github/actions/vcpkg-cache
-        with:
-          id: ${{ inputs.upload-package-base-name }}
       - name: Check out repository code
         uses: actions/checkout@v4
         with:
           submodules: recursive
+      - name: Configure vcpkg caching
+        uses: ./.github/actions/vcpkg-cache
+        with:
+          id: ${{ inputs.upload-package-base-name }}
       - name: Install Ninja and patchelf
         run: |
           sudo apt install -y ninja-build patchelf


kring:
Fixes #1611

kring:
The package ZIP files still contains libtinyxml2.a for Linux at `CesiumForUnreal/Source/ThirdParty/lib/Linux-x86_64-Release/libtinyxml2.a`. As far as I can tell, based on the size, this is _not_ the copy of this file from Unreal Engine. The corresponding Windows file is not included in the ZIP, so I'm not sure of the reason for this difference. Any ideas?

I also checked that the tinyxml2 files are still included for macOS, iOS, and Android. They are!

And I packaged a game with a TMS overlay on Windows. Works great!

So all looks good to me except for the extra Linux lib.

timoore:
> The package ZIP files still contains libtinyxml2.a for Linux at `CesiumForUnreal/Source/ThirdParty/lib/Linux-x86_64-Release/libtinyxml2.a`. As far as I can tell, based on the size, this is _not_ the copy of this file from Unreal Engine. The corresponding Windows file is not included in the ZIP, so I'm not sure of the reason for this difference. Any ideas?
> 
> I also checked that the tinyxml2 files are still included for macOS, iOS, and Android. They are!
> 
> And I packaged a game with a TMS overlay on Windows. Works great!
> 
> So all looks good to me except for the extra Linux lib.

I found that it was necessary to blow away my EZVCPKG directory, for me located in my home directory, for the changes to take effect. Are you seeing this locallly, or in CI?

kring:
> Are you seeing this locallly, or in CI?

CI. You can download the CI-built packages yourself and take a look if you want. If you haven't done that before, there are instructions for it here: https://cesium.com/learn/cesium-unreal/ref-doc/using-prerelease.html

timoore:
> > Are you seeing this locallly, or in CI?
> 
> CI. You can download the CI-built packages yourself and take a look if you want. If you haven't done that before, there are instructions for it here: https://cesium.com/learn/cesium-unreal/ref-doc/using-prerelease.html

My bad, there was a typo in the Linux logic. One of those "how did it ever work?" things.

kring:
@timoore that didn't seem to fix it. `libtinyxml2.a` is still found in `CesiumForUnreal/Source/ThirdParty/lib/Linux-x86_64-Release` in the CI-built release package.

I think the problem might be that vcpkg caching on Linux on CI is not working correctly. If you look at the [Windows57 build for your latest commit](https://github.com/CesiumGS/cesium-unreal/actions/runs/21945708222/job/63382589399?pr=1790), in the "Configure vcpkg caching" section, you'll see this line:

> Cache not found for input keys: vcpkg-CesiumForUnreal-57-windows-4316e590bcf20032265d6ca759fa0dde528d5a259c6f3577b60608bba4a73495-45d51b51a978ea98965db17fd123306ae3706f0a38fdb7b6f306064e24e3d2f5, vcpkg-CesiumForUnreal-57-windows-4316e590bcf20032265d6ca759fa0dde528d5a259c6f3577b60608bba4a73495-

But the [Linux57 build](https://github.com/CesiumGS/cesium-unreal/actions/runs/21945708222/job/63382589352?pr=1790) instead has this:

> Cache hit for: vcpkg-CesiumForUnreal-57-linux--

Two things about this:

1. The cache hit is probably the reason the libtinyxml2.a file still exists, because it existed last build due to the typo.
2. Why is there no hash in the Linux cache key? That hash is what's supposed to make sure the cache gets invalidated when overlay ports (etc) change.

kring:
I just pushed a fix to the cache key problem, which will hopefully eliminate the extra file, too. The linux build was cleverly computing a hash from source files... before it had the source files. 🤦

kring:
Everything looks to be working well now. Thanks @timoore!

Issue #1509, opened 2024-08-28
Issue compiling cesium-native in plugin


Blyron:
Platform: Ubuntu 22.04 
UE 5.3.2

Log:
user@user:~$ cd arenderer/Unreal/MyProjectUE5/Plugins/Cesium/extern
user@user:~/arenderer/Unreal/MyProjectUE5/Plugins/Cesium/extern$ cmake -B build -S . -DCMAKE_TOOLCHAIN_FILE="unreal-linux-toolchain.cmake" -DCMAKE_POSITION_INDEPENDENT_CODE=ON -DCMAKE_BUILD_TYPE=Release
-- The CXX compiler identification is Clang 16.0.6
-- The C compiler identification is Clang 16.0.6
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /home/user/UE5_engine/Engine/Extras/ThirdPartyNotUE/SDKs/HostLinux/Linux_x64/v22_clang-16.0.6-centos7/x86_64-unknown-linux-gnu/bin/clang++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /home/user/UE5_engine/Engine/Extras/ThirdPartyNotUE/SDKs/HostLinux/Linux_x64/v22_clang-16.0.6-centos7/x86_64-unknown-linux-gnu/bin/clang - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Using Unreal Engine installation at: /home/user/UE5_engine
-- VCPKG_TRIPLET x64-linux-unreal
-- EZVCPKG v0.1 starting up
	Website: https://github.com/jherico/ezvcpkg
-- EZVCPKG_BASEDIR envrionment variable not found and basedir not set, using default /home/user/.ezvcpkg
-- EZVCPKG initializing
	commit:     2024.07.12
	repository: https://github.com/microsoft/vcpkg.git
	local dir:  /home/user/.ezvcpkg/2024.07.12
-- EZVCPKG Building/Verifying package asyncplusplus using triplet x64-linux-unreal
-- EZVCPKG Building/Verifying package expected-lite using triplet x64-linux-unreal
-- EZVCPKG Building/Verifying package fmt using triplet x64-linux-unreal
-- EZVCPKG Building/Verifying package glm using triplet x64-linux-unreal
[...]

PR #1792, opened 2026-02-04
Fix OpenSSL directory search

mvollrath:
## Description

Filter out a "Patches" directory which found its way into my engine sources.  It looks like I'm not the only one who ended up here, so it seems worthy of a fix for posterity.

Context is I'm building v2.22.1 against Unreal 5.7.1 on Linux amd64.  For what it's worth, the faulty patch file is "darwin64-arm64-cc.patch" which shouldn't be here for OpenSSL 1.1.1t on my platform.

## Issue number or link

#1509 

## Author checklist

## Testing plan

* `mkdir -p "${UNREAL_ENGINE_ROOT}/Engine/Engine/Source/ThirdParty/OpenSSL/Patches"`
* Run cmake

### DIFF
--- a/extern/vcpkg-overlays/openssl/portfile.cmake
+++ b/extern/vcpkg-overlays/openssl/portfile.cmake
@@ -28,6 +28,11 @@ foreach(OPENSSL_POSSIBLE_ROOT_DIR IN LISTS OPENSSL_POSSIBLE_ROOT_DIRS)
   endif()
 endforeach()
 
+# Filter out directories which are not version numbers.
+# Must match e.g. "1.1.1t"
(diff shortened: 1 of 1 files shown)


kring:
I downloaded the UE 5.7.1 distribution for Linux, and I don't see an `Engine/Source/ThirdParty/OpenSSL/Patches` folder. Maybe it only pops up when building UE from source? In any case, this seems like a reasonable change. Filtering the directories down to only those that look like legitimate versions may help avoid some problems and it won't do any harm. Thanks for the PR @mvollrath!

mvollrath:
Seems like it only appears after the `Setup.sh` step.  I couldn't find where that happens in the Unreal sources, maybe part of whatever dependency blobs they are serving.

PR #1748, opened 2025-10-14
Implement Gaussian Splats

azrogers:
Closes #1670. This PR adds support for the `KHR_gaussian_splatting` extension in Cesium for Unreal (and, by virtue of CesiumGS/cesium-native#1262, SPZ as well). It does this using the Niagara particle system to handle rendering the splats from a custom NiagaraDataInterface. A subsystem is used to collect the splats into the singleton Niagara system. 

This is very very experimental! Even when this gets merged I think we'll need to call it an experimental feature. But some things we should take care of *before* this makes it out of a draft PR:
1. Sometimes the assertion `ensure(GroupCount.X <= GRHIMaxDispatchThreadGroupsPerDimension.X);` fails in `RenderGraphUtils.h` as a result of Niagara trying to generate too many sort keys at once. This is maybe an inevitable result of trying to pack so many splats into a single Niagara system, but packing them all into a single system is necessary so they sort together (else we'd get some awful looking seams between adjacent systems). I've implemented a really dirty system to try to mitigate this (trying to avoid resetting the system more often than every five ticks), but a better solution is almost certainly needed. Ideally we can find a way to avoid this assertion that doesn't involve splitting the scene into multiple Niagara systems.
- ⚠️ This has been confirmed to be an Unreal bug and a fix should be shipping in Unreal 5.8. We'll have to consider whether we want to find a workaround before 5.8 is out.
2. ~The Niagara system graph needs to be cleaned up. There's a lot of different code from a lot of different false-starts, rewrites, failed experiments, and other implementations I referenced (open-source implementations, license information for those implementations is included in the Niagara asset as a block comment). It would be good to streamline it into a finished implementation. Actually, it *might* be better just to do away with using the Niagara graph system entirely - just have the DataInterface provide a single function that takes in the index and matrices and returns all the splat information. That way we can have all of the compute shader code in the cpp file instead of putting it in a uasset that's opaque to git.~ ✅ Fixed in [c0a0ce9](https://github.com/CesiumGS/cesium-unreal/pull/1748/commits/c0a0ce90cbdcbffbf3a65867a447cd535f3dd8aa).
[...]

PR #1795, opened 2026-02-10
Tweaks to improve Gaussian splat rendering

j9liu:
Reflects much of the findings / discussion in #1748.

The most notable changes in this PR include:

- Removing the `OVERSCALE` fudge factor by fixing some coefficients earlier in the math
- Transform and sprite scale adjustments to better match CesiumJS's shader (incorporating some of @azrogers's changes)
- Sorting particles by depth (not distance)

### To Decide: World Space Computation?

I hacked putting the position / covariance matrix computation on the CPU with double-precision math. It didn't seem to help immediately, but I wonder if it's an improvement worth keeping.

On the one hand, it cleans up the shader quite nicely because it requires less parameters to be passed in. On the other hand, putting more computational load in the render command may affect performance. I had notable frame stutters before I created a separate "visibility" buffer so that the Editor didn't freeze whenever tile visibility was toggled. That seemed to be the culprit, but it's possible that performance will still regress in cases where we update the transform of the tileset.


j9liu:
Based on offline discussion with @azrogers, I'm going to revert the changes that compute world space covariance/position on the CPU because:

1. They do not improve appearance when the splats are far from the georeferenced origin.
2. They do not have a notable impact on the splats' appearance when they are close to the georeferenced origin.
3. The constant recalculation would have a heavy performance impact when used with origin shifting.

j9liu:
Restored the tile matrix + splat scale/rotation buffers in the original implementation. I see that there's a conflict with `main`, I can deconflict it tomorrow, or if it's easy for you to do it @azrogers, feel free to push to this PR as needed

j9liu:
Resolved the conflict, so this should be mergeable @azrogers!

azrogers:
Closing this because the changes have been integrated into #1748.

Issue #1472, opened 2024-07-03
Easy way to clamp objects to the ground


csciguy8:
We need a convenient way to force game objects to be at ground elevation of their current lat / long.

Came up on forum [here](https://community.cesium.com/t/stick-to-the-ground/13576).

There are many ways to achieve this with blueprints (as the forum post can show), but something easy would be ideal.

csciguy8:
We might be able to make a new cesium component that does exactly this, or add functionality to an existing component (like `CesiumGlobeAnchor`)

Solution would probably depend on [this cesium-native PR](https://github.com/CesiumGS/cesium-native/pull/783).

j9liu:
Came up on the forum here. https://community.cesium.com/t/is-there-a-way-to-detect-when-the-highest-lod-of-a-tileset-is-loaded/33540/

myepicspace:
As mentioned in [link](https://community.cesium.com/t/stick-to-the-ground/13576/6?u=epic_space) I solved this issue (or feature). I have tested it many times and it works very nice. even create a checkbox btn in my menu, to stick to ground actor or move free in air.

j9liu:
CC https://github.com/CesiumGS/cesium-native/issues/72

j9liu:
Ideas from dicsussion with @kring:

> UI-wise, I think a simple and useful place to start is with a new "Height Reference" on CesiumGlobeAnchor. The default value, "Ellipsoid", works as today. If you set it to "Tileset" instead, and specify a Tileset, then the height is a height above the current tileset surface, and the object moves as the tileset LOD changes.
> 
> We can initially implement that by doing a physics line trace every tick. That shouldn't be too difficult, and is fast enough for a handful of objects. From there it's just a matter of optimizing by doing the line traces only when necessary (i.e., the LOD changes).

david-lively:
Updated PR here: https://github.com/CesiumGS/cesium-unreal/pull/1797

PR #1797, opened 2026-02-13
Fix height relative to terrain

david-lively:
## Description

When an object is moving, or the camera moves resulting in tileset LOD changes, objects may float above or sink below the terrain. This PR adjusts the actual height as needed to maintain the given distance to the tileset. Properties to control this are added to `UCesiumGlobeAnchorComponent`. 

`HeightReference` is an enum indicating whether the height of the actor is fixed relative to the tileset or to the ellipsoid. `ECesiumHeightReferenceMode::Ellipsoid` is identical to the behavior before this PR; no height adjustment will be made. 

`ReferencedTileset` specifies an `ACesium3DTileset` actor, relative to which `Height` will be interpreted when `HeightReference` is `ECesiumHeightReferenceMode::Tileset`. If `ReferencedTileset` is `nullptr`, no adjustments will be made. 

`HeightUpdateInterval` how frequently, in `TickComponent()` calls, for the height to be checked and the actor moved. A value of `1` indicates that this will happen on every call. A value of `2` indicates every second call, and so forth. 

image

https://github.com/user-attachments/assets/098820d1-55b6-4487-a277-18669444b6f8

## Issue number or link

https://github.com/CesiumGS/cesium-unreal/issues/1472

## Author checklist

~~- [ ] I have submitted a [Contributor License Agreement](https://github.com/CesiumGS/community/tree/main/CLAs) (only needed once).~~

~~- [ ] I have added or updated unit tests to ensure consistent code coverage as necessary.~~
~~- [ ] I have updated the documentation as necessary.~~

## Testing plan

0. Open a test scene.
1. Add a cube to the scene. 
2. Set the cube's `Mobility` to `Movable`. 
3. Add a `CesiumGlobeAnchor` component. 
4. Set the `HeightReference` value to `Tileset`. 
5. Set the `ReferencedTileset` value to `CesiumWorldTerrain` or another `ACesium3DTileset` actor. 
6. Run the scene. 
7. In the Outliner, select the cube, and modify it's Transform to move it above a hill or other area where the tileset height is different. 
8. Observe that the cube adjusts its height to stay the same distance from the tileset. 

It might be easier to set these fields on Dynamic Pawn's `CesiumGlobeAnchor`, and fly over some terrain while observing that the pawn will remain a fixed height above the tileset.

### DIFF
--- a/Source/CesiumRuntime/Private/CesiumGlobeAnchorComponent.cpp
+++ b/Source/CesiumRuntime/Private/CesiumGlobeAnchorComponent.cpp
@@ -656,7 +656,7 @@ void UCesiumGlobeAnchorComponent::OnRegister() {
   }
 
   bool detectTransformChanges =
-      this->DetectTransformChanges &&
+      this->DetectTransformChanges ||
       this->HeightReference == ECesiumHeightReference::Tileset;
 
 #if WITH_EDITOR


Source/CesiumRuntime/Public/CesiumGlobeAnchorComponent.h
```diff
@@ -45,6 +60,43 @@ class CESIUMRUNTIME_API UCesiumGlobeAnchorComponent : public UActorComponent {
...
+   * the actor's Transform is updated.
+   */
+  UPROPERTY(
+      EditAnywhere,
+      BlueprintReadWrite,
+      // BlueprintGetter = GetInitialHeightAboveTerrain,
+      // BlueprintSetter = SetInitialHeightAboveTerrain,
+      Category = "Cesium",
+      Meta = (AllowPrivateAccess))
+  float FixedHeightAboveTerrain = 0;
```
j9liu: I'm wondering how this will interact with the existing **Height** property and the user expectations around it...

Going back to [@kring's thoughts](https://github.com/CesiumGS/cesium-unreal/issues/1472#issuecomment-3807562486) in the original issue: 

> UI-wise, I think a simple and useful place to start is with a new "Height Reference" on CesiumGlobeAnchor. The default value, "Ellipsoid", works as today. If you set it to "Tileset" instead, and specify a Tileset, then the height is a height above the current tileset surface, and the object moves as the tileset LOD changes.

Although it's true that we have traditionally used **Height** to mean height about the WGS84 ellipsoid, a separate "height" parameter could be confusing. I think it's possible to carefully + intentionally repurpose the existing **Height** property, and have the new **Height Reference** property determine its meaning. What do you think?

Source/CesiumRuntime/Public/CesiumGlobeAnchorComponent.h
```diff
@@ -455,6 +520,14 @@ class CESIUMRUNTIME_API UCesiumGlobeAnchorComponent : public UActorComponent {
...
   void Sync();
 
+  /**
+   * @brief Calculate the height distance from the actor to the terrain.
+   * @param GroundIntersection Calculated point on the terrain directly below
+   * the actor.
+   * @return Whether an intersection was found.
+   */
+  UFUNCTION(BlueprintCallable, Category = "Cesium")
+  bool QueryPositionOnTileset(FVector& GroundIntersection);
```
j9liu: This name is slightly inaccurate because we're not filtering out tileset vs. non-tileset intersections.

Also for the parameter name, see our naming guidelines in the [Style Guide](https://cesium.com/learn/cesium-unreal/ref-doc/style-guide.html#clarity).

```suggestion
  bool FindGroundIntersection(FVector& LongitudeLatitudeHeight);
```

Source/CesiumRuntime/Public/CesiumGlobeAnchorComponent.h
```diff
@@ -45,6 +60,43 @@ class CESIUMRUNTIME_API UCesiumGlobeAnchorComponent : public UActorComponent {
       Meta = (AllowPrivateAccess))
   TSoftObjectPtr<ACesiumGeoreference> Georeference = nullptr;
 
+  UPROPERTY(
```
j9liu, line 80: Needs a comment.

Source/CesiumRuntime/Public/CesiumGlobeAnchorComponent.h
```diff
@@ -9,6 +9,21 @@
 
 class ACesiumGeoreference;
 
+/**
+ * Mode for height reference of
+ */
```
j9liu, line 16: Something happened to this comment here?

Source/CesiumRuntime/Public/CesiumGlobeAnchorComponent.h
```diff
@@ -9,6 +9,21 @@
...
+UENUM(BlueprintType)
+enum class ECesiumHeightReferenceMode : uint8 {
+  /**
+   * Height offset relative to the ellipsoid.
+   */
+  RelativeToEllipsoid UMETA(DisplayName = "Relative to ellipsoid"),
```
j9liu:
```suggestion
  /**
   * Height is interpreted relative to the ellipsoid.
   */
  Ellipsoid UMETA(DisplayName = "Relative to Ellipsoid"),
```

Source/CesiumRuntime/Public/CesiumGlobeAnchorComponent.h
```diff
@@ -9,6 +9,21 @@
...
+  /**
+   * Reference height is relative to the terrain.
+   */
+  RelativeToTileset UMETA(DisplayName = "Clamp to terrain"),
```
j9liu: **Edit:** I wrote this before making this comment here: https://github.com/CesiumGS/cesium-unreal/pull/1797#discussion_r2806076750. But we can change things so that this name will be accurate.

<hr>

`RelativeToTileset` might be misleading because we're not actually testing if our intersection is with a tileset or not. It could be any object underneath it, unless we want to specifically look for intersections with 3D Tiles (`CesiumGltfPrimitiveComponent`s). So we may need to distinguish between "Tileset" and "Ground".

```suggestion
  /**
   * Height is interpreted relative to the surface of a 3D Tileset. 
   */
  Tileset UMETA(DisplayName = "Relative to Tileset"),
```

Source/CesiumRuntime/Public/CesiumGlobeAnchorComponent.h
```diff
@@ -45,6 +60,43 @@ class CESIUMRUNTIME_API UCesiumGlobeAnchorComponent : public UActorComponent {
...
   TSoftObjectPtr<ACesiumGeoreference> Georeference = nullptr;
 
+  UPROPERTY(
+      EditAnywhere,
+      BlueprintReadWrite,
+      BlueprintGetter = GetHeightReference,
+      BlueprintSetter = SetHeightReference,
+      Category = "Cesium",
+      meta = (AllowPrivateAccess))
+  ECesiumHeightReferenceMode HeightReference =
```
j9liu: `HeightReference` should be `Ellipsoid` by default to preserve existing `CesiumGlobeAnchor` behavior.

Source/CesiumRuntime/Public/CesiumGlobeAnchorComponent.h
j9liu: A couple comments:

1. Should we make this function private? This might be confusing as public API given the existence of `SampleHeightMostDetailed`.
2. We're not filtering out tileset vs. non-tileset intersections here, but if we're going to advertise this as "relative to tileset" then I think we should. Per [Kevin's comment](https://github.com/CesiumGS/cesium-unreal/issues/1472#issuecomment-3807562486): 
> If you set it to "Tileset" instead, and specify a Tileset, then the height is a height above the current tileset surface, and the object moves as the tileset LOD changes.

Source/CesiumRuntime/Public/CesiumGlobeAnchorComponent.h
```diff
@@ -45,6 +60,43 @@ class CESIUMRUNTIME_API UCesiumGlobeAnchorComponent : public UActorComponent {
...
+   * How many frames to skip between height reference updates.
+   */
+  UPROPERTY(
+      EditAnywhere,
+      BlueprintReadWrite,
+      BlueprintGetter = GetHeightReferenceUpdateInterval,
+      BlueprintSetter = SetHeightReferenceUpdateInterval,
+      Category = "Cesium",
+      Meta = (AllowPrivateAccess))
+  int HeightReferenceUpdateInterval = 1;
```
j9liu: I like that you added this setting, just put an `EditCondition` so it's only be active when height reference =/= ellipsoid. Ellipsoid height doesn't require raycasting to compute.

I might tweak the name too to reflect this:

```suggestion
  int TilesetHeightUpdateInterval = 1;
```

Source/CesiumRuntime/Public/CesiumGlobeAnchorComponent.h
david-lively: Renamed and made it private.

Source/CesiumRuntime/Public/CesiumGlobeAnchorComponent.h
david-lively: Implemented in a later commit.

Source/CesiumRuntime/Public/CesiumGlobeAnchorComponent.h
david-lively: Fixed in latest commit.

CHANGES.md
```diff
@@ -2,6 +2,11 @@
 
 ### ???
 
+##### Additions :tada:
+
+- Added `UCesiumGlobeAnchor::HeightReference`. Setting this to `ECesiumHeightReferenceMode::Tileset` will cause the actor to maintain its height above the tileset or geometry during LOD transitions or object movement.
+- Added `UCesiumGlobeAncor::TilesetHeightUpdateInterval` to specify how often the actor's height is updated when `HeightReference` is `Tileset`.
```
```suggestion
- Added `UCesiumGlobeAnchorComponent::TilesetHeightUpdateInterval` to specify how often the actor's height is updated when `HeightReference` is `Tileset`.
```

CHANGES.md
```diff
@@ -2,6 +2,11 @@
 
 ### ???
 
+##### Additions :tada:
+
+- Added `UCesiumGlobeAnchor::HeightReference`. Setting this to `ECesiumHeightReferenceMode::Tileset` will cause the actor to maintain its height above the tileset or geometry during LOD transitions or object movement.
```
```suggestion
- Added `UCesiumGlobeAnchorComponent::HeightReference`. Setting this to `ECesiumHeightReferenceMode::Tileset` will cause the actor to maintain its height above the tileset or geometry during LOD transitions or object movement.
```

Source/CesiumRuntime/Public/CesiumGlobeAnchorComponent.h
```diff
@@ -526,8 +585,32 @@ class CESIUMRUNTIME_API UCesiumGlobeAnchorComponent : public UActorComponent {
...
+      float DeltaTime,
+      ELevelTick TickType,
+      FActorComponentTickFunction* ThisTickFunction) override;
+
 #pragma region Implementation Details
 private:
+  int _heightReferenceUpdateCounter = 0;
+
+  /**
+   * When HeightReference is ::Tileset, this is the target height
```
j9liu:
```suggestion
   * When HeightReference is Tileset, this is the target height
```

CHANGES.md
```diff
@@ -8,6 +8,8 @@
 
 ##### Additions :tada:
 
+- Added `UCesiumGlobeAnchor::HeightReference`. Setting this to `ECesiumHeightReferenceMode::Tileset` will cause the actor to maintain its height above the tileset or geometry during LOD transitions or object movement.
+- Added `UCesiumGlobeAncor::TilesetHeightUpdateInterval` to specify how often the actor's height is updated when `HeightReference` is `Tileset`.
```
```suggestion
- Added `UCesiumGlobeAnchorComponent::TilesetHeightUpdateInterval` to specify how often the actor's height is updated when `HeightReference` is `Tileset`.
```

CHANGES.md
```diff
@@ -8,6 +8,8 @@
 
 ##### Additions :tada:
 
+- Added `UCesiumGlobeAnchor::HeightReference`. Setting this to `ECesiumHeightReferenceMode::Tileset` will cause the actor to maintain its height above the tileset or geometry during LOD transitions or object movement.
```
```suggestion
- Added `UCesiumGlobeAnchorComponent::HeightReference`. Setting this to `ECesiumHeightReferenceMode::Tileset` will cause the actor to maintain its height above the tileset or geometry during LOD transitions or object movement.
```

Source/CesiumRuntime/Public/CesiumGlobeAnchorComponent.h
```diff
@@ -45,6 +61,29 @@ class CESIUMRUNTIME_API UCesiumGlobeAnchorComponent : public UActorComponent {
...
+   *
+   * ELLIPSOID indicates that the object will maintain a fixed height above the
+   * ellipsoid. No special handling is performed.
+   *
+   * TILESET causes the object to maintain its height above the tileset. The
+   * object may move vertically to maintain a fixed height above the terrain or
+   * other geometry below the object.
+   */
+  UPROPERTY()
+  ECesiumHeightReferenceMode HeightReference =
```
j9liu: This is looking better, but one thing I want to bring up is specifying which `Tileset` the height is being referenced against:

>  If you set it to "Tileset" instead, and specify a Tileset, then the height is a height above the current tileset surface, and the object moves as the tileset LOD changes.

Right now the logic is being done against any surface in Unreal, which is not a bad option, but it isn't faithful to the implications of "Tileset". Would this be a heavy lift to implement? In the `Physics` raytrace it may be possible to get the owner of the object that was intersected, so you could test if it was the same tileset specified.

Source/CesiumRuntime/Public/CesiumGlobeAnchorComponent.h
```diff
@@ -45,6 +61,29 @@ class CESIUMRUNTIME_API UCesiumGlobeAnchorComponent : public UActorComponent {
...
+      ECesiumHeightReferenceMode::Ellipsoid;
+
+  /**
+   * Interval, in Ticks, to update object's height when HeightReference is set
```
j9liu:
```suggestion
   * The interval, in Ticks, in which to update the object's height when HeightReference is set
```

Source/CesiumEditor/Private/CesiumGlobeAnchorCustomization.h
```diff
@@ -108,6 +109,35 @@ class UCesiumGlobeAnchorDerivedProperties : public UObject,
...
+   *
+   * A value of 1 causes Height to be updated on every Tick.
+   */
+  UPROPERTY(
+      EditAnywhere,
+      Category = "Cesium",
+      meta =
+          (EditCondition =
+               "HeightReference == ECesiumHeightReferenceMode::Tileset",
+           EditConditionHides,
```
j9liu: We don't usually use `EditConditionHides`, as written in our [API design guide](https://cesium.com/learn/cesium-unreal/ref-doc/api-design.html#conditional-properties). We let things show up as greyed out.

Source/CesiumRuntime/Public/CesiumGlobeAnchorComponent.h
```diff
@@ -333,7 +385,8 @@ class CESIUMRUNTIME_API UCesiumGlobeAnchorComponent : public UActorComponent {
       BlueprintPure,
       Category = "Cesium",
       Meta = (ReturnDisplayName = "LongitudeLatitudeHeight"))
-  FVector GetLongitudeLatitudeHeight() const;
+  FVector
+  GetLongitudeLatitudeHeight(const bool IgnoreHeightReference = false) const;
```
j9liu: Would it make more sense to explicitly pass in a `HeightReference` instead? So this could look something like:

```suggestion
  GetLongitudeLatitudeHeight(ECesiumHeightReferenceMode HeightReference = ECesiumHeightReferenceMode::Ellipsoid) const;
```

j9liu: Should something like this be added to `MoveToLongitudeLatitudeHeight` too (and potentially override the internal height reference)?

Source/CesiumRuntime/Public/CesiumGlobeAnchorComponent.h
```diff
@@ -567,8 +624,34 @@ class CESIUMRUNTIME_API UCesiumGlobeAnchorComponent : public UActorComponent {
...
+  /**
+   * Find the location of the component directly above or below the component.
+   * Returns whether an intersection was found.
+   *
+   * alternateActorPosition, if specified, performs the intersection test
+   * starting at a different location.
+   */
+  bool _queryLongitudeLatitudeHeightPositionOnTileset(
+      FVector& groundIntersection,
+      const std::optional<FVector>& alternateActorPosition = std::nullopt);
```
j9liu:
```suggestion
      const std::optional<FVector>& alternateStartPosition = std::nullopt);
```

Source/CesiumEditor/Private/CesiumGlobeAnchorCustomization.cpp
```diff
@@ -238,8 +268,9 @@ void UCesiumGlobeAnchorDerivedProperties::PostEditChangeProperty(
                             Latitude) ||
         propertyName == GET_MEMBER_NAME_CHECKED(
                             UCesiumGlobeAnchorDerivedProperties,
-                            Height)) {
+                            HeightReference)) {
       this->GlobeAnchor->Modify();
+      this->GlobeAnchor->SetHeightReference(this->HeightReference);
```
j9liu: Just wondering, why did `Height` get swapped out for `HeightReference` here?

Source/CesiumRuntime/Private/CesiumGlobeAnchorComponent.cpp
```diff
@@ -757,13 +824,63 @@ void UCesiumGlobeAnchorComponent::_updateFromNativeGlobeAnchor(
...
     this->_setCurrentRelativeTransform(VecMath::createTransform(anchorToLocal));
   } else {
     this->_lastRelativeTransformIsValid = false;
   }
 }
 
+bool UCesiumGlobeAnchorComponent::
+    _queryLongitudeLatitudeHeightPositionOnTileset(
+        FVector& groundIntersection,
+        const std::optional<FVector>& alternateActorPosition) {
```
j9liu:
```suggestion
        const std::optional<FVector>& alternateStartPosition) {
```

Source/CesiumRuntime/Private/CesiumGlobeAnchorComponent.cpp
```diff
@@ -772,7 +889,21 @@ void UCesiumGlobeAnchorComponent::_onActorTransformChanged(
...
+      this->GetHeightReference() == ECesiumHeightReferenceMode::Tileset) {
+    // update the height-above-reference based on the new transform.
+    FVector positionOnTileset;
+    if (this->_queryLongitudeLatitudeHeightPositionOnTileset(
+            positionOnTileset)) {
+      FVector realLLH = this->GetLongitudeLatitudeHeight(true);
+      this->_fixedHeightAboveHeightReference = realLLH.Z - positionOnTileset.Z;
+    }
+  }
+#endif
```
j9liu, line 956: Why is this Editor only?

Source/CesiumRuntime/Private/CesiumGlobeAnchorComponent.cpp
```diff
@@ -816,3 +947,24 @@ void UCesiumGlobeAnchorComponent::_onGeoreferenceChanged() {
...
+
+  if (--this->_heightReferenceUpdateCounter > 0) {
+    return;
+  }
+
+  this->_heightReferenceUpdateCounter = this->TilesetHeightUpdateInterval;
+  FVector llh = this->GetLongitudeLatitudeHeight(true);
+  llh.Z = 0;
+
+  this->MoveToLongitudeLatitudeHeight(llh);
```
j9liu: I'm confused by what's happening here... but maybe the code will get clearer if we factor in explicit `HeightReference` specification?

Source/CesiumRuntime/Private/CesiumGlobeAnchorComponent.cpp
david-lively: @j9liu  This resets the Height relative to the reference when the actor is dragged around in the editor. This event doesn't even fire when, for instance, the transform is updated via Blueprint. The idea is to set the Height value automagically when the user is placing or moving the actor within the level. This is the only place we write to `_fixedHeightAboveHeightReference` other than the explicit `::SetHeight` property setter.

Source/CesiumRuntime/Public/CesiumGlobeAnchorComponent.h
david-lively: Added a new `HeightReferenceTileset` field which specifies the tileset to use for distance calculations.

Source/CesiumRuntime/Private/CesiumGlobeAnchorComponent.cpp
david-lively: The call to MoveToLongitudeLatitudeHeight causes the actor's position to be updated relative to the tileset. I added a comment explaining this.

Source/CesiumRuntime/Public/CesiumGlobeAnchorComponent.h
```diff
@@ -364,22 +444,27 @@ class CESIUMRUNTIME_API UCesiumGlobeAnchorComponent : public UActorComponent {
       BlueprintPure,
       Category = "Cesium",
       Meta = (ReturnDisplayName = "Height"))
-  double GetHeight() const { return this->GetLongitudeLatitudeHeight().Z; }
+  double GetHeight(const bool IgnoreHeightReference = true) const;
```
j9liu: See my comment under `GetLongitudeLatitudeHeight`

Source/CesiumRuntime/Public/CesiumGlobeAnchorComponent.h
```diff
@@ -2,12 +2,30 @@
...
 #include "Components/ActorComponent.h"
 #include "Delegates/IDelegateInstance.h"
 #include "CesiumGlobeAnchorComponent.generated.h"
 
 class ACesiumGeoreference;
+class ACesium3DTileset;
+
+/**
+ * When updating actor's height during movement or LOD transitions, controls
+ * whether to maintain height relative to the ellipsoid or tileset.
```
j9liu: **NIT**: Sounds more general + succinct?

```suggestion
 * The frame of reference from which to interpret a given height value.
```

Source/CesiumRuntime/Public/CesiumGlobeAnchorComponent.h
```diff
@@ -2,12 +2,30 @@
...
+ */
+UENUM(BlueprintType)
+enum class ECesiumHeightReferenceMode : uint8 {
+  /**
+   * Height is interpreted relative to ellipsoid.
```
j9liu:
```suggestion
   * Height is interpreted relative to the ellipsoid.
```

Source/CesiumRuntime/Public/CesiumGlobeAnchorComponent.h
```diff
@@ -45,6 +63,43 @@ class CESIUMRUNTIME_API UCesiumGlobeAnchorComponent : public UActorComponent {
...
 
+  /**
+   * Reference used when updating height during movement or LOD transitions.
+   *
+   * ELLIPSOID indicates that the object will maintain a fixed height above the
+   * ellipsoid. No special handling is performed.
+   *
+   * TILESET causes the object to maintain its height above the tileset. The
+   * object may move vertically to maintain a fixed height above the terrain or
+   * other geometry below the object.
```
j9liu: **NIT**:

```suggestion
   * The frame of reference in which to interpret the object's height.
   *
   * "Ellipsoid" indicates the object's height above the ellipsoid set on the CesiumGeoreference. The object will remain at this height unless it is otherwise changed.
   *
   * "Tileset" indicates a height above a given tileset. The
   * object will move vertically to maintain the specified height.
```

Source/CesiumRuntime/Public/CesiumGlobeAnchorComponent.h
j9liu, line 1: I'm running into some weird behavior where when you add a CesiumGlobeAnchor, then change the reference to Tileset`, the fields are grayed out (but the names aren't??):

Image

But if I click on another actor, then click back, the fields become editable:

Image

Do you see this too?

Source/CesiumRuntime/Public/CesiumGlobeAnchorComponent.h
```diff
@@ -45,6 +63,43 @@ class CESIUMRUNTIME_API UCesiumGlobeAnchorComponent : public UActorComponent {
...
+   */
+  UPROPERTY()
+  ECesiumHeightReferenceMode HeightReference =
+      ECesiumHeightReferenceMode::Ellipsoid;
+
+  /**
+   * The tileset actor used as a height reference when moving or during LOD
+   * transitions. Only used when HeightReference is set to
+   * ECesiumHeightReferenceMode::Tileset.
```
j9liu: **NIT**:

```suggestion
   * The tileset actor to use as a height reference for the object. The object
   * will maintain its height relative to this tileset even through movement or level-of-detail transitions. Only used when HeightReference is set to
   * ECesiumHeightReferenceMode::Tileset.
```

Source/CesiumRuntime/Public/CesiumGlobeAnchorComponent.h
```diff
@@ -45,6 +63,43 @@ class CESIUMRUNTIME_API UCesiumGlobeAnchorComponent : public UActorComponent {
...
+   */
+  UPROPERTY(
+      EditAnywhere,
+      BlueprintReadWrite,
+      BlueprintGetter = GetHeightReferenceTileset,
+      BlueprintSetter = SetHeightReferenceTileset,
+      Category = "Cesium",
+      Meta = (AllowPrivateAccess))
+  TSoftObjectPtr<ACesium3DTileset> HeightReferenceTileset = nullptr;
```
j9liu: **NIT**: What do you think of this name instead?

```suggestion
  TSoftObjectPtr<ACesium3DTileset> ReferencedTileset = nullptr;
```

Source/CesiumRuntime/Public/CesiumGlobeAnchorComponent.h
```diff
@@ -45,6 +63,43 @@ class CESIUMRUNTIME_API UCesiumGlobeAnchorComponent : public UActorComponent {
...
+
+  /**
+   * The interval, in Ticks, in which to update the object's height when
+   * HeightReference is set to Tileset.
+   *
+   * A value of 1 causes Height to be updated on every Tick.
+   */
+  UPROPERTY()
+  int TilesetHeightUpdateInterval = 1;
```
j9liu: **NIT**: Think it's okay to omit `Tileset` here.

```suggestion
  int HeightUpdateInterval = 1;
```

Source/CesiumRuntime/Public/CesiumGlobeAnchorComponent.h
j9liu: Bumping this comment. "IgnoreHeightReference" is kind of misleading here; you're basically using this to always get the height above the WGS84 ellipsoid, right? Why not take a HeightReference enum instead, with the default being `Ellipsoid`?

Source/CesiumRuntime/Public/CesiumGlobeAnchorComponent.h
```diff
@@ -364,22 +444,27 @@ class CESIUMRUNTIME_API UCesiumGlobeAnchorComponent : public UActorComponent {
...
+   * set, the Z height value is the desired height above the reference tileset.
+   *
    * If `AdjustOrientationForGlobeWhenMoving` is enabled, the Actor's
    * orientation will also be adjusted to account for globe curvature.
    */
   UFUNCTION(BlueprintCallable, Category = "Cesium")
-  void MoveToLongitudeLatitudeHeight(const FVector& LongitudeLatitudeHeight);
+  void MoveToLongitudeLatitudeHeight(
+      const FVector& LongitudeLatitudeHeight,
+      const bool IgnoreHeightReference = false);
```
j9liu: See my comment under `GetLongitudeLatitudeHeight`

Source/CesiumRuntime/Public/CesiumGlobeAnchorComponent.h
```diff
@@ -567,8 +652,38 @@ class CESIUMRUNTIME_API UCesiumGlobeAnchorComponent : public UActorComponent {
...
+
+  /**
+   * Determines if HeightReference is Tileset and HeightReferenceTileset is set.
+   */
+  bool _isUsingHeightReference() const;
+  /**
+   * Project the component's position onto the tileset specified by
+   * HeightReferenceTileset, Returns whether an intersection was found.
+   *
+   * alternateActorPosition, if specified, performs the intersection test
```
j9liu:
```suggestion
   * alternateStartPosition, if specified, performs the intersection test
```

CHANGES.md
```diff
@@ -8,6 +8,9 @@
 
 ##### Additions :tada:
 
+- Added `UCesiumGlobeAnchorComponent::HeightReferenceTileset`. Set this field to a `ACesium3DTileset` actor reference to maintain a fixed height relative to the tileset.
+- Added `UCesiumGlobeAnchorComponent::HeightReference`. Setting this to `ECesiumHeightReferenceMode::Tileset` will cause the actor to maintain its height above the specified tileset LOD transitions or object movement.
```
j9liu: **NIT**: This could be consolidated.

```suggestion
- Added `UCesiumGlobeAnchorComponent::HeightReference`. Setting this to `ECesiumHeightReferenceMode::Tileset` will cause the actor to maintain its height above the given `UCesiumGlobeAnchorComponent::ReferencedTileset`, even during movement or level-of-detail transitions.
```

Source/CesiumEditor/Private/CesiumGlobeAnchorCustomization.cpp
```diff
@@ -100,6 +105,46 @@ void FCesiumGlobeAnchorCustomization::CustomizeDetails(
...
+  Group.AddPropertyRow(HeightReferenceProperty.ToSharedRef());
+
+  TSharedPtr<IPropertyHandle> TilesetHeightUpdateIntervalProperty =
+      DetailBuilder.AddObjectPropertyData(
+          View,
+          GET_MEMBER_NAME_CHECKED(
+              UCesiumGlobeAnchorDerivedProperties,
+              TilesetHeightUpdateInterval));
+
+  Group.AddPropertyRow(TilesetHeightUpdateIntervalProperty.ToSharedRef());
```
j9liu: I would move this below `TilesetProperty`; it's more immediate for the user to have to set the Tileset than to adjust the update interval.

Source/CesiumEditor/Private/CesiumGlobeAnchorCustomization.h
```diff
@@ -108,6 +113,46 @@ class UCesiumGlobeAnchorDerivedProperties : public UObject,
...
+   * Reference used when updating height during movement or LOD transitions.
+   *
+   * ELLIPSOID indicates that the object will maintain a fixed height above the
+   * ellipsoid. No special handling is performed.
+   *
+   * TILESET causes the object to maintain its height above the tileset. The
+   * object may move vertically to maintain a fixed height above the terrain or
+   * other geometry below the object.
+   */
+  UPROPERTY(EditAnywhere, Category = "Cesium")
```
j9liu: Can you update the docs here after your respective updates to the docs on `UCesiumGlobeAnchorComponent`?

Source/CesiumRuntime/Public/CesiumGlobeAnchorComponent.h
j9liu: Hm, almost wonder if you need a `None` option and call this parameter `HeightReferenceOverride`. Or `Use Existing Reference` or something. Would need to namesmith.

Source/CesiumRuntime/Public/CesiumGlobeAnchorComponent.h
j9liu, line 1: Another UX thing: I like that the object stays in place when you first switch to `Tileset` reference. But after you set `HeightReferenceTileset`, it'd be nice if the object could compute its height above the tileset and reflect that in the `Height` property.

Source/CesiumRuntime/Private/CesiumGlobeAnchorComponent.cpp
```diff
@@ -325,20 +364,48 @@ void UCesiumGlobeAnchorComponent::InvalidateResolvedGeoreference() {
...
+  FVector realLongitudeLatitudeHeight = TargetLongitudeLatitudeHeight;
+  FVector tilesetPosition{};
+
+  if (!IgnoreHeightReference && this->_isUsingHeightReference() &&
+      _queryLongitudeLatitudeHeightPositionOnTileset(
+          tilesetPosition,
+          TargetLongitudeLatitudeHeight)) {
+    _fixedHeightAboveHeightReference = TargetLongitudeLatitudeHeight.Z;
+    realLongitudeLatitudeHeight.Z =
+        tilesetPosition.Z + _fixedHeightAboveHeightReference;
```
j9liu: Per style guide, prefer using `this->` before member variables/functions:

```suggestion
    this->_fixedHeightAboveHeightReference = TargetLongitudeLatitudeHeight.Z;
    realLongitudeLatitudeHeight.Z =
        tilesetPosition.Z + this->_fixedHeightAboveHeightReference;
```

Source/CesiumRuntime/Public/CesiumGlobeAnchorComponent.h
david-lively: Fixed; changing the reference mode or reference tileset is now correctly handled.

Source/CesiumRuntime/Private/CesiumGlobeAnchorComponent.cpp
david-lively: Fixed in next push

Source/CesiumEditor/Private/CesiumGlobeAnchorCustomization.cpp
david-lively: Fixed in next push.

Source/CesiumRuntime/Public/CesiumGlobeAnchorComponent.h
```diff
@@ -436,6 +437,10 @@ class CESIUMRUNTIME_API UCesiumGlobeAnchorComponent : public UActorComponent {
    * Gets the height in meters above the ellipsoid or tileset, depending on
    * the value of HeightReference.
    *
+   * When OverrideHeightReference is true, this returns the height above the
+   * reference tileset, if set. Otherwise, it will return the height above
+   * the ellipsoid.
```
j9liu: `OverrideHeightReference` seems to be mainly in service of the `Tileset` option, and not `Ellipsoid`. It doesn't do anything in the latter situation, which is a bit confusing.

I still advocate for erring on the side of over-clarifying here, and I think this function (and the others with `OverrideHeightReference`) should instead take in an explicit `HeightReference` parameter, to specify which reference they want to interpret the height from. This is more informative for users and we won't have to invent special rules under-the-hood on their behalf. The default value can be `Ellipsoid` for backwards compatibility, since the ellipsoid was the only previously supported option.

If they want to just use what the tileset is using, then they can pass in `GetHeightReference()`. What do you think about that approach?

Source/CesiumRuntime/Public/CesiumGlobeAnchorComponent.h
```diff
@@ -15,15 +15,20 @@ class ACesium3DTileset;
...
-  Ellipsoid UMETA(DisplayName = "Ellipsoid"),
+  Ellipsoid,
   /**
    * Height is interpreted relative to the surface of a 3D tileset.
    */
-  Tileset UMETA(DisplayName = "Tileset")
+  Tileset,
+  /**
+   * Height is interpreted based on the HeightReference and ReferencedTileset
+   * values of the associated CesiumGlobeAnchor component.
```
j9liu:
```suggestion
   * No specified height reference. Used only for HeightReferenceOverride
   *  in the CesiumGlobeAnchor Blueprints. Do not use otherwise. 
```

Source/CesiumRuntime/Public/CesiumGlobeAnchorComponent.h
```diff
@@ -435,11 +454,11 @@ class CESIUMRUNTIME_API UCesiumGlobeAnchorComponent : public UActorComponent {
 
   /**
    * Gets the height in meters above the ellipsoid or tileset, depending on
-   * the value of HeightReference.
+   * the value of HeightReferenceOverride.
```
j9liu:
```suggestion
   * Gets the height relative to the specified height reference.
```

Source/CesiumRuntime/Public/CesiumGlobeAnchorComponent.h
j9liu: Break this off from the rest of the comment, and move the "Do not confuse" passage after the comments about Height Reference.

```suggestion
   * The Height (Z) is measured relative to the specified HeightReference. Do not
```

j9liu:
```suggestion
   * and height relative to the specified height reference.
```

Source/CesiumRuntime/Public/CesiumGlobeAnchorComponent.h
```diff
@@ -405,15 +423,16 @@ class CESIUMRUNTIME_API UCesiumGlobeAnchorComponent : public UActorComponent {
    * mean sea level, which can be tens of meters higher or lower depending on
    * where in the world the object is located.
```
j9liu: Move this below the comments about height reference

Source/CesiumRuntime/Public/CesiumGlobeAnchorComponent.h
```diff
@@ -405,15 +423,16 @@ class CESIUMRUNTIME_API UCesiumGlobeAnchorComponent : public UActorComponent {
...
    *
-   * When HeightReference is Tileset, and a valid ReferencedTileset is
+   * When HeightReferenceOverride is Tileset, and a valid ReferencedTileset is
```
j9liu: How does this sound?
  
```suggestion
   * When HeightReferenceOverride is None, this function will use the HeightReference
   * that is set on the component.
   * 
   * Otherwise, when HeightReferenceOverride is Tileset and a valid ReferencedTileset is
   * set, the returned value is the height above the reference tileset.
   * 
   * Likewise, when HeightReferenceOverride is Ellipsoid, the returned value is the
   * height above the ellipsoid in meters.
```

Source/CesiumRuntime/Public/CesiumGlobeAnchorComponent.h
```diff
@@ -435,11 +454,11 @@ class CESIUMRUNTIME_API UCesiumGlobeAnchorComponent : public UActorComponent {
...
    * Gets the height in meters above the ellipsoid or tileset, depending on
-   * the value of HeightReference.
+   * the value of HeightReferenceOverride.
    *
-   * When OverrideHeightReference is true, this returns the height above the
-   * reference tileset, if set. Otherwise, it will return the height above
-   * the ellipsoid.
+   * When HeightReferenceOverride is Tileset or None, this returns the height
+   * above the reference tileset, if set. Otherwise, it will return the height
+   * above the ellipsoid.
```
j9liu:
```suggestion
   * When HeightReferenceOverride is None, this function will use the HeightReference
   * that is set on the component.
   * 
   * Otherwise, when HeightReferenceOverride is Tileset and a valid ReferencedTileset is
   * set, the returned value is the height above the reference tileset.
   * 
   * Likewise, when HeightReferenceOverride is Ellipsoid, the returned value is the
   * height above the ellipsoid in meters.
```

Source/CesiumRuntime/Public/CesiumGlobeAnchorComponent.h
```diff
@@ -460,7 +481,7 @@ class CESIUMRUNTIME_API UCesiumGlobeAnchorComponent : public UActorComponent {
    * level, which can be tens of meters higher or lower depending on where in
    * the world the object is located.
    *
-   * When HeightReference is Tileset, and a valid ReferencedTileset is
+   * When HeightReferenceOverride is Tileset, and a valid ReferencedTileset is
    * set, the Z height value is the desired height above the reference tileset.
```
j9liu: Can you make similar changes to the comment here?

j9liu (approved):
@david-lively I found a minor bug that I pushed a fix for -- a hard case to catch, but it was happening after you edited the Z-value in the Actor's `Location`; the Height above the `Tileset` wouldn't update. I ended up also doing some cleanup while I was in there, so those other commits are for that too.

I think this is good to go! Sadly GitHub actions is down, and I'd like to make sure the build works before merging, but there's nothing actionable on your part.

Source/CesiumRuntime/Private/CesiumGlobeAnchorComponent.cpp
```diff
@@ -667,7 +666,7 @@ void UCesiumGlobeAnchorComponent::OnRegister() {
 #endif
 
   USceneComponent* pOwnerRoot = pOwner->GetRootComponent();
-  if (pOwnerRoot && detectTransformChanges) {
+  if (pOwnerRoot && DetectTransformChanges) {
```
j9liu: Maybe you missed the `detectTransformChanges` above:

```
bool detectTransformChanges =
      this->DetectTransformChanges &&
      this->HeightReference != ECesiumHeightReference::Tileset;
```

You can change the naming, but I think this is necessary, because detecting when the transform changes is necessary for  `HeightReference == Tileset` to work at all.

Source/CesiumRuntime/Private/CesiumGlobeAnchorComponent.cpp
```diff
@@ -192,11 +192,10 @@ bool UCesiumGlobeAnchorComponent::GetDetectTransformChanges() const {
 }
 
 void UCesiumGlobeAnchorComponent::SetDetectTransformChanges(bool Value) {
-  this->DetectTransformChanges = Value;
-
-  if (this->HeightReference == ECesiumHeightReference::Tileset) {
+  if (Value == this->DetectTransformChanges) {
     return;
   }
+  this->DetectTransformChanges = Value;
```
j9liu, line 198: The reason I added this originally is because `DetectTransformChanges` would prevent any of the fields on the globe anchor from updating, and that's not good for the `HeightReference == Tileset` behavior. I'm not sure what the expected iteration would be otherwise.

Source/CesiumRuntime/Private/CesiumGlobeAnchorComponent.cpp
david-lively: With this change, the cube is not moved at all. Perhaps this should be 
```
this->HeightReference == ECesiumHeightReference::Tileset
```
 instead of 
```
this->HeightReference != ECesiumHeightReference::Tileset
```

PR #1799, opened 2026-02-18
Detect transform changes

### LINKED ISSUE #1231: Add option to disable "Detect Transform Changes" on CesiumGlobeAnchorComponent
In Cesium for Unity, there is an option on CesiumGlobeAnchor to disable "Detect Transform Changes". Disabling it is a pretty big performance win because it avoids the need to run a coroutine to poll for those transform changes.

In Unreal, detecting those changes is simply a matter of subscribing to `TransformUpdated` on the root component, which is much cheaper. Still, it's not free, so we should add a similar option to disable it in Cesium for Unreal.

timoore:
Following the example of Cesium for Unity, add a DetectTransformChanges property to CesiumGlobeAnchor. The default (true) is to track the motion of the owning actor; when false, the CesiumGlobeAnchor does not update.

Resolves #1231

### DIFF
--- a/Source/CesiumRuntime/Private/Tests/CesiumGlobeAnchor.spec.cpp
+++ b/Source/CesiumRuntime/Private/Tests/CesiumGlobeAnchor.spec.cpp
@@ -153,7 +153,7 @@ void FCesiumGlobeAnchorSpec::Define() {
     this->pActor->SetActorLocation(FVector(1000.0, 2000.0, 3000.0));
 
     this->pGlobeAnchor->Sync();
-    // Globe position doesn't update while unsubscribed
+    // Globe position has been updated even though GlobeAnchor is unsubscribed.
     TestNotEqual(
         "globe position when DetectTransformChanges is false before Sync()",
         this->pGlobeAnchor->GetLongitudeLatitudeHeight(),


Source/CesiumRuntime/Public/CesiumGlobeAnchorComponent.h
j9liu, line 1: Remember to update `CHANGES.md`!

Source/CesiumRuntime/Private/CesiumGlobeAnchorComponent.cpp
```diff
@@ -225,6 +248,9 @@ void UCesiumGlobeAnchorComponent::SnapToEastSouthUp() {
 void UCesiumGlobeAnchorComponent::Sync() {
   // If we don't have a actor -> ECEF matrix yet, we must update from the
   // actor's root transform.
+  if (!DetectTransformChanges) {
+    return;
+  }
```
j9liu: This prevents the first calibation that `CesiumGlobeAnchor` does updating when `ResolveGeoreference`, which is necessary for it to work in the first place. So we should remove this early return. _After_ `Sync()`, though, then no further changes should be picked up.

Cesium for Unity remarks on its own `Sync` function like so:

> If the game object's `Transform` has changed since the last time `OnEnable`, or if a position setter was called, this method updates all of this instance's position properties from the current transform. This works even if <see cref="detectTransformChanges"/> is disabled. It will also update the object's orientation if <see cref="adjustOrientationForGlobeWhenMoving"/> is enabled.

So we should have this work similarly -- have it work regardless of `DetectTransformChanges`, and mention so in the function's documentation.

Source/CesiumRuntime/Public/CesiumGlobeAnchorComponent.h
```diff
@@ -128,6 +128,19 @@ class CESIUMRUNTIME_API UCesiumGlobeAnchorComponent : public UActorComponent {
       Meta = (AllowPrivateAccess))
   FMatrix ActorToEarthCenteredEarthFixedMatrix;
 
+  /**
+   * Detect changes in the root component's transform, or not. This defaults to
+   * true, but may be useful to disable.
```
j9liu: Adapted from the Unity documentation:

```suggestion
   * Whether to automatically detect changes in the Actor's root transform and
   * update the precise globe coordinates accordingly.
   *
   * When this is property is false, the precise coordinates may become out of
   * sync with the Actor's root transform. However, you can still directly set
   * the precise coordinates, and its transform will update accordingly. You can
   * also call Sync after changing the Actor's transform  to manually update the
   * precise coordinates.
```

Source/CesiumRuntime/Private/Tests/CesiumGlobeAnchor.spec.cpp
```diff
@@ -121,6 +121,31 @@ void FCesiumGlobeAnchorSpec::Define() {
         beforeLLH);
   });
 
+  It("doesn't update anchor when DetectTransformChanges is false", [this]() {
```
j9liu, line 124: Can you also add a test for `Sync()` working even when `DetectTransformChanges` is false?

Source/CesiumRuntime/Public/CesiumGlobeAnchorComponent.h
```diff
@@ -128,6 +128,19 @@ class CESIUMRUNTIME_API UCesiumGlobeAnchorComponent : public UActorComponent {
...
+   */
+  UPROPERTY(
+      EditAnywhere,
+      BlueprintReadWrite,
+      BlueprintGetter = GetDetectTransformChanges,
+      BlueprintSetter = SetDetectTransformChanges,
+      Category = "Cesium",
+      Meta = (AllowPrivateAccess))
+  bool DetectTransformChanges = true;
```
j9liu: In order to match Cesium for Unity's UI + property ordering, could you move this to appear right after `AdjustOrientationForGlobeWhenMoving`?

Source/CesiumRuntime/Public/CesiumGlobeAnchorComponent.h
```diff
@@ -276,6 +289,27 @@ class CESIUMRUNTIME_API UCesiumGlobeAnchorComponent : public UActorComponent {
   UFUNCTION(BlueprintSetter, Category = "Cesium")
   void SetAdjustOrientationForGlobeWhenMoving(bool Value);
 
+  /**
+   * Gets a flag indicating whether to update the globe anchor when the owner's
+   * postition and transform change.
```
j9liu:
```suggestion
   * Gets a flag indicating whether to update the globe anchor when the Actor's
   * transform changes.
```

Source/CesiumRuntime/Public/CesiumGlobeAnchorComponent.h
```diff
@@ -276,6 +289,27 @@ class CESIUMRUNTIME_API UCesiumGlobeAnchorComponent : public UActorComponent {
...
+   *
+   * This property should usually be enabled, but it may be useful to disable it
+   * for performance reasons.
+   */
+  UFUNCTION(BlueprintGetter, Category = "Cesium")
+  bool GetDetectTransformChanges() const;
+
+  /**
+   * Sets a flag indicating whether to update the globe anchor when the owner's
+   * postition and transform change.
```
j9liu:
```suggestion
   * Sets a flag indicating whether to update the globe anchor when the Actor's
   * transform changes.
```

j9liu (changes requested):
Thanks @timoore ! This seems to be working well, just a matter of documentation + some behavior consistency with Unity.

CHANGES.md
```diff
@@ -8,9 +8,10 @@
 
 ##### Additions :tada:
 
+- Added support for loading tilesets with the `KHR_gaussian_splatting` extension.
 - Added support for styling with tileset statistics on `UCesiumFeaturesMetadataComponent`.
 - Added `UCesiumPropertyArrayBlueprintLibrary::ToString` to print the contents of a `FCesiumPropertyArray` as a string.
-- Added support for loading tilesets with the `KHR_gaussian_splatting` extension.
+- Added a DetectTransformChanges property on CesiumGlobeAnchor, which can be used to disable updating the globe anchor.
```
j9liu: Rewording for consistency with other changelog entries:

```suggestion
- Added `DetectTransformChanges` to `UCesiumGlobeAnchorComponent` to toggle whether the globe anchor should update itself after changes in its Actor's transform.
```

Source/CesiumRuntime/Public/CesiumGlobeAnchorComponent.h
```diff
@@ -276,6 +295,27 @@ class CESIUMRUNTIME_API UCesiumGlobeAnchorComponent : public UActorComponent {
   UFUNCTION(BlueprintSetter, Category = "Cesium")
   void SetAdjustOrientationForGlobeWhenMoving(bool Value);
 
+  /**
+   * Gets a flag indicating whether to update the globe anchor when the Actors's
```
j9liu:
```suggestion
   * Gets a flag indicating whether to update the globe anchor when the Actor's
```

Source/CesiumRuntime/Public/CesiumGlobeAnchorComponent.h
```diff
@@ -276,6 +295,27 @@ class CESIUMRUNTIME_API UCesiumGlobeAnchorComponent : public UActorComponent {
...
+   * transform changes.
+   *
+   * This property should usually be enabled, but it may be useful to disable it
+   * for performance reasons.
+   */
+  UFUNCTION(BlueprintGetter, Category = "Cesium")
+  bool GetDetectTransformChanges() const;
+
+  /**
+   * Sets a flag indicating whether to update the globe anchor when the Actors's
```
j9liu:
```suggestion
   * Sets a flag indicating whether to update the globe anchor when the Actor's
```

Source/CesiumRuntime/Private/Tests/CesiumGlobeAnchor.spec.cpp
```diff
@@ -121,6 +121,46 @@ void FCesiumGlobeAnchorSpec::Define() {
...
+    this->pActor->SetActorLocation(FVector(1000.0, 2000.0, 3000.0));
+
+    this->pGlobeAnchor->Sync();
+    // Globe position doesn't update while unsubscribed
+    TestNotEqual(
+        "globe position when DetectTransformChanges is false before Sync()",
+        this->pGlobeAnchor->GetLongitudeLatitudeHeight(),
+        beforeLLH);
+    this->pGlobeAnchor->SetDetectTransformChanges(true);
+  });
```
j9liu, line 162: Are there missing lines here? Seems like this should be the intended test:

```suggestion
    FVector beforeLLH = this->pGlobeAnchor->GetLongitudeLatitudeHeight();

    this->pGlobeAnchor->SetDetectTransformChanges(false);
    this->pActor->SetActorLocation(FVector(1000.0, 2000.0, 3000.0));

    // Globe position doesn't update while unsubscribed
    TestEqual(
        "globe position when DetectTransformChanges is false before Sync()",
        this->pGlobeAnchor->GetLongitudeLatitudeHeight(),
        beforeLLH);
        
    this->pGlobeAnchor->Sync();
    
    // Globe position is explicitly updated by Sync
    TestNotEqual(
        "globe position when DetectTransformChanges is false after Sync()",
        this->pGlobeAnchor->GetLongitudeLatitudeHeight(),
        beforeLLH);
  });
```

Source/CesiumRuntime/Public/CesiumGlobeAnchorComponent.h
```diff
@@ -91,6 +91,25 @@ class CESIUMRUNTIME_API UCesiumGlobeAnchorComponent : public UActorComponent {
...
+   * precise coordinates.
+   */
+  UPROPERTY(
+      EditAnywhere,
+      BlueprintReadWrite,
+      BlueprintGetter = GetDetectTransformChanges,
+      BlueprintSetter = SetDetectTransformChanges,
+      Category = "Cesium",
+      Meta = (AllowPrivateAccess))
+  bool DetectTransformChanges = true;
```
j9liu, line 111: I just realized that the reason this appears out-of-order on the Details panel is because we have our own customization 😱 

So in `CesiumGlobeAnchorCustomization.cpp`, could you look for:
```
 CesiumCategory.AddProperty(GET_MEMBER_NAME_CHECKED(
      UCesiumGlobeAnchorComponent,
      AdjustOrientationForGlobeWhenMoving));
```
And add the same thing but for `DetectTransformChanges`? Sorry I missed that!

j9liu (commented):
Thanks @timoore ! This is looking good, just a few minor comments. (I missed something earlier, sorry about that!) I think this can get merged soon after.

Source/CesiumRuntime/Public/CesiumGlobeAnchorComponent.h
timoore: That grammatical mistake is pervasive in CesiumGlobeAnchorComponent's comments. I probably cut-and-pasted it 🙄

Source/CesiumRuntime/Private/Tests/CesiumGlobeAnchor.spec.cpp
timoore: The test was intended to test whether or not Sync() causes an update when `DetectTransformChanges` is false. I think the previous case tests that update doesn't happen while unsubscribed?

j9liu: The comment (`Globe position doesn't update while unsubscribed`) doesn't match what the `TestNotEqual` is testing, hence the confusion. If it was a copy paste error, then the original test is fine but I would change the comment to be more informative.

PR #1802, opened 2026-02-25
Fix documentation of {y} and {reverseY}

timoore:
See https://github.com/CesiumGS/cesium-native/issues/1183


j9liu:
Thanks @timoore !

PR #1806, opened 2026-03-05
Add call to dispatchMainThreadTasks for new Cesium Native behavior

timoore:
This includes the change required to work with https://github.com/CesiumGS/cesium-native/pull/1315 in cesium-native.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,7 +9,6 @@
 - Added `UCesiumGlobeAnchorComponent::HeightUpdateInterval` to specify how often the actor's height is updated when `HeightReference` is `Tileset`.
 
 ### v2.24.1 - 2026-03-02
->>>>>>> main
 
 ##### Fixes :wrench:
 


CHANGES.md
```diff
@@ -1,5 +1,11 @@
 # Change Log {#changes}
 
+### ? - ?
+
+##### Fixes :wrench:
+
+- Added a call to `AsyncSystem::dispatchMainThreadTasks` to account for a change in Cesium Native, which no longer calls that function itself.
```
j9liu: This is a rare occasion where we don't need a changelog entry since this is a private change (as opposed to a public-facing one).

Source/CesiumRuntime/Private/Cesium3DTileset.cpp
```diff
@@ -2116,6 +2116,7 @@ void ACesium3DTileset::Tick(float DeltaTime) {
   }
 
   const Cesium3DTilesSelection::ViewUpdateResult* pResult;
+  getAsyncSystem().dispatchMainThreadTasks();
```
j9liu: So there's actually a call to `getAsyncSystem().dispatchMainThreadTasks();` earlier in this function, but it's gated by this if statement:

```c++
  if (this->_pFeaturesMetadataComponent.IsValid() &&
      this->_pFeaturesMetadataComponent->IsSyncing()) {
    // Styling may require the tileset's metadata to be loaded first (for schema
    // and/or statistics) before streaming tiles. But continue to dispatch tasks
    // so that the metadata future resolves.
    // This will not add much overhead since tilesets will have no metadata or
    // otherwise embed it in the tileset.json anyway.
    getAsyncSystem().dispatchMainThreadTasks();
    return;
  }
```

It would be better to move this call _before_ the if statement, then have the if return early. Then this second call is no longer needed.

CHANGES.md
timoore: My thought was that cesium-unreal is probably the main example of how to write a cesium-native integration, and for that reason we should highlight breaking changes caused by cesium-native. Thoughts?

j9liu: We have not done this in the past. For example, cesium-native 0.47.0 had immensely breaking changes as noted in the [changelog](https://github.com/CesiumGS/cesium-native/blob/main/CHANGES.md#v0470---2025-05-01), but those were never mentioned in the Unreal [changelog](https://github.com/CesiumGS/cesium-unreal/blob/main/CHANGES.md#v2161---2025-06-02) that included that version. Instead we call out the change in cesium-native version at the bottom of the changelog.

Please update this accordingly.

PR #1822, opened 2026-04-13
Refactor Gaussian splats for improved memory performance

j9liu:
## Description

This PR refactors our Gaussian splats system to fix various issues with memory and visual artifacts as LODs changed.

I'll post a more comprehensive overview of the changes after I've squashed one more issue.

## Issue number or link

Closes #1814.

## Author checklist

  - There may be a way to test subsystems/rendering, but none currently set up. 
  - No outward facing docs, but added internal docs as needed.

## Remaining Tasks

- [X] Figure out why additional sprites are spawning in play-in-editor mode (which inevitably triggers the `groupCount` condition that happens after too many sprites are spawned).
  - This appears to be general behavior with `ACesium3DTileset` - should definitely look to mitigate that, but will not be in scope for this PR.
- [x] General cleanup

## Testing plan

This is best tested with the **Microsoft Campus** dataset (see this [Sandcastle](https://sandcastle.cesium.com/?id=3d-tiles-gaussian-splats-with-lod)), which contains multiple LODs.

You can view memory stats by going to this menu in the viewport:

Before this PR, navigating around the scene would accumulate a giant number of sprites. After this PR, the number of sprites should maintain a reasonable number. You can see this behavior more aggressively if you set the tileset's **Maximum Cached Bytes** to 0, which will unload/load tiles rather than caching them when they go invisible.

## Reviewer checklist


j9liu:
This PR is ready for review; here's a summary of changes:

## Niagara System Asset Changes

Since it's impossible to view the diff on GitHub, I'll include images of the changes here.

### Loop behavior

Previously, the Niagara System was set to loop once and always spawn `NumSplats` particles. Now, the Niagara System loops infinitely with an interval of 0.5, dynamically spawning/killing particles based on the changed number.

image

This removes the need for `RESET_FRAME_COUNT` and calling `ResetSystem()` in `CesiumGaussianSplatSubsystem::Tick()`.

### Spawn/Kill modules

The full emitter stack looks like so:

image

Where `ParticlesToSpawn` = `SplatCount - NumParticles`.

This results in particles only spawning when necessarily. The Kill Particles stage ensures extraneous particles are killed as `SplatCount` changes.

For this to work, particles must use the built-in `Execution Index` rather than Unique ID (which continuously increments as particles spawn).

## Niagara Data Interface/Proxy Refactor

`UCesiumGaussianSplatDataInterface` was refactored to model the behavior of other data interfaces. Any renames here are an intentional result of this to match Niagara's other patterns. (See `UNiagaraDataInterfaceSpline` for example).

Most notably, the interface now provides a `PerInstanceTick` function that runs on the game thread. This tick is responsible for queueing the render commands to the GPU when necessary. This lets us take advantage of `FRenderCommandFence`, which can only be created from the game thread.

Additionally, `FScopeLock` was unfortunately not always preventing the subsystem's components array from being changed mid-execution. During development, I was still triggering breakpoints (and sometimes crashes!) where the array would change mid-iteration. The safest workaround was to create a copy of the array pointers, and we have to iterate over anyways to ensure we only pick up components belonging to the same `UWorld`.

## Subsystem Tick Changes

The biggest change here is in how the subsystem interacts with the data interface. It:

- Tracks whether or not the interface is marked as dirty.
- If so, it pauses the Niagara component to ensure that the simulation doesn't continue running on outdated data. This otherwise would cause unpleasant LOD pop-ins as the data was rewritten under the hood.
- Then, it waits for the `FRenderCommandFence`s are complete. The fences are used to determine when the system should be unpaused.

Most other changes in `CesiumGaussianSplatSubsystem.cpp` are organizational.

Source/CesiumRuntime/Private/CesiumGaussianSplatDataInterface.cpp
```diff
@@ -61,29 +62,32 @@ void UploadTileTransforms(
...
 
-  Buffer.Initialize(
-      RHICmdList,
-      TEXT("FNDIGaussianSplatProxy_TileTransformsBuffer"),
-      sizeof(FVector4f),
-      NumComponents * VectorsPerComponent,
-      EPixelFormat::PF_A32B32G32R32F,
-      BUF_Static);
+  const int32 totalVectorCount = Components.Num() * VectorsPerComponent;
+  const uint32 requiredByteLength = totalVectorCount * sizeof(FVector4f);
```
azrogers, line 66: Are we using `UpperCamelCase` or `lowerCamelCase` for local variables? The Unreal style is UpperCamelCase, the Cesium Native style is lowerCamelCase. If going with lowerCamelCase, change `VectorsPerComponent` to match.

Source/CesiumRuntime/Private/CesiumGaussianSplatDataInterface.cpp
```diff
@@ -147,260 +151,218 @@ void UploadTileTransforms(
...
-        RHICmdList.UnlockBuffer(this->OrientationsBuffer.Buffer);
-        RHICmdList.UnlockBuffer(this->ColorsBuffer.Buffer);
-        if (TotalCoeffsCount > 0) {
-          RHICmdList.UnlockBuffer(this->SHNonZeroCoeffsBuffer.Buffer);
-        }
-        RHICmdList.UnlockBuffer(this->TileIndicesBuffer.Buffer);
-        RHICmdList.UnlockBuffer(this->SplatSHDegreesBuffer.Buffer);
-      }
+  int32 CoeffCountWritten = 0;
+  int32 SplatCountWritten = 0;
```
azrogers: More UpperCamelCase local variables. Just check to make sure we're consistent.

Source/CesiumRuntime/Private/CesiumGaussianSplatDataInterface.cpp
```diff
@@ -585,29 +545,149 @@ void UCesiumGaussianSplatDataInterface::PostInitProperties() {
...
+
+void UCesiumGaussianSplatDataInterface::DestroyPerInstanceData(
+    void* PerInstanceData,
+    FNiagaraSystemInstance* SystemInstance) {
+  check(SystemInstance);
+  this->_worldToProxyData.Remove(SystemInstance->GetWorld());
+
+  FNDICesiumGaussianSplats_InstanceData* pInstData =
+      static_cast<FNDICesiumGaussianSplats_InstanceData*>(PerInstanceData);
+  pInstData->~FNDICesiumGaussianSplats_InstanceData();
```
azrogers, line 570: It doesn't feel like we're doing things right if we're manually calling the destructor. Maybe I'm missing something here, or maybe the intended Unreal behavior really is that weird, but presumably the per-instance data gets `delete`d at some point and that would call the destructor?

azrogers:
Great set of changes here! Not much on the review from me. I do still see the `GroupCount` assertion show up on testing with the Microsoft tileset, but the tileset still seems to work anyways and considering the size of the dataset I'm not too concerned about it. There's a good amount of pop-in still but presumably this is what CesiumGS/cesium-native#1345 is trying to fix.

Source/CesiumRuntime/Private/CesiumGaussianSplatDataInterface.cpp
j9liu: We're technically supposed to use `lowerCamelCase` for private code, I was just trying to avoid bloating the diff with a big style sweep.

Source/CesiumRuntime/Private/CesiumGaussianSplatDataInterface.cpp
j9liu: Hmm I can't claim I fully understand the lifetime here 😅 I just decided to closely follow what Niagara seems to do for its other classes. This is what it looks like in `UNiagaraDataInterfaceSpline`:

```c++
void UNiagaraDataInterfaceSpline::DestroyPerInstanceData(void* PerInstanceData, FNiagaraSystemInstance* SystemInstance)
{
	SystemInstancesToProxyData_GT.Remove(SystemInstance->GetId());
	
	FNDISpline_InstanceData* InstData = static_cast<FNDISpline_InstanceData*>(PerInstanceData);
	InstData->~FNDISpline_InstanceData();

	FNiagaraDataInterfaceProxySpline* RT_Proxy = GetProxyAs<FNiagaraDataInterfaceProxySpline>();
	ENQUEUE_RENDER_COMMAND(FNiagaraDIDestroyInstanceData) (
		[RT_Proxy, InstanceID=SystemInstance->GetId()](FRHICommandListImmediate& CmdList)
		{
#if STATS
			if (FNDISpline_InstanceData_RenderThread* TargetData = RT_Proxy->SystemInstancesToProxyData_RT.Find(InstanceID))
			{
				TargetData->Reset();
			}
#endif
			RT_Proxy->SystemInstancesToProxyData_RT.Remove(InstanceID);
		}
	);
}
```

j9liu:
Thanks for the review @azrogers! I addressed the style inconsistencies you pointed out, let me know if I missed anything else!

Source/CesiumRuntime/Private/CesiumGaussianSplatDataInterface.cpp
azrogers: If Epic's doing it that's good enough for me, even if it's weird!

azrogers:
Thanks @j9liu!

PR #1824, opened 2026-04-16
Add `UnloadEditorTilesInPlayMode` setting

j9liu:
## Description

This PR addresses an issue noted in #1822 re: degraded performance in Play-in-Editor (PIE) mode due to splats being doubly-loaded. Although this PIE behavior is expected, it's undesired when dealing with very intensive datasets.

The `UnloadEditorTilesInPlayMode` setting provides a workaround for this, causing the Editor instance of the tileset to unload itself before PIE starts. Tiles will reload in Editor after PIE ends. Since this could potentially result in consuming more sessions (I think?), I have this setting turned off by default.

## Issue number or link

N/A

## Author checklist

  - Unclear how to set up a unit test that switches between Editor and PIE mode. 

## Testing plan

This could be testable with any tileset, but you can measure the impact most prominently with the [Microsoft Campus](https://sandcastle.cesium.com/?id=3d-tiles-gaussian-splats-with-lod) Gaussian splat dataset. Use the Niagara stat panel to compare the number of sprites with and without the setting.

image

## Reviewer checklist


Source/CesiumRuntime/Private/Cesium3DTileset.cpp
```diff
@@ -2436,6 +2447,12 @@ void ACesium3DTileset::RuntimeSettingsChanged(
     this->DestroyTileset();
   }
 }
+
+void ACesium3DTileset::OnPreBeginPIE(bool bIsSimulating) {
```
david-lively, line 2451: Doesn't really matter in this implementation, but I believe the parameter should be `const`? Unreal will figure it out regardless.

Source/CesiumRuntime/Public/Cesium3DTileset.h
```diff
@@ -1400,6 +1414,8 @@ class CESIUMRUNTIME_API ACesium3DTileset : public AActor {
   void RuntimeSettingsChanged(
       UObject* pObject,
       struct FPropertyChangedEvent& changed);
+
+  void OnPreBeginPIE(bool bIsSimulating);
```
david-lively, line 1418: See `const` review note on the implementation.

david-lively (commented):
I have a few notes, but looks good overall.

Source/CesiumRuntime/Private/Cesium3DTileset.cpp
j9liu: I used to be proactive about adding `const` myself, but our [C++ style guidelines](https://cesium.com/learn/cesium-native/ref-doc/style-guide.html#const) suggest to avoid `const` for by-value parameters. Is that fair? 😅

PR #1835, opened 2026-05-11
Prevent physics log spam for line/point meshes

j9liu:
## Description

Working with scale in Unreal has continually resulted in headaches due to the aggressive epsilon checks in code. The loudest complaints are from Chaos (the physics engine), which we previously merged #1668 for.

However, we were still coming across log spam due to some models with small glTF line primitives:

image

It turns out the workaround in `UCesiumGltfPrimitiveComponent::OnCreatePhysicsState` is not sufficient for line primitives, because `BodyInstance.InitBody` does not result in a valid body instance even with a scale of 1. You can confirm this if you add and put a breakpoint on this `return`:

```c++
  if (!BodyInstance.IsValidBodyInstance() &&
      BodyTransform.GetScale3D().IsNearlyZero()) {
    FTransform BodyTransformWithoutScale = BodyTransform;
    BodyTransformWithoutScale.SetScale3D(FVector(1.0));

    UBodySetup* BodySetup = GetBodySetup();
    BodyInstance.InitBody(
        BodySetup,
        BodyTransformWithoutScale,
        this,
        GetWorld()->GetPhysicsScene());
    if(!BodyInstance.IsValidBodyInstance())
        return;
    BodyInstance.UpdateBodyScale(BodyTransform.GetScale3D());
  }
```

So when `Super::OnCreatePhysicsState` is called, it seems the `BodyInstance` is not valid, so it will try to `BodyInstance.InitBody` with the original small scale. That results in the "Scale3D is (nearly) zero" warning spam.

The most elegant solution seemed to be to avoid this call chain entirely. But we need to set the `bPhysicsStateCreated` variable from `UActorComponent` to true, otherwise a `check` somewhere else in the code complain.

I also added this check to our points component, since I imagine something similar could happen later down the line.

## Issue number or link

N/A

## Author checklist

  - We DO have unit tests for physics, but I'm not sure how to test whether or not a log was printed to the console.

## Testing plan

Reach out to me for a testing dataset.

1. Open `06_CesiumMetadata` in the Cesium Samples.
2. Add a tileset with many small line components to the level (best examples are from AEC).
3. Reposition the CesiumGeoreference origin and the DynamicPawn to view the new tileset.
4. Press Play.

The physics interactions should still work with everything else, and merely ignore the lines.

## Reviewer checklist


david-lively (approved):
LGTM.

PR #1837, opened 2026-05-13
Add support for `EXT_mesh_primitive_edge_visibility`

j9liu:
## Description

~Depends on #1836, so merge that first.~

This PR supports rendering edges from the [`EXT_mesh_primitive_edge_visibility`](https://github.com/CesiumGS/glTF/tree/pmc/edge-visibility/extensions/2.0/Vendor/EXT_mesh_primitive_edge_visibility) glTF extension, which is often used to enhance the appearance of BIM/CAD models, shown below:

| Without Edges | With |
| - | - |
|HighresScreenshot00000 | HighresScreenshot00003 |
| HighresScreenshot00001 | HighresScreenshot00004 |
| HighresScreenshot00002 | HighresScreenshot00005 |

## Issue number or link

Closes #1809.

## Author checklist

- [X] I have submitted a [Contributor License Agreement](https://github.com/CesiumGS/community/tree/main/CLAs) (only needed once).
- [x] I have done a full self-review of my code.
- [X] I have updated `CHANGES.md` with a short summary of my change (for user-facing changes).
- ~[ ] I have added or updated unit tests to ensure consistent code coverage as necessary.~
- ~[ ] I have updated the documentation as necessary.~

## Remaining Tasks

- [x] Parse the edge material to color edges correctly.
- [X] Support `lineString` + material overrides - #1842 
- ~[ ] Investigate if multi-pass approach mitigates z-fighting artifacts:~
image

Aside from these cases where a bunch of tiny edges are bunched together, they look pretty good as is, so we'll hold off on the multi-pass approach for now.

## Testing plan

Reach out to me for the pictured test dataset! It should be as simple as loading any other tileset into Unreal, just confirm that the edges show up.

## Reviewer checklist

Thank you for taking the time to review this PR. By approving a PR you are taking as much responsibility for these changes as the author.

As you review, please go through the checklist below:

- [ ] Review and run all parts of the test plan on this branch and verify it matches expectations.
    - [ ] If the issue is a bug please make sure you can reproduce the bug in the main branch and then checkout this branch to make sure it actually solved the issue.
- [ ] Review the code and make sure you do not have any remaining questions or concerns. You should understand the code change and the chosen approach. If you are not confident or have doubts about the code, please do not hesitate to ask questions.
    - [ ] Code should follow the [C++ Style Guide](https://cesium.com/learn/cesium-native/ref-doc/style-guide.html)
- [ ] Review the unit tests and make sure there are no missing tests or edge cases.
[...]

PR #1842, opened 2026-05-22
Support line strings from `EXT_mesh_primitive_edge_visibility`

j9liu:
## Description

Merges into #1837. This PR adds support for the `lineStrings` property in the `EXT_mesh_primitive_edge_visibility` extension, as well as the material overrides that each line string may have.

Instead of creating a separate line component / material for each line string, I've decided to batch them together for performance reasons, letting the material determine the vertex colors. This is similar to the implementation in https://github.com/CesiumGS/cesium/pull/13002.

## Issue number or link

N/A

## Author checklist

- ~[ ] I have updated `CHANGES.md` with a short summary of my change (for user-facing changes).~
  - This has already been noted in #1837 

## Testing plan

Test tileset:
[edge3.zip](https://github.com/user-attachments/files/28162748/edge3.zip)

The edges won't show up on #1837. With this PR, the two triangles will render edges with one triangle's being green and the other blue.

Screenshot 2026-05-22 163324

## Reviewer checklist


Source/CesiumRuntime/Private/CesiumGltfPrimitiveEdges.cpp
```diff
@@ -37,57 +84,181 @@ CesiumGltfPrimitiveEdges::createInWorkerThread(
...
+  if (extension.visibility >= 0) {
+    CesiumGltf::AccessorView<uint8_t> visibilityView(
         model,
-        positionView,
-        indicesView,
-        extension);
+        extension.visibility);
+
+    // Remake the index accessor view here to avoid templates in the .h file.
+    switch (pIndexAccessor->componentType) {
```
timoore, line 96: I don't really understand the comment and actually think that createAccessorView(), using a template, would be cleaner here. You might think about revising at least the comment at a later date.

j9liu: This comment was present in the #1837, it's not specific to this PR.

We construct an accessor view for the indices in `CesiumGltfComponent.cpp`, so it would be nice to just pass that in as a parameter to `createOnWorkerThread`. But if we define the function as templated in `CesiumGltfPrimitiveEdges.h`, all of the implementation would _also_ have to be in `CesiumGltfPrimitiveEdges.h`. This is just because of how templated functions are resolved; I've run into similar issues working with the metadata stuff as well. I do like when templates make things cleaner, but it would bloat the `.h` file here, and I want to avoid that where possible.

timoore: As far as I can see, `indicesView` is only passed in to `extractVisibleEdges`, which is already a template function.  I think we're talking at cross purposes here 😄 I now understand that the comment is explaining why `indicesView` is not a parameter to `createInWorkerThread`. My remark about `createAccessorView` was directed at this chunk of code, where I think it would be neater than the switch statement.

j9liu: I see what you mean now! Yeah that would probably be an improvement. I copied this unthinkingly from `CesiumGltfComponent.cpp`, and it was probably written before `createAccessorView` existed.

PR #1845, opened 2026-05-26
Support for `CesiumVectorOverlays::VectorTilesRasterOverlay`

azrogers:
Depends on CesiumGS/cesium-native#1365. This PR adds support for the new `VectorTilesRasterOverlay` to allow vector 3D Tiles tilesets to be draped over other tilesets. It also in the process adds support for displaying points from GeoJSON documents. Because the vector tiles PR moved the GeoJSON overlay to the new `CesiumVectorOverlays` namespace, this PR also applies that change here.


Source/CesiumRuntime/Public/CesiumVectorStyle.h
```diff
@@ -135,6 +135,49 @@ struct FCesiumVectorPolygonStyle {
   FCesiumVectorLineStyle OutlineStyle;
 };
 
+USTRUCT(BlueprintType)
+struct FCesiumVectorPointStyle {
+  GENERATED_BODY()
+
+  /**
+   * The radius of the point.
```
j9liu:
```suggestion
   * The radius of the point in pixels.
```

Source/CesiumRuntime/Private/CesiumVectorTilesRasterOverlay.cpp
```diff
@@ -0,0 +1,50 @@
...
+        std::string(TCHAR_TO_UTF8(*this->CesiumIonServer->ApiUrl)) + "/",
+        vectorOptions,
+        options);
+  }
+
+  return std::make_unique<CesiumVectorOverlays::VectorTilesRasterOverlay>(
+      TCHAR_TO_UTF8(*this->MaterialLayerKey),
+      TCHAR_TO_UTF8(*this->Url),
+      vectorOptions,
+      options);
```
j9liu, line 61: Could we add early bails if `IonAssetID = 0` or `Url.IsEmpty()`, similar to other overlays? Just so it doesn't attempt to create this if nothing is specified.

Source/CesiumRuntime/Private/CesiumVectorStyle.cpp
```diff
@@ -79,6 +92,30 @@ FCesiumVectorStyle::fromNative(const CesiumVectorData::VectorStyle& style) {
...
   }
 
+  FCesiumVectorLineStyle PointOutlineStyle;
+  if (style.point.outline) {
+    PolygonOutlineStyle = FCesiumVectorLineStyle{
+        FColor(
+            (uint8)style.point.outline->color.r,
+            (uint8)style.point.outline->color.g,
+            (uint8)style.point.outline->color.b,
+            (uint8)style.point.outline->color.a),
```
j9liu: Technically our [style guide](https://cesium.com/learn/cesium-native/ref-doc/style-guide.html#integers) says to use constructor notation for safe integer conversions... I see that the parentheses were already a pattern in this file, but if it's not too much trouble to change, that'd be nice.

j9liu (commented):
Thanks @azrogers for the PR! I still have to finish the cesium-native PR review and run some test data through this, but just wanted to leave the few small comments I had so far.

Source/CesiumRuntime/Public/CesiumVectorTilesRasterOverlay.h
```diff
@@ -0,0 +1,119 @@
...
+   * The ID of the Cesium ion asset to use.
+   */
+  UPROPERTY(
+      EditAnywhere,
+      BlueprintReadWrite,
+      Category = "Cesium",
+      meta =
+          (EditCondition =
+               "Source == ECesiumVectorTilesRasterOverlaySource::FromCesiumIon"))
+  int64 IonAssetID;
```
j9liu, line 53: This is missing an ion access token parameter as well

j9liu:
Something else I found... the Unreal Editor unfortunately gets stuck on quit when a vector overlay is in the scene because `IsReadyForFinishDestroy` never resolves to true. It seems like the async destruction event is never called, maybe something is stuck or not properly destroyed?

azrogers:
@j9liu Looks like the problem was I set the destructor of `VectorTilesRasterOverlay` to `default` when it's the destructor that resolves the promise... should be fixed now. Also, had to fix the CI that broke due to the linux image changing.

j9liu (approved):
Thanks @azrogers!

I'm just doing a fresh build to verify that the destructor is working properly, but if all goes well, this should be good to merge.

PR #1849, opened 2026-06-02
Fix Gaussian splats not rendering in Standalone game

UmarMukhtar0039:
## Description

Gaussian splats never renders in the Standalone game session. There are two main causes:

- `GetPrimaryWorld()` lacked proper handling for picking the right world context for all play modes (PIE, Standalone etc.) it only checked `GEditor`, which is null at runtime in a game session. This fix ensures that the code gracefully falls through to `GEngine` and prefers the Game mode world context.

- The splat actor was spawned with `bTemporaryEditorActor`, making it an editor-preview actor. This supresses `BeginPlay` in a game world, so the attached Niagara system never initialized its per instance data. It's now only marked as a preview actor in a non-game (editor) worlds.

## Issue number or link

N/A

## Author checklist

- ~~[ ] I have added or updated unit tests to ensure consistent code coverage as necessary.~~
- ~~[ ] I have updated the documentation as necessary.~~

## Testing plan

1. Launch an Unreal project with Gaussian splats enabled (preferrably cesium-unreal-samples).
2. Open the map with Gaussian splats
3. Launch in Standalone mode - splats now should render. 

## Reviewer checklist

### DIFF
--- a/Source/CesiumRuntime/Private/CesiumGaussianSplatSubsystem.cpp
+++ b/Source/CesiumRuntime/Private/CesiumGaussianSplatSubsystem.cpp
@@ -76,9 +76,9 @@ UWorld* GetPrimaryWorld() {
 #if WITH_EDITOR
   if (IsValid(GEditor)) {
     Contexts = &GEditor->GetWorldContexts();
-  } else
+  }
 #endif
-      if (IsValid(GEngine)) {
+  if (Contexts == nullptr && IsValid(GEngine)) {
     Contexts = &GEngine->GetWorldContexts();
   }
 
@@ -174,9 +174,9 @@ void UCesiumGaussianSplatSubsystem::initializeForWorld(UWorld& InWorld) {
 #if WITH_EDITOR
   // Only mark the splat system actor as a temp editor-preview actor when
   // spawning into a non-game worlds. In a game world bTemporaryEditorActor sets
-  // bIsEditorPreviewActor, which suppresses BeginPlay and the Niagrara component
-  // never runs its per-instance data lifecycle hence the splat data interface
-  // is never initialized for the world, therefore no splats render.
+  // bIsEditorPreviewActor, which suppresses BeginPlay and the Niagrara
+  // component never runs its per-instance data lifecycle hence the splat data
+  // interface is never initialized for the world, therefore no splats render.
   if (!InWorld.IsGameWorld()) {
     ActorParams.bTemporaryEditorActor = true;
     // Set this to false if you need the actor to be visible in the outliner to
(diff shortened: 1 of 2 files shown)


Source/CesiumRuntime/Private/CesiumGaussianSplatSubsystem.cpp
```diff
@@ -65,31 +65,40 @@ FBox CalculateBounds(
...
-  }
-
-  const TIndirectArray<FWorldContext>& Contexts = GEditor->GetWorldContexts();
-#else
-  if (!IsValid(GEngine)) {
-    return nullptr;
-  }
-
-  const TIndirectArray<FWorldContext>& Contexts = GEngine->GetWorldContexts();
+  if (IsValid(GEditor)) {
```
azrogers, line 77: It seems like the change is that, in the editor, if `GEditor` is invalid `GEngine` will be used instead if possible, whereas before in editor `GEngine` was never checked. Is there ever a time where this will be called when `WITH_EDITOR` is defined, `GEngine` is valid, but `GEditor` is not?

Anyways, I'm fine with leaving in the fallback to `GEngine` just in case, but I'm a little wary of the dangling `else` here. Instead I think it would be better to remove the `else` and add a `Contexts == nullptr &&` clause to the `if (IsValid(GEngine))` statement.

azrogers:
@UmarMukhtar0039 Thanks for the fix! Could you fix the issues on CI? At the moment it's just formatting, an `npm install` and `npm run format` needs to be done, but if other jobs come back red those should be fixed too. Besides that it's just the one comment and suggested edit above.

Source/CesiumRuntime/Private/CesiumGaussianSplatSubsystem.cpp
UmarMukhtar0039: Yes, that's exactly the case this fixes. "Is there ever a time where this will be called" - A standalone game session launched from Editor runs as a -game child process, where `GEditor` is null at runtime but `GEngine` is valid. And without this `GetPrimaryWorld()` returns nullptr and splats system never recieves the world.

And you are right about the  dangling `else` sorry for an oversight. I'll make the fix in a little while.

UmarMukhtar0039:
@azrogers this is ready

azrogers:
@UmarMukhtar0039 Thank you! I'll just wait for CI to complete again and if it all looks good I'll get it merged in.

UmarMukhtar0039:
> @UmarMukhtar0039 Thank you! I'll just wait for CI to complete again and if it all looks good I'll get it merged in.

Happy to help

PR #1874, opened 2026-07-23
Support billboarding via the `KHR_billboard` extension

j9liu:
## Description

Credit to @david-lively for most of this code.

This PR adds support for billboarding meshes with the `KHR_billboard` extension.

Note that this is really simple support, and we don't properly handle `rotationAxis`, `overlay`, etc. properties. The code to make billboarding work with an arbitrary up-axis was non-trivial, so the `rotationAxis` stuff could sadly become a potential nightmare.

If need be, these can be handled in a future PR.

Here's the show of the material function code:

image

Notably this is toggled with an `if`. Typically `if`s aren't ideal, but all the points of the same mesh will go down the same branch, so this shouldn't be too costly.

## Issue number or link

N/A

## Author checklist

## Testing plan

Reach out to me for a sample tileset featuring the billboard data we've used for testing.

I was hoping to use the [BillboardDemo sample](https://github.com/KhronosGroup/glTF-External-Reference/blob/main/demo/Models/BillboardDemo/glTF/BillboardDemo.gltf) to showcase this, but turns out that the model uses the `rotationAxis` property we have intentionally omitted as part of this PR. So the billboards are not rotating the intended way.

| Good | Bad |
| - | - |
| image | image |

## Reviewer checklist


Source/CesiumRuntime/Private/CesiumGltfComponent.cpp
```diff
@@ -3139,6 +3147,19 @@ PRAGMA_ENABLE_DEPRECATION_WARNINGS
...
+    UMaterialInstanceDynamic* pMaterial,
+    const std::optional<CesiumGltf::ExtensionKhrBillboard>& billboard,
+    const EMaterialParameterAssociation association,
+    const int32 index) {
+
+  pMaterial->SetScalarParameterValueByInfo(
+      FMaterialParameterInfo(FName("isBillboard"), association, index),
+      billboard ? 1.0 : 0.0);
+
+  // Future TODO: support other properties in the KHR_billboard extension.
```
j9liu, line 3159: There are additional properties in `KHR_billboard` ([PR for reference](https://github.com/KhronosGroup/glTF/pull/2522/changes)) that are potentially non-trivial to implement given that we have to work with a varying up-axis across the world. @david-lively put a lot of effort into getting this to work for that already. I think deferring this to a later PR if there is community need / interest is the way to go.

azrogers:
That if statement could be turned into a multiplication, but to be honest I don't know enough about GPU scheduling to say whether that would be faster. It definitely seems like performance isn't a problem right now anyways so nothing to worry about. Thanks @j9liu!

PR #1876, opened 2026-07-30
Allow `VectorTilesRasterOverlay` elements to be styled from Blueprints or C++

azrogers:
This PR builds on the `VectorStylingProvider` introduced in CesiumGS/cesium-native#1430 to allow elements loaded from the vector tiles tileset to be styled using either a specified Blueprint class or from C++.

Because of the peculiarities of how Blueprint works, all the work done by Blueprint has to happen on the main thread. This is a significant hit to performance with large datasets, so the choice was made to allow the user to also specify a `VectorStylingProvider` directly. It would be nice to work this in to the Unreal class system, but because UObjects are not thread-safe, the easiest way I could think of to make this work was to just have the user specify a `TFunction` (the Unreal standard library equivalent to a `std::function`) that returns a `shared_ptr` to a `VectorStylingProvider`.


Source/CesiumRuntime/Public/CesiumVectorTilesRasterOverlay.h
```diff
@@ -122,10 +213,53 @@ class CESIUMRUNTIME_API UCesiumVectorTilesRasterOverlay
...
+      TFunction<std::shared_ptr<CesiumVectorOverlays::VectorStylingProvider>()>>
+      LambdaStylingProvider;
+
 protected:
   virtual std::unique_ptr<CesiumRasterOverlays::RasterOverlay> CreateOverlay(
       const CesiumRasterOverlays::RasterOverlayOptions& options = {}) override;
+
+private:
+  UPROPERTY(Transient)
+  UObject* _pStylingInterfaceObject;
```
j9liu, line 276: Should this be an Unreal smart pointer (e.g., `TUniquePtr`)? Are we confident that as a raw pointer it will be properly garbage collected?

Source/CesiumRuntime/Private/CesiumVectorTilesRasterOverlay.cpp
```diff
@@ -7,13 +7,237 @@
...
+
+      --this->_pendingStylingJobs;
+      return result;
+    };
+
+    return asyncSystem.runInMainThread(lambda);
+  }
+
+  CesiumAsync::Future<std::vector<std::optional<CesiumVectorData::VectorStyle>>>
+  onStylePolylines(
```
j9liu, line 112: Just copying what I shared in DM...

I'm triggering a ton of assert fails in the generated file (`CesiumVectorTilesRasterOverlay.gen.cpp`) when I first create a new Blueprint styling provider:

Image

I believe it's because I'm using "Create New Blueprint", which is creating an empty class and immediately assigning it to the overlay:

Image

Maybe we need to add a check for `ImplementsInterface(UCesiumVectorTilesStylingCallbacks::StaticClass())` and prevent the task from being queued if this is not true.

Source/CesiumRuntime/Public/CesiumVectorTilesRasterOverlay.h
```diff
@@ -2,15 +2,106 @@
...
+/**
+ * Determines the styling provider to be used to provide styling information for
+ * individual features.
+ */
+UENUM()
+enum class ECesiumVectorStylingProviderType : uint8 {
+  None = 0 UMETA(ToolTip = "Only the default style will be used."),
+  Blueprint = 1 UMETA(
+      ToolTip =
+          "Uses the Blueprint class implementing the ICesumVectorTilesStylingCallbacks interface specified in BlueprintStylingProvider."),
```
j9liu:
```suggestion
          "Uses the Blueprint class implementing the ICesiumVectorTilesStylingCallbacks interface specified in BlueprintStylingProvider."),
```

Source/CesiumRuntime/Public/CesiumVectorTilesRasterOverlay.h
j9liu, line 1: Remember to update CHANGES.md!

j9liu (commented):
Thanks for the PR @azrogers! This is looking pretty good -- aside from the issue I found with "Create New Blueprint", the Blueprint styling seems to work well.

Just a few small comments otherwise!

Source/CesiumRuntime/Public/CesiumVectorTilesRasterOverlay.h
azrogers: All `UObject`s are garbage collected and cannot be used with smart pointers.

azrogers:
@j9liu *Hopefully* CI succeeds this time around, and if it does, this and the Cesium Native PR should be ready to go.

j9liu (approved):
Looks great, thanks @azrogers !

PR #1877, opened 2026-07-31
Revert "Support billboarding via the `KHR_billboard` extension"

j9liu:
Reverts CesiumGS/cesium-unreal#1874.

Unfortunately, the changes to `CesiumGltfFunction` results in significant artifacts when used with Temporal Anti-Aliasing / Super Resolution. This is because the way "World Position Offset" works in Unreal does not interact well with the TAA velocity buffer.

[Helpful excerpt from Unreal tutorial](https://dev.epicgames.com/documentation/unreal-engine/temporal-super-resolution-in-unreal-engine?lang=en-US#animationworldpositionoffset):

> Materials with parameters that are animated present a problem for TSR and other temporal accumulations of the renderer, such as a material using World Position Offset (WPO) with [vertex animation](https://dev.epicgames.com/documentation/unreal-engine/vertex-animation-tool-in-unreal-engine). For these types of materials, the WPO needs to evaluate the current and previous frame in order to retrieve proper motion vectors. Because the previous frame evaluation of WPO logic only has the current frame's value, the previous frame result ends up being incorrect. For materials that draw velocity, it's important that they specify the value of what the previous frame was.

This is most egregiously noticeable on point cloud meshes (try viewing `11_CesiumPointClouds`). Part of it is an outdated check on `FCesiumGltfPointsSceneProxy` when we set `bVelocityRelevance` in `GetViewRelevance`. However, the artifacts still occur even with that check fixed. They don't appear for non-temporal AA methods.

We previously have never run into this b/c as long as WPO is unused in the material function, the relevant parts of the shader are excluded. Once it is connected to a potentially non-zero value, it will enable a flag: `USES_WORLD_POSITION_OFFSET` and potentially `ALWAYS_EVALUATE_WORLD_POSITION_OFFSET`, in `LocalVertexFactory.ush`. This makes the artifacts appear even if the returned WPO value is 0.

There is a `PreviousFrameSwitch` node that can supposedly be used to mitigate this effect, but I don't understand how we can use it in our context. We're not even applying a non-zero WPO to the mesh unless it is billboarded. Warrants some investigation.

image


j9liu:
Self-merging to get a start on the release.

j9liu:
For future reference, my attempted fixes are on the [`points-wpo` branch](https://github.com/CesiumGS/cesium-unreal/tree/points-wpo).

Issue #1872, opened 2026-07-22
Crash in Gaussian Splat Subsystem


Darkere:
### What happened?

We've recently updated Cesium to 2.27.0 (from 2.19.1) and are now experiencing a somewhat random crash in the UCesiumGaussianSplatSubsystem::Tick.

### Environment

Cesium for Unreal version: 2.27.0
Unreal Engine Version: 5.6
Operating System: Windows

Crash randomly appears when changing maps. 

CallStack: 
std::_Atomic_storage<T>::load (atomic:1006)
FWindowsPlatformAtomics::AtomicRead_Relaxed (WindowsPlatformAtomics.h:329)
UObjectBase::GetFlagsInternal (UObjectBase.h:338)
UObjectBase::GetFlags (UObjectBase.h:257)
UObjectBaseUtility::HasAnyFlags (UObjectBaseUtility.h:90)
FInternalUObjectBaseUtilityIsValidFlagsChecker::CheckObjectValidBasedOnItsFlags (Object.h:1831)
IsValid (Object.h:1858)
UCesiumGaussianSplatSubsystem::Tick (CesiumGaussianSplatSubsystem.cpp:249)
FTickableGameObject::TickObjects (Tickable.cpp:196)

There does not appear to be a way to disable this subsystem. We don't make use of cesium outside the editor so this would be a nice option to have.

j9liu:
Hi @Darkere, we saw this on our side and added a fix for it in v2.28.0. Can you upgrade and let us know if that resolves things?

Darkere:
Ah. I misread the commit and thought we had that already. 

Makes sense. We'll be unable to test this unfortunately since we decided to disable the subsystem entirely.

Fsbzsszqs21:
v2.29.0
I have the same problem：
crash when openlevel 



[2026.08.26-11.11.05:469][185]LogGlobalStatus: UEngine::LoadMap Load map complete /Game/core/StartMap
[2026.08.26-11.11.06:078][185]LogWindows: Could not start crash report client using ../../../Engine/Binaries/Win64/CrashReportClient-Win64-Debug.exe
[2026.08.26-11.11.06:078][185]LogMemory: Platform Memory Stats for Windows
[2026.08.26-11.11.06:080][185]LogMemory: Process Physical Memory: 2074.31 MB used, 2309.20 MB peak
[...]

PR #1886, opened 2026-08-26
Fix crash on editor exit in UCesiumGaussianSplatSubsystem::Deinitialize

### LINKED ISSUE #1879: Crash on every editor exit in UCesiumGaussianSplatSubsystem::Deinitialize() (dangling raw pointers at shutdown)
### What happened?

The Unreal Editor crashes on **every editor exit** when Cesium for Unreal is enabled — even in sessions where no Gaussian splat content is ever loaded. The crash is in `UCesiumGaussianSplatSubsystem::Deinitialize()`.

We see two flavors of the same crash across repeated exits (3 crashes in one day, 100% repro on exit for this project):

**Flavor 1 — access violation (2 of 3 crashes):**
```
Unhandled Exception: EXCEPTION_ACCESS_VIOLATION reading address 0x0000000000009704

UnrealEditor_CesiumRuntime!UCesiumGaussianSplatSubsystem::Deinitialize() [CesiumGaussianSplatSubsystem.cpp:129]
```

**Flavor 2 — assertion (1 of 3 crashes):**
```
Assertion failed: Index >= 0 [File:.../Engine/Source/Runtime/CoreUObject/Public/UObject/UObjectArray.h] [Line: 1083]

UnrealEditor_CesiumRuntime!FUObjectArray::IndexToObject() [UObjectArray.h:1083]
UnrealEditor_CesiumRuntime!UCesiumGaussianSplatSubsystem::Deinitialize() [CesiumGaussianSplatSubsystem.cpp:125]
```

### Root cause analysis

`UCesiumGaussianSplatSubsystem` stores its Niagara actor/component as **raw C++ pointers** (not `UPROPERTY`, `TObjectPtr`, or `TWeakObjectPtr`):

```cpp
UNiagaraComponent* _pNiagaraComponent = nullptr;
ACesiumGaussianSplatActor* _pNiagaraActor = nullptr;
UWorld* _pLastCreatedWorld = nullptr;
```

Because these are invisible to the garbage collector, nothing nulls them when the objects they point to are destroyed. During editor shutdown, the editor world (and the transient `CesiumGaussianSplatSystemActor` spawned into it) is torn down and GC'd **before** the engine subsystem's `Deinitialize()` runs. `Deinitialize()` then calls `IsValid(this->_pNiagaraActor)` on a dangling pointer:

- `IsValid()` reads the freed object's `InternalIndex` and calls `FUObjectArray::IndexToObject()` with garbage
- garbage index negative → the `Index >= 0` assertion (flavor 2, line 125)
- garbage index positive → out-of-bounds/dangling chunk read → access violation (flavor 1, attributed to line 129 by optimized line-mapping)

Two contributing factors make this hit **every** session, not just splat-using ones:

1. `Tick()` unconditionally spawns `CesiumGaussianSplatSystemActor` + the Niagara system into the editor world (`initializeForWorld`) even when zero splat tilesets/components exist — so every editor session has an actor to dangle at exit.
2. There is no setting, CVar, or `ShouldCreateSubsystem` override to disable the subsystem for projects that don't use splats (same ask as the reporter in #1872).

This appears to be the shutdown-time sibling of #1872 (`Tick` crash via `IsValid` on the same raw pointers, fixed in 2.28.0) — `Deinitialize()` has the same dangling-pointer pattern but was not covered by that fix. As of `main` today (post-v2.29.0), `Deinitialize()` and the raw pointer members are unchanged, so upgrading does not resolve it.

### Suggested fix

Any of:
- Make `_pNiagaraActor` / `_pNiagaraComponent` `TWeakObjectPtr` (or `UPROPERTY(Transient)` `TObjectPtr`) so shutdown-time GC nulls them, or
- Early-return from `Deinitialize()` when `IsEngineExitRequested()` — everything is being torn down anyway, or
- Don't spawn the splat system actor until the first splat component registers (also avoids the cost for non-splat projects), and/or expose a project setting to disable the subsystem.

### Reproduction

1. Enable Cesium for Unreal 2.28.0 in a UE 5.8 project (no splat content required; project has Cesium 3D Tiles terrain levels, none using `KHR_gaussian_splatting`)
2. Open the editor, work normally
3. Exit the editor → crash reporter fires with one of the two stacks above

### Environment

Cesium for Unreal version: 2.28.0 (Fab/Marketplace install)
Unreal Engine Version: 5.8.0
Operating System: Windows 11 Home 10.0.26200

Viandanze:
Fixes #1879.

## Problem

The Unreal Editor crashes on every editor exit when Cesium for Unreal is enabled, even in sessions where no Gaussian splat content is loaded (see #1879 for full crash logs). Two crash flavors, same root cause:

- `Assertion failed: Index >= 0` in `FUObjectArray::IndexToObject`, from `IsValid(this->_pNiagaraActor)` at `CesiumGaussianSplatSubsystem.cpp:125`
- `EXCEPTION_ACCESS_VIOLATION` from `Deinitialize()` at line 129, via `reset()` → `makeInterfaceDirty()` → `getDataInterface()` dereferencing the dangling `_pNiagaraComponent`

## Root cause

The subsystem holds the Niagara actor/component and the world as **raw pointers**, which are invisible to the garbage collector. During engine shutdown, the editor world and the transient `ACesiumGaussianSplatActor` spawned into it are destroyed **before** the engine subsystem's `Deinitialize()` runs. Both `IsValid()` (which reads `InternalIndex` from potentially-freed memory) and `reset()` (which dereferences `_pNiagaraComponent` through `UNiagaraFunctionLibrary::GetDataInterface`) are then unsafe to call on the dangling pointers.

Since `Tick()` spawns the splat system actor unconditionally (even with zero splat tilesets), every editor session has an actor to dangle at exit - hence the 100% repro rate.

## Fix

Early-return from the cleanup when the engine is exiting. Everything the cleanup would touch is either already destroyed or about to be, so we simply drop the references without calling `IsValid()` or `reset()`. Non-shutdown paths (plugin unload, hot reload) keep the existing behavior with valid pointers.

## Alternatives considered

- **`TWeakObjectPtr` members** would be the more thorough fix, but requires touching ~20 usage sites across the file. Since this subsystem does not own the actor (the world does), weak pointers are the semantically correct long-term direction - happy to do that as a follow-up if preferred.
- **Lazy actor spawn** (only when the first splat component registers) would additionally avoid spawning the Niagara system in projects that never use splats, but that is a behavioral change beyond the scope of this crash fix (also mentioned in #1872).

## Testing

Not compiled locally (no UE 5.8 environment available to me); relying on CI and review. The change is confined to a single function and only adds a guard branch ahead of the existing logic.

### DIFF
--- a/Source/CesiumRuntime/Private/CesiumGaussianSplatSubsystem.cpp
+++ b/Source/CesiumRuntime/Private/CesiumGaussianSplatSubsystem.cpp
@@ -122,6 +122,20 @@ void UCesiumGaussianSplatSubsystem::Initialize(
 void UCesiumGaussianSplatSubsystem::Deinitialize() {
   this->_isTickEnabled = false;
 
+  // The splat actor, its Niagara component, and the world they live in are
+  // owned elsewhere, and the raw pointers to them are invisible to the
+  // garbage collector. During engine shutdown the world and everything in it
+  // may already be destroyed by the time this function runs, leaving the
+  // pointers dangling and even IsValid() unsafe to call on them (#1879).
+  // Everything is being torn down anyway, so simply drop the references.
+  if (IsEngineExitRequested()) {
+    this->_pNiagaraActor = nullptr;
+    this->_pNiagaraComponent = nullptr;
+    this->_pLastCreatedWorld = nullptr;
+    this->SplatComponents.Empty();
+    return;
+  }
+
   if (IsValid(this->_pNiagaraActor)) {
     this->_pNiagaraActor->Destroy();
   }


j9liu:
Hi @Viandanze, thanks for the PR! Before we can review it, we need a signed [Contributor License Agreement (CLA)](https://github.com/CesiumGS/community/tree/main/CLAs) from you. Once we have that in our records, we can add this to our queue. Thank you!

j9liu:
HI @Viandanze, friendly reminder to sign a CLA per my comment above! Otherwise we won't be able to take a look at this PR. :(

Viandanze:
CLA signed, thanks!

PR #1887, opened 2026-08-27
Fix typo in ACesium3DTileset::CreateViewStateFromViewParameters

### LINKED ISSUE #1873: Wrong resolution used with Movie Render Pipeline
### What happened?

When rendering a video, the resolution is retrieved with pPlayerController->GetViewportSize(sizeX, sizeY); inside ACesium3DTileset::GetPlayerCameras() method.
But it's the window resolution, not the rendering resolution.
That has side effects: 
- on aspect ratio, some tiles may be missing on the video border 
- and probably the LOD selection.

We have a quick fix but not universal:
Adding a method:
```
void ACesium3DTileset::PlayMovieSequencerWithResolution(FIntPoint RenderResolution) {
  this->PlayMovieSequencer();
  if (this->_captureMovieMode) {
    this->_captureMovieRenderResolution = RenderResolution;
    UE_LOG(LogCesium, Log,
        TEXT("Cesium capture mode: render resolution set to %dx%d"),
        RenderResolution.X, RenderResolution.Y);
  }
}
```
And in ACesium3DTileset::GetPlayerCameras()
```
    int32 sizeX, sizeY;
    // During MRP capture the OS window stays at its original (smaller) size.
    // GetViewportSize returns the window size, not the render target resolution,
    // so the frustum aspect ratio and pixel count are wrong. Use the render
    // resolution that was stored when capture mode was activated instead.
    if (this->_captureMovieMode &&
        this->_captureMovieRenderResolution.X > 0 &&
        this->_captureMovieRenderResolution.Y > 0) {
      sizeX = this->_captureMovieRenderResolution.X;
      sizeY = this->_captureMovieRenderResolution.Y;
    } else {
      pPlayerController->GetViewportSize(sizeX, sizeY);
    }
    if (sizeX < 1 || sizeY < 1) {
      continue;
    }
```
Then reset the _captureMovieRenderResolution in StopMovieSequencer().

It would be better to retrieve the resolution from Movie Render Pipeline but that add some module dependencies.

Also, pPlayerController->GetPlayerViewPoint shouldn't be use with MRP too.
Something like this should be used:
```
// During movie capture the Movie Render Pipeline resets the player
// controller's view target to the gameplay camera between every captured
// frame (pre-animated state restore). GetPlayerViewPoint() goes through the
// camera manager and therefore returns the gameplay camera position for one
// Cesium tick per frame. This causes the frustum used for tile selection to
// be wrong, and low-priority tiles (bottom of the render camera's frustum)
// are evicted before they can load. When in capture mode, bypass the player
// controller and read the position directly from the current view target's
// camera component instead.
if (this->_captureMovieMode) {
  AActor* ViewTarget = pPlayerCameraManager->GetViewTarget();
  UCameraComponent* CamComp = ViewTarget
      ? ViewTarget->FindComponentByClass<UCameraComponent>()
      : nullptr;
  if (CamComp) {
    location = CamComp->GetComponentLocation();
    rotation = CamComp->GetComponentRotation();
    fov = CamComp->FieldOfView;
  } else if (ViewTarget) {
    location = ViewTarget->GetActorLocation();
    rotation = ViewTarget->GetActorRotation();
  } else {
    pPlayerController->GetPlayerViewPoint(location, rotation);
  }
} else {
  pPlayerController->GetPlayerViewPoint(location, rotation);
}
```

### Environment

Cesium for Unreal version: 2.28.0
Unreal Engine Version: 5.6
Operating System: Windows

### Reproduction steps

1. the bug was detected in iTwin Engage 
2. We just render the video with 1920x1080. The issue was not detected by everyone probably because of the window layout.
3. The issue is probably more noticeable when the aspect ratio is different (square vs 16/9).
...

### Supporting evidence

Image

laurentcau:
Fix typo in ACesium3DTileset::CreateViewStateFromViewParameters, it should use cameraRotation/cameraLocation instead of camera.Rotation/camera.Location, especially when camera.ParameterSource != ECameraParameterSource::Manual.

### DIFF
--- a/Source/CesiumRuntime/Private/Cesium3DTileset.cpp
+++ b/Source/CesiumRuntime/Private/Cesium3DTileset.cpp
@@ -1679,12 +1679,12 @@ ACesium3DTileset::CreateViewStateFromViewParameters(
   double verticalFieldOfView =
       atan(tan(horizontalFieldOfView * 0.5) / actualAspectRatio) * 2.0;
 
-  FVector direction = camera.Rotation.RotateVector(FVector(1.0f, 0.0f, 0.0f));
-  FVector up = camera.Rotation.RotateVector(FVector(0.0f, 0.0f, 1.0f));
+  FVector direction = cameraRotation.RotateVector(FVector(1.0f, 0.0f, 0.0f));
+  FVector up = cameraRotation.RotateVector(FVector(0.0f, 0.0f, 1.0f));
 
   glm::dvec3 tilesetCameraLocation = glm::dvec3(
       unrealWorldToTileset *
-      glm::dvec4(camera.Location.X, camera.Location.Y, camera.Location.Z, 1.0));
+      glm::dvec4(cameraLocation.X, cameraLocation.Y, cameraLocation.Z, 1.0));
   glm::dvec3 tilesetCameraFront = glm::normalize(glm::dvec3(
       unrealWorldToTileset *
       glm::dvec4(direction.X, direction.Y, direction.Z, 0.0)));


j9liu:
Thanks for the PR @laurentcau! I'd like @timoore to verify the correctness of this change, but just to move things along, could you please update `CHANGES.md` to reflect this bugfix?

timoore (approved):
This is the correct change.

j9liu:
Thanks @laurentcau !

PR #1890, opened 2026-09-02
Area paging test

timoore:
## Description

An interface for testing area-based paging by adding an area of interest and LOD to a tileset.

## Author checklist

## Testing plan

Add an area of interest and LOD to an existing tileset.

## Reviewer checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,11 @@
 # Change Log {#changes}
 
+### ? -?
+
+##### Additions :tada:
+
+- Added provisional support for testing area-based paging at fixed LOD.
+
 ### v2.29.1 - 2026-09-01
 
 This release updates [cesium-native](https://github.com/CesiumGS/cesium-native) from v0.63.0 to v0.64.0. See the [changelog](https://github.com/CesiumGS/cesium-native/blob/main/CHANGES.md) for a complete list of changes in cesium-native.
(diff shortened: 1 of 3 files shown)


j9liu:
Thanks for opening this for review, @timoore!

Not a bad start, but I think we'd want the parameters to be less baked into `ACesium3DTileset`. For starters, I think the long/lat region would benefit from being a separate struct, and perhaps with the ability to register multiple of these regions on the tileset.

But going a step further... there's no fundamental difference derived from a `ViewState` derived from a camera, versus a `ViewState` derived from a bounding volume. Should `CesiumCameraManager` become `CesiumViewManager` and allow for defining more arbitrary selection volumes? Or should the bounding volumes be a per-tileset array to use for selection? (And if so, why aren't cameras treated this way? Maybe `CesiumCameraManager` is a good shortcut, but could toggle between letting the user use it as a "default" versus more granularly adding/removing cameras to a `Cesium3DTileset`. Food for thought...) 

Additionally, I can't help but consider `CesiumTileExcluder`, an interface that lets users filter out tiles from selection and rendering. In a way, setting up the `ViewState` with a bounding volume and geometric error threshold feels like a spiritual opposite of `CesiumTileExcluder`. Given some metric to **include** tiles -- e.g., bounding volume, geometric error, tile ID / level -- we grant the user to control over whether or not a tile should be selected. (Perhaps `CesiumTileSelector` is a decent name?)

We were also talking about an abstractable tile selection interface, perhaps with methods like `meetsSelectionCriteria()`, `shouldRenderTile()`, `shouldRefineTile()`, etc. to more granularly guide selection. But I think this is definitely more heavy-handed than passing a simple region / camera frustum to the tileset and saying "get the tiles here" and maybe "exclude these tiles from the original selection".

All of this to say, there's broader implications for how we can improve user control of tileset selection as per https://github.com/CesiumGS/cesium-native/issues/1409. But we don't have to implement all of this now. Instead I think we should:

- Decide whether to continue developing this PR, having "selection regions" as structs that are registered to the tileset.
  - Would help to know if there is immediate community interest for the above (please chime in if so!) 
- Defer implementation, aiming for a broader, more robust selection system that would require more work but is generalizable, covering more use cases.

